PR #149, opened 2017-11-27
chore: remove all vendored dependencies

b5:


b5:
Ok, so this is a bit insane, but I think a step in the right direction. I want to vendor things, but gx + vendoring just isn't working.

PR #199, opened 2018-01-26
feat: add core dataset differ function

### LINKED ISSUE #198: Refactor DatasetDiffer to include jsondiffer and write core method for cli
#### dataset differ
- [x] open WIP pull request
- [ ] update unexporeted sub-component differs (structure, data, meta, etc)
  - [x] to be exported
  - [ ] to take arguments consistent with their scope (structure differ should take structures etc, instead of datasets)
- [ ] update Diff and DiffList structs to align with jsondiff
- [ ] update DiffDataset to output a map of its subcomponents rather than a flat slice of diffs
#### add diff function to core
  - [ ] should be a method on DatasetRequest struct
  - [x] should take a diffParams struct that includes a left and right dataset ref
  - [ ] should also take an empty result struct  and return an error
  - [ ] should allow flexibility to just diff structure or just diff data etc.
  - [ ] should allow for multiple output formats (or this may belong in the cli or api fucntions that call the core function

osterbit:
closes #198

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - checkout
       - run: mkdir -p $TEST_RESULTS
-      - run: go get github.com/jstemmer/go-junit-report github.com/golang/lint/golint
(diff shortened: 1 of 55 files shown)


osterbit:
@b5 I think this should pass if it is rebuilt

PR #236, opened 2018-02-06
feat(diff): added new diff display types

osterbit:


b5:
once this is ready to merge, I think we should merge this as a feat, not a chore. I'd call it:
`feat(diff): added new diff display types`, and add a big description of what this enhancement does. in the commit message. It's an exciting feature, and people need to know about it!

It's worth taking a look at the [changelog](https://github.com/qri-io/qri/blob/master/CHANGELOG.md) from our releases to get a feel for what shows up in change logs.

cmd/diff.go
```diff
@@ -8,17 +8,22 @@ import (
...
 
 var datasetDiffCmd = &cobra.Command{
 	Use:   "diff",
 	Short: "diff two datasets",
 	Long: `
-Diff diffs two datasets`,
+Diff compares two datasets from your repo and prints a represntation 
+of the differences between them.  You can specifify the datasets
+either by name or by their hash`,
 	Example: `todo`,
```
b5: let's make an issue to add examples once this merges.

PR #244, opened 2018-02-07
feat(export): export to zip file and export directory using dataset name

### LINKED ISSUE #212: docs(cmd): `qri export` name-spaced folder inconsistent with description
from the help text for qri export: 
```
Export gets datasets out of qri. By default it exports only a dataset’s data to
the path [current directory]/[peername]/[dataset name]/[data file].
```
however when I run 
```
qri export me/cities
```
or 
```
qri export osterbit3/cities
```
the data is exported into my current working directory rather than to `./osterbit3/`

not sure which behavior we prefer but just wanted to flag the mismatch

### LINKED ISSUE #213: docs(cmd): `qri export` --zip flag returns no output

osterbit:
single line addition I think gets almost everything we wanted working
closes #212 
closes #213

### DIFF
--- a/cmd/export.go
+++ b/cmd/export.go
@@ -59,6 +59,18 @@ To export everything about a dataset, use the --dataset flag.`,
 		ds := res.Dataset
 
 		path = filepath.Join(path, dsr.Name)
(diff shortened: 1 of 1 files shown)


b5:
nice! which issues does this close? should be added to PR description

PR #247, opened 2018-02-07
feat(cmd.Add): display validation error count and detail with optional flag when dataset is added

### LINKED ISSUE #191: Add Validation warnings to add, update
_moved from qri-io/dataset#53_

qri should accept all sorts of data, so long as it parses, but we should _warn_ users when the data they pass doesn't match the schema for that dataset. a simple yellow warning line printed after successful execution of add / update listing the number of errors should work.

Bonus points:
add a link users can click or a command users can run so they can get info on how to fix those validation errors.

osterbit:
closes #191


osterbit:
@b5 let me know what you think about the `--verbose flag`/ if there is a clear, less resource-intensive option that might be better

cmd/add.go
```diff
@@ -115,6 +117,18 @@ func initDataset(name *repo.DatasetRef) {
 	ref := &repo.DatasetRef{}
 	err = req.Init(p, ref)
 	ExitIfErr(err)
+	if ref.Dataset.Structure.ErrCount > 0 {
+		printWarning(fmt.Sprintf("this dataset has %d validation errors", ref.Dataset.Structure.ErrCount))
+		if addDsVerbose {
```
b5: love it. let's keep it and hope it doesn't have bugs 😉

PR #250, opened 2018-02-08
refactor(DatasetRef): use dataset refs whenever dealing with a repo

### LINKED ISSUE #220: Move `/dataset.json` ending deeper in the stack
repo.ParseDatasetRef() takes a string and gives back a datasetRef. This parser also removes the "/dataset.json" ending from the hash.

We need this ending to resolve hashes, however. At some point before the hash gets resolved, we need to make sure this "/dataset.json" ending gets added back on.

### LINKED ISSUE #246: add from peer won't work if the dataset name matches one already in your repo
So qri won't let you add a peer's dataset to your repo, when that dataset has the same name as a dataset already in your repo.

I'm assuming we just give the user a place to add their preferred name. 

Does changing the name of this repo somehow mess something up? We will still have the history of the dataset, as long as we don't change the PreviousPath, correct?

If they don't add their preferred name, we should error, correct? Not save the repo as [datasetname]_1 or [peername]_[datasetname]?

Thoughts?

b5:
Our version of dataset names that we've used so far landed pretty late in the course of shipping 0.1.0. We've set the external expectation that `peername/datasetname` is a proper way to refer to a dataset, but haven't enforced that expectation across the codebase.

The most obvious place this shows up is the `repo.Namestore` interface, which makes less sense in a world where we use `repo.DatasetRef` to move dataset names around.

With that in mind, I've renamed `Namestore` to `Refstore` and changed it's interface to reflect this. The most notable change is `Refstore.GetRef`, which could be renamed to `CompleteRef` in the future, as that's a more accurate description of what it does. If given a peername/name combo,
`GetRef` adds the path, if given a path, `GetRef` adds the peername/name.
This removes a *lot* of code before that used to be `if ref.Path == ""` before calls to `dsfs.LoadDataset`.

In the future we can consider making GetRef an internal method to be used by higher-order repo interface methods. For example repo.GetDataset would internally call repo.GetRef, allowing callers of
repo.GetDataset to just pass in incomplete references, expecting the repo to work it out for them. This'll de-duplicate a bunch of code in core.

closes #220, closes #246.


b5:
pointer logic was getting in the way, and I suspect it'll get us some microscopic performance gains (not having to deref all dem pointers).

Generally, I feel like the go code I've written to date is a little pointer happy, and this seemed like a prime candidate for using `DatasetRef` directly. Not using pointers means as you won't have to box around the ref being nil anymore, which is one less thing that can go wrong.

PR #266, opened 2018-02-13
feat(API tests, save): add api tests and refactor save

### LINKED ISSUE #221: Add API happy path tests for all endpoints that hit local server
* [x] status
* [x] /profile
 * [x] /me
* [x] /profile/photo
* [x] /profile/poster
* [x] /list
* [x] /save
* [x] /remove
* [x] /me/[datasetname]
* [x] /add via url
* [x] /add via file upload
* [x] /rename
* [x] /export
* [x] /history/me/[datasetname]

### LINKED ISSUE #222: Ensure we can use dataset hashes in API endpoints
Once #220 and #273 is closed, add tests for hashes/fix bugs for:

works for all hashes:
* [x] /me/datasetname/at/[datasethash]
* [x] /at/[datasethash]
* [x] /export/peername/datasetname/at/[datasethash]
* [x] /export/at/[datasethash]
* [x] /add/peername/datasetname/at/[datasethash]
* [x] /history/peername/datasetname/at/[datasethash]

only for most recent hash:
* [x] /history/at/[datasethash]
* [x] /remove/peername/datasetname/at/[datasethash]
* [x] /remove/at/[datasethash]

Does not work:
* /add/at/[datasethash], will return error "need peername and dataset name: '/add/[peername]/[datasetname]'"

### LINKED ISSUE #263: Inconsistency with how we handle/load files over the api endpoints
Some endpoints assume the file was sent over the wire. Others load a file from an absolute path.

Not exactly sure what the right way to go with this.

With a typical API, you are pinging a remote server that has no access to your directories. In order to send a file, you would need to send the actual file in the request. 

With our API, we assume the server is local and has access to your directories. The paths must be absolute, because your working directory may be different from the directory the qri node uses as its working directory.

But does it makes sense for the server to be able to load files via the API? Seems like it shouldn't. I had been modeling everything on the API after the CLI, which we can safely assume is local. What if we somehow are requesting a qri node that is ours, but is remote? We need to be able to send the file over the wire.

Okay, it seems like what we should really do, is accept both. If the data fields are filled, don't load files. If they aren't, load files. If those files/directories don't exist, error.

### LINKED ISSUE #268: the core Save function should be refactored to match the patterns found in the core Init function
The Init function has some properties that the save function does not:

We are able to initialize a dataset, by pulling the data from a given URL.
We should be able to save an update to a dataset by pulling data from a given URL as well!

The Initparams for meta, structure, and data are all io.Readers, let's model the SaveParams after that.

Let's fix some bugs that cause the datasetDiffer to only allow a save if data has been changed, but doesn't take into account the changes of meta and structure.

Woo!

ramfox:
closes #221 Add API Tests
closes #263 inconsistency with how we handle/load files over the api endpoints
closes #268 refactor of the core save function to better match the patterns set up in the init function, and also fix bugs that don't allow saves when changes to meta and structure (and not data) have been made


cmd/save.go
```diff
@@ -31,7 +29,7 @@ var saveCmd = &cobra.Command{
 	Short: "save changes to a dataset",
 	Long: `
 Save is how you change a dataset, updating one or more of data, metadata, and 
-structure. Every time you run save, an entry is added to your dataset’s log 
+structure. You can also update your data via url. Every time you run save, an entry is added to your dataset’s log
```
b5: These strings are tuned to use an 80 character ruler for cleaner CLI output. Sublime text has view > ruler > 80 to put a ruler in yo editor.

api/server_test.go
```diff
@@ -136,22 +45,239 @@ func TestServerRoutes(t *testing.T) {
...
+		if res.StatusCode != 200 {
+			t.Errorf("testMimeMultipart case %d, %s - %s:\nstatus code mismatch. expected: %d, got: %d", i, c.method, c.endpoint, 200, res.StatusCode)
+			continue
+		}
+	}
+}
+
+// MakeMimeMultipartBody takes a map of filepaths and params and add thems to a writer
+// adds that writer to *bytes.Buffer that can then be used as the body for a request
+func MakeMimeMultipartReq(method, endpoint, url string, filePaths, params map[string]string) (*http.Request, error) {
```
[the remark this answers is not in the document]
b5: consider using `NewFilesRequest`. Also, I would reword the comment:
```golang
// NewFilesRequest creates a mime/multipart http.Request with files specified by a map of param : filepath,
// and form values specified by a map, params
```
When writing comments, try to think about the consumer of the function instead of it's internals.

cmd/save.go
ramfox: omg thank you I've been stressing about this

PR #270, opened 2018-02-16
refactor(profile): profile has separate function for saving peername

### LINKED ISSUE #257: Validation for profile info
Users shouldn't be able to update:
id, created, updated, profile, poster, thumb, type

But currently can

Also, need to validate email, name, description, homeUrl, color, & twitter before saving

Should users be able to change their peername??
Edit: as of right now, no. Peername is currently our source of truth as to who our peer is (will change to peerID, but for right now let's cover our butts and not let the user change their peername after setup)

Also, should be able to update only one field without writing over the entire profile.

Profile is important, because it is the source of truth for peernames. If a peername is overridden, we lose a lot of functionality

### LINKED ISSUE #259: Changing your peername messes everything up
If you change your peername, using the API `/profile` endpoint or the `qri profile` command, you mess up all `gets` on past datasets. Datasets keep a record of their peername, and when profile is updated with a new peername, the peername on each of these datasets does not change.

So, does this mean we just don't allow peers to change their peername, once set?
Or do we all changes to a peername, and traverse through a peer's datasets and change the peername?

ramfox:
In the future, when we move away from peername as a source of truth and toward ID, we can probably get rid of this. For now, however, let's make sure that, for safety, the only way to save a peername is via the SavePeername function

We now have a ValidateProfile function that uses the qri-io/jsonschema package to validate the profile before saving.

closes #259 
closes #257 Validation for profiles


core/profile.go
```diff
@@ -106,35 +108,137 @@ func (r *ProfileRequests) GetProfile(in *bool, res *Profile) error {
...
 }
 
-// SaveProfile stores changes to this peer's profile
-func (r *ProfileRequests) SaveProfile(p *Profile, res *Profile) error {
-	if r.cli != nil {
-		return r.cli.Call("ProfileRequests.SaveProfile", p, res)
-	}
-	if p == nil {
-		return fmt.Errorf("profile required for update")
+// AssignEditable - help with this name plz
```
b5: I think AssignEditable is fine for now. We can revisit on a larger code review later on.

PR #280, opened 2018-02-21
WIP - feat(api): updates to API for webapp

### LINKED ISSUE #285: feat: local qri frontend
Finallllly ready to sew our frontend into the qri to ship as a webapp on `localhost:2505`. While most of the planning is happening over on qri-io/frontend#242, there are steps to complete on this repo to ship in time.

* [x] Close qri-io/frontend#242
* [ ] Provision a canonical qri node & add an IPNS entry that points to the hash of the bundled webapp
* [ ] Have qri grab this hash from IPFS & serve it
* [ ] add a note to the boot sequence of `qri connect` to let users know they can see the webapp at `localhost:2505`

### LINKED ISSUE #294: Add diff API endpoint
Let's add a diff api endpoint so we can get some fancy diffing views for our datasets on the frontend!!

To proceed:

* [x] on server.go, add handler for /diff
* [x] to be able to marshal response body json to a go struct, create diffAPIParams struct:
```
type diffAPIParams struct {
    Left, Right  String
    Format       String
}
```
Left and Right are the datasetRef paths

The Format field is a string. If it matches the possible formats that we can mine from dsdiff, diffHandler should return a formatted string (using dsdiff.MapDiffsToString). If it's an empty string, the handler should return a json object of the diffs.

* [x] write diffHandler, that either gets the core.DiffParams by decoding from json or grabbing the FormValues. Call to core.Diff and return either a formatted string (if the diffAPIParams.Format field is not empty) or the diffs.

* [x] test /diff endpoint with and without format option

b5:
Resurrecting our frontend as a webapp has brought a number of issues to light that this PR is intended to address.

* add CORS availability on `localhost:2505`, the port we're intending to use for serving the webapp.
* The repo/test package now uses `qri-io/dataset/dstest`. While in there, cdxj support is slated for removal from the dataset package in the near future, so this PR removes the 'archive' dataset used in a few tests.
* add back `/data` endpoint, using a refactored `core.StructuredData` method

closes #285

@ramfox edit:
closes #294


b5:
lol once again I can't self review, but this looks great, LGTM.

🌮 🍥 🇲🇬

PR #311, opened 2018-03-12
fix(structuredData): data should return DataPath not dataset Path

### LINKED ISSUE #310: bug at /data api endpoint -> hash of dataset returned, not hash of data
When we send data via the API, we send two pieces -> the data itself and a hash of the data

However, right now, we are actually sending the hash of the dataset itself, not the data

ramfox:
closes #310

### DIFF
--- a/api/datasets.go
+++ b/api/datasets.go
@@ -5,11 +5,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/qri-io/dataset"
 	"net/http"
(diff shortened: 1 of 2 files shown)


ramfox:
not sure what is causing the errors :/ not getting them on my end, and I can only see what tests have failed, not what the failing tests are returning.

Got the same errors before I pulled from the qri-io/dataset master branch, but circleci seems to be up-to-data

Any thoughts, I'm lost?

PR #323, opened 2018-03-23
feat(read-only): add read-only mode to qri server

### LINKED ISSUE #312: feat(api): add option for read only api
* [ ] add readonly flag to cli
* [ ] add to api middleware, if readonly, don't let POST, PUT, DELETE requests through
* respond with 403 on GET calls to:
  * [ ] to dataHandler
  * [ ] to /me /profile

Only endpoints that need to respond with 200:
GET /list/
GET / endpoint (/peername/datasetname, /at/path, /peername)
GET /connect/path
GET /history/

endpoints that need blocking:
/ipfs/ 
/ipns/ 
/profile 
/me 
/profile/photo 
/profile/poster 
/peers 
/peers/ 
/connections 
/list 
/save 
/save/ 
/remove/ 
/me/ 
/add 
/add/ 
/rename 
/export/ 
/diff 
/data/ 

So initially the thought was to filter out anything that isn't a GET, but actually we want to scope much less than that

ramfox:
closes #312 

Additionally:
* adds tests for read-only server

### DIFF
--- a/api/config.go
+++ b/api/config.go
@@ -73,6 +73,8 @@ type Config struct {
 	RPCPort string
 	// set to to disable webapp
 	WebappPort string
+	// read-only mode
+	ReadOnly bool
 	// list of addresses to bootsrap qri peers on
 	BoostrapAddrs []string
 	// WebappScripts is a list of script tags to include the webapp page, useful for using alternative / beta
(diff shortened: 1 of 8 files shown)


ramfox:
@b5, this is almost done, my only question is should I be letting OPTIONS through on all endpoints, even when in ReadOnly?
Or just letting OPTIONS through on endpoints that aren't forbidden when in read-only mode?

Can you tell I haven't decided which way punctuate readonly?

b5:
looks great. As for OPTIONS requests, I think they should all be allowed through for now to make our CORS story simpler. In the future we may need to revisit this, but we'll deal with that later. So, let's leave as is :)

PR #331, opened 2018-04-04
Dockerfile: Use unambiguous adduser arguments

hexagon6:
-D & -h are amgibuous arguments and could mean different things on different operating systems.

I assumed that with `-D` `--disabled-password` was meant. `-h` usually resolves to `--help`, but `--home` was meant.

These changes fix errors when building the docker image.


b5:
I'm going squash-merge this to be consistent with our [commit conventions](https://github.com/qri-io/qri/blob/master/DEVELOPERS.md#-git-commit-guidelines).

Thanks so much for your contribution @hexagon6!

PR #372, opened 2018-04-25
test(core.Profile): testcase for SaveProfile

dustmop:
testcase for core.SaveProfile, specifically, the assumptions it makes and idiosyncrasies of how it
currently works.

### DIFF
--- a/core/profile_test.go
+++ b/core/profile_test.go
@@ -3,9 +3,12 @@ package core
 import (
 	"os"
 	"path/filepath"
+	"reflect"
 	"testing"
+	"time"
 
 	"github.com/ipfs/go-datastore"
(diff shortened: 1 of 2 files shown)


core/profile_test.go
```diff
@@ -73,6 +76,105 @@ func TestProfileRequestsSave(t *testing.T) {
 	}
 }
 
+func TestSaveProfile(t *testing.T) {
+	// Mock data for the global Config's Profile, used to create new profile.
+	// TODO: Remove the randomly built Profile that config.DefaultProfile creates.
```
b5: I like this TODO. It'd be great if we put it in `profile.NewRandomProfile()` or something

PR #384, opened 2018-05-08
refactor(p2p): extract NewTestNetwork into its own p2ptest package

dustmop:
Will enable easier testing of p2p package. Also move NewTestRepo to repo/test, and create
NewDatasetRequests constructor that can take a QriNode parameter.

### DIFF
--- a/p2p/test/p2ptest.go
+++ b/p2p/test/p2ptest.go
@@ -1,0 +1,75 @@
+package p2ptest
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+	"github.com/qri-io/qri/config"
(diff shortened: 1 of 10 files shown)


p2p/test/p2ptest.go
```diff
@@ -0,0 +1,75 @@
+package p2ptest
+
+import (
+	"context"
```
b5: So this is super nitpickie, but I prefer separate three categories of imports with a blank line, in the form
```golang
import (
  [standard library packages]

  [third party packages]
  
  [gx packages]
)
```
I find it makes it easier to read what a file is importing this way, no need to change this for this PR, we can catch it later on, just wanted to mention.

PR #386, opened 2018-05-08
test(p2p): test for DatasetRequests.List over p2p

dustmop:
Test DatasetRequest.List over p2p by building a test network and calling List on each peer.
Move the rest of p2p/p2p_test into p2p/test/p2ptest so that NewTestDirNetwork and ConnectNodes
can be used in other tests. NewTestRepoFromProfileID will populate the test repo with data if
given an index as its second parameter.


b5:
@dustmop mind rebasing, we should be able to get this merged now

dustmop:
Rebased and pushed.

Issue #381, opened 2018-05-07
Connected peers meta-issue


b5:
## Goal
Reduce the ambiguity around "who is online" by only showing online profiles by default. 

## History
Currently qri has been fudging the notion of online & offline peers. For example the p2p `profiles` method returns cached profiles, and no mechanism for distinguishing "online peers only please". This means that when asking for peers we sometimes get back datasets that aren't in fact accessible.
I think the source of this is the very human desire to show users all the neat data that exists on qri, but we're kinda setting them up for failure if we show data that isn't available.
When the network is small, with very little reinforcement.

## Is this going backwards?
In the immediate future yes, it is. The net change here is peers will by default tell each other about fewer peers, and the network will (by default) be limited to who is online at a given moment. I prefer to think of this change as moving toward a "more honest" representation of the network, because we're exposing more information about who is and isn't online.

Long term, I think this is where registries come into play. Registries allow us to draw an explicit box around "federated behaviours", which will allow us to make the network healthier. To that end, I think it'll be the job of registries to keep track of _who has a copy of which datasets_ (keeping in mind that peers can add each-other's datasets, making them available). With registries carrying this info, we can show users currently unavailable datasets, and spec out features that allow users to say stuff like "if you see this dataset anywhere/anytime, grab it for me". Because this behaviour is more advanced than a simple "this is what's available right now", we'll do it after clarifying the online/offline problem.

## Changes
In the abstract, this issue can be closed when we:
* add an online flag to profile representations
* any request that deals with collections of peers should be changed to only return information from online peers
* parameterize any of those requests to add back offline / cached profile information
* have tests that prove the online requests work as expected, and a plan for adding tests for requests that list offline/cached profiles

b5:
ok, this is largely closed save for one point I'm going to spin off to it's own issue:
* have tests that prove the online requests work as expected, and a plan for adding tests for requests that list offline/cached profiles

Other than that I'd call this closed by #388, #390, and #392

PR #388, opened 2018-05-09
feat(p2p): transition to online-first peer display, p2p networking fixes, tag qri peers in p2p.ConnManager, merge "qri connections" command into "qri peers list"

b5:
### transition to online-first peer display
In an ongoing effort to make our network more functional, we're switching to only showing connected peers by default. This brings a bunch of changes, with the most important being adding in a mechanism for peers joining into the network.

All of these minor changes make doing things like node.ConnectedQriPeers much easier & more consistent. The tougher change comes from the paradigm shift from gossip-oriented to live connections, and how new peers are discovered.

In previous iterations, peers simply passed around their gossip-oriented list of known profiles, which
meant that if peer A connects to peer B, peer B stored a gossip-reference to peer A, if peer C came along and asked B for it's "profiles", A would be included in that list, now C can discover A without
ever connecting to A.

Now that we're favouring only showing peers that are currently online, this is tougher. To address this,
I've added an AnnounceConnect method that fires after initial boostrapping. The annoouncer (message initiator) adds details on how to connect to them to a message, and broadcasts this message to all currently connected peers (aka, whoever this peer *just* finished bootstrapping to). Receivers who are seeing this message for the first time echo this message to all peers *they* are connected to, and each of those peers attempts to connect to the recently-joined peer.

It's worth noting that at scale this is going to be a little nuts. So eventually we'll add rate limiting
and a "hops" count that stops propagating the message after a little while. These two things combined with a connections high-water mark should do the trick.

### qri & `node.Host.ConnManager`
When operating on "real" IPFS networks, we need to let the connection manager know
that peers supporting the qri protocol have more value to us. I've added a default
value of 6 as a starting point.

One day when we have time to deal with stuff like this I'd love a test that checks for priortization of connetions amongst a number of qri & non-qri peers. There should be situations where we choose to deprioritize a qri peer. One thing we may do is use reputation coming from registries we've opted into to tag bad-actor peers with a *negative* tag value.

### merge `qri connections` into `qri peers list`
`qri peers list` now does much nicer things, showing online peers by default when connected, and cached peers by default when disconnected. Less commands == better.

Makes significant inroads on #381.


api/peers.go
```diff
@@ -88,9 +88,19 @@ func (h *PeerHandlers) ConnectionsHandler(w http.ResponseWriter, r *http.Request
 
 func (h *PeerHandlers) listPeersHandler(w http.ResponseWriter, r *http.Request) {
 	args := core.ListParamsFromRequest(r)
-	args.OrderBy = "created"
+	// args.OrderBy = "created"
+	cached, err := util.ReqParamBool("cached", r)
```
ramfox: If you just hit /peers with no "cached" param, you get this error: strconv.ParseBool: parsing "": invalid syntax

I don't think we need to care about the error? assuming that ReqParamBool will only have cached == true if we supply "true" or "t" anything else should be false anyway

b5: ah shit yeah I forgot about this. Will fix. At some point we should PR `datatogether/api/apiutil` to add a default param to `ReqParamBool` so that it returns the default value when no value is present, instead of erroring.

PR #392, opened 2018-05-10
feat(Profile Info, Connections): verbose profile info, better explicit conn mgmt

b5:
Added a series of tweaks & refinements surrounding `qri peers connect` & `qri peers disconnect`:
* Made disconnect close *all* open connections (so disconnect actually works) 
* removed dial backoff checks from explicit connections
* added a -verbose flag to `qri peers info` that'll yank network addresses from the host Peerstore, which is useful for debugging how peers found each other.

### DIFF
--- a/cmd/peers.go
+++ b/cmd/peers.go
@@ -26,17 +26,17 @@ var peersInfoCmd = &cobra.Command{
 	PreRun: func(cmd *cobra.Command, args []string) {
 		loadConfig()
 	},
+	Args: cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		if len(args) < 1 {
-			ErrExit(fmt.Errorf("peer name is required"))
-		}
+		req, err := peerRequests(false)
+		ExitIfErr(err)
 
-		printInfo("searching for peer %s...", args[0])
-		req, err := peerRequests(true)
+		v, err := cmd.Flags().GetBool("verbose")
 		ExitIfErr(err)
 
 		p := &core.PeerInfoParams{
 			Peername: args[0],
+			Verbose:  v,
 		}
 
 		res := &config.ProfilePod{}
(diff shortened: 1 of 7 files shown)


config/profile.go
```diff
@@ -48,6 +48,9 @@ type ProfilePod struct {
 	// where QmSy... is a peer identifier on the IPFS peer-to-peer network
 	// Should not serialize to config.yaml
 	PeerIDs []string `json:"peerIDs,omitempty" yaml:"peerIDs,omitempty"`
+	// NetworkAddrs keeps a list of locations for this profile on the network as multiaddr strings
+	// Should not serialize to config.yaml
+	NetworkAddrs []string `json:"networkAddrs,omitempty" yaml:"networkAddrs,omitempty"`
```
dustmop: Could you modify the test TestWriteToFileWithAddresses in config/config_test.go so that it assigns to NetworkAddrs without causing the test to fail? Also, we should rename TestWriteToFileWithAddresses to TestWriteToFileWithExtraData or something, since it's not about "Addresses" being non-serializable.

cmd/peers.go
```diff
@@ -26,17 +26,17 @@ var peersInfoCmd = &cobra.Command{
 	PreRun: func(cmd *cobra.Command, args []string) {
 		loadConfig()
 	},
+	Args: cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		if len(args) < 1 {
-			ErrExit(fmt.Errorf("peer name is required"))
-		}
+		req, err := peerRequests(false)
```
dustmop: Just to double-check, this is changing to work for offline peers?

b5: kinda, this `false` value is a bit of a misnomer, I'm planning on removing it entirely in the future, it currently only sets if a command _must_ be connected to the p2p network to work. But generally speaking this value is stupid, and should be ignored until we kill it.

PR #401, opened 2018-05-15
refactor(commit message): new default first commit message

b5:
replaced "initial commit" with "created dataset", which is just easier to understand


b5:
ok we gotta ship so I'ma merge

PR #413, opened 2018-05-22
perf(Add, Save): Overhaul Dataset Add & Save Methods

b5:
This is a relatively large PR that overhauls the guts of dataset Init & Add


dustmop:
There's a ton changing here, only some of which I'm familiar with, but after a quick read, nothing bad seems to stand out. Approved.

PR #428, opened 2018-06-01
refactor(yaml): changes to make yaml our default format

### LINKED ISSUE #425: `qri export` should give the option to export as yaml or json
Our new `qri add --file=dataset.yaml` and `qri export --blank` both use .yaml files
However, when we export structure or meta from qri (`qri export --structure`) it exports as json. This makes it harder for the user to use a structure from different but similarly formatted dataset on a new dataset.

Let's add a `--format` flag that allows you to export as either yaml or json

### LINKED ISSUE #426: yaml should be our default export format for everything that isn't the body/data fight me
Since our `qri add --dataset` flag accepts a yaml file, and our qri config lives naturally in yaml, yaml should be our default format for everything except for the dataset body/data.

### LINKED ISSUE #427: replace `gopkg.in/yaml.v2` with `github.com/ghodss/yaml`
The ghodss/yaml package converts from yaml to json to go, and back.

The `gopkg.in/yaml.v2` package was creating weird field name and structures from our more complicated json and structs (namely `structure.schema` struct)

We only want to use one yaml package. Going through each file that imports `gopkg.in/yaml.v2` and seeing what has to changing what we need in order for `github.com/ghodss/yaml` to play nice.

ramfox:
closes #427 
closes #425 
closes #426


cmd/export.go
```diff
@@ -157,18 +160,38 @@ To export everything about a dataset, use the --dataset flag.`,
 			}
 
 			metaPath := filepath.Join(path, dsfs.PackageFileMeta.Filename())
-			mdBytes, err := json.MarshalIndent(md, "", "  ")
+			mdBytes := []byte{}
```
b5: This should be `var mdBytes []byte`. 

`mdBytes := []byte{}` allocates a byte slice with a length & cap of 0 and assigns it to `mdBytes`. This initial allocation is never used b/c we're only ever assigning `mdBytes` to other newly-allocated byte-slices below. 

`var mdBytes []byte` declares "I need memory for a slice of bytes", but does no actual byte-slice allocation.

cmd/export.go
```diff
@@ -270,3 +303,18 @@ structure:
 # or a URL that leads to the raw data
 # dataPath:
 `
+
+// FilenameToYAML takes a filename, removes any format ending (if there is one), and replaces it with yaml
+// should probably be generalized to
+// FilenameFormatChange(filename string, newEnding string) string
+func FilenameToYAML(filename string) string {
```
b5: I think instead of this custom function we could use a `filepath` one-liner whenever you need to manually construct a filename with extension:
```golang
fmt.Sprintf("%s.yaml", filepath.Base(filename))
```

there's also:
```golang
// dataset data formats have a string that matches their extension name:
fmt.Sprintf("%s.%s", filepath.Base(filename), dataset.JSONDataFormat.String())

// if you have a dataset with a structure, you can do this:
fmt.Sprintf("%s.%s", filepath.Base(filename), ds.Structure.Format.String())
```

This carries the added safety of working a little better cross-platform where path strings can get weird.

config/profile.go
```diff
@@ -42,15 +42,15 @@ type ProfilePod struct {
 	Twitter string `json:"twitter"`
 	// Online indicates if the user is currently connected to the qri network
 	// Should not serialize to config.yaml
-	Online bool `json:"online,omitempty" yaml:"online,omitempty"`
+	Online bool `json:"online,omitempty"`
```
b5: oh that's nice

PR #430, opened 2018-06-01
fix(cmd): changes to add, diff, and global config flag

### LINKED ISSUE #429: `qri add` bug: `message` and `title` flags don't do anything
The `message` and `title` flags are there in case someone wants to overwrite the message in the dataset.commit, or if they don't add a dataset but want a custom initial commit message.

Right now, they don't do anything.

### LINKED ISSUE #431: `qri diff` soooo slow on large-ish datasets
Also frustrating because in these two particular versions of the dataset, the data should be the same. Seems like we should be able to check hashes and know if the data is the same or not.

The dataset I'm trying this on using 9MBs (ramfox/lolstats2018)

ramfox:
closes #429 
closes #431

### DIFF
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -19,6 +19,18 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	bite = 1 << (10 * iota)
+	kilobyte
+	megabyte
+	gigabyte
+	terabyte
+	petabyte
+	exabyte
+	zettabyte
+	yottabyte
+)
+
 var printPrompt = color.New(color.FgWhite).PrintfFunc()
 var spinner = sp.New(sp.CharSets[24], 100*time.Millisecond)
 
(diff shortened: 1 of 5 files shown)


cmd/print.go
```diff
@@ -46,6 +58,48 @@ func printNotYetFinished(cmd *cobra.Command) {
...
 }
 
+func printByteInfo(l int) string {
+	length := struct {
+		name  string
+		value int
+	}{"", 0}
+
+	switch {
+	// yottabyte and zettabyte overflow int
```
b5: This is the best comment in our codebase.

cmd/print.go
```diff
@@ -46,6 +58,48 @@ func printNotYetFinished(cmd *cobra.Command) {
 	color.Yellow("%s command is not yet implemented", cmd.Name())
 }
 
+func printByteInfo(l int) string {
```
b5: `printByteInfo` should probably be `byteInfoString`

PR #432, opened 2018-06-04
refactor(datasets): Use Copy and PagedReader. Rename StructuredData.

dustmop:
Use new utilities Copy and PagedReader instead of copying entries manually. Rename StructuredData (the method) to LookupBody, rename StructuredData (the struct) to LookupResult.

### DIFF
--- a/api/datasets.go
+++ b/api/datasets.go
@@ -655,7 +655,7 @@ func (h DatasetHandlers) dataHandler(w http.ResponseWriter, r *http.Request) {
 		err = nil
 	}
 
-	p := &core.StructuredDataParams{
+	p := &core.LookupParams{
 		Path:   d.Path,
 		Format: dataset.JSONDataFormat,
 		Limit:  limit,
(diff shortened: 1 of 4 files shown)


api/datasets.go
```diff
@@ -655,24 +655,24 @@ func (h DatasetHandlers) dataHandler(w http.ResponseWriter, r *http.Request) {
 		err = nil
 	}
 
-	p := &core.StructuredDataParams{
+	p := &core.LookupParams{
```
b5: nice. `lookup` > `structuredData`.

core/datasets.go
```diff
@@ -576,32 +576,32 @@ func (r *DatasetRequests) Remove(p *repo.DatasetRef, ok *bool) (err error) {
...
-// StructuredData combines data with it's hashed path
-type StructuredData struct {
+// LookupResult combines data with it's hashed path
+type LookupResult struct {
 	Path string `json:"path"`
+	// TODO: Rename to Body
 	Data []byte `json:"data"`
 }
 
 // StructuredData retrieves dataset data
```
b5: should be `LookupBody retrieves dataset data`

core/datasets.go
```diff
@@ -576,32 +576,31 @@ func (r *DatasetRequests) Remove(p *repo.DatasetRef, ok *bool) (err error) {
...
 	FormatConfig  dataset.FormatConfig
 	Path          string
 	Limit, Offset int
 	All           bool
 }
 
-// StructuredData combines data with it's hashed path
-type StructuredData struct {
+// LookupResult combines data with it's hashed path
+type LookupResult struct {
```
b5: This is great for now. In the future I think we should:
* land a change to `dataset.DatasetPod` that renames "Data" to "Body"
* remove `LookupResult` & replace with a `repo.DatasetRef` that has `ref.Dataset.Body` populated

PR #438, opened 2018-06-04
feat(cmd/export): specify data format using --data-format flag

### LINKED ISSUE #435: CLI export command: add --data-format flag
Add flag that will allow users to export data in csv, json, or cbor.

ramfox:
closes #435 

Also replaces redundant --data flag with --no-data flag (data is always exported by default, and if there are no flags then just the data will export)

### DIFF
--- a/cmd/export.go
+++ b/cmd/export.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"encoding/json"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -12,6 +11,7 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/qri-io/dataset"
 	"github.com/qri-io/dataset/dsfs"
+	"github.com/qri-io/dataset/dsio"
 	"github.com/qri-io/dataset/dsutil"
 	"github.com/qri-io/qri/core"
 	"github.com/qri-io/qri/repo"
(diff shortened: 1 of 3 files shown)


cmd/export.go
```diff
@@ -49,6 +49,10 @@ To export everything about a dataset, use the --dataset flag.`,
 		requireNotRPC(cmd.Name())
 		path := cmd.Flag("output").Value.String()
 		format := cmd.Flag("format").Value.String()
+		dataFormat := cmd.Flag("data-format").Value.String()
+		if dataFormat != "" && !(dataFormat == "json" || dataFormat == "csv" || dataFormat == "cbor") {
```
b5: In the future we should think about adding a function to the dataset package that makes this a litle less verbose, but that doesn't exist so this is great :)

cmd/export.go
```diff
@@ -195,14 +199,40 @@ To export everything about a dataset, use the --dataset flag.`,
 		}
 
 		if exportCmdData {
+
```
b5: we should use `core.LoadDataset` to get the path, then `core.LookupBody` to get the data of duplicating this code here.

cmd/export.go
ramfox: This comment makes me happy, because my instinct was to do just that but I didn't follow it. So next time I'll follow it.

cmd/export.go
```diff
@@ -194,15 +196,32 @@ To export everything about a dataset, use the --dataset flag.`,
...
 		}
 
-		if exportCmdData {
-			src, err := dsfs.LoadData(r.Store(), ds)
+		if !exportCmdNoData {
+			if dataFormat == "" {
+				dataFormat = ds.Structure.Format.String()
+			}
+
+			df, err := dataset.ParseDataFormatString(dataFormat)
```
b5: 👌

PR #440, opened 2018-06-06
WIP feat: add search

osterbit:


cmd/print.go
```diff
@@ -100,6 +101,31 @@ func printByteInfo(l int) string {
 	return fmt.Sprintf("%v %s", length.value, length.name)
 }
 
+func printSearchResult(i int, result core.Result) {
+	white := color.New(color.FgWhite).SprintFunc()
+	green := color.New(color.FgGreen).SprintFunc()
+	// NOTE: in the future we need to switch based on result.Type
```
b5: try to use `TODO` for anything that we need to do moving forward. codeclimate picks up on `TODO` automatically & helps us keep track of tech debt.

core/search.go
```diff
@@ -30,56 +27,60 @@ func NewSearchRequests(r repo.Repo, cli *rpc.Client) *SearchRequests {
...
+	for i, result := range results {
+		searchResults[i].Type = result.Type
+		searchResults[i].ID = result.ID
+		searchResults[i].Value = result.Value
+	}
+	*res = searchResults
+	return nil
 }
 
 // ReindexSearchParams defines parmeters for
```
b5: Any particular reason this is being removed? If so we should delete it outright

core/search_test.go
```diff
@@ -3,69 +3,109 @@ package core
 import (
 	"testing"
 
-	"github.com/qri-io/qri/repo"
+	// "github.com/qri-io/qri/repo"
```
b5: let's delete this. Apologies for introducing the bad habit of commenting out imports, I'm hoping we can all avoid it in the future 😃

core/search_test.go
```diff
@@ -3,69 +3,109 @@ package core
...
-	mr, err := testrepo.NewTestRepo(nil)
+func TestSearch(t *testing.T) {
+	rc, _ := regmock.NewMockServer()
+	mr, err := testrepo.NewTestRepo(rc)
 	if err != nil {
 		t.Errorf("error allocating test repo: %s", err.Error())
 		return
 	}
 
+	cases := []struct {
```
b5: If there's only 1 search case we can drop the table test pattern & just run a raw test without the `cases` struct & loop.

cmd/print.go
```diff
@@ -100,6 +101,31 @@ func printByteInfo(l int) string {
...
+	// For now we are taking a shortcut and assuming a dataset struct
+	dsname := result.ID
+	resultStruct := &registry.Dataset{}
+	resultString := dsname
+	bytes, err := json.Marshal(result.Value)
+	if err == nil {
+		err = json.Unmarshal(bytes, resultStruct)
+		if err == nil {
+			var title, desc string
+			if resultStruct != nil {
```
dustmop: No need to check for nil here, resultStruct is both constructed a few lines above and also passed as an out parameter to Unmarshal.

core/search.go
```diff
@@ -35,51 +36,57 @@ func NewSearchRequests(r repo.Repo, cli *rpc.Client) *SearchRequests {
 	}
 }
 
+// Result struct
+type Result struct {
```
dustmop: Would recommend calling this SearchResult, both to make it clear in other places what role this struct serves, and also to bring it closer in line with SearchParams, and other structs like LookupParams / LookupResults in core/datasets.go

core/search.go
osterbit: just postponed dealing with it to try to get this pr out sooner.  can file an issue or work on here depending on your preference

osterbit:
superseded by #460

PR #442, opened 2018-06-06
chore(transform): Copy util for handling output from transformation

dustmop:


b5:
but who cares let's merge

PR #444, opened 2018-06-06
feat(registry cmd): add commands for working with registries

b5:
I've refactored a bunch of the logic around working with registries to make the process more explicit & controllable.

### DIFF
--- a/cmd/registry.go
+++ b/cmd/registry.go
@@ -38,7 +38,7 @@ $ qri config set registry.location ""`,
 	},
 }
 
-// publishCmd represents the export command
+// publishCmd represents the publish command
 var publishCmd = &cobra.Command{
 	Use:   "publish",
 	Short: "publish dataset info to the registry",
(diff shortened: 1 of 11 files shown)


cmd/registry.go
```diff
@@ -0,0 +1,97 @@
...
+
+You can opt out of registries entirely by running:
+$ qri config set registry.location ""`,
+
+	Annotations: map[string]string{
+		"group": "network",
+	},
+}
+
+// publishCmd represents the export command
```
dustmop: "the publish command"

cmd/registry.go
```diff
@@ -0,0 +1,97 @@
...
+			ExitIfErr(err)
+
+			err = req.Publish(&ref, &res)
+			ExitIfErr(err)
+			printInfo("published dataset %s", ref)
+		}
+	},
+}
+
+// unpublishCmd represents the export command
```
dustmop: "the unpublish command"

PR #455, opened 2018-06-11
feat: second pass on core search and command line search implementations

osterbit:
- still need to refactor and move the contents of core/search down to repo/actions/registry so that core/search is a wrapper and searches both the registry and your own repo, but will put that in a separate PR if that's okay.


osterbit:
superseded by #460

PR #460, opened 2018-06-12
feat(Search): added implementations for core search and command line search

osterbit:


cmd/search.go
```diff
@@ -54,33 +53,28 @@ func (o *SearchOptions) Complete(f Factory, args []string) (err error) {
...
+	p := &core.SearchParams{
+		QueryString: o.Query,
+		Limit:       100,
+		Offset:      0,
 	}
-	res := []repo.DatasetRef{}
+	results := &[]core.SearchResult{}
 
-	if err = o.SearchRequests.Search(p, &res); err != nil {
+	if err = o.SearchRequests.Search(p, results); err != nil {
```
b5: I think it's better to have `results` be a `[]core.SearchResult{}` instead of a pointer-to `[]core.SearchResult`, as there's some danger that others will use `results` later in this function in a way where the pointer messes things up. Instead I'd prefer we do this:
```golang
results := []core.SearchResult{}
if err = o.SearchRequests.Search(p, &results); err != nil {
// ...
```

cmd/search.go
```diff
@@ -54,33 +53,28 @@ func (o *SearchOptions) Complete(f Factory, args []string) (err error) {
...
 		return err
 	}
 
 	switch o.Format {
 	case "":
-		for i, ref := range res {
-			printDatasetRefInfo(o.Out, i, ref)
+		fmt.Printf("showing %d results for '%s'\n", len(*results), o.Query)
```
b5: After our recent refactoring, all output should be routed to `o.Out`, so this should be:
```golang
printInfo(o.Out, "showing %d results for '%s'", len(results), o.Query)
```
printInfo will add in the newline for you :)

cmd/print.go
```diff
@@ -129,6 +130,31 @@ func printDatasetRefInfo(w io.Writer, i int, ref repo.DatasetRef) {
 	fmt.Fprintln(w)
 }
 
+func printSearchResult(i int, result core.SearchResult) {
```
b5:
```golang
func printSearchResult(i int, result core.SearchResult) {
  white := color.New(color.FgWhite).SprintFunc()
  green := color.New(color.FgGreen).SprintFunc()
  // NOTE: in the future we need to switch based on result.Type
  // For now we are taking a shortcut and assuming a dataset struct
  ds := &registry.Dataset{}
  resultString := result.ID
  if data, err := json.Marshal(ds.Value); err == nil {
    if err = json.Unmarshal(data, ds); err == nil && ds.Meta != nil {
      resultString = fmt.Sprintf("%s\n   %s\n   %s\n", white(result.ID), green(ds.Meta.Title), desc)
    }
  }
  fmt.Printf("%s. %s", white(i+1), resultString)
}
```

PR #461, opened 2018-06-12
fix(CreateDataset): user-specified data should override transforms

### LINKED ISSUE #456: things specified in `dataset.yaml` should override transforms
So we've made transforms, dope. Transforms have sweet convenience methods like `qri.set_meta("title", "this is a title")`, also dope.

But this creates ambiguity. When I set a `meta.title` in both the transform _and_ in `dataset.yaml`, what's supposed to happen? This is effectively the same thing as supplying input data on the frontend via & API call to a dataset that also includes a transform.

I'd like to vote for things specified in dataset.yaml overriding everything. To me this places a greater burden on those writing transformations (who now need to think about weather or not a user has provided field _foo_ already). But at least in that case they conditional statements to work around this stuff.

What's important to me here is to prioritize declarative statements put forth by the end user. If I explicitly add `meta.title: "my dataset"`, I don't think the transform should get to override me. On the other hand, if I provide nothing and the transform auto-sets a title for me, I'm delighted. If this irritates me, I have a way to modify the output without knowing how the transform works.

Holler if you think this is silly

b5:
ugh I screwed up my commits & don't have time to de-couple them. This commit also refactors our tests in the cmd package to make them more readable / sensible.

On the transforms side this uses an `Apply` to reinstate user-supplied data after the transform has run. I'm working on a pseudo-spec for how this should evolve over time, but this is a good start.

closes #456


b5:
This additional force-push is just to replace a hard-coded value in `dataset_test.go` with `skytf.Version`, so that we don't have to worry about constantly updating that value

PR #462, opened 2018-06-13
refactor(dataset.Body): rename dataset.Data to Body

### LINKED ISSUE #422: New Vocabulary for Datasets
There's unavoidable confusion whenever we talk about data (what is "data", what does the word mean, how is it distinct from similar terms) but even still, I believe we can improve some of the terms used in Qri to help tell better stories. One particular example involves Dataset, which is distinct from DatasetPod, which itself has a field called Data. DatasetPod is part of the recent move to naming things "Pod" when they are Plain-Old-Data, using all basic types, that can be trivially serialized. Yet, it and Dataset are still distinct from the field Data, and in fact it's possible to have either Dataset or DatasetPod without any Data, which is strange.

Furthermore, there are other terms like Meta and Structure that both talk about Datasets, but in different ways. It's important to have good explanations for what each of these terms mean relative to one another.

My suggestion is to strongly embrace analogies to html, a well-understood and often talked about technology. The main change is to rename the field Data to Body, matching how an html document may have meta tags, but that the body is the main thing a consumer cares about reading. This leads to the following comparisons in Qri:

- Structure :: XSD (xml schema definition, describes an xml / html file's structure)
- Meta :: meta tags (contains descriptions, keywords, search engine stuff)
- Body :: body tag (the document's real contents)
- Dataset :: webpage

Most importantly, this last point reveals we can talk about Datasets that don't have bodies, perhaps because they haven't been loaded yet, or because we only have a reference to them. This is similar to the HTTP method "head", which retrieves information about a webpage without getting the full body. No more talking nonsensically about "Datasets without Data"; saying instead "Datasets without Bodies" makes much more sense.

b5:
Ok this is going to break lots of stuff, but it'll get the switch to calling data "Body" over with. Things that'll probably break:
* Getting Dataset Data
* The API
* Our brainz

Given that we're in move-fast-and-settle-interfaces mode, I'm not going to bother adding logic to support "dataPath" alternate lookups, so this is going to mean we all need to re-create our datasets if we want them to work properly.

closes #422.


api/datasets.go
```diff
@@ -344,16 +344,16 @@ func (h *DatasetHandlers) initHandler(w http.ResponseWriter, r *http.Request) {
...
+			util.WriteErrResponse(w, http.StatusBadRequest, fmt.Errorf("if adding dataset using json, body of request must have 'bodyPath' field"))
 			return
 		}
 
 	default:
 		dsp = &dataset.DatasetPod{
 			Peername: r.FormValue("peername"),
 			Name:     r.FormValue("name"),
-			DataPath: r.FormValue("data_path"),
+			BodyPath: r.FormValue("data_path"),
```
dustmop: "body_path"? Or does this require electron changes?

api/datasets.go
```diff
@@ -633,7 +633,7 @@ type DataResponse struct {
 	Data json.RawMessage `json:"data"`
 }
 
-func (h DatasetHandlers) dataHandler(w http.ResponseWriter, r *http.Request) {
+func (h DatasetHandlers) bodyHandler(w http.ResponseWriter, r *http.Request) {
 	d, err := DatasetRefFromPath(r.URL.Path[len("/data"):])
```
dustmop: "/body"

cmd/add.go
```diff
@@ -51,7 +51,7 @@ changes to qri.`,
 	}
 
 	cmd.Flags().StringVarP(&o.File, "file", "f", "", "dataset data file in either yaml or json format")
-	cmd.Flags().StringVarP(&o.DataPath, "data", "d", "", "path to file or url to initialize from")
+	cmd.Flags().StringVarP(&o.BodyPath, "body", "b", "", "path to file or url to contents of dataset")
```
dustmop: This is very pedantic of me, but I parse this sentence like "path : file OR url : contents of dataset" when I think it's supposed to be "path : (file or url) : contents of dataset". Maybe change to "path to file or url for contents of dataset".

core/datasets.go
```diff
@@ -409,7 +409,7 @@ func (r *DatasetRequests) Save(p *SaveParams, res *repo.DatasetRef) (err error)
 		return fmt.Errorf("peername & name are required to update dataset")
 	}
 
-	// if dsp.DataPath == "" && dsp.DataBytes == nil && dsp.Transform == nil {
+	// if dsp.BodyPath == "" && dsp.BodyBytes == nil && dsp.Transform == nil {
 	// 	return fmt.Errorf("either dataBytes, dataPath, or a transform is required to create a dataset")
```
dustmop: Are we updating all commented out code? This should be "either bodyBytes, bodyPath, ...

core/datasets.go
```diff
@@ -431,7 +431,7 @@ func (r *DatasetRequests) Save(p *SaveParams, res *repo.DatasetRef) (err error)
 		return fmt.Errorf("error getting previous dataset: %s", err.Error())
 	}
 
-	if dsp.DataBytes != nil || dsp.DataPath != "" {
+	if dsp.BodyBytes != nil || dsp.BodyPath != "" {
 		dataFile, err = repo.DatasetPodDataFile(dsp)
```
dustmop: DatasetPodBodyFile?

core/datasets_test.go
```diff
@@ -69,20 +69,20 @@ func TestDatasetRequestsInit(t *testing.T) {
...
 	}{
 		{nil, nil, "dataset is required"},
-		{&dataset.DatasetPod{}, nil, "either dataBytes, dataPath, or a transform is required to create a dataset"},
-		{&dataset.DatasetPod{DataPath: "/bad/path"}, nil, "opening file: open /bad/path: no such file or directory"},
-		{&dataset.DatasetPod{DataPath: jobsDataPath, Commit: &dataset.CommitPod{Qri: "qri:st"}}, nil, "decoding dataset: invalid commit 'qri' value: qri:st"},
-		{&dataset.DatasetPod{DataPath: "http://localhost:999999/bad/url"}, nil, "fetching data url: Get http://localhost:999999/bad/url: dial tcp: address 999999: invalid port"},
-		{&dataset.DatasetPod{Name: "bad name", DataPath: jobsDataPath}, nil, "invalid name: error: illegal name 'bad name', names must start with a letter and consist of only a-z,0-9, and _. max length 144 characters"},
-		{&dataset.DatasetPod{DataPath: badDataS.URL + "/data.json"}, nil, "determining dataset schema: invalid json data"},
-		{&dataset.DatasetPod{DataPath: "testdata/q_bang.svg"}, nil, "invalid data format: unsupported file type: '.svg'"},
+		{&dataset.DatasetPod{}, nil, "either dataBytes, bodyPath, or a transform is required to create a dataset"},
```
dustmop: is "dataBytes" staying the same?

core/datasets_test.go
```diff
@@ -157,7 +157,7 @@ func TestDatasetRequestsSave(t *testing.T) {
 		return
 	}
 
-	citiesDataPath, err := dstest.DataFilepath("testdata/cities_2")
+	citiesBodyPath, err := dstest.DataFilepath("testdata/cities_2")
```
dustmop: dstest.BodyFilepath

api/datasets.go
```diff
@@ -153,8 +153,8 @@ func (h *DatasetHandlers) RenameHandler(w http.ResponseWriter, r *http.Request)
 	}
 }
 
-// DataHandler gets a dataset's data
-func (h *DatasetHandlers) DataHandler(w http.ResponseWriter, r *http.Request) {
+// BodyHandler gets a dataset's body
```
dustmop: I prefer the phrasing used elsewhere in the file:
// BodyHandler gets the contents of a dataset

or

// BodyHandler gets the contents of a dataset's body

It's less tautological sounding.

api/datasets.go
```diff
@@ -163,7 +163,7 @@ func (h *DatasetHandlers) DataHandler(w http.ResponseWriter, r *http.Request) {
 			readOnlyResponse(w, "/data/")
```
dustmop: "/body/"

PR #465, opened 2018-06-18
feat(use, get): added use and get commands

b5:
Two things showed up while working on tutorials:
1. We haven't built a way to actually _get_ dataset info from the command line. Currently we have `qri info`, which only displays a summary.
2. constantly retyping dataset names has officially gotten too irritating.

### Reference Selection & `qri use`
For the retyping problem, I think we should steal the concept of "USE" from database land to set the current dataset. For example, in postgres's `psql` tool, the command `USE [database_name]` sets the current database you're interacting with. So for us I've first plumbed the concept of a "Reference Selection" as an opt-in interface in the repo package:

```golang
// RefSelector is an interface for supporting reference selection
// a reference selection is a slice of references intended for using
// in dataset operations
type RefSelector interface {
	SetSelectedRefs([]DatasetRef) error
	SelectedRefs() ([]DatasetRef, error)
}
```

It's a simple setter-getter interface. Notice reference selections are a a slice of references. Having this be an opt-in interface means not all interfaces may implement reference selection. This PR adds reference selection for both `MemRepo` and `Fs` repo.

Now that we have a way to store a selection of datasets, I've plumed this up into a new `qri use` command that takes any number of dataset references as arguments (a raw `qri use` with no args clears the selection). Once a user has set a selection with `qri use`, they become the default dataset names when no selection is made. So:
```shell
$ qri use osterbit/list_of_shas
osterbit/list_of_shas

$ qri info
1  osterbit/list_of_shas
    /ipfs/QmNhUwFjFa1j6kKS9LebDPS7KLr6ovhYgRgLEHjfNB9GAX
    113 Shorthand Abstractions/Mental Models
    source: https://www.fs.blog/mental-models

$ qri use

$ qri info
requires at least 1 arg(s), only received 0
```

From the CLI this'll dramatically cut down on typing, and abate the need for tab completions. We can also plumb this up into the API & frontend as needed.

### `qri get`

Ok, so I'm _really_ into `qri config get`. There's something so lovely about being able to hold a mental model of the configuration file in my head and using dot selectors to isolate bits of configuration info. This PR adds a new command `qri get` that does the same thing for datasets:

```bash
$ qri get meta osterbit/list_of_shas
osterbit/list_of_shas@QmRGAvfLoRcx2nTExZLaPajh6KNH8bDSUZvYDVsUVJSxEe/ipfs/QmNhUwFjFa1j6kKS9LebDPS7KLr6ovhYgRgLEHjfNB9GAX:
  citations: []
  contributors: []
  description: "source: https://www.fs.blog/mental-models/\nadded because \n- lists
    of these are sorta interesting\n- example of a simple dataset that could be organized/represented
    a few different different hierarchical formats or as a table\n\ncitations:\nCharlie
    Munger, https://www.fs.blog/mental-models/\n\nNOTE - have noticed some errors
    in the data from the original parsing\nTODO: fix"
  homePath: https://www.fs.blog/mental-models/
  language:
  - english
  qri: md:0
  readmePath: https://www.fs.blog/mental-models/
  theme:
  - special interest
  title: '113 Shorthand Abstractions/Mental Models '
```

Default output is YAML, switchable with the same `--format` flag that `qri config` uses. But the more interesting thing shows up when `qri get` is used on _multiple datasets_:
```shell
$ qri get meta osterbit/census_acs5_2010_tag b5/schools
b5/schools@QmSyDX5LYTiwQi861F5NAwdHrrnd1iRGsoEvCyzQMUyZ4W/ipfs/QmSies9e7YEWUpT27RniNULJ4rxSD5zZd1aAe5d8BBEfji:
  accrualPeriodicity: R/P1W
  downloadPath: https://code.org/schools.json
  qri: md:0
osterbit/census_acs5_2010_tag@QmRGAvfLoRcx2nTExZLaPajh6KNH8bDSUZvYDVsUVJSxEe/ipfs/QmbDbPMbnsYHYtTPXE2Jv1EfwzY3y28hu93By9UiYpikCk:
  accrualPeriodicity: R/P1W
  downloadPath: https://api.census.gov/data/2010/acs5/tags.json
  qri: md:0
```

combined with `qri use` this makes for what I think is a pleasant experience:
```shell
$ qri use osterbit/census_acs5_2010_tag b5/schools
osterbit/census_acs5_2010_tag 
b5/schools

$ qri get meta.title -f json
{
 "b5/schools@QmSyDX5LYTiwQi861F5NAwdHrrnd1iRGsoEvCyzQMUyZ4W/ipfs/QmSies9e7YEWUpT27RniNULJ4rxSD5zZd1aAe5d8BBEfji": "",
 "osterbit/census_acs5_2010_tag@QmRGAvfLoRcx2nTExZLaPajh6KNH8bDSUZvYDVsUVJSxEe/ipfs/QmbDbPMbnsYHYtTPXE2Jv1EfwzY3y28hu93By9UiYpikCk": ""
}
```

To me we clearly need some work on the lengthy name that's output, but this points to the _inherent compatibility_ of qri datasets. I'm very excited to fold some limit/offset tools that apply to everything, but especially `qri body`, where datasets that are meant to kinda "go together" can now be conjoined with `qri use` to return data that combines a number of dataset bodies quickly & easily. I'm also very excited about reduced complexity of a mental model that `qri get` implies.

### Next Steps
* merge the `qri body` command into `qri get`
* Let's consider the format of the top level output. Should it just be an array of arrays? I'm assuming we should add parameters for this
* Let's consider adding some sort of selection syntax that extends this dot notation with pattern matching (let's take our time with this one)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: /go/src/github.com/qri-io/qri
     docker:
-      - image: circleci/golang:1.10.1
+      - image: circleci/golang:1.10.3
         environment:
           GOLANG_ENV: test
           PORT: 3000
--- a/cmd/body.go
+++ b/cmd/body.go
@@ -54,6 +54,7 @@ type DataOptions struct {
 
 	UsingRPC        bool
 	DatasetRequests *lib.DatasetRequests
+	Repo            repo.Repo
 }
 
 // Complete adds any missing configuration that can only be added just before calling Run
@@ -61,6 +62,10 @@ func (o *DataOptions) Complete(f Factory, args []string) (err error) {
 	o.Ref = args[0]
 	o.UsingRPC = f.RPC() != nil
 	o.DatasetRequests, err = f.DatasetRequests()
+	if err != nil {
+		return
+	}
+	o.Repo, err = f.Repo()
 	return
 }
 
@@ -72,6 +77,10 @@ func (o *DataOptions) Run() error {
 
 	dsr, err := repo.ParseDatasetRef(o.Ref)
 	if err != nil {
+		return err
+	}
+
+	if err = lib.DefaultSelectedRef(o.Repo, &dsr); err != nil {
 		return err
 	}
 
--- a/cmd/factory.go
+++ b/cmd/factory.go
@@ -29,6 +29,7 @@ type Factory interface {
 	ProfileRequests() (*lib.ProfileRequests, error)
 	SearchRequests() (*lib.SearchRequests, error)
 	RenderRequests() (*lib.RenderRequests, error)
+	SelectionRequests() (*lib.SelectionRequests, error)
 }
 
 // PathFactory is a function that returns paths to qri & ipfs repos
--- a/cmd/get.go
+++ b/cmd/get.go
@@ -1,0 +1,77 @@
+package cmd
+
+import (
+	"github.com/qri-io/qri/lib"
+	"github.com/qri-io/qri/repo"
+	"github.com/spf13/cobra"
+)
+
+// NewGetCommand creates a new `qri search` command that searches for datasets
+func NewGetCommand(f Factory, ioStreams IOStreams) *cobra.Command {
+	o := &GetOptions{IOStreams: ioStreams}
+	cmd := &cobra.Command{
+		Use:   "get",
+		Short: "get elements of qri datasets",
+		Long:  ``,
+		Annotations: map[string]string{
+			"group": "dataset",
+		},
+		Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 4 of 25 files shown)


ramfox:
I think we need a `qri use list` command that will remind the user the datasets that he or she has set in `qri use`

I'm thinking of the way `git stash` works, you use `git stash list` to list what you have stashed, and `git stash clear` to remove everything from your stash.

My other other thought, is that `qri use` should actually list the datasets that are being referenced, and `qri use clear` clears them.

b5:
ok ok sounds great. @ramfox I've added changes to how use works to work more like the rest of our commands, and totally agreed @osterbit we should get on that aliasing stuff.

PR #473, opened 2018-06-26
fix(cmd): Wrap how DatasetRefs are parsed for better cmd behavior.

dustmop:
This command-line invocation:

  qri add --body=some_rows.json new_dataset

would previously act as follows:

  added new dataset me/some_rowsjson@QmBlahBlah/ipfs/QmBlahBlah

but now acts like:

  added new dataset me/new_dataset@QmBlahBlah/ipfs/QmBlahBlah

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -86,7 +86,7 @@ func parseSecrets(secrets ...string) (map[string]string, error) {
 
 // parseCmdLineDatasetRef parses DatasetRefs, assuming peer "me" if none given.
 func parseCmdLineDatasetRef(ref string) (repo.DatasetRef, error) {
-	if strings.Index(ref, "@") == -1 && strings.Index(ref, "/") == -1 {
+	if !strings.ContainsAny(ref, "@/") {
 		ref = "me/" + ref
 	}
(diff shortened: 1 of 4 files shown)


cmd/cmd.go
```diff
@@ -80,3 +83,11 @@ func parseSecrets(secrets ...string) (map[string]string, error) {
 	}
 	return s, nil
 }
+
+// parseCmdLineDatasetRef parses DatasetRefs, assuming peer "me" if none given.
+func parseCmdLineDatasetRef(ref string) (repo.DatasetRef, error) {
+	if strings.Index(ref, "@") == -1 && strings.Index(ref, "/") == -1 {
```
b5: how about [`strings.ContainsAny`](https://godoc.org/strings#ContainsAny) instead?
```golang
  if strings.ContainsAny(ref, "@/") {
  }
```

b5:
Erm, shoot, I take that back, we should probably make the same update for the `save` and `delete` commands, which are all "local edit commands" that assume you control the dataset in question.

dustmop:
Added this functionality to commands `save` and `remove` as well. Using strings.ContainsAny.

PR #475, opened 2018-06-28
feat(api/registry, api/search, api/version): add api endpoints to get closer to command line functionality

### LINKED ISSUE #474: Bring API functionality closer to the command line
We want the frontend to have the same functionality as the command line, or at least, as close as possible. 
To do that, we first need to update the API!

Here are the commands that don't have API endpoint counterparts:
* config (although we have the `/profile` endpoint that changes the profile config)
* registry
* search
* validate
* version

I propose we start with:
### Registry
We need this so we can publish and unpublish.
`/registry/publish`
`/registry/unpublish`
`/registry/check`
or should it be
`/registry/` and then the payload says whether to publish, unpublish, or check status. This fits closer to our usual api patterns, so going to start with that
### Search
we had search up once, should check commits and see how it operated before
### Version
This is for a little UI detail that rico suggested, when qri loads, it should show what version you are using. Perhaps this should also exist somewhere else on the app, as a reference

ramfox:
close #474


repo/actions/registry.go
```diff
@@ -42,6 +42,21 @@ func (act Registry) Unpublish(ref repo.DatasetRef) (err error) {
 	return cli.DeleteDataset(ref.Peername, ref.Name, ds.Encode(), pub)
 }
 
+// Status checks to see if a dataset is published to a repo's specific registry
+func (act Registry) Status(ref repo.DatasetRef) (err error) {
+	cli, _, _, err := act.dsParams(&ref)
+	if err != nil {
+		return err
+	}
+	if err = act.permission(ref); err != nil {
```
b5: the "permission" check should be dropped. "permission" checks to see if this peer owns the dataset in question, but peers should be able to check the status of datasets they don't own.

repo/actions/registry.go
```diff
@@ -42,6 +42,21 @@ func (act Registry) Unpublish(ref repo.DatasetRef) (err error) {
 	return cli.DeleteDataset(ref.Peername, ref.Name, ds.Encode(), pub)
 }
 
+// Status checks to see if a dataset is published to a repo's specific registry
+func (act Registry) Status(ref repo.DatasetRef) (err error) {
```
b5: I like it. In the near future registries are going to get an upgrade where they can also pin, and we'll have to modify this func to return the status of two things:
* if a given ref is published to the registry (what it's currently doing)
* if that given ref is _pinned_ to the registry (commng soon)

For now this is a great starting point.

PR #476, opened 2018-06-29
feat: Pin on publish

b5:
Over on the registry repo we've added interfaces for registries to accept requests to pin data. Profiles that are connected to the registry can sign a request that the registry copy the data.

This PR adapts `qri registry publish` to also submit a request to pin if a registry is present. The process works like this:
1. User needs a local dataset they've created. They run `qri registry publish me/foo`
2. the user isn't connected to the p2p network, qri connects, and write's down it's local addresses
3. qri adds the full datasetRef of `me/foo` (most importantly, the `/ipfs/QmFooHash` bit), and those local addresses into a call to `regclient.Pin`. Those are sent to the registry server. In a request that'll take a little time...
4. Over on the registry side, an ipfs node is asked to pin the hash `/ipfs/QmFooHash`. To make the process quicker, that ipfs node is told to connect directly to the peer addresses that came with the request.
5. Once the hash is pinned, the request will finish, and the `regclient.Pin` request will return `200`.
6. Data is now pinned in a remote, always on location.

Lots of work to do here, but this is the general gist.

### DIFF
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -105,11 +105,14 @@ func TestCommandsIntegration(t *testing.T) {
 	_, registryServer := regmock.NewMockServer()
 
 	path := filepath.Join(os.TempDir(), "qri_test_commands_integration")
+	fmt.Println(path)
 	t.Logf("test filepath: %s", path)
 
 	//clean up if previous cleanup failed
 	if _, err := os.Stat(path); os.IsNotExist(err) {
-		os.RemoveAll(path)
+		if err := os.RemoveAll(path); err != nil {
+			t.Fatalf("failed to cleanup from previous test execution: %s", err.Error())
+		}
 	}
 	if err := os.MkdirAll(path, os.ModePerm); err != nil {
 		t.Errorf("error creating test path: %s", err.Error())
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -105,7 +105,6 @@ func TestCommandsIntegration(t *testing.T) {
 	_, registryServer := regmock.NewMockServer()
 
 	path := filepath.Join(os.TempDir(), "qri_test_commands_integration")
-	fmt.Println(path)
 	t.Logf("test filepath: %s", path)
 
 	//clean up if previous cleanup failed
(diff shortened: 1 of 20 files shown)


cmd/cmd_test.go
```diff
@@ -105,11 +105,14 @@ func TestCommandsIntegration(t *testing.T) {
 	_, registryServer := regmock.NewMockServer()
 
 	path := filepath.Join(os.TempDir(), "qri_test_commands_integration")
+	fmt.Println(path)
```
ramfox: is this supposed to be here?

b5: ah shoot. yeah should remove. thx!

repo/actions/registry.go
```diff
@@ -42,6 +43,37 @@ func (act Registry) Unpublish(ref repo.DatasetRef) (err error) {
...
+	cli, _, ds, err := act.dsParams(&ref)
+	if err != nil {
+		return err
+	}
+	if err = act.permission(ref); err != nil {
+		return
+	}
+
+	// TODO - better test for this
+	if ds.Structure.Length > 1000000*250 {
```
dustmop: Make a const instead

cmd/qri.go
```diff
@@ -213,7 +213,11 @@ func (o *QriOptions) RegistryRequests() (*lib.RegistryRequests, error) {
 	if err := o.init(); err != nil {
 		return nil, err
 	}
-	return lib.NewRegistryRequests(o.repo, o.rpc), nil
+	req := lib.NewRegistryRequests(o.repo, o.rpc)
+	if o.node != nil {
```
dustmop: Instead of this, copy what is going on with lib/datasets.go, where there's NewDatasetRequestsWithNode, which takes a third parameter for the *QriNode. Doesn't matter if you pass a null pointer, there's the default anyway.

lib/registry.go
```diff
@@ -30,12 +34,65 @@ func NewRegistryRequests(r repo.Repo, cli *rpc.Client) *RegistryRequests {
...
+			// if we don't have an online node, create one and connect
+			node, err = p2p.NewQriNode(r.repo.Repo, func(c *config.P2P) {
+				*c = *Config.P2P
+				c.Enabled = true
+			})
+			if err != nil {
+				return err
+			}
+
+			if err := node.StartOnlineServices(func(string) {}); err != nil {
```
dustmop: Extract this common code and combine with the next condition, change from "else if" to "if".

PR #483, opened 2018-07-13
feat(viz): add viz api endpoint, viz refactor

b5:
okok this makes viz work. it reworks the `lib/render` logic to work in a new order, using the first template from this order:
* use a template provided by params
* load the dataset & check for a render template
* fall back to the default template

I've also added a new `/render/peername/dataset_name` that'll dynamically execute the template & pipe back HTML (note, this is an HTML response, not a JSON response, so frontend stuff can just directly reference the url instead of flowing through an API action)

real on the real this code needs work, but tests are passing and I'd like to get this in ASAP & iterate from there.

bonus points: local tests for api b/c I'm not going to have an internet connection on this plane.


api/testdata/addRequestFromURL.json
```diff
@@ -1,5 +1,5 @@
 {
   "peername":"peer",
   "name":"family_relationships",
-  "bodyPath":"http://insight.dev.schoolwires.com/HelpAssets/C2Assets/C2Files/C2ImportFamRelSample.csv"
+  "bodyPath":"http://127.0.0.1:55555/C2ImportFamRelSample.csv"
```
ramfox: yes this is so much betterrrrrr

api/testdata/statusResponse.json
```diff
@@ -1 +1 @@
-{ "meta": { "code": 200, "status": "ok", "version":"0.5.0" }, "data": [] }
\ No newline at end of file
+{ "meta": { "code": 200, "status": "ok", "version":"0.5.1-dev" }, "data": [] }
```
ramfox: this is sort of a side thing, but can we have a doc or issue somewhere that states where all the different places that tell us where all the "versions" live. For example, I know that for the backend, we have a `VersionNumber` const that lives in lib, but I'm unsure where the same thing exists on the frontend. Is it just the `package.json` file?

I know you've also linked to info that describes the way you want to version things, before, but I'd be nice to have a mini style guide for that. Eg, before it's released we call it `-dev`

Happy to make a paper doc or issue if we think that's the best place to keep that knowledge.

PR #487, opened 2018-07-14
If you clone the repo directly, go install doesn't work; run `go build`

zsck:


b5:
Hey Zack! Thanks for this, but we'd prefer others use the makefile to generate the qri binary, which will (hopefully) work better across platforms like so:
```shell
$ cd $GOPATH/github.com/qri-io/qri
$ make install
```

I'm going to close this in favour of a PR that adds instructions along those lines. But thanks anyways for your contributions!

PR #489, opened 2018-07-18
fix(rpc): Register types with gob to avoid "reading body EOF" errors

dustmop:
By registering interface array and interface map types with gob, we should hopefully avoid all future gob encoding erorrs. Revert commit d9f78058adb which added a work-around for this error; it isn't needed anymore.

### DIFF
--- a/lib/lib.go
+++ b/lib/lib.go
@@ -4,6 +4,7 @@
 package lib
 
 import (
+	"encoding/gob"
 	golog "github.com/ipfs/go-log"
 	"github.com/qri-io/qri/p2p"
 )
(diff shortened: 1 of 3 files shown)


lib/lib.go
```diff
@@ -20,6 +21,11 @@ type Requests interface {
 	CoreRequestsName() string
 }
 
+func init() {
+	gob.Register([]interface{}{})
```
b5: Can we get a quick comment on what's going on here? maybe:
```golang
// fields like dataset.Structure.Schema contain arbitrary data structures,
// registring with the gob package mitigates errors when sending over net/rpc calls
```

PR #499, opened 2018-07-24
test(cmd/factory): added a test for EnvPathFactory

crazcalm:


cmd/factory_test.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func TestEnvPathFactory(t *testing.T) {
+	path := "fake_path"
+	home, err := homedir.Dir()
+	if err != nil {
+		t.Errorf("Failed to find the home directory: %s", err.Error())
+	}
+
+	tests := []struct {
```
b5: It may be worth re-structuring this test to _always_ set `QRI_PATH` and `IPFS_PATH` env variables, and make use of the `path` element of your test struct

```golang
  tests := []struct{
    qriPathEnv string
    ipfsPathEnv string
    pwd string // renamed from "path" to clarify 
    qriAnswer string
    ipfsAnswer string
  }{
    // an example refactoring of your first test case:
    {"", "", "fake_path", filepath.Join(home, ".qri"), filepath.Join(home, ".ipfs")},
  }

  for i, c := rance tests {
    os.SetEnv("QRI_PATH", c.qriPathEnv)
    os.SetEnv("IPFS_PATH", c.ipfsPathEnv)
    if err := os.Chdir(c.pwd); err != nil {
      t.Fatalf("case %d error setting process directory: %s", i, err.Error())
    }
    
    qriResult, ipfsResult := EnvPathFactory()
    // ... rest of test as you have it written
  }
```

this way we can simplify the test logic, generating clearer table tests. 

Also, it would be great to add test cases that _explicitly_ set `QRI_PATH` and `IPFS_PATH` env variables to empty strings will increase test coverage by getting into the `if qriRepoPath == "" { // ...` clauses in the base function.

You can always check the coverage of a test by running:
```shell
$ cd $GOPATH/github.com/qri-io/qri/cmd
$ go test --coverprofile=coverage.txt
$ go tool cover --html=coverage.txt
```

cmd/factory_test.go
```diff
@@ -0,0 +1,58 @@
...
+			err := os.Setenv("IPFS_PATH", test.path)
+			if err != nil {
+				t.Errorf("Failed to set up IPFS_PATH: %s", err.Error())
+			}
+		}
+
+		qriResult, ipfsResult := EnvPathFactory()
+
+		if !strings.EqualFold(qriResult, test.qriAnswer) {
+			t.Errorf("Expected qri path to be %s, but got %s", test.qriAnswer, qriResult)
```
b5: I'm a big fan of noting the case iteration number on these:
```golang
    t.Errorf("case %d Expected qri path to be %s, but got %s", i, test.qriAnswer, qriResult)
```
it makes our tests a little easier to debug 🙂

cmd/factory_test.go
```diff
@@ -0,0 +1,58 @@
...
+		path       string
+		qriAnswer  string
+		ipfsAnswer string
+	}{
+		{false, false, path, filepath.Join(home, ".qri"), filepath.Join(home, ".ipfs")},
+		{true, false, path, path, filepath.Join(home, ".ipfs")},
+		{false, true, path, path, path},
+	}
+
+	//Note: Variables created with os.Setenv exist for the rest of the process
```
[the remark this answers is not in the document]
crazcalm: ## defer
Using defer to clean up after myself makes sense. Will do.

## the note
The defer makes sense, but I think it is unrelated to the note because the note is about the state change.

os.Getenv(key string) returns the value of the key, but, if the key does not exist, it returns an empty string.

On my machine, the environment variables "QRI_PATH" and "IPFS_PATH" do not exist. Once I set them with os.Setenv, they then existed for the rest of the running process. The defer will change the values of these environment variables to their original state, which, in my case, in an empty string, but the created environment variables will still exist, which is a temporary change of state (because they did not exist before this process) that could affect other tests (or test cases).

I can elaborate the note so that it is clearer?

## os.Getwd
I am not sure why you would want to change, or keep track of, the working directory. As far as I can tell, the EnvPathFactory function does not change or modify the working directory. Thus, adding those lines should not affect the test, but I do not understand why it is needed. Can you explain?

cmd/factory_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"testing"
+
+	"github.com/mitchellh/go-homedir"
+)
+
+func TestEnvPathFactory(t *testing.T) {
+	path := "fake_path"
+	home, err := homedir.Dir()
+	if err != nil {
+		t.Errorf("Failed to find the home directory: %s", err.Error())
```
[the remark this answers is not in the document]
crazcalm: Understood. Will do

cmd/factory_test.go
crazcalm: ## My oversight
After reading your comments, I realized that you are your test machines may have QRI_PATH and IPFS_PATH already set as environmental values. My computer does not, thus, that line was always implicitly always tested.

I will change the test so that the environment variables are explicitly set during each test case.

cmd/factory_test.go
crazcalm: Okay. Will do

cmd/factory_test.go
b5: I had made the same implicit oversight as well, nice catch! Having test test cases for each will mean both get worked in all contexts 🙂

cmd/factory_test.go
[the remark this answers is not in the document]
b5: In regard to the `os.Getwd`, oh my I'm sorry. This bit is coming from a long-since removed code path that used to be working-directory dependent. You're absolutely right that we don't need to do anything that changes the working directory for these tests. Apologies for the mixup

crazcalm:
Let me know if I missed anything

crazcalm:
I did not see your reply to the os.Chdir unitl now. I will remove it from the PR.

crazcalm:
@b5 No thanks needed. I told you that I would take a look at the repo (Sorry it took so long...).

PR #501, opened 2018-07-24
feat(profile): Detect peername renames when listing datasets.

dustmop:


repo/ref_test.go
```diff
@@ -461,3 +462,42 @@ func TestCanonicalizeProfile(t *testing.T) {
 		}
 	}
 }
+
+func TestCanonicalizeProfileWithRename(t *testing.T) {
```
b5: 👌

repo/ref.go
```diff
@@ -344,7 +354,7 @@ func CanonicalizeDatasetRef(r Repo, ref *DatasetRef) error {
 
 // CanonicalizeProfile populates dataset DatasetRef ProfileID and Peername properties,
 // changing aliases to known names, and adding ProfileID from a peerstore
-func CanonicalizeProfile(r Repo, ref *DatasetRef) error {
+func CanonicalizeProfile(r Repo, ref *DatasetRef, need *NeedPeernameRenames) error {
```
b5: Adding `*NeedPeernameRenames` feels like overloading the purpose of `CanonicalizeProfile`, but I think we should merge this as-is.

repo/ref.go
```diff
@@ -298,6 +298,16 @@ func isBase58Multihash(hash string) bool {
 	return true
 }
 
+type NeedPeernameRenames struct {
```
b5: Golint is sad

repo/ref.go
```diff
@@ -298,6 +298,16 @@ func isBase58Multihash(hash string) bool {
...
+	Renames map[string]string
+}
+
+func NewNeedPeernameRenames() NeedPeernameRenames {
```
b5: other golint complaint comes from here.

PR #504, opened 2018-07-25
test(repo/events): Added tests for MemEventLog

crazcalm:


b5:
This is so wonderful @crazcalm, I'm all jammed up with other work stuff, but will try to get to a review tomorrow morning. After a quick look this PR already looks _fantastic_, I'd like to take the time to give it proper consideration :)

crazcalm:
@b5 No worries. Take your time.

b5:
Ok, so after review, I think it's worth refactoring the way table tests are composed to make these tests more readable. To me the upside of a table test should be be to make them relatively comparable

Below I've refactored `TestEventsSince`, separating out the two major cases your tests are (very rightfully) checking: an empty log, and a "hydrated" event log. In this refactoring instead of populating a log with a for loop, I'm just manually constructing one for brevity. This also means that we can test events that are technically "invalid" (because they don't 'have things like `peer.ID` specified), but these "invalid" qualities aren't relevant to all tests.

I've also added two new functions `compareEventSlices` and `compareEvents`. These functions are meant to be reusable pieces to get deep comparison of returned events from tests, and return rich errors when there's a mismatch. We can reuse these functions all over out test suite, including in the other tests you've written!

```golang
// compareEventSlices confirms two slices of events are equal
func compareEventSlices(a, b []*Event) error {
	if len(a) != len(b) {
		return fmt.Errorf("length mismatch: %d != %d", len(a), len(b))
	}

	for i, ae := range a {
		be := b[i]
		if err := compareEvents(ae, be); err != nil {
			return fmt.Errorf("event index %d error: %s", i, err.Error())
		}
	}

	return nil
}

// compareEvents checks to see if all elements of an event are the same,
// returning a rich error message on mismatch
func compareEvents(a, b *Event) error {
	if a.Type != b.Type {
		return fmt.Errorf("type mismatch. %s != %s", a.Type, b.Type)
	}
	if !a.Time.Equal(b.Time) {
		return fmt.Errorf("timestamp mismatch. %s != %s", a.Time, b.Time)
	}
	if !a.Ref.Equal(b.Ref) {
		return fmt.Errorf("ref mismatch. %s != %s", a.Ref, b.Ref)
	}
	if a.PeerID != b.PeerID {
		return fmt.Errorf("peerID mismatch: %s != %s", a.PeerID, b.PeerID)
	}
	if !reflect.DeepEqual(a.Params, b.Params) {
		return fmt.Errorf("params mismatch")
	}
	return nil
}

func TestEventsSince(t *testing.T) {

	emptyLog := MemEventLog{}
	evts, err := emptyLog.EventsSince(time.Unix(1, 0))
	if err != nil {
		t.Fatalf("emptyLog unexpected error: %s", err.Error())
	}
	if len(evts) != 0 {
		t.Errorf("emptyLog returned events?")
	}

	// example log to run tests against
	log := MemEventLog{
		&Event{Type: ETDsAdded, Time: time.Unix(5, 0)},
		&Event{Type: ETDsCreated, Time: time.Unix(10, 0)},
		&Event{Type: ETDsDeleted, Time: time.Unix(15, 0)},
	}

	cases := []struct {
		since  time.Time
		expect []*Event
	}{
		{time.Unix(0, 0),
			[]*Event{
				&Event{Type: ETDsDeleted, Time: time.Unix(15, 0)},
				&Event{Type: ETDsCreated, Time: time.Unix(10, 0)},
				&Event{Type: ETDsAdded, Time: time.Unix(5, 0)},
			}},
		{time.Unix(10, 0),
			[]*Event{
				&Event{Type: ETDsDeleted, Time: time.Unix(15, 0)},
			}},
		{time.Unix(20, 0), nil},
	}

	for i, c := range cases {
		got, err := log.EventsSince(c.since)
		if err != nil {
			t.Errorf("case %d unexpected error: %s", i, err.Error())
			continue
		}
		if err := compareEventSlices(got, c.expect); err != nil {
			t.Errorf("case %d unexpected error: %s", i, err.Error())
		}
	}
}
```

In my opinion by separating out the major cases of an empty and a hydrated log, and running tests against the same hydrated log, we get a more readable, maintainable set of tests. 

To get this merged I'd love to see the above refactoring incorporated for any of the tests where this pattern is relevant.

Thanks! 🛴

crazcalm:
Thanks for the feedback and getting back to me so quickly!

## compareEvent
The other tests I wrote are in a different package, so they will not be able to use these helper functions (unless we decide to export them).

At first glance, compareEvent seems like overkill. None of the code being tested modifies the Events. If events cannot be modified, and we are checking a collection of events, then I think it is safe to assume that if we have the right number of events, then we have all the events that we want.

The next question we have to ask is if the collection of Events are in the correct order. Given that the collection is sorted by time, verifying that it is sorted by time should suffice.

Also, I am pretty sure that the compareEvent does not account for two distinct Events with the same data. Given that the Events are pointers, it would probably make more sense to compare the pointer reference. That way we will know that the two events are the same event.

## LogEventDetails
I do not believe compareEvents or campareEvents slice can be used here because the Events are created within the function. As a result, I will not have access to the original Event. Thus, if i wanted to compare the results of LogEventDetails with my own MemEventLog, I would have to create duplicate Events and check if these two different Events match. Though this can be done, I am not sure if this what you want. 

## Closing thoughts
For any function that does not populate its own MemEventLog, I can definitely use the following to make the test more readable:

     MemEventLog{
		&Event{Type: ETDsAdded, Time: time.Unix(5, 0)},
		&Event{Type: ETDsCreated, Time: time.Unix(10, 0)},
		&Event{Type: ETDsDeleted, Time: time.Unix(15, 0)},
	}


In terms of the test cases that check an empty MemEventLog, I might create separate test functions for them because they all have more than one case that should be check. If I were to write out two or three tests for the empty case in the same test function with everything else, readability will then be killed (again).

I could create a test function with the following signature and pass in the wanted MemEventLog:
    func TestEvents(memEventLog MemEventLog, t testing.T)

I have not fully thought through how that would work, but it is an option.

## Time
I am pretty busy today and I have a weekend wedding that I will be leaving for tomorrow. So, I will try to get this done before I leave for the wedding, but there is a slight chance that I will not get to this until Tuesday.

crazcalm:
I redid TestEventsSince and TestEvents with using the compareEventSlices and creating my own MemEventLog structs. Let me know if I missed anything.

b5:
Amaze. Thanks again @crazcalm!

PR #507, opened 2018-07-26
fix(cmd): Split old `add` command into `new` and `add`.

dustmop:


b5:
any idea y tests r fail? I'll have a look at code now.

cmd/add.go
```diff
@@ -1,82 +1,40 @@
...
-
-Once you’ve added data, you can use the export command to pull the data out of 
-qri, change the data outside of qri, and use the save command to record those 
-changes to qri.`,
-		Example: `  add a new dataset named annual_pop:
-  $ qri add --body data.csv me/annual_pop
-
-create a dataset with a dataset data file:
-  $ qri add --file dataset.yaml --body comics.csv me/comic_characters`,
+Add retrieves a dataset owned by another peer and adds it to your repo.`,
```
b5: muuuch cleaner.

cmd/new.go
```diff
@@ -0,0 +1,200 @@
...
+
+
+const providingSecretWarningMessage = `
+Warning: You are providing secrets to a dataset transformation.
+Never provide secrets to a transformation you do not trust.
+continue?`
+
+
+// NewNewCommand creates a new command
+func NewNewCommand(f Factory, ioStreams IOStreams) *cobra.Command {
```
b5: lol `NewNewCommand`

dustmop:
> any idea y tests r fail? I'll have a look at code now.

Ack, something I need to dig into. We're using cobra.Command.ExecuteC, which I believes uses os.Exit(1) when it fails, meaning there's no error message reporting at all. I'll try and replace that before going forward with this.

b5:
That would be delightful. I've come across similar issues that came when we switched away from `panic` to `os.Exit`. You're much better at determining what should happen in this context, so if you need to change either the behavior of qri (bring back panic), _or_ the tests, feel free. I'd like to do whichever feels more "Unix idiomatic"

PR #511, opened 2018-07-31
fix(cmd): Execute commands using RunE instead of Run

dustmop:
In order to better understand why cmd tests fail, use RunE. This returns an err and avoids calling os.Exit, which is bad since it silences test output. Other behavior, like failing when a command is given invalid arguments, stays the same even with this change.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -33,9 +33,15 @@ func Execute() {
 	}
 
 	root := NewQriCommand(EnvPathFactory, os.Stdin, os.Stdout, os.Stderr)
+	// If the subcommand hits an error, don't show usage or the error, since we'll show
+	// the error message below, on our own. Usage is still shown if the subcommand
+	// is missing command-line arguments.
+	root.SilenceUsage = true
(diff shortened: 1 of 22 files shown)


cmd/cmd.go
```diff
@@ -33,9 +33,15 @@ func Execute() {
...
+	// If the subcommand hits an error, don't show usage or the error, since we'll show
+	// the error message below, on our own. Usage is still shown if the subcommand
+	// is missing command-line arguments.
+	root.SilenceUsage = true
+	root.SilenceErrors = true
+	// Execute the subcommand
 	if err := root.Execute(); err != nil {
 		printErr(os.Stdin, err)
-		os.Exit(-1)
+		os.Exit(1)
```
dustmop: Note: The intention here is that, if a subcommand returns an error, bail out with os.Exit, but I don't believe this code was *ever* being run. Each subcommand was itself calling ExitIfErr, which calls os.Exit and kills the process. This entire change, which makes commands actually return an error, brings this code back into line with the apparent intention.

Issue #509, opened 2018-07-30
`qri get` cannot retrieve a peer's dataset, only those in your own repo


dustmop:
While investigating https://github.com/qri-io/qri/issues/508, I hit this problem:

Running `qri get some_peer/dataset` fails to retrieve data, instead displaying this unhelpful error: `error loading dataset: error getting file bytes: error resolving hash: selected encoding not supported`.

The reason is that DatasetRequests.Select calls CanonicalizeDatasetRef, which then calls repo.GetRef, but GetRef only resolves DatasetRef's in your own repo. The err from GetRef is incorrectly ignored, which results in CanonicalizeDatasetRef returning a ref with no path set, which is then passed to repo.Select. Select turns this empty path into the string "/", which passes it to dataset.dsfs.LoadDatasetRefs as "/ipfs/dataset.json", wherein fileBytes tries to read from this path.

The short-term fix here is to have CanonicalizeDatasetRef not ignore the error from GetRef, which shows the more accurate message `repo: not found`. The better solution would be to have DatasetRequests.Select follow up by requesting peers' paths after CanonicalizeDatasetRef fails, so that we can resolve remote references.

This may also be related to https://github.com/qri-io/qri/issues/479.

dustmop:
Unfortunately, the command-line and api are using different lib calls. `qri get ...` on the command-line calls lib/DatasetRequests.Select while the api calls lib/DatasetRequests.Get. A single fix won't solve both this and #479, two fixes are needed.

PR #516, opened 2018-08-10
fix(get): Can get another user's repo, as long as node is connected.

dustmop:
Fixed the command-line version of #509.

### DIFF
--- a/lib/datasets.go
+++ b/lib/datasets.go
@@ -198,11 +198,20 @@ func (r *DatasetRequests) Select(p *SelectParams, res *[]byte) (err error) {
 
 	encode := map[string]interface{}{}
 	for _, ref := range p.Refs {
-		if err = repo.CanonicalizeDatasetRef(r.repo.Repo, &ref); err != nil {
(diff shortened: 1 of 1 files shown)


lib/datasets.go
```diff
@@ -198,11 +198,20 @@ func (r *DatasetRequests) Select(p *SelectParams, res *[]byte) (err error) {
...
 	encode := map[string]interface{}{}
 	for _, ref := range p.Refs {
-		if err = repo.CanonicalizeDatasetRef(r.repo.Repo, &ref); err != nil {
+		err = repo.CanonicalizeDatasetRef(r.repo.Repo, &ref)
+		if err != nil && err != repo.ErrNotFound {
 			log.Debug(err.Error())
 			return err
 		}
-
+		if err == repo.ErrNotFound {
```
b5: we've had an offline conversation about having a better / more explicit error response here, so I'm ok with `repo.ErrNotFound` for now. Let's get that one cleaned up ASAP.

PR #521, opened 2018-08-23
refactor(actions): Move actions to its own top-level dir, out of repo.

dustmop:
Logically, actions is a higher-level concept than repo, and should not be
in the same directory. There are still some parts of repo that depend on
actions, but they are in subdirectories (such as fs and test), that can be
cleaned up later.

Deleted repo/test/test_dataset_actions.go, which was already disabled and
not being run. More tests should be added in the near future.


b5:
I'm going to merge this b/c @dustmop is at the 🏖

PR #523, opened 2018-08-24
feat(cmd/*): expand `qri use` functionality to other commands

### LINKED ISSUE #515: Expand `qri use` to work on all commands that examine a dataset
Right now `qri use` works only with the `qri get` command.

I propose that we expand that usage to work for:
* [x] log
* [x] body
* [x] diff
* [x] export
* [x] info
* [x] render
* [x] validate

I don't think this should extend to the `save` command.

At first glance, it seems that this would take two changes from each command:
1) allow 0 arguments and allow empty Refs to pass to the lib function that underlies each command
2) add the lib.DefaultSelectedRefs() function to each underlying lib command, if it doesn't already exist

ramfox:
closes #515 

Commands now compatible with `qri use`:
* get
* info
* log
* validate
* diff
* body
* export
* render


api/server_test.go
```diff
@@ -121,7 +121,7 @@ func TestServerRoutes(t *testing.T) {
 		{"GET", "/profile", "", "profileResponse.json", 200},
 		{"GET", "/me", "", "profileResponse.json", 200},
 
-		{"POST", "/add", "addRequestFromURL.json", "addResponseFromURL.json", 200},
+		{"POST", "/new", "addRequestFromURL.json", "addResponseFromURL.json", 200},
```
dustmop: Let's also rename the test files to `newRequestFromURL.json` and `newResponseFromURL.json`.

api/server_test.go
```diff
@@ -393,7 +393,7 @@ func testMimeMultipart(t *testing.T, server *httptest.Server, client *http.Clien
 			map[string]string{},
 			map[string]string{},
 		},
-		{"POST", "/add", "testdata/addResponsePrivate.json", 500,
+		{"POST", "/new", "testdata/addResponsePrivate.json", 500,
```
dustmop: Also rename, please.

api/server_test.go
```diff
@@ -405,7 +405,7 @@ func testMimeMultipart(t *testing.T, server *httptest.Server, client *http.Clien
 				"private":  "true",
 			},
 		},
-		{"POST", "/add", "testdata/addResponseFromFile.json", 200,
+		{"POST", "/new", "testdata/addResponseFromFile.json", 200,
```
dustmop: Same.

cmd/info.go
```diff
@@ -90,30 +89,46 @@ func (o *InfoOptions) Run() error {
 		}
 	}
 
+	if len(o.Refs) < 1 {
```
dustmop: Just a small nit, but please make this `== 0` instead of `< 1`, which reads a bit clearer.

cmd/info.go
```diff
@@ -90,30 +89,46 @@ func (o *InfoOptions) Run() error {
...
 			return err
 		}
+	}
 
-		if ref.IsPeerRef() {
-			printWarning(o.Out, "please specify a dataset for peer %s", ref.Peername)
+	return nil
+}
+
+// info pulls out the common info code so it can be re-run easily
```
dustmop: This comment should describe what the function does when executed, not what this pull request is doing to change the code, since the latter won't make sense when reading this in the future. Change it to something like "// info prints terse information about a single dataset"

lib/datasets.go
```diff
@@ -670,12 +670,16 @@ func (r *DatasetRequests) Validate(p *ValidateDatasetParams, errors *[]jsonschem
...
 	}
 
-	if err = DefaultSelectedRef(r.repo, &p.Ref); err != nil {
+	if err = DefaultSelectedRef(r.repo.Repo, &p.Ref); err != nil {
 		return
 	}
 
-	if p.Ref.IsEmpty() && p.Data == nil {
-		return fmt.Errorf("either data or a dataset reference is required")
+	// TODO: restore
```
dustmop: Please expand on this comment. What is different about this commented out code from the code being added? Is it the ability to use a URL? If so, make the comment about that. Alternatively, just delete it instead of disabling it, and add a TODO like "allow validation from a URL"

Also, use tabs instead of spaces.

lib/error.go
```diff
@@ -26,3 +28,6 @@ func NewError(err error, msg string) Error {
 		msg: msg,
 	}
 }
+
+// ErrBadArgs is a custom error for when a user provides bad arguments
```
dustmop: Drop the word "custom", that is self-evident.

Issue #524, opened 2018-08-28
Build error


aborruso:
Hi,
when I run make build I have the below error.

Ho to solve it??

Thank you

```
1/5 install non-gx deps:

go get -v -u github.com/briandowns/spinner github.com/datatogether/api/apiutil github.com/fatih/color github.com/ipfs/go-datastore github.com/olekukonko/tablewriter github.com/qri-io/skytf github.com/qri-io/bleve github.com/qri-io/dataset github.com/qri-io/doggos github.com/qri-io/dsdiff github.com/qri-io/varName github.com/qri-io/registry/regclient github.com/sergi/go-diff/diffmatchpatch github.com/sirupsen/logrus github.com/spf13/cobra github.com/spf13/cobra/doc github.com/theckman/go-flock github.com/ugorji/go/codec github.com/beme/abide github.com/ghodss/yaml
github.com/briandowns/spinner (download)
github.com/fatih/color (download)
github.com/datatogether/api (download)
github.com/ipfs/go-datastore (download)
github.com/google/uuid (download)
github.com/jbenet/goprocess (download)
github.com/olekukonko/tablewriter (download)
github.com/mattn/go-runewidth (download)
github.com/qri-io/skytf (download)
github.com/google/skylark (download)
github.com/qri-io/cafs (download)
github.com/jbenet/go-base58 (download)
github.com/multiformats/go-multihash (download)
github.com/gxed/hashland (download)
github.com/minio/blake2b-simd (download)
github.com/minio/sha256-simd (download)
github.com/mr-tron/base58 (download)
github.com/spaolacci/murmur3 (download)
Fetching https://golang.org/x/crypto/blake2s?go-get=1
Parsing meta tags from https://golang.org/x/crypto/blake2s?go-get=1 (status code 200)
get "golang.org/x/crypto/blake2s": found meta tag main.metaImport{Prefix:"golang.org/x/crypto", VCS:"git", RepoRoot:"https://go.googlesource.com/crypto"} at https://golang.org/x/crypto/blake2s?go-get=1
get "golang.org/x/crypto/blake2s": verifying non-authoritative meta tag
Fetching https://golang.org/x/crypto?go-get=1
Parsing meta tags from https://golang.org/x/crypto?go-get=1 (status code 200)
golang.org/x/crypto (download)
Fetching https://golang.org/x/sys/cpu?go-get=1
Parsing meta tags from https://golang.org/x/sys/cpu?go-get=1 (status code 200)
[...]

PR #526, opened 2018-08-28
To update the go requirement

aborruso:
As in #524


dustmop:
Sorry, I just merged #525 at almost the same time you opened this. Thank you for sending, but it's already fixed.

PR #528, opened 2018-08-28
fix(cmd): Set number of open files limit at startup.

dustmop:
If the user is running qri without having a sufficiently large enough value for the "open files limit", their IPFS instance can start to fill with the error "too many open files". Setting this value at startup avoids this issue.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -14,7 +14,6 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
-	"syscall"
 
 	golog "github.com/ipfs/go-log"
 	"github.com/qri-io/qri/lib"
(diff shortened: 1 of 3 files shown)


cmd/cmd.go
```diff
@@ -122,3 +125,30 @@ func currentPath() (string, bool) {
 	}
 	return path.Dir(filename), true
 }
+
+const PreferredNumOpenFiles = 10000
```
b5: * Need a comment here to make golint happy.
* Have you experimented with this number at all? Any chance we can get away with, say 5000?

cmd/cmd.go
```diff
@@ -122,3 +125,30 @@ func currentPath() (string, bool) {
...
+
+// ensureLargeNumOpenFiles ensures that user can have a large number of open files
+func ensureLargeNumOpenFiles() {
+	// Get the number of open files currently allowed.
+	var rLimit syscall.Rlimit
```
b5: I'm delighted to see this is all

cmd/cmd.go
b5: Erm, seems like others use the 10000 figure, which may be arbitrary, but we don't have time to tinker with _everything_ 😄, maybe we can revisit this const later on

cmd/system.go
```diff
@@ -0,0 +1,39 @@
+// +build !windows
```
b5: OoooOOOoooOh 👍

PR #534, opened 2018-08-31
test(api): overhaul API tests

b5:
our API tests are brittle & coupled up in a few mega-functions. Time to break those apart &
provide an easy update mecahnism for when breaking API changes are intentional.
I've *heavily* refactored the api tests, working individual handlers & comparing responses
to abide records instead of json files.


api/api_test.go
```diff
@@ -0,0 +1,446 @@
...
+	// we use the dsfs.Timestamp func variable to override
+	// the actual time
+	prevTs := dsfs.Timestamp
+	dsfs.Timestamp = func() time.Time { return time.Date(2001, 01, 01, 01, 01, 01, 01, time.UTC) }
+
+	if r, err = test.NewTestRepo(rc); err != nil {
+		t.Fatalf("error allocating test repo: %s", err.Error())
+	}
+
+	lib.Config = config.DefaultConfig()
```
dustmop: Please use DefaultConfigForTesting from config/test_config.go : https://github.com/qri-io/qri/blob/31be72d4e30742ac7ea731600f66f769ead3afb1/config/test_config.go#L8

This avoids generating private keys, using pregenerated keys instead.

api/api_test.go
```diff
@@ -0,0 +1,446 @@
...
+
+	client := &http.Client{}
+
+	r, err := test.NewTestRepo(nil)
+	if err != nil {
+		t.Errorf("error allocating test repo: %s", err.Error())
+		return
+	}
+
+	s, err := New(r, func(opt *config.Config) {
```
dustmop: This will conflict with #536, btw.

api/peers_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/qri-io/qri/config"
+	"github.com/qri-io/qri/p2p"
+)
+
+func TestPeerHandlers(t *testing.T) {
+	r, teardown := newTestRepo(t)
+	defer teardown()
+
+	node, err := p2p.NewQriNode(r, func(o *config.P2P) {
```
dustmop: Another conflict

api/api_test.go
```diff
@@ -0,0 +1,446 @@
...
+	defer teardown()
+
+	h := NewRootHandler(NewDatasetHandlers(r, false), NewPeerHandlers(r, nil, false))
+	rootCases := []handlerTestCase{
+		{"OPTIONS", "/", nil},
+		{"GET", "/", nil},
+	}
+	runHandlerTestCases(t, "root", h.Handler, rootCases)
+
+	healthCheckCases := []handlerTestCase{
```
dustmop: These two arrays are the same, will they always be the same? How about just using the same one twice.

api/render_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"testing"
+)
+
+func TestRenderHandler(t *testing.T) {
+	r, teardown := newTestRepo(t)
+	defer teardown()
+
+	cases := []handlerTestCase{
```
dustmop: Please add a comment why this is disabled.

api/api_test.go
b5: They should diverge over time, just wanted to get them in place as such for basic coverage.

b5:
LGTM emoji-acronym courtesy of @ramfox

PR #536, opened 2018-09-04
fix(p2p): NewQriNode parameter cleanup, and factory for TestableQriNode.

dustmop:
First, NewQriNode does not any longer generate its own private keys and PeerID. The default behavior for NewQriNode used to be it would generate private keys and PeerIDs, which is very expensive and hard to test. This was a consequence of using the "options" pattern, but nothing in the code base was really taking advantage of this convention. Instead, just take a config.P2P struct, constructed however the caller needs it to be.

Now, the only place that generates p2p keys should be the command `setup`, which explicitly calls GeneratePrivateKeyAndPeerID. There may still be callers to the old DefaultConfig in the codebase, they should be removed in the future and DefaultConfigWithoutKeys should be renamed to DefaultConfig (same with DefaultProfileWithoutKeys and DefaultP2PWithoutKeys).

The constructor to api/ also went through a similar transformation, because it was also taking an "options" list, and the default behavior was to generate keys.

Finally, p2ptest now has a TestNodeFactory that is used everywhere, removing some ugly hacks and generally unifying the usage of TestableNodes in test, though it is a bit more verbose now.

This change speeds up api/ and cmd/ tests considerably.

Before

```
ok  	github.com/qri-io/qri/api	11.543s
ok  	github.com/qri-io/qri/cmd	55.545s
ok  	github.com/qri-io/qri/config	1.306s
ok  	github.com/qri-io/qri/lib	10.766s
ok  	github.com/qri-io/qri/p2p	2.660s
```

After

```
ok  	github.com/qri-io/qri/api	6.970s
ok  	github.com/qri-io/qri/cmd	21.122s
ok  	github.com/qri-io/qri/config	2.384s
ok  	github.com/qri-io/qri/lib	11.376s
ok  	github.com/qri-io/qri/p2p	2.109s
```


api/server.go
```diff
@@ -33,23 +33,15 @@ type Server struct {
...
-// calling New() with no options will return the default configuration
-// as specified in DefaultConfig
-func New(r repo.Repo, options ...func(*config.Config)) (s *Server, err error) {
-	cfg := config.DefaultConfig()
-	for _, opt := range options {
-		opt(cfg)
-	}
-
+// New creates a new qri server from a configuration
+func New(r repo.Repo, cfg *config.Config) (s *Server, err error) {
```
b5: Yeah I think this is the way to go. I was hoping in the long run this options pattern would help make configuration management easier, but this is just much cleaner.

lib/lib_test.go
```diff
@@ -4,36 +4,28 @@ import (
...
 
+	node := n.(*p2p.QriNode)
 	reqs := Receivers(node)
 	if len(reqs) != 8 {
 		t.Errorf("unexpected number of receivers returned. expected: %d. got: %d\nhave you added/removed a receiver?", 8, len(reqs))
 		return
 	}
 }
-
-func testQriNode(cfgs ...func(c *config.P2P)) (*p2p.QriNode, error) {
```
b5: Nice to see this go

PR #552, opened 2018-09-24
feat(dry run): add dry-run flag to new & save dataset

b5:
When working on a dataset, it's nice to be able to see what qri will output without actually committing the results. For this I've added a `dry-run` flag on both the API & command line


b5:
Codecov is dropped a bit, but considering this fixes currently-broken tests on master, we should merge if it's good

PR #554, opened 2018-10-01
feat(ConvertBodyFile): new action to convert body data

b5:
Ok so this isn't much of a feature, but people like the word "feat" in change logs, so whatever.

*fixes*:
* API adjustment so that `POST`ing to `/new` and `/save` with form data behave the same way
* fixes an error returning body data that isn't JSON from POST to `/new` and `/save`

*refactors*:
* updates to new skytf.ExecFile signature, removes a reduntant read through body data in the process

ugh I need to get better at splittin' up my commits.

### DIFF
--- a/actions/transform.go
+++ b/actions/transform.go
@@ -51,7 +51,5 @@ func ExecTransform(node *p2p.QriNode, ds *dataset.Dataset, infile cafs.File, sec
 		return
 	}
 
-	// ds.Structure = st
-	// return cafs.NewMemfileBytes(fmt.Sprintf("data.%s", st.Format.String()), buf.Bytes()), nil
 	return
 }
--- a/actions/transform.go
+++ b/actions/transform.go
@@ -1,12 +1,10 @@
 package actions
 
 import (
-	"fmt"
 	"os"
 
 	"github.com/qri-io/cafs"
 	"github.com/qri-io/dataset"
-	"github.com/qri-io/dataset/dsio"
 	"github.com/qri-io/qri/p2p"
 	"github.com/qri-io/qri/repo"
 	"github.com/qri-io/skytf"
@@ -15,7 +13,7 @@ import (
 // ExecTransform executes a designated transformation
 func ExecTransform(node *p2p.QriNode, ds *dataset.Dataset, infile cafs.File, secrets map[string]string) (file cafs.File, err error) {
 	filepath := ds.Transform.ScriptPath
-	rr, err := skytf.ExecFile(ds, filepath, infile, skytf.AddQriNodeOpt(node), func(o *skytf.ExecOpts) {
+	file, err = skytf.ExecFile(ds, filepath, infile, skytf.AddQriNodeOpt(node), func(o *skytf.ExecOpts) {
 		if secrets != nil {
 			// convert to map[string]interface{}, which the lower-level skytf supports
 			// until we're sure map[string]string is going to work in the majority of use cases
@@ -28,25 +26,6 @@ func ExecTransform(node *p2p.QriNode, ds *dataset.Dataset, infile cafs.File, sec
 	})
 	if err != nil {
 		return nil, err
-	}
-
-	st := &dataset.Structure{
-		Format: dataset.JSONDataFormat,
-		Schema: ds.Structure.Schema,
-	}
-
-	buf, err := dsio.NewEntryBuffer(st)
-	if err != nil {
-		return nil, fmt.Errorf("error allocating result buffer: %s", err)
-	}
-
-	err = dsio.Copy(rr, buf)
-	if err != nil {
-		return nil, err
-	}
-
-	if err := buf.Close(); err != nil {
-		return nil, fmt.Errorf("error closing row buffer: %s", err.Error())
 	}
 
 	// TODO - adding here just to get the content-addressed script path for the event.
@@ -72,6 +51,7 @@ func ExecTransform(node *p2p.QriNode, ds *dataset.Dataset, infile cafs.File, sec
 		return
 	}
 
-	ds.Structure = st
-	return cafs.NewMemfileBytes(fmt.Sprintf("data.%s", st.Format.String()), buf.Bytes()), nil
+	// ds.Structure = st
+	// return cafs.NewMemfileBytes(fmt.Sprintf("data.%s", st.Format.String()), buf.Bytes()), nil
+	return
 }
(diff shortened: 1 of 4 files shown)


api/datasets.go
```diff
@@ -497,6 +410,141 @@ func (h *DatasetHandlers) initHandler(w http.ResponseWriter, r *http.Request) {
 	json.NewEncoder(w).Encode(env)
 }
 
+func formFileDataset(dsp *dataset.DatasetPod, r *http.Request) (cleanup func(), err error) {
+	var rmFiles []*os.File
+	cleanup = func() {
+		for _, f := range rmFiles {
+			// TODO - log error?
+			os.Remove(f.Name())
```
dustmop: This line is kind of terrifying. Manual inspection convinces me that it's safe, but I can imagine a scenario in the future where that's not the case. It appears that, currently, all elements added to rmFiles are in the temporary directory, which means their paths should start with the value returned by "TempDir". How about, outside this loop, get the value of TempDir, then only remove files that match this prefix.

api/datasets.go
```diff
@@ -497,6 +410,141 @@ func (h *DatasetHandlers) initHandler(w http.ResponseWriter, r *http.Request) {
...
+	tfFile, _, err := r.FormFile("transform")
+	if err != nil && err != http.ErrMissingFile {
+		err = fmt.Errorf("error opening transform file: %s", err)
+		return
+	}
+	if tfFile != nil {
+		// TODO - this assumes a skylark / starlark transform file
+		f, e := ioutil.TempFile("", "transform")
+		if e != nil {
+			err = e
```
dustmop: Instead of needing this assignment, just call the error variable "err" 2 lines up.

api/datasets.go
```diff
@@ -497,6 +410,141 @@ func (h *DatasetHandlers) initHandler(w http.ResponseWriter, r *http.Request) {
...
+	vizFile, _, err := r.FormFile("viz")
+	if err != nil && err != http.ErrMissingFile {
+		err = fmt.Errorf("error opening viz file: %s", err)
+		return
+	}
+	if vizFile != nil {
+		// TODO - this assumes an html viz file
+		f, e := ioutil.TempFile("", "viz")
+		if e != nil {
+			err = e
```
dustmop: Same thing here.

api/datasets.go
```diff
@@ -497,6 +410,141 @@ func (h *DatasetHandlers) initHandler(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func addBodyFile(res *repo.DatasetRef) error {
+	if res.Dataset.Structure.Format == dataset.JSONDataFormat.String() {
+		// TODO - this'll only work for JSON responses
+		data, err := ioutil.ReadAll(res.Dataset.Body.(io.Reader))
+		if err != nil {
+			return err
+		}
+		res.Dataset.Body = json.RawMessage(data)
```
dustmop: return here, get rid of the `else`, indent by one level less

api/datasets.go
```diff
@@ -497,6 +410,141 @@ func (h *DatasetHandlers) initHandler(w http.ResponseWriter, r *http.Request) {
...
+				Schema: in.Schema,
+			})
+
+			data, err := actions.ConvertBodyFile(file, in, st, 0, 0, true)
+			if err != nil {
+				return fmt.Errorf("converting body file to JSON: %s", err)
+			}
+			res.Dataset.Body = json.RawMessage(data)
+		} else {
+			log.Error("response body isn't a cafs.File")
```
dustmop: This error handling is too far from its cause. Move it up to where the conversion happens. Put the success case underneath it.

api/datasets.go
```diff
@@ -497,6 +410,141 @@ func (h *DatasetHandlers) initHandler(w http.ResponseWriter, r *http.Request) {
...
+		rmFiles = append(rmFiles, f)
+		io.Copy(f, bodyfile)
+		f.Close()
+		dsp.BodyPath = path
+	}
+
+	return
+}
+
+func addBodyFile(res *repo.DatasetRef) error {
```
dustmop: I don't understand what this function does. It looks like it reads the body to a string, then wraps it as a "raw" message, but I don't understand how that is different from being a no-op. Could you add a comment?

api/datasets.go
```diff
@@ -497,6 +410,141 @@ func (h *DatasetHandlers) initHandler(w http.ResponseWriter, r *http.Request) {
...
+		}
+	}
+
+	tfFile, _, err := r.FormFile("transform")
+	if err != nil && err != http.ErrMissingFile {
+		err = fmt.Errorf("error opening transform file: %s", err)
+		return
+	}
+	if tfFile != nil {
+		// TODO - this assumes a skylark / starlark transform file
```
dustmop: Does this warrant a "TODO", or just a regular comment? In other words, is there actually any follow-up work related to the fact that it assumes a transform file?

api/datasets.go
```diff
@@ -497,6 +410,141 @@ func (h *DatasetHandlers) initHandler(w http.ResponseWriter, r *http.Request) {
...
+		f.Close()
+		dsp.BodyPath = path
+	}
+
+	return
+}
+
+func addBodyFile(res *repo.DatasetRef) error {
+	if res.Dataset.Structure.Format == dataset.JSONDataFormat.String() {
+		// TODO - this'll only work for JSON responses
```
dustmop: This comment is just restating what the if conditional is checking. Either rephrase or remove.

actions/transform.go
```diff
@@ -72,6 +51,7 @@ func ExecTransform(node *p2p.QriNode, ds *dataset.Dataset, infile cafs.File, sec
 		return
 	}
 
-	ds.Structure = st
-	return cafs.NewMemfileBytes(fmt.Sprintf("data.%s", st.Format.String()), buf.Bytes()), nil
+	// ds.Structure = st
```
dustmop: Remove these lines instead of commenting them out.

api/datasets.go
```diff
@@ -420,7 +420,10 @@ func formFileDataset(dsp *dataset.DatasetPod, r *http.Request) (cleanup func(),
 	}
 
 	datafile, dataHeader, err := r.FormFile("file")
-	if err != nil && err != http.ErrMissingFile {
```
dustmop: I think I prefer the old style. Making `err` into mutable state feels weird.

api/datasets.go
b5: Oh man you're so right. This is like, way under-considered. We're on a mad time crunch so I'm going to slap a huge TODO on this & we should fix this in a patch by Friday. Thankfully this won't to anything bad in it's current incarnation, but totally agreed on the dangers of this kind of code down the road #rememberUltima

api/datasets.go
b5: Yes I think the assmuption warrents a TODO. in the future we'll come up with a fix that either requires or infers based on things like file extension. At that point we should update this code.

PR #556, opened 2018-10-01
refactor(crypto): Extract key generation into repo/gen. Speeds up tests.

dustmop:
The only place that generates private keys and peerIDs is now in repo/gen. The new struct contained within, `RealCryptoSource`, is now passed in from a high-level, close to the executable's entry point, which makes testing easier. Delete old constructors like DefaultConfig and DefaultP2P which used to generate private keys when they really shouldn't have.

Speeds up unit tests from approximately 60 seconds to 40 seconds, about 33% faster. Basically eliminates the runtime of config/ and repo/fs/, and reduces lib/ from 14 to 4 seconds. Only cmd/ is still slow, and it's because it creates new ipfs repositories; a TODO has been left about this identifying a possible fix.

Before this change:
```
ok    github.com/qri-io/qri/actions       6.525s
ok    github.com/qri-io/qri/api           1.605s
ok    github.com/qri-io/qri/cmd          22.111s
ok    github.com/qri-io/qri/config        5.250s
ok    github.com/qri-io/qri/lib          14.455s
ok    github.com/qri-io/qri/p2p           3.814s
ok    github.com/qri-io/qri/repo          0.202s
ok    github.com/qri-io/qri/repo/fs       6.910s
ok    github.com/qri-io/qri/repo/profile  0.142s
ok    github.com/qri-io/qri/repo/test     0.315s
TOTAL:                                   61.329
```

After this change:
```
ok    github.com/qri-io/qri/actions       5.592s
ok    github.com/qri-io/qri/api           2.379s
ok    github.com/qri-io/qri/cmd          23.092s
ok    github.com/qri-io/qri/config        0.336s
ok    github.com/qri-io/qri/lib           3.967s
ok    github.com/qri-io/qri/p2p           3.686s
ok    github.com/qri-io/qri/repo          0.237s
ok    github.com/qri-io/qri/repo/fs       0.139s
ok    github.com/qri-io/qri/repo/profile  0.154s
ok    github.com/qri-io/qri/repo/test     0.309s
TOTAL:                                   39.891
```


b5:
_daaaamn_ pumped to read this.

repo/gen/crypto_generator.go
```diff
@@ -0,0 +1,53 @@
...
+
+	return privKey, peerID
+}
+
+// GenerateNickname returns a nickname using a peerID as a seed
+func (g *RealCryptoSource) GenerateNickname(peerID string) string {
+	return doggos.DoggoNick(peerID)
+}
+
+// TODO: Add a method that wraps the behavior of cafs/ipfs/init, which also performs
```
b5: we might also want consider implementing a mock ipfs/cafs repo as it's own package instead of placing that code here, or maybe proxy such a package here.

repo/gen/crypto_generator.go
```diff
@@ -0,0 +1,53 @@
...
+	"encoding/base64"
+
+	"github.com/libp2p/go-libp2p-crypto"
+	"github.com/libp2p/go-libp2p-peer"
+	"github.com/qri-io/doggos"
+)
+
+// CryptoGenerator is an interface for generating cryptographic info like private keys and peerIDs
+type CryptoGenerator interface {
+	GeneratePrivateKeyAndPeerID() (string, string)
```
b5: Can we get a comment here that indicates the returned private key value must be a base64-encoded private key string? In the future I'd love a test that checks a given implementation for such a thing, but we don't have time for such a thing atm

dustmop:
All comments addresses, pushing once tests complete...

PR #562, opened 2018-10-04
feat(RequestDatasetLog): add back functionality for getting a peer's …

### LINKED ISSUE #561: Get peer's commit history
Right now, we have the p2p DatasetLog request and handle functions commented out.

Let's add this functionality back in, modeling the request and response messages off the `p2p.RequestDataset` and `p2p.handleDataset` functions.

ramfox:
…dataset log over p2p

Modelled after the `RequestDataset` and `handleDataset` functions.

There is some repeat code used between `actions.log` and `handleDatasetLog`, this spawned a discussion about needing another layer of functions, perhaps called `core` that would be shared between actions and p2p. Just noting this here for posterity.

Commit also adds back clearing the "Backoff" indicator when explicitly connecting to another peer.

Closes #561


p2p/log.go
```diff
@@ -9,100 +14,131 @@ const (
 	MtDatasetLog = MsgType("dataset_log")
 )
 
+// DatasetLogRequestBody encapsulates options for requesting dataset history
```
b5: Agree we need this struct, name feels a little verbose. Maybe just `DatasetLogRequest`?

p2p/log.go
```diff
@@ -9,100 +14,131 @@ const (
...
+type DatasetLogRequestBody struct {
+	Ref    repo.DatasetRef
+	Limit  int
+	Offset int
+}
+
+// DatasetLogResponseBody encapsulates option for responding to a dataset history request
```
b5: Ditto here, maybe we consider renaming to `DatasetLogResponse`

p2p/log.go
```diff
@@ -9,100 +14,131 @@ const (
...
-	resref := []repo.DatasetRef{}
-	// err = json.Unmarshal(data, &resref)
-	// if len(resref) == 0 && err != nil {
-	// 	err = fmt.Errorf("no log found")
-	// }
-
-	return resref, nil
+
+	// get a list of peers to whom we will send the request
+	pids := n.ClosestConnectedPeers(ref.ProfileID, 15)
```
dustmop: Please create a constant and use that instead of `15`.

p2p/log.go
```diff
@@ -9,100 +14,131 @@ const (
...
+	if len(pids) == 0 {
+		return nil, fmt.Errorf("no connected peers")
+	}
+
+	// create a channel on which to send the requests and receive the responses
```
dustmop: The comment on this line, and the 3 comments that follow don't really add anything, they're just restating what the code already says in a redundant manner. Please remove.

p2p/log.go
```diff
@@ -9,100 +14,131 @@ const (
...
+				dataset, err := dsfs.LoadDataset(n.Repo.Store(), datastore.NewKey(ref.Path))
+				if err != nil {
+					reqBody.Err = err
+					break
+				}
+				ref.Dataset = dataset.Encode()
+
+				offset--
+				if offset > 0 {
+					continue
```
dustmop: Is this supposed to be setting ref.Path to ref.Dataset.PreviousPath? Otherwise a non-zero offset will just keep loading the same ref.Path over and over.

p2p/log.go
```diff
@@ -9,100 +14,131 @@ const (
...
+
+		if err = n.SendMessage(req, replies, pid); err != nil {
+			log.Debugf("%s err: %s", pid, err.Error())
+			continue
+		}
+
+		res := <-replies
+		dlrb := DatasetLogResponseBody{}
+		if err = json.Unmarshal(res.Body, &dlrb); err == nil {
+			if dlrb.Err != nil {
```
dustmop: Here I think a comment would be helpful, as I'm a little lost on the semantics. If a peer doesn't own the dataset, it'll return with an empty history list (len == 0), but not an error, right? And we expect only a single peer to ever return a non-empty history list, which is why we return immediately upon finding one, correct?

p2p/log.go
```diff
@@ -9,100 +14,131 @@ const (
...
+			return
+		}
+
+		ref := resBody.Ref
+		limit := resBody.Limit
+		offset := resBody.Offset
+
+		local := true
+
+		err := n.ResolveDatasetRef(&ref)
```
dustmop: This seems incorrect, but maybe I'm misunderstanding. There's two functions, ResolveDatasetRef in actions/dataset_ref.go which returns (local bool, error), and then there's ResolveDataRef in p2p/resolve_ref.go which returns (error). This is calling the latter, which goes out and contacts peers in order to resolve the ref, meaning it's return value does *not* tell you if the dataset is local. If that's the intention, this should be using actions/dataset_ref.go's function.

p2p/log.go
```diff
@@ -9,100 +14,131 @@ const (
...
+		if err != nil {
+			local = false
+		}
+
+		reqBody := DatasetLogResponseBody{}
+		history := []repo.DatasetRef{}
+
+		if local {
```
dustmop: In the non-local case we just want to send a response with an empty history list. I would do that here, removing the `local` variable and return early, getting rid of the next level of indention. It might help to refactor the calls below the `for` loop (`req.UpdateJSON(reqBody)`, `res.WithHeaders("phase", "response")`, `ws.sendMessage(res)`) into their own function.

PR #568, opened 2018-10-08
fix(export): Blank yaml file should have correct fields

### LINKED ISSUE #567: qri export --blank produces invalid yaml
The yaml produced by `qri export --blank` has some issues:

* `bodyPath` is wrong, it should be `bodypath`
* `tags` is wrong, the closest I could find in Meta is `keywords`

However, due to https://github.com/qri-io/dataset/issues/149, the Yaml parser ignores these fields entirely. This could lead to datasets that silently drop their tags, and also errors also body not being provided.

dustmop:
Fixes #567, as long as my assumption that `tags` should be `keywords` is correct.


b5:
run `api` tests like this to fix the CI error:
`go test -v -- -u`

PR #572, opened 2018-10-09
feat(export): Always export by zip for now, until full RFC is implemented

dustmop:


dustmop:
Unit test is failing because this PR is dependent upon https://github.com/qri-io/dataset/pull/151

Issue #510, opened 2018-07-31
`qri peers connect` doesn't work across separate networks


dustmop:
So @osterbit and I both run `qri connect`. Then, he sends me his address in the proper format: /ip4/<ip>/tcp/<port>/ipfs/Qm<hash>. When I run `qri peers connect <address>`, I can see in my connected terminal the message: `12:31:39.962 DEBUG     qrip2p: connect to peer: {  <peer.ID > /ip4/<ip>/tcp/<port>/ipfs/Qm<hash>} peers.go:170`, however the command never completes, it just hangs forever.

It occurs to me that we're using separate networks; I'm on WiFiA, he's on WiFiB. Once I get onto WiFiB, running `qri peers connect <addres>` completes quickly.

1) If `connect` cannot make a connection, it should fail fast instead of hanging, or at least use a reasonable timeout, like a few seconds.
2) Using `qri peers connect` should be doable across different networks. Unclear if this currently requires the same LAN, or if there's just NAT / firewall issues we need to figure out.

b5:
A few potential solutions:
`qri peers connect [osterbit's qri profile ID]`
`qri peers connect [osterbit's IPFS peerID]`

One of the issues is a multiaddr is as specific as possible, which may not survive NAT traversal. Using the more general profileID / IPFS peer ID is _supposed_ to fall back to dialing _multiple_ known/suspected multiaddrs to establish a connection.

b5:
Adding a docs label, we should at bare minimum provide greater detail on how `qri peers connect` works, and notify users of the differences between each technique:
* using a qri profile ID is the most general, can work even when not running qri on IPFS
* IPFS ID's are the most _practical_. IPFS has tricks for finding peers that don't exist elsewhere
* multiaddrs are _deterministic_. it either dials or it doesn't.

We should also provide baseline definitions of what a profile.ID, ipfs peer ID, and multiaddr is, and how to find each of those things. `qri connect` prints most of this info, just in obscure ways.

b5:
Ok, update on this:
**tl;dr; NAT Travsersal is really, really hard. This is going to be a long standing issue for some time**

For a long time I've implicitly assumed that all nodes on a peer to peer network are dialable, which isn't really true at the moment (at least, not in practice). There are, however, some very exciting developments in libp2p land that make inroads on this problem, [autorelay](https://github.com/libp2p/go-libp2p/pull/454) being one of the biggies.

[...]

PR #573, opened 2018-10-10
fix(peer sharing): fix peers not sharing info with each other

b5:
Trying to make inroads on #510, I've re-worked the way peers share info about each other. Instead of asking for profile information, a new request method "qri_peers" asks for just the list of peer & profile id's a connected peer is connected to. From there the local peer examines the list, and attempts to connect to any addresses it hasn't seen before. This process repeats until no new peers are found.

Lots of work left to do here, but this should be a massive improvement, and may address #510 for the moment


dustmop:
I don't know enough about peer discovery, but at a cursory glance this looks good. I'm a big fan of `asQriNodes`, there's definitely a lot of test functions that could make use of it in a future cleanup.

PR #580, opened 2018-10-20
feat(p2p.Reputation): sew reputation checks in, adjust `HandlePeerFound`

### LINKED ISSUE #578: feat(p2p): check for a peer's reputation before returning any data
qri-io/registry#15 and qri-io/registry_redis#16 must be merged before the ci will allow this issue to go through

## summary:

Before you exchange data with a peer, check their reputation by pinging the registry. If they have a negative reputation, close the connection.

In order to achieve this, the p2p layer will need to know the where to call to the registry, which means the registry config will also need to be surfaced in the QriNode.

#### add registryConfig to QriNode

I made the choice to add `registryConfig *config.Registry` to the QriNode struct, and I __did not__ change the `cfg` name (which houses the `*config.P2P` struct) to `p2pConfig`. If we think I should, I am happy to change it. 

let's also add a helper function for the RegistryLocation, that returns the registry location uri:
```
func (n *QriNode) RegistryLocation() string
```

`NewQriNode` should be modified to take a `*config.Config`, rather than `*config.P2P`. Ensure that anything that uses `NewQriNode` is also properly modified.

#### adding reputation to our processes:

After doing a deep dive on p2p, we learned a few things.

First, the ConnManager prioritizes connections based on the Tag values. We knew this, and this is why we tag connections with Qri peers higher than normal connections. However, and this makes so much sense in retrospect, that when a connection is closed we loose the ConnManager's tag for that connection/peer. 

So if we ever check the reputation of a peer, tag the peer as a bad actor, and then close the connection, we dont have any persistant knowledge of that we should not communicate with this peer. This means any time we create a new connection with a peer, we have to call to the registry server, to find the reputation, even if we have already done so and closed the connection previously.

Instead, let's store the reputation of a peer in the Peerstore, which allows peer related storage based on key value pairs (`peerstore.Put(peerid, keyString, value )`, and won't remove a peer from a store when a connection is closed.

Let's store the reputation with a key called `qriReputationKey`

Let's change the variable `qriConnManagerValue` to `qriBaselineReputation`. 

Let's create a new function 
`func (n *QriNode) TagReputation(pid peer.ID) error` 
- look up the reputation in the peerstore
- if it does not exists, add the baseline reputation
- if it does exist, save to a variable
- Get tag from ConnManager
- If it is the incorrect reputation:
- Untag the connection and Tag it with the reputation.

`func  (n *QriNode) UpdateReputation(pid peer.ID, reputation int) error`
- get's the reputation from the peerstore
- if one does not exist, old reputation == baseline rep
- adds the new reputation to the old reputation
- puts the reputation into the peerstore
- untags the ConnManager, and Tags with the new reputation

In `HandlePeerFound`, when we find out that a peer is indeed a Qri peer, `TagReputation`.

In AddQriPeer, once we have the profile, `checkReputation(profile *profile.Profile)`

In check Reputation, get the reputation from the registry. If no response, do nothing.

func  (n *QriNode) checkReputation(profile *profile.Profile) error
- for each peerID (found in profile.PeerIDs):
- check peerstore for reputation
- if reputation does not equal baseline reputation and is negative, call `DisconnectFromPeer` and return (since we are going to require registries to provide a reputation that is positive or negative, we know that if we have checked the reputation before, we will not get the baseline reputation back. This will prevent unnecessary calls to the registry)
- get or create regClient
- call GetReputation(profile.ID)
- if no response, return
- get the reputation from the response
- call `updateReputation(reputation)`
- get reputation from peerstore, if negative => `DisconnectFromPeer` and return

`func (n *QriNode) DisconnectFromPeer(pid peer.ID)`
- Get all connections to peer
- for each connection, close it!

Phew! very exciting

### remove unnecessary tagging and adding of Addrs.

Currently we see this in `SendMessage`, but I've noticed this in a few other places as well. We should remove these as we see them.

__last important thing to not forget: add to registry notation that we NEEEED the registry to respond with non zero numbers in order for registry to do anything effective!__

## Edit:
After doing a pass on this, we realized that we are missing something.

We need a `ProfileStore` that is an ephemeral store of a `repo.profile`. Also, `repo.Profile` needs to store a Reputation, that does not travel with the `Profile` once you convert to a `repo.ProfilePod`

To better plan for the future, we want to add a TTL to the registry reputation response. This will get discarded for now.

We need to create a `ProfileStore`:

```
type ProfileStore struct {
    profiles map[repo.ID]*repo.Profile
    peers map[peer.ID]*repo.Profile
    profilesLock sync.Mutex
    peersLock sync.Mutex
}

// Profile must have a profile.ID
func (ps *ProfileStore) SetProfile(p *repo.Profile, ttl time.Duration) error

func (ps *ProfileStore) GetProfile(id *repo.ID) *repo.Profile

func (ps *ProfileStore) GetProfileFromPeerID(id peer.ID) *repo.Profile

var ErrNotFound = fmt.Error("item not found")
```

ramfox:
closes #578

### DIFF
--- a/p2p/connected_test.go
+++ b/p2p/connected_test.go
@@ -2,6 +2,7 @@ package p2p
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"testing"
 
@@ -18,15 +19,18 @@ func TestAnnounceConnected(t *testing.T) {
 		return
 	}
 
-	if err := p2ptest.ConnectQriPeers(ctx, testNodes); err != nil {
-		t.Error(err.Error())
-		return
-	}
-
 	// Convert from test nodes to non-test nodes.
 	nodes := make([]*QriNode, len(testNodes))
 	for i, node := range testNodes {
 		nodes[i] = node.(*QriNode)
+	}
+
+	for i, a := range nodes {
+		for _, b := range nodes[i+1:] {
+			bpi := b.SimplePeerInfo()
+			a.Host.Connect(ctx, bpi)
+			a.HandlePeerFound(bpi)
+		}
 	}
 
 	// create a new, disconnected node
@@ -35,7 +39,7 @@ func TestAnnounceConnected(t *testing.T) {
 		t.Error(err.Error())
 		return
 	}
-	node := nds[0]
+	node := nds[0].(*QriNode)
 	wg := sync.WaitGroup{}
 
 	// TODO - this logic needs some precise-ifying to make this test more robust
@@ -57,19 +61,17 @@ func TestAnnounceConnected(t *testing.T) {
 				break
 			}
 		}
-	}(node.(*QriNode))
+	}(node)
 
 	// connected that node to only one member of the network
-	if err := p2ptest.ConnectQriPeers(ctx, []p2ptest.TestablePeerNode{node, testNodes[0]}); err != nil {
-		t.Error(err.Error())
-		return
-	}
+	npi := nodes[0].SimplePeerInfo()
+	node.Host.Connect(ctx, npi)
+	node.HandlePeerFound(npi)
 
 	// have that node announce connection
-	if err := node.(*QriNode).AnnounceConnected(); err != nil {
+	if err := node.AnnounceConnected(); err != nil {
 		t.Error(err.Error())
 		return
 	}
-
 	wg.Wait()
 }
--- a/p2p/peers.go
+++ b/p2p/peers.go
@@ -6,7 +6,7 @@ import (
 
 	"github.com/qri-io/qri/config"
 	"github.com/qri-io/qri/repo/profile"
-
+	"github.com/qri-io/registry/regclient"
 	ma "gx/ipfs/QmYmsdtJ3HsodkePE3eU3TsCaP2YvPZJ4LoXnNkDE5Tpt7/go-multiaddr"
 	pstore "gx/ipfs/QmZR2XWVVBCtbgBWnQhWk2xcQfaR3W8faQPriAiaaj7rsr/go-libp2p-peerstore"
 	peer "gx/ipfs/QmdVrMn1LhB4ybb8hMVaMLXnA8XRSewMnK6YqXKXoTcRvN/go-libp2p-peer"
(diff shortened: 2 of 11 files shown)


p2p/peers.go
```diff
@@ -111,18 +111,100 @@ func (n *QriNode) PeerInfo(pid peer.ID) pstore.PeerInfo {
...
+}
+
 // AddQriPeer negotiates a connection with a peer to get their profile details
-// and peer list.
+// and peer list. Checks reputation, updates conn manager with reputation, closes
+// connection if resulting reputation is negative.
 func (n *QriNode) AddQriPeer(pinfo pstore.PeerInfo) error {
-	// add this peer to our store so libp2p has the provided addresses of
-	// the peer in the next call
-	n.Host.Peerstore().AddAddrs(pinfo.ID, pinfo.Addrs, pstore.TempAddrTTL)
```
ramfox: Okay. So. Having a hard time with this one line. 

the `p2ptest.ConnectQriPeers` is used throughout our codebase as the way to connect nodes. It basically wraps `AddPeers` which wraps `AddQriPeers`. However, AddQriPeers doesn't do anything with connecting two peers, it just on-boards a peer if it is a Qri peer, after we have established a connection. 

This line here, where we add the peerinfo to the AddressBook is why `p2ptest.ConnectQriPeers` works. To connect two qri peers, we should really be calling `n.Host.Connect`, which takes care of adding the Addrs to the address book. However, I'm not totally sure how much we want to add to our `TestablePeerNode` interface and what the logic behind keeping it small might be. Another option would be to call `n.Host.Connect(ctx, peerinfo)` in`n.AddPeer` before calling `AddQriPeer`. 

Basically, removing this line breaks all the tests that use `p2p.ConnectQriPeers`, and I'm struggling with figuring out the best way to resolve.

p2p/connected_test.go
```diff
@@ -18,24 +19,27 @@ func TestAnnounceConnected(t *testing.T) {
...
-		return
-	}
-
 	// Convert from test nodes to non-test nodes.
 	nodes := make([]*QriNode, len(testNodes))
 	for i, node := range testNodes {
 		nodes[i] = node.(*QriNode)
 	}
 
+	for i, a := range nodes {
```
dustmop: We shouldn't be changing this, the semantics between ConnectQriPeers and Host.Connect are completely separate. See my other comments in p2p/peers.go

p2p/peers.go
dustmop: First off, apologies because the whole situation with ConnectNodes and ConnectQriPeers is messy and needs a major overall. I'll describe what I think the current reality is, but I may have certain details incorrect (I'm still constantly confused by it, and had to just read through enough of the code to remember what's going on).

> the p2ptest.ConnectQriPeers is used throughout our codebase as the way to connect nodes

This is not accurate. `p2ptest.ConnectNodes` actually creates the live communication connection channel between nodes. `p2ptest.ConnectQriPeers` is doing something else.

The problem is that we have at least 3 meanings of the term "connect" right now:

* libp2p connect - This is `Host.Connect` and `Network.DialPeer`, the thing you're talking about; a live communication channel between nodes.
* peer list & profile - This is what `p2p.ConnectQriPeers` is doing right now. It's trading profile data between nodes. Most of our p2p tests are depending on this happening right now, but don't actually care about network connection.
* mapstore's mock ipfs links - This is call to `m0.AddConnection` seen in some tests, a way to immitate ipfs's global data so that filenames work across nodes

My vote is to modify `p2ptest.ConnectQriPeers` as little as possible right now, and definitely don't change its semantics. If the reputation implementation needs nodes to be connected over a communication channel, it should be using `p2ptest.ConnectNodes` instead. Then, in the near future, I'll try making a change that cleans all of this up by having a single method "connect" nodes, with an options struct defining exactly what kind of connection we need (libp2p, peer&profile, mock ifps, others...).

ramfox:
This dive into p2p made us realize a bunch of cleaning up/refactoring needed to happen before we could get to reputation. Implementing reputation has been pushed back, closing this PR for now, since I am not currently working on it.

PR #581, opened 2018-10-23
feat(save): Merge command `new` into `save`.

dustmop:
TODO:
* The api needs more work / more code deleted / handling of refs
* Tests are failing, probably need to delete / rename some things
* Might have broken the spinner? I don't understand how it was working before.
* The way bodyPaths are made into absolute paths has a bug (is breaking a test)


lib/datasets.go
```diff
@@ -161,12 +116,24 @@ func (r *DatasetRequests) Save(p *SaveParams, res *repo.DatasetRef) (err error)
...
+		}
+	} else {
+		ds, bodyFile, secrets, err = actions.UpdateDataset(r.node, p.Dataset)
+		if err != nil {
+			return err
+		}
 	}
 
-	ref, body, err := actions.CreateDataset(r.node, p.Dataset.Name, ds, body, secrets, p.DryRun, true)
+	ref, body, err := actions.CreateDataset(r.node, p.Dataset.Name, ds, bodyFile, secrets, p.DryRun, true)
```
b5: Let's rename `actions.CreateDataset` -> `actions.SaveDataset`

lib/datasets.go
```diff
@@ -161,12 +116,24 @@ func (r *DatasetRequests) Save(p *SaveParams, res *repo.DatasetRef) (err error)
 		return fmt.Errorf("option to make dataset private not yet implimented, refer to https://github.com/qri-io/qri/issues/291 for updates")
 	}
 
-	ds, body, secrets, err := actions.UpdateDataset(r.node, p.Dataset)
```
b5: The bulk of the work that needs to be checked here is improving the logic performed by `actions.NewDataset` and `actions.UpdateDataset`, with the [rfc](https://github.com/qri-io/rfcs/blob/master/text/0017-define_dataset_creation.md#steps) as the source of truth. 

I think a solid starting point would be to declare the new core package and move:
* `actions.NewDataset` -> `core.PrepareDatasetNew`
* `actions.UpdateDataset` -> `core.PrepareDatasetUpdate`. 
The `action` func that saves a dataset should then check for a `ds.Prev` value, and call one of those two core functions based on the presence of `ds.Prev`. This is kinda laid out in the [rfc reference section](https://github.com/qri-io/rfcs/blob/master/text/0017-define_dataset_creation.md#reference-level-explanation), but feel free to consider it a loose guide. For example, I think we can get away without `actions.prepareDataset`.

The logic in `NewDataset` and `UpdateDataset` is awful, but I think we'll have an easier time if we keep those calls & have this PR accomplish the task of merging new & save. If we get those methods moved into the new core package we can then do the work of cleaning them up in subsequent PRs.

These changes are a good start. The failing API tests that show missing/altered `bodyPath` and `path` values are an indication that logic from prior `actions.NewDataset` and `actions.UpdateDataset` are missing.

b5:
Oh also, rebasing will make some things work better. You'll have to `go get github.com/qri-io/startf` (we've finished the skytf rename)

cmd/save.go
```diff
@@ -164,7 +159,22 @@ func (o *SaveOptions) Run() (err error) {
...
+		// Get the absolute path to the body file. Especially important if we are running
+		// `qri connect` in a different terminal, and that instance is in a different directory;
+		// that instance won't correctly find the body file we want to load if it's not absolute.
+		dsp.BodyPath, err = filepath.Abs(dsp.BodyPath)
+		if err != nil {
+			return err
+		}
+	}
+	// If dataset name is not given, derive it from the --body value.
+	if dsp.Name == "" {
```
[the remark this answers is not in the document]
dustmop: I moved this back to the actions/ location. Also, the --body flag is converted by concating the basename and extension without the period, so there would be multiple default names, like "me/bodyjson" and "me/bodycsv". But yes, I agree, we should make sure the "standard" names used in zip files aren't turned into the dataset name, an import should always contain the name of the desired dataset.

lib/datasets.go
dustmop: Speaking in person, @b5 and I decided to hold off on the new `core` functions for a later PR. I added comments describing the intention for these new functions.

Failing API tests were mostly a consequence of the /new API route not being deleted yet.

PR #582, opened 2018-10-23
refactor(p2p): p2p and p2ptest need a refactor to clarify and slim down what we do to connect with a peer

ramfox:
* renamed `n.Connect` => `n.GoOnline`, made `Host` field & and `StartOnlineServices` private
* added `Host()` func that returns the QriNodes underlying host
* added `Host()` to`p2ptest.TestablePeerNode` interface
* renamed `AddPeer` to `UpgradeQriConnection`, renamed `Connect` to `GoOnline` in p2p and p2ptest
* removed `Keys()` and `Addrs()` from `p2ptest.TestablePeerNode` interface
* copied 4 peers testdata from the `p2p` package to the `p2ptest` package. This notably a bad way to do this.
* wrote `TestableNode` struct that satisfies the `p2ptest.TestablePeerNode` interface
* wrote test for `ConnectNodes` using `TestableNode`s


api/api.go
```diff
@@ -54,18 +54,6 @@ func (s *Server) Serve() (err error) {
 	go s.ServeRPC()
 	go s.ServeWebapp()
 
-	peerBootstrapped := func(peerId string) {
```
b5: Note for others: I asked @ramfox to remove this. It's a carry-over from older libp2p days that we no longer need. The contents of `node.StartOnlineServices` have been incorporated into the new `GoOnline` call

p2p/node.go
```diff
@@ -167,12 +181,12 @@ func (n *QriNode) Connect() (err error) {
 		n.Online = true
 		go n.echoMessages()
 	}
-	return nil
+	return n.startOnlineServices()
```
b5: it's important that `n.startOnlineServices` is only called when tansitioning from offline to online state. To do that we should move `return n.startOnlineServices()` up and into the `if !n.Online {` closure, and restore the `return nil` that should happen otherwise.

Come to think of it, it'd be great if we removed the giant `if !n.Online` and instead wrote:
```go
if n.Online {
  return nil 
}

// Then to all the go-online stuff here
```

p2p/node.go
```diff
@@ -294,16 +316,16 @@ func (n *QriNode) SendMessage(msg Message, replies chan Message, pids ...peer.ID
...
 		if err != nil {
 			return fmt.Errorf("error opening stream: %s", err.Error())
 		}
 		defer s.Close()
 
 		// now that we have a confirmed working connection
 		// tag this peer as supporting the qri protocol in the connection manager
-		n.Host.ConnManager().TagPeer(peerID, qriConnManagerTag, qriConnManagerValue)
-		n.Host.Peerstore().AddAddr(peerID, s.Conn().RemoteMultiaddr(), pstore.TempAddrTTL)
+		n.Host().ConnManager().TagPeer(peerID, qriConnManagerTag, qriConnManagerValue)
```
b5: I think all of these should use the internal `n.host` instead of `n.Host()`, it'll save a function call.

p2p/test/p2ptest.go
```diff
@@ -13,20 +13,18 @@ import (
...
 type TestablePeerNode interface {
-	Keys() pstore.KeyBook
-	Addrs() pstore.AddrBook
-	HostNetwork() net.Network
+	// add host here?
+	Host() host.Host
 	SimplePeerInfo() pstore.PeerInfo
-	AddPeer(pstore.PeerInfo) error
-	Connect() error
+	UpgradeQriConnection(pstore.PeerInfo) error
```
b5: Any opposition to naming this `UpgradeToQriConnection`, adding "To" implies transitioning from one state to another. As written I think it's possible to get confused and think we're going `QriConnection -> QriConnection` instead of `BasicConnection -> QriConnection`

ramfox: that's so funny, I literally did that after pushing and was going to ask how you felt about it.

Yes I totally agree.

p2p/node.go
```diff
@@ -106,73 +107,88 @@ func NewQriNode(r repo.Repo, p2pconf *config.P2P) (node *QriNode, err error) {
...
-		n.Online = true
-		go n.echoMessages()
+	p, err := n.Repo.Profile()
+	if err != nil {
+		log.Errorf("error getting repo profile: %s\n", err.Error())
+		return err
 	}
-	return nil
+	p.PeerIDs = []peer.ID{n.Host().ID()}
+	// add listen addresses to profile store
```
b5: let's delete this commented-out code

p2p/node.go
```diff
@@ -106,73 +107,88 @@ func NewQriNode(r repo.Repo, p2pconf *config.P2P) (node *QriNode, err error) {
...
+		return err
+	}
+
+	n.Online = true
+	go n.echoMessages()
+
+	return n.startOnlineServices()
 }
 
 // StartOnlineServices bootstraps the node to qri & IPFS networks
```
b5: lowercase `startOnlineServices`

p2p/node.go
```diff
@@ -181,11 +197,9 @@ func (n *QriNode) StartOnlineServices(bootstrapped func(string)) error {
 	// need a call here to ensure boostrapped is called at least once
```
b5: let's remove this comment

p2p/node.go
```diff
@@ -181,11 +197,9 @@ func (n *QriNode) StartOnlineServices(bootstrapped func(string)) error {
...
 	// TODO - this is an "original node" problem probably solved by being able
 	// to start a node with *no* qri peers specified.
-	defer bootstrapped("")
 
 	go func() {
-		pInfo := <-bsPeers
-		bootstrapped(pInfo.ID.Pretty())
+		_ = <-bsPeers
```
b5: this can be refactored to just `<-bsPeers` without the assignment. It'd be great to add a comment that says: `block until we have at least one successful bootstrap connection`

p2p/test/testable_node.go
```diff
@@ -0,0 +1,108 @@
+package p2ptest
```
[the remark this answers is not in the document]
ramfox: oooo interesting. kk, i didn't understand before, i get it now.

p2p/bootstrap.go
```diff
@@ -25,7 +25,7 @@ func (n *QriNode) Bootstrap(boostrapAddrs []string, boostrapPeers chan pstore.Pe
 	for _, p := range randomSubsetOfPeers(pinfos, 4) {
 		go func(p pstore.PeerInfo) {
 			log.Infof("boostrapping to: %s", p.ID.Pretty())
-			if err := n.Host.Connect(context.Background(), p); err == nil {
+			if err := n.Host().Connect(context.Background(), p); err == nil {
```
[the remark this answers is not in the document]
ramfox: kk got it

b5:
🎉

PR #584, opened 2018-10-25
feat(DatasetRef.Published): add "published" flag to dataset ref, support for storing published in fsrepo

b5:
This PR is the first of a few working towards qri-io/rfcs#28. I've put some time into cleaning up our testing situation in the repo packages, found a few bugs in the process, and even managed to fix one or two. I'm structuring this as a separate PR to keep the noise lower when we change things higher up the stack. 

While working through the RFC Dustin mentioned just storing a "published" flag in our existing refstore, which to me feels like the right thing to do. In this case the path of least resistance to moving "Published" information around our stack is to include it as a field on `DatasetRef`. I want to take this approach for now, but if we find ourselves wanting to attach any more fields beyond `Published`, I think we should rethink this approach, potentially implementing some additional struct of ref-metadata, or a generic key-value store associated with each Reference (similar to how libp2p.Peerstore works).

I don't want to jump there yet b/c we don't know for sure that we need such a thing, and think this is a nice solution while we figure out weather we need more features here.


repo/ref.go
```diff
@@ -68,6 +74,8 @@ type DatasetRef struct {
 	Path string `json:"path,omitempty"`
 	// Dataset is a pointer to the dataset being referenced
 	Dataset *dataset.DatasetPod `json:"dataset,omitempty"`
+	// Published indicates weather this reference is listed as an available dataset
```
dustmop: "weather" -> "whether"

repo/ref.go
```diff
@@ -15,9 +15,15 @@ type Refstore interface {
 	// PutRef adds a reference to the store. References must be complete with
 	// Peername, Name, and Path specified
 	PutRef(ref DatasetRef) error
+	// GetRef "completes" a passed in partial reference, filling in missing fields
```
dustmop: Does "partial" here mean that the input is allowed to be missing Peername or Name or Path? Or does it mean that those are required, but that GetRef will fill in the Dataset field?

repo/test/test_repo.go
```diff
@@ -240,53 +240,3 @@ var BadStructureFile = cafs.NewMemfileBytes("badStructure.csv", []byte(`
 colA, colB, colB, colC
 1,2,3,4
 1,2,3,4`))
-
-// BodyFilepath returns the absolute path to a data file for a given test case name
-// func BodyFilepath(testcase string) (string, error) {
-// 	gp := os.Getenv("GOPATH")
```
dustmop: Big fan of deleting all this 👍

PR #586, opened 2018-10-25
feat(base): defining base package

b5:
For a while now we've wanted to carve a "local" package out of actions that both p2p and actions can use. (Okay, we might get right of actions entirely, I'm not sure yet). This starts that process, and already it's feelin' like a great move.

I want to call it "base" instead of "core" b/c IPFS already has very developed terminology around the word "core".


b5:
I'd like to get this PR landed after #585. It looks huge, but is mainly a refactor, moving code out of `actions` and into `base`. Little-to-no functionality should change.

PR #590, opened 2018-10-29
feat(Publish): add publish command, basic API endpoints

b5:


lib/datasets.go
```diff
@@ -162,6 +162,12 @@ func (r *DatasetRequests) Save(p *SaveParams, res *repo.DatasetRef) (err error)
 	return nil
 }
 
+// SetPublishStatus updates the pubilicity of a reference in the peer's namespace
```
dustmop: pubilicity -> publicity

cmd/qri.go
```diff
@@ -56,8 +56,12 @@ https://github.com/qri-io/qri/issues`,
 		NewInfoCommand(opt, ioStreams),
 		NewListCommand(opt, ioStreams),
 		NewLogCommand(opt, ioStreams),
+		NewPublishCommand(opt, ioStreams),
 		NewPeersCommand(opt, ioStreams),
-		NewRegistryCommand(opt, ioStreams),
+		// TODO - registry command is currently removed in favor of the newer "qri publish" command
+		// we should consider refactoring this code (espcially it's documentation) &
+		// use it for restistry-specific publication & search interaction
```
dustmop: restistry -> registry

cmd/registry.go
```diff
@@ -8,6 +8,9 @@ import (
 )
 
 // NewRegistryCommand creates a `qri registry` subcommand for working with configured registries
+// TODO - registry command is currently removed in favor of the newer "qri publish" command
+// we should consider refactoring this code (espcially it's documentation) &
+// use it for restistry-specific publication & search interaction
```
dustmop: same spelling change

base/ref.go
```diff
@@ -0,0 +1,30 @@
...
+	p, err := r.Profile()
+	if err != nil {
+		return false
+	}
+
+	return p.ID == ref.ProfileID
+}
+
+// SetPublishStatus updates the Published field of a dataset ref
+func SetPublishStatus(r repo.Repo, ref *repo.DatasetRef) error {
```
dustmop: This comment doesn't appear to match what the function body is doing. Nothing here is modifying the Publish field. Is the expectation that the caller sets the field and then calls this to serialize the value? Rather, I would expect that this would take a bool parameter and assign it to the Published field.

base/datasets.go
```diff
@@ -9,14 +9,26 @@ import (
 )
 
 // ListDatasets lists datasets from a repo
-func ListDatasets(r repo.Repo, limit, offset int, RPC bool) (res []repo.DatasetRef, err error) {
+func ListDatasets(r repo.Repo, limit, offset int, RPC, publishedOnly bool) (res []repo.DatasetRef, err error) {
```
dustmop: About this filename, there's only one function here, `ListDatasets`. Can it be moved into the very similarly named file "dataset.go"? Or vice versa?

base/ref_test.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"testing"
+
+	"github.com/qri-io/qri/repo"
+	"github.com/qri-io/qri/repo/profile"
+)
+
+func TestInLocalNamespace(t *testing.T) {
+	r := newTestRepo(t)
+	reff := addCitiesDataset(t, r)
```
dustmop: Name this variable `ds` or `cities` instead.

cmd/publish.go
```diff
@@ -37,14 +37,17 @@ to a published dataset will be immideately visible to connected peers.
...
 	return cmd
 }
 
 // PublishOptions encapsulates state for the publish command
 type PublishOptions struct {
 	ioes.IOStreams
 
-	Refs []string
+	Refs      []string
+	Unpublish bool
```
dustmop: I'd recommend keeping this consistent with the other flags, not using the negated form, and calling it "Publish" instead, or maybe "PublishState".

base/ref.go
```diff
@@ -26,6 +30,9 @@ func SetPublishStatus(r repo.Repo, ref *repo.DatasetRef, published bool) error {
 		return fmt.Errorf("can't publish datsets that are not in your namespace")
 	}
 
+<<<<<<< HEAD
```
dustmop: Merge conflicts need to be resolved in this commit.

b5: yeah I made a bunch of garbage errors on a rebase, force-pushed the last commit

b5:
alrite, solid review @dustmop, look good to merge?

PR #592, opened 2018-10-30
refactor(save): large refactoring for saving datasets

dustmop:


dustmop:
I don't understand this refactoring well enough to provide great feedback. I realize it's mostly moving functions, but there's also some subtle changes in behavior as well. It would be great to have more comments that explain the context around code being commented out or slightly altered as things get rearranged so that we don't lose that information.

b5:
Agreed, there are lots of little things I'll make a note of, most of which should be noted before the next release.

dustmop:
Looks good, go ahead.

PR #596, opened 2018-11-02
feat(zip): Import a zip file using save on the command-line

dustmop:


b5:
We ended up landing this in another PR!

PR #603, opened 2018-11-08
refactor: Mutually Exclusive Manual & Scripted Tranforms

b5:
When finished, this PR should implement the bulk of [RFC0020](https://github.com/qri-io/rfcs/blob/master/text/0020-distingush_manual_vs_scripted_transforms.md), excluding field-level exclusion checks

### TODOS:
* ~~[ ] lib:~~
  * ~~[ ] an integration test on lib that runs the tutorial from RFC 0020~~ _(this is better covered as a unit test on `actions.SaveDataset`)_
* [x] cmd: 
  * [x] update: add `--recall-tf` flag
  * [x] save
    * [x] add `--recall tf` flag
    * [ ] ~~add command documentation explaining transforms run only the first time they're provided~~ _(punt to later)_
* [x] startf
  * [x] adjust startf/dataset to accept a MutationCheckFunc
  * [x] lib version needs to to go 0.1.0
  * [x] startf/dataset infile -> bodyFile
* [x] actions:
  * [x] adjust logic of `SaveDataset` and `UpdateDataset` transform exec to match RFC
  * [x] test to ensure previous dataset is provided to transform
  * [x] confirm manual transforms run before scripted transforms
  * [x] generate a MutationCheckFunc from the result of user-provided values, pass to transform
  * [x] test to ensure tf is removed by save that doesn't run
  * [ ] ~~update: add no-transform error that uses history check, reports actionable message~~ _(punt to later)_
* [x] base:
  * [x] add capacity to recall components from history

### DIFF
--- a/actions/dataset.go
+++ b/actions/dataset.go
@@ -3,6 +3,7 @@ package actions
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/ipfs/go-datastore"
 	"github.com/qri-io/cafs"
@@ -64,12 +65,22 @@ func SaveDataset(node *p2p.QriNode, dsp *dataset.DatasetPod, dryRun, pin bool) (
 	if ds.Transform != nil {
 		mutateCheck := mutatedComponentsFunc(dsp)
 		if ds.Transform.Script == nil {
-			var f *os.File
-			f, err = os.Open(ds.Transform.ScriptPath)
-			if err != nil {
-				return
+			if strings.HasPrefix(ds.Transform.ScriptPath, "/ipfs") || strings.HasPrefix(ds.Transform.ScriptPath, "/map") || strings.HasPrefix(ds.Transform.ScriptPath, "/cafs") {
+				var f cafs.File
+				f, err = node.Repo.Store().Get(datastore.NewKey(ds.Transform.ScriptPath))
+				if err != nil {
+					return
+				}
+				ds.Transform.Script = f
+			} else {
+				var f *os.File
+				f, err = os.Open(ds.Transform.ScriptPath)
+				if err != nil {
+					return
+				}
+				ds.Transform.Script = f
 			}
-			ds.Transform.Script = f
+
 		}
 		// TODO - consider making this a standard method on dataset.Transform
 		script := cafs.NewMemfileReader(ds.Transform.ScriptPath, ds.Transform.Script)
--- a/actions/dataset.go
+++ b/actions/dataset.go
@@ -12,49 +12,57 @@ import (
 	"github.com/qri-io/qri/base"
 	"github.com/qri-io/qri/p2p"
 	"github.com/qri-io/qri/repo"
+	"github.com/qri-io/qri/repo/profile"
 )
 
 // SaveDataset initializes a dataset from a dataset pointer and data file
 func SaveDataset(node *p2p.QriNode, dsp *dataset.DatasetPod, dryRun, pin bool) (ref repo.DatasetRef, body cafs.File, err error) {
 	var (
-		ds       *dataset.Dataset
-		bodyFile cafs.File
-		secrets  map[string]string
-		// NOTE - struct fields need to be instantiated to make assign set to
-		// new pointer values
-		userSet = &dataset.Dataset{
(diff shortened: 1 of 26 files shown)


lib/datasets.go
```diff
@@ -138,11 +150,19 @@ func (r *DatasetRequests) Save(p *SaveParams, res *repo.DatasetRef) (err error)
...
 	}
 
+	if ds.Name == "" {
+		return fmt.Errorf("name is required")
+	}
+	if ds.BodyPath == "" && ds.Body == nil && ds.BodyBytes == nil && ds.Structure == nil && ds.Meta == nil && ds.Viz == nil && ds.Transform == nil {
+		return fmt.Errorf("no changes to save")
+	}
+
 	ref, body, err := actions.SaveDataset(r.node, ds, p.DryRun, true)
```
ramfox: Because of the naming, it's sort of unclear that what you are passing is a datasetPod, and a Pod of only the changes that we want to make on the previous dataset.

actions/dataset.go
```diff
@@ -12,56 +13,74 @@ import (
...
+		prevPath                 string
+		bodyFile, changeBodyFile cafs.File
+		secrets                  map[string]string
+		pro                      *profile.Profile
+		changes                  = &dataset.Dataset{}
+		r                        = node.Repo
 	)
 
+	// set ds to dataset head, or empty dataset if no history
+	ds, bodyFile, prevPath, err = base.PrepareDatasetSave(r, dsp.Peername, dsp.Name)
```
ramfox: Not totally clear that ds is the fully hydrated previous dataset, with just the PreviousPath field updated. It makes it a bit hard to understand the actions we take going forward without knowing this.

actions/dataset.go
```diff
@@ -12,56 +13,74 @@ import (
...
+	}
+
+	if err = changes.Decode(dsp); err != nil {
+		return
+	}
+	ds.Assign(changes)
+	clearPaths(ds)
 
 	if ds.Transform != nil {
+		mutateCheck := mutatedComponentsFunc(dsp)
```
ramfox: Would be great to have a note here what exactly we are doing: making a record of all the parts that the datasetPod is changing, that way we can make sure the transform is not also changing them.

Part of the confusion, is that it isn't totally clear that `dsp` is a dataset made up of only the changes. `changes` makes sense, but maybe calling dsp `changesPod` might make this more clear?

actions/dataset.go
```diff
@@ -72,10 +91,29 @@ func SaveDataset(node *p2p.QriNode, dsp *dataset.DatasetPod, dryRun, pin bool) (
 			return
 		}
 		node.LocalStreams.Print("✅ transform complete\n")
-		ds.Assign(userSet)
 	}
 
-	return base.CreateDataset(node.Repo, node.LocalStreams, dsp.Name, ds, bodyFile, dryRun, pin)
+	// let's make history, if it exists:
+	ds.PreviousPath = prevPath
```
ramfox: This also happens in `PrepareDatasetSave`. Is this in case `changes` ends up smashing the `PreviousPath` field?

base/dataset.go
```diff
@@ -219,7 +201,7 @@ func UnpinDataset(r repo.Repo, ref repo.DatasetRef) error {
 // * dsp.BodyPath being a url
 // * dsp.BodyPath being a path on the local filesystem
 // TODO - consider moving this func to some other package. maybe actions?
-func DatasetPodBodyFile(dsp *dataset.DatasetPod) (cafs.File, error) {
+func DatasetPodBodyFile(store cafs.Filestore, dsp *dataset.DatasetPod) (cafs.File, error) {
```
ramfox: With all the `else`s used in this function, there is something that seems off about this function.

Perhaps, to be more clear,  we bail if `BodyPath` is empty, before we start checking all the different `BodyPath` cases? That way they can be all if statements that return at the end of each statement.

actions/dataset.go
b5: nice catch. I'm going to remove the assignment in `PrepareDatasetSave`

b5:
For the record, this one was a lot of work, and very much worth it.

Issue #612, opened 2018-11-16
Problems with saving a dataset with a format different than its previous version


dustmop:
Say we have a dataset in `csv` format. Running `qri save` using a new format like `json` leads to two different outcomes right now:
* The new format overwrites the old format, thus changing the structure
* The old format is used to parse the new body, leading to parse errors

This mismatch also manifests in the frontend, when modifying the body of a csv dataset with a headerRow: the frontend hits the API using a "csv-ish" POST body, and includes the structure with headerRow set, but without including the header row in the POST body. The backend handles this by stripping the first row.

The solution here is to change the frontend to always POST json data, and add a flag to SaveParams that requests automatic conversion from the given dataset body format, to the format of the previous version. Attempting to save a body with a different format should be an error if this flag is not set to true.

In addition, the code in base.CreateDataset calls InferValues, which uses the body's filename extension to guess a format if not is provided. However, this function is called *after* the previous pod and changes pod are merged (in actions.SaveDataset), which is incorrect. This also needs to be fixed before this entire issue can be solved.

Here is the current state of the world, showing what happens when a dataset is first newly created, and then has an updated save that uses a different format. Note that all branches end in an undesirable state.

```
use `qri save` first on a csv body file, then a json body file
+=========================================================================================+
| Method           | Body       | prev    | changes | merged  | infer   | Result          |
+-----------------------------------------------------------------------------------------+
| save --body=csv  | bodyPath   | <nil>   | <nil>   | <nil>   | csv     | csv             |
+-----------------------------------------------------------------------------------------+
| save --body=json | bodyPath   | csv     | <nil>   | csv     | csv !   | parse error     |
+=========================================================================================+

use `qri save` first on a json body file, then a csv body file
+=========================================================================================+
| Method           | Body       | prev    | changes | merged  | infer   | Result          |
+-----------------------------------------------------------------------------------------+
| save --body=json | bodyPath   | <nil>   | <nil>   | <nil>   | json    | json            |
+-----------------------------------------------------------------------------------------+
| save --body=csv  | bodyPath   | json    | <nil>   | json    | json !  | parse error     |
+=========================================================================================+

use `qri save` first on dataset.yaml with csv no structure, then dataset.yaml with json no structure
+=========================================================================================+
| Method            | Body       | prev    | changes | merged  | infer   | Result         |
+-----------------------------------------------------------------------------------------+
| save --file(csv)  | bodyPath   | <nil>   | <nil>   | <nil>   | csv     | csv            |
[...]

PR #613, opened 2018-11-16
fix(format): Detect format change when saving, either error or rewrite

dustmop:
When a save command is run, and the new format is different from the format of the previous version, either return an error, or if the ConvertFormatToPrev flag is set on SaveParams, rewrite the body to match the previous version’s format. The API always sets ConvertFormatToPrev, but the command line does not (yet).

### DIFF
--- a/actions/dataset.go
+++ b/actions/dataset.go
@@ -17,7 +17,7 @@ import (
 )
 
 // SaveDataset initializes a dataset from a dataset pointer and data file
-func SaveDataset(node *p2p.QriNode, changesPod *dataset.DatasetPod, secrets map[string]string, dryRun, pin bool) (ref repo.DatasetRef, body cafs.File, err error) {
+func SaveDataset(node *p2p.QriNode, changesPod *dataset.DatasetPod, secrets map[string]string, dryRun, pin bool, convertFormatToPrev bool) (ref repo.DatasetRef, body cafs.File, err error) {
 	var (
 		prev                     *dataset.Dataset
 		prevPath                 string
@@ -33,12 +33,13 @@ func SaveDataset(node *p2p.QriNode, changesPod *dataset.DatasetPod, secrets map[
 		return
 	}
 
+	pro, err = r.Profile()
+	if err != nil {
+		return
+	}
+
 	if dryRun {
 		node.LocalStreams.Print("🏃🏽‍♀️ dry run\n")
-		pro, err = r.Profile()
-		if err != nil {
-			return
-		}
 		// dry-runs store to an in-memory repo
 		r, err = repo.NewMemRepo(pro, cafs.NewMapstore(), profile.NewMemStore(), nil)
 		if err != nil {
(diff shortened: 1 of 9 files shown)


dustmop:
Addresses https://github.com/qri-io/qri/issues/612.

actions/dataset.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 // SaveDataset initializes a dataset from a dataset pointer and data file
-func SaveDataset(node *p2p.QriNode, changesPod *dataset.DatasetPod, secrets map[string]string, dryRun, pin bool) (ref repo.DatasetRef, body cafs.File, err error) {
+func SaveDataset(node *p2p.QriNode, changesPod *dataset.DatasetPod, secrets map[string]string, dryRun, pin bool, convertFormatToPrev bool) (ref repo.DatasetRef, body cafs.File, err error) {
```
b5:
```suggestion
func SaveDataset(node *p2p.QriNode, changesPod *dataset.DatasetPod, secrets map[string]string, dryRun, pin, convertFormatToPrev bool) (ref repo.DatasetRef, body cafs.File, err error) {
```

base/dataset.go
```diff
@@ -265,3 +263,32 @@ func DatasetPodBodyFile(store cafs.Filestore, dsp *dataset.DatasetPod) (cafs.Fil
 
 	return cafs.NewMemfileReader(filepath.Base(dsp.BodyPath), file), nil
 }
+
+// ConvertBodyFormat rewrites a body from a source format to a destination format.
+func ConvertBodyFormat(bodyFile cafs.File, fromSt *dataset.Structure, toSt *dataset.Structure) (cafs.File, error) {
```
b5:
```suggestion
func ConvertBodyFormat(bodyFile cafs.File, fromSt, toSt *dataset.Structure) (cafs.File, error) {
```

base/dataset.go
```diff
@@ -265,3 +263,32 @@ func DatasetPodBodyFile(store cafs.Filestore, dsp *dataset.DatasetPod) (cafs.Fil
...
+		return nil, err
+	}
+	err = w.Close()
+	if err != nil {
+		return nil, err
+	}
+
+	// Set the new format on the structure.
+	fromSt.Format = toSt.Format
+	return cafs.NewMemfileReader("", buffer), nil
```
b5: For the sake of consistency the name of MemFileReader should be `fmt.Sprintf("body.%s", toSt.Format)`.

base/dataset.go
b5: We should have a happy-path test for this at a bare minimum:

```golang
csv := cafs.NewMemFileBytes([]byte("a,b,c"))
// Convert to JSON...
expect := `["a","b","c"]`
```

base/dataset.go
```diff
@@ -265,3 +263,32 @@ func DatasetPodBodyFile(store cafs.Filestore, dsp *dataset.DatasetPod) (cafs.Fil
...
+// ConvertBodyFormat rewrites a body from a source format to a destination format.
+func ConvertBodyFormat(bodyFile cafs.File, fromSt *dataset.Structure, toSt *dataset.Structure) (cafs.File, error) {
+	// Reader for entries of the source body.
+	r, err := dsio.NewEntryReader(fromSt, bodyFile)
+	if err != nil {
+		return nil, err
+	}
+
+	// Writes entries to a new body.
+	buffer := bytes.NewBufferString("")
```
b5:
```suggestion
	buffer := &bytes.Buffer{}
```

PR #620, opened 2018-11-20
feat(registry): create api endpoint to request list of available datasets from the registry

### LINKED ISSUE #616: Add API endpoints to talk to registry
* [x] revive registry status endpoint
*  need an endpoint that gets a list of paginated datasets from the registry
    * [x] in api: `listRegistryHandler`
            - call `RegistryRequests.List`
            - return slice of repo.DatasetRefs (double check how this gets returned in `DatasetRequests.List` func)
    * [x] in lib: `RegistryRequests.List` => 
            - call `actions.RegistryList(node)`
    * [x] in actions: `actions.RegistryList` => 
            - call to the registry
            - get a slice of `registry.Dataset`
            - convert those to `repo.DatasetRefs`
            - return

This also requires adding limit and offset to registry's list datasets function.

ramfox:
closes #616 
Also changes `ClosestConnectedPeer` to `ClosestConnectedQriPeer`, which gives a list of Qri Peers, rather than arbitrary peers we happen to be connected to.


actions/registry.go
```diff
@@ -145,6 +145,27 @@ func Unpin(node *p2p.QriNode, ref repo.DatasetRef) (err error) {
 	}
 
 	return reg.Unpin(ref.Path, pk)
+
+}
+
+// RegistryList gets a list of the published datasets of a repo's specific registry
+func RegistryList(node *p2p.QriNode, limit, offset int) (datasets []*repo.DatasetRef, err error) {
```
dustmop: Could you swap the order of the `limit` and `offset` parameters? Everywhere else (nearly?) in the codebase uses the order of `offset` then `limit`, and since both are int's, we can't use the compiler to make sure the right values are passed in, we have to stick to a consistent convention.

p2p/peers.go
```diff
@@ -57,31 +57,38 @@ func (n *QriNode) ConnectedQriPeerIDs() []peer.ID {
...
-				break
+			peerID := conn.RemotePeer()
+			protocols, err := n.host.Peerstore().SupportsProtocols(peerID, string(QriProtocolID))
+			if err != nil {
+				continue
+			}
+			if len(protocols) != 0 {
+				pid = append(pid, peerID)
+				added++
+				if added == max {
```
dustmop: Please change this to `added >= max`. It's functionally the same thing, but it makes it a little clearer what's going on. Plus, if the code changes in the future (like, another condition ends up doing `added++`), this condition will handle the case of going above the max length.

actions/registry.go
ramfox: I'm not seeing any cases in where `offset` is before `limit`. I was following the convention that was set in `lib.List`, `actions.listDatasets`, `base.libDatasets` basically our local dataset listing.

Totally agree it should be uniform!

p2p/peers.go
ramfox: Doooope.

actions/registry.go
dustmop: My mistake, you're totally right. Nevermind then!

ramfox:
Closing w/o @b5's review because he is off the grid atm.

api/api.go
```diff
@@ -238,6 +238,7 @@ func NewServerRoutes(s *Server) *http.ServeMux {
 	m.Handle("/history/", s.middleware(lh.LogHandler))
 
 	rgh := NewRegistryHandlers(s.qriNode)
+	m.Handle("/registry/list", s.middleware(rgh.RegistryListHandler))
```
b5: So glad to see this got merged! I do have one small nit here. I think this should be `/registry/datasets`, because the registry can also list users (more abstractly, the registry can `list` more than one type of thing)

ramfox: Got it, will make the change!

PR #624, opened 2018-11-27
refactor(api): endpoint changed from `registry/list` to `registry/datasets`

ramfox:


b5:
oh wait, will that affect tests?

b5:
meh, we'll deal with it it later. LGTM!

ramfox:
> oh wait, will that affect tests?

Only cosmetically (the name saved in abide), which I'm changing right now, and will push before merging!

PR #625, opened 2018-11-28
fix(webapp): restore webapp

b5:
Our webapp has gotten too complicated to reasonably serve as a single file. This PR adjusts the way we serve frontend stack files. We can compile our frontend down to a directory, and serve that CAFS directory behind the `/webapp/` endpoint. webapp code can come from the dweb and always be served consistently behind the `/webapp/` endpoint from api's webapp server. neat!

### DIFF
--- a/api/transports.go
+++ b/api/transports.go
@@ -37,7 +37,7 @@ func StartServer(c *config.API, s *http.Server) error {
 	}
 
 	// Attempt to boot a port 80 https redirect
-	go func() { HTTPSRedirect() }()
+	go HTTPSRedirect(":80")
 
 	s.TLSConfig = &tls.Config{
 		GetCertificate: certManager.GetCertificate,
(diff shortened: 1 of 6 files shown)


api/transports.go
```diff
@@ -54,14 +54,14 @@ func StartServer(c *config.API, s *http.Server) error {
 	return s.ListenAndServeTLS(cert, key)
 }
 
-// HTTPSRedirect listens on port 80, redirecting HTTP requests to https
-func HTTPSRedirect() {
-	ln, err := net.Listen("tcp", ":80")
+// HTTPSRedirect listens pver TCP on addr, redirecting HTTP requests to https
```
dustmop: pver -> over

api/webapp.go
```diff
@@ -20,20 +21,25 @@ func (s *Server) ServeWebapp() {
...
 
-// WebappJSHandler attempts to resolve an update
-func (s *Server) WebappJSHandler() http.Handler {
+// FrontendHandler resolves the current webapp hash, (fetching the compiled frontend in the process)
+// and serves it up as a traditional HTTP endpoint, transparently redirecting requests
+// for [prefix]/foo.js to [CAFS Hash]/foo.js
+// prefix is the path prefix that should be stripped from the request URL.Path
+func (s *Server) FrontendHandler(prefix string) http.Handler {
 	var webappPath = s.cfg.Webapp.EntrypointHash
 	go s.resolveWebappPath(&webappPath)
```
dustmop: I'm not a fan of this line for multiple reasons: it's confusing, it's mutating its input parameter, it has a race condition with the code below, there's no error handling, and really the data could be owned by `Server` and initialized by it, as there's nothing that necessitates it being within FrontendHandler. Could you add a comment explaining some of what's going on, or maybe add a TODO to fix it later?

b5: Totally agreed. Don't have time to fix now, but I've documented your comments here in a TODO

Issue #577, opened 2018-10-11
fix(read-only): edits to api read-only


ramfox:
Changes to qri to get frontend read-only functionality to work:

* [ ] get requests on profile picture and profile poster should be allowed through
* [x] `lib.Info` function should add the `online` field to the profilePod before returning. (The `/peers` list endpoint calls to the `lib.Peers`, which calls `action.ListPeers`. `action.ListPeers` adds the `online` fields, which is why we get the correct online-ness when we are not in read-only mode: the frontend usually gets all its profile info from this `/peers` endpoint)
* [ ] perhaps allow GET requests  to the `/peers` endpoint, this depends on what our `read-only` frontend landing page is going to be

b5:
I think this is closable. can you confirm @ramfox?

PR #629, opened 2018-11-29
feat(online): Set Online flag for peers when using lib.Info. Cleanups.

dustmop:
When getting Info for a peer, set the Online flag appropriately, to match the behavior of lib.List (peers.go). Do this by checking the other connected peers that are known to be online, and also checking if you yourself are online.

Cleanup some of the usage around ConnectedQriProfiles.

Fixes part 2 of #577.

### DIFF
--- a/lib/peers.go
+++ b/lib/peers.go
@@ -77,8 +77,20 @@ func (d *PeerRequests) ConnectedQriProfiles(limit *int, peers *[]*config.Profile
 		return d.cli.Call("PeerRequests.ConnectedQriProfiles", limit, peers)
 	}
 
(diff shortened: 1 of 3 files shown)


lib/peers.go
```diff
@@ -197,6 +209,7 @@ func (d *PeerRequests) Info(p *PeerInfoParams, res *config.ProfilePod) error {
 		return d.cli.Call("PeerRequests.Info", p, res)
 	}
 
+	// TODO: Move most / all of this to actions package, perhaps.
```
b5: agreed.

PR #636, opened 2018-12-12
fix(config migration): add config migration to update p2p.QriBootstrapAddrs

b5:
This PR adds a `Version` field the config file, and migrates any configuration that _doesn't_ have

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -15,12 +15,13 @@ import (
 	"github.com/qri-io/jsonschema"
 )
 
-// CurrentConfigVersion is the latest configuration number
-const CurrentConfigVersion = 1
+// CurrentConfigRevision is the latest configuration revision configurations
+// that don't match this revision number should be migrated up
+const CurrentConfigRevision = 1
 
 // Config encapsulates all configuration details for qri
 type Config struct {
-	Version  int
+	Revision int
 	Profile  *ProfilePod
 	Repo     *Repo
 	Store    *Store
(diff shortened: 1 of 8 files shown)


config/config.go
```diff
@@ -15,8 +15,12 @@ import (
 	"github.com/qri-io/jsonschema"
 )
 
+// CurrentConfigVersion is the latest configuration number
+const CurrentConfigVersion = 1
```
dustmop: So this might be a minor quibble, but I could see an issue with calling this "version". These days, version tends to imply something like semver, and is expected to rarely change its major value, incrementing a minor number more often. Of course we're only using the term "version" internally, but it's still a visible term inside the textual config file. As an alternative, we could use "revision", which is more often used as a monotonically increasing value, like for example a build number.

I don't feel super strongly about this, just a thought.

lib/config.go
```diff
@@ -4,6 +4,9 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"github.com/qri-io/ioes"
+	"github.com/qri-io/qri/config/migrate"
+
```
dustmop: Extra blank line?

lib/config_test.go
```diff
@@ -7,6 +7,8 @@ import (
 	"os"
 	"testing"
 
+	"github.com/qri-io/ioes"
+
```
dustmop: Extra blank line?

PR #649, opened 2019-01-07
fix(Save): add tests to cmd that test you can save a transform and viz

ramfox:
Tests ensure you can save a Viz or Transform to a dataset without having to add other dataset components as well.

Also, fixes to `prepareViz()` to ensure that we have a `Viz.Script` before creating the datasets

need to close qri-io/dsdiff#13 in order for these tests to pass


b5:
... after conflict fix

PR #656, opened 2019-01-11
fix(export): Check for unsupported export flags, recognize --format

dustmop:
Export was left in a half-finished state: it was always assuming Zipped was true, yet not checking for the presense of the flag. It accepts --format and --body-format, yet was using neither. Many other flags were being ignored entirely. Add errors for these unused flags, require that --zipped
is always used, and recognize the --format flag so that people may at least export their files using the preferred format.


dustmop:
This depends upon https://github.com/qri-io/dataset/pull/167. Also, the build is currently broken until https://github.com/qri-io/qri/pull/653 is merged.

There's a lot of commented out code in this change, but nearly all of that code wasn't being running anyway, due to Zipped being assigned true causing an early return. Export needs a lot of work, but it needs to be planned out in an RFC before these TODOs can be resolved.

api/datasets.go
```diff
@@ -247,7 +247,11 @@ func (h *DatasetHandlers) zipDatasetHandler(w http.ResponseWriter, r *http.Reque
 
 	w.Header().Set("Content-Type", "application/zip")
 	w.Header().Set("Content-Disposition", fmt.Sprintf("filename=\"%s.zip\"", "dataset"))
-	dsutil.WriteZipArchive(h.repo.Store(), ds, res.String(), w)
+	err = dsutil.WriteZipArchive(h.repo.Store(), ds, "json", res.String(), w)
```
b5: remove the `"json"` hardcode in favour of a `r.FormValue("format")`, also, I think the default should be `"yaml"` considering we use it exclusively in our docs ATM. We should start getting into the habit of keeping parity across our API and CLI for all changes moving forward to avoid tech debt.

cmd/cmd_test.go
```diff
@@ -172,8 +172,8 @@ func TestCommandsIntegration(t *testing.T) {
 		fmt.Sprintf("qri save --body=%s -t=commit_1 me/movies", movies2FilePath),
 		"qri log me/movies",
 		"qri diff me/movies me/movies2 -d=detail",
-		fmt.Sprintf("qri export -o=%s me/movies", path),
-		fmt.Sprintf("qri export -o=%s --format=cbor --body-format=json me/movies", path),
+		fmt.Sprintf("qri export -o=%s me/movies --zipped", path),
```
b5: exciting!

cmd/export.go
```diff
@@ -60,6 +53,7 @@ new dataset, use --blank.`,
 	cmd.Flags().StringVarP(&o.BodyFormat, "body-format", "", "", "format for dataset body. default is the original data format. options: json, csv, cbor")
 	cmd.Flags().BoolVarP(&o.NoBody, "no-body", "b", false, "don't include dataset body in export")
 	cmd.Flags().BoolVarP(&o.PeerDir, "peer-dir", "d", false, "export to a peer name namespaced directory")
+	cmd.Flags().BoolVarP(&o.Zipped, "zipped", "z", false, "export as a zip file")
```
b5: `zipped` or `zip`? Is there prior art for `zipped`?

cmd/export.go
```diff
@@ -155,22 +169,23 @@ func (o *ExportOptions) Run() error {
...
-		return dst.Close()
+	// TODO (dlong): Use --body-format here to convert the body and ds.Structure.Format, before
+	// passing ds to WriteZipArchive.
+	if err = dsutil.WriteZipArchive(o.Repo.Store(), ds, format, res.String(), dst); err != nil {
+		return err
 	}
+	return dst.Close()
 
-	if path != "" {
+	// TODO (dlong): Document the full functionality of export, and restore this code below. Allow
```
b5: I think we should elevate this to an issue and block the next release until it's shipped. Getting export in a better place has jumped up our list of priorities. Mind filing an issue so others can see what's going on there? No need to block merging tho

cmd/export.go
```diff
@@ -155,22 +169,23 @@ func (o *ExportOptions) Run() error {
...
+	dst, err := os.Create(fmt.Sprintf("%s.zip", path))
+	if err != nil {
+		return err
+	}
 
-		if err = dsutil.WriteZipArchive(o.Repo.Store(), ds, res.String(), dst); err != nil {
-			return err
-		}
-		return dst.Close()
+	// TODO (dlong): Use --body-format here to convert the body and ds.Structure.Format, before
```
b5: I'm SO pumped for this to land, would love it if we landed a small placeholder for this then started to think about piped / streaming use-cases for export with conversions.

The issue of converting to CSV has been on my mind for a whioe. We'll have to take care about down-depth conversions (say, JSON of depth 5 to csv). I think we should error on conversions to CSV when dataset depth is either not present or above 2. On the upside we can rely on depth now to be smarter

api/datasets.go
dustmop: The previous behavior was using "json", because WriteZipArchive was only working that way. I'm fine with changing it, but it'll include updating the api response data.

cmd/export.go
dustmop: I'm fine with `zip`, went with `zipped` only because that's what the old Options struct was using, but again I'm okay with changing it.

cmd/export.go
dustmop: Agreed, will do.

PR #662, opened 2019-01-14
fix(remove): Remove requires the --revisions to specify what to delete

dustmop:
New flag --revisions is now required by remove. If set to an integer greater than zero, will remove that many revisions, starting from newest. If set to "all", will remove the dataset entirely (the old behavior).


ramfox:
what are your thoughts about having an `--all` flag that proxies `--revisions=all`?

was just using the new command and `qri rm me/test --all` felt intuitive.

dustmop:
I actually like it a lot. Was hesitant at first, but I think it's a great idea. Adding...

b5:
Big fan. I won’t have a chance to check actual code, but the interface is lovely. Into the —all flag as well.

Issue #661, opened 2019-01-14
Qri connect & Qri export 


WaughB:
While using `qri connect` in another terminal I downloaded a dataset from another user. When I tried to run `qri export` it gave me `repo not available (are you running qri in another terminal?)`. 
To `qri export` I had to first go to the other terminal and get out of `qri connect`, then `qri export`, then `qri connect` again.

PR #667, opened 2019-01-15
fix(export): Export can be used while running `qri connect`

dustmop:
Move the main functionality of export down to lib/. Invoke the RPC call if needed. Remove the Repo method from Factory, and make the Node method more explicit (only `qri connect` needs it). Now all commands can be run while connected.

Fixes https://github.com/qri-io/qri/issues/661.

### DIFF
--- a/cmd/export.go
+++ b/cmd/export.go
@@ -4,13 +4,10 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path/filepath"
 
-	"github.com/qri-io/dataset/dsutil"
 	"github.com/qri-io/ioes"
 	"github.com/qri-io/qri/lib"
 	"github.com/qri-io/qri/repo"
-	"github.com/qri-io/qri/repo/profile"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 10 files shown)


cmd/export.go
```diff
@@ -71,10 +68,8 @@ type ExportOptions struct {
 	BodyFormat string
 	NoBody     bool
 
-	UsingRPC        bool
-	Repo            repo.Repo
-	Profile         *profile.Profile
-	DatasetRequests *lib.DatasetRequests
+	UsingRPC       bool
+	ExportRequests *lib.ExportRequests
```
ramfox: I dig this!

cmd/qri.go
```diff
@@ -208,25 +208,10 @@ func (o *QriOptions) RPC() *rpc.Client {
...
 }
 
-// Repo returns from internal state
-// TODO (dlong): This function is deprecated. Don't add new calls. Remove it soon.
-func (o *QriOptions) Repo() (repo.Repo, error) {
-	if err := o.Init(); err != nil {
-		return nil, err
-	}
+// ConnectionNode returns the internal QriNode, if it is available
+func (o *QriOptions) ConnectionNode() (*p2p.QriNode, error) {
```
ramfox: dope

PR #670, opened 2019-01-21
docs: update package definition documentation

b5:
starting into a documentation cleanup. This should complete the overview of qri at https://godoc.org/github.com/qri-io/qri


lib/test/test_crypto_generator.go
```diff
@@ -1,3 +1,5 @@
+// Package test defines utilities for testing the lib package, inlcuding caches
```
dustmop: inlcuding -> including

PR #674, opened 2019-01-25
fix(get): Datasets whose name contains a field should work with get

dustmop:


cmd/get_test.go
```diff
@@ -28,6 +28,7 @@ func TestGetComplete(t *testing.T) {
 		{[]string{"peer/ds_two", "peer/ds"}, "", []string{"peer/ds_two", "peer/ds"}, ""},
 		{[]string{"foo", "peer/ds"}, "", []string{"foo", "peer/ds"}, ""},
 		{[]string{"structure"}, "structure", []string{}, ""},
+		{[]string{"peer/human_body_facts"}, "", []string{"peer/human_body_facts"}, ""},
```
b5: yay test case

PR #677, opened 2019-01-27
refactor: use string values instead of datastore.Key

b5:
implements changes made in qri-io/cafs#22 (worth reading that PR). When interacting with a cafs.Filestore, use strings instead of datastore.Key.

depends on qri-io/cafs#22. depends on qri-io/dataset#168. depends on qri-io/startf#29


repo/fs/files.go
```diff
@@ -74,14 +73,8 @@ var paths = map[File]string{
 	FileChangeRequests: "/change_requests.json",
 }
 
-// Filepath gives the relative filepath to a repofile
+// Filepath gives the relative filepath to a repofiles
```
dustmop: `a repofiles` -> `the repofiles`?

b5: agreed! will address in a subsequent PR

PR #680, opened 2019-01-31
fix(get): Turn `body` into a selector of the `get` command

dustmop:
This is a fairly sizable change:
* Change `lib.Get` to take LookupParams and LookupResult, like `LookupBody`
* Move code out of `cmd/get` and into `lib.Get`
* Unify flags between Get and LookupBody
* Change semantics of lib.Get so that it either uses a ref or parses a path
* Deprecate `cmd/body`, add TODO to remove it soon


cmd/body.go
```diff
@@ -76,55 +71,6 @@ func (o *BodyOptions) Complete(f Factory, args []string) (err error) {
...
-	}
-
-	if o.Output != "" {
-		ioutil.WriteFile(o.Output, data, os.ModePerm)
-	} else {
-		fmt.Fprintln(o.Out, string(data))
-	}
-
-	return nil
+	// TODO (dlong): Delete `body` command in a later change. Update docs.
```
b5: 👌

cmd/get_test.go
```diff
@@ -50,8 +50,8 @@ func TestGetComplete(t *testing.T) {
 			continue
 		}
 
-		if c.path != opt.Path {
-			t.Errorf("case %d, opt.Path not set correctly. Expected: '%s', Got: '%s'", i, c.path, opt.Path)
+		if c.selector != opt.Selector {
+			t.Errorf("case %d, opt.Path not set correctly. Expected: '%s', Got: '%s'", i, c.selector, opt.Selector)
```
b5: path var name has changed: `case %d opt.Selector not set correctly...`

lib/datasets.go
```diff
@@ -65,24 +69,63 @@ func (r *DatasetRequests) List(p *ListParams, res *[]repo.DatasetRef) error {
...
 }
 
-// Get retrieves a dataset head (commit, structure, meta, etc) for a given reference, either
-// from the local repo or by asking peers for it. The res parameter will be populated upon success.
-func (r *DatasetRequests) Get(ref *repo.DatasetRef, res *repo.DatasetRef) error {
+// Get retrieves datasets and components for a given reference. If p.Ref is provided, it is
+// used to load the dataset, otherwise p.PathString is parsed to create a reference. The
+// dataset will be loaded from the local repo if available, or by asking peers for it.
+// Using p.Selector will control what components are returned in res.Bytes. The default,
+// a blank selector, will also fill the entire dataset at res.Data.
```
b5: does that dataset include the body?

lib/datasets.go
```diff
@@ -400,39 +443,68 @@ func (r *DatasetRequests) Remove(p *RemoveParams, numDeleted *int) error {
...
 
-	*data = LookupResult{
-		Path: bodyPath,
-		Data: bufData,
+	bodyPath, bufData, err := actions.LookupBody(r.node, &res.Data, df, p.FormatConfig, p.Limit, p.Offset, p.All)
+	if err != nil {
+		return err
 	}
+
+	// TOOD: --format flag
```
b5: halp my find & replace doesn't know TOOD 😛

cmd/get_test.go
dustmop: Good catch, done.

lib/datasets.go
dustmop: Expanded the comment to clarify that if p.Selector == "body", then the body is the in res.Bytes.

lib/datasets.go
dustmop: This TOOD / TODO is out of date anyway, removed.

b5: even better!

PR #685, opened 2019-02-07
docs: Change contributor link

il-dar:
Current contributor link in the heading of README goes to a 404 page, this fixes it


dustmop:
This is great, thanks very much @il-dar for the contribution!

PR #686, opened 2019-02-08
fix(get): Apply format to get when using a selector

dustmop:
Example: `qri get commit.title me/my_ds --format yaml`


lib/datasets_test.go
```diff
@@ -384,103 +385,118 @@ func TestDatasetRequestsGet(t *testing.T) {
...
-				continue
-			}
+func setDatasetName(ds *dataset.Dataset, name string) *dataset.Dataset {
+	parts := strings.Split(name, "/")
+	ds.Peername = parts[0]
+	ds.Name = parts[1]
+	return ds
+}
+
+func componentToString(component interface{}, format string) string {
```
b5: just thinking out loud here, but `ComponentToString` might make a nice exported function on `github.com/qri-io/dataset/dstest`. If we come across a second use for this let's do it.

lib/datasets_test.go
```diff
@@ -384,103 +385,118 @@ func TestDatasetRequestsGet(t *testing.T) {
...
+		if err != nil {
+			return err.Error()
 		}
+		return string(bytes)
+	default:
+		return "Unknown format"
+	}
+}
+
+func bodyToString(component interface{}) string {
```
b5: ditto re: `ComponentToString`

lib/lib.go
```diff
@@ -13,7 +13,7 @@ import (
 var log = golog.Logger("lib")
 
 // VersionNumber is the current version qri
-const VersionNumber = "0.7.0"
+const VersionNumber = "0.7.1-dev"
```
b5: 🙏 tyty

lib/datasets_test.go
```diff
@@ -384,103 +385,118 @@ func TestDatasetRequestsGet(t *testing.T) {
...
+		if err != nil {
+			if err.Error() == "EOF" {
+				break
+			}
+			t.Fatal(err.Error())
+		}
+		moviesBody = append(moviesBody, ent.Value)
+	}
+
 	cases := []struct {
```
b5: muuuuuuuch better. in a later PR we should add tests for the only missing branches of the `Selector ==` if statement: `transform.script` and `viz.script`. Thankfully these tests are doing something meaningful :)

PR #687, opened 2019-02-11
feat(lib.Datasets.Save): add force flag to skip empty-commit checking

b5:
adds a new option to `lib.DatasetRequests().Save()` that forces new commit creation even if `no meaningful changes` are detected.


b5:
Many thanks to @WaughB, who pointed out the need for this feature!

PR #688, opened 2019-02-11
fix(export): Overhaul export command

dustmop:
Many changes to the export command to get it working more closely in line with the recent rfc https://github.com/qri-io/rfcs/blob/master/text/0021-export_behavior.md.

Changes include:

* Export as a single file yaml
* Respect `use` when exporting
* Derive the filename when -o flag isn't used, using the repo name and timestamp
* Derive the format if -o is used, by looking at the file extension
* Calculate absolute path for where export should write, without uglyfying console output
* Introduce new base.ReadEntries utility for turning EntryReader into body
* Switch default export format to single file json

Some things that are broken now, to be fixed later:

* --zipped is broken for now
* Raw block "native" export format
* Needs tests

### DIFF
--- a/base/entries.go
+++ b/base/entries.go
@@ -1,16 +1,13 @@
 package base
 
 import (
-	"fmt"
-
 	"github.com/qri-io/dataset/dsio"
 )
 
 // ReadEntries reads entries and returns them as a native go array or map
-func ReadEntries(reader dsio.EntryReader, all bool, limit int, offset int) (interface{}, error) {
+func ReadEntries(reader dsio.EntryReader) (interface{}, error) {
 	obj := make(map[string]interface{})
 	array := make([]interface{}, 0)
-	numRead := 0
 
 	tlt, err := dsio.GetTopLevelType(reader.Structure())
 	if err != nil {
@@ -25,39 +22,15 @@ func ReadEntries(reader dsio.EntryReader, all bool, limit int, offset int) (inte
 			}
 			return nil, err
 		}
-		if !all && i < offset {
-			continue
-		}
-
 		if tlt == "object" {
 			obj[val.Key] = val.Value
 		} else {
 			array = append(array, val.Value)
-		}
-
-		numRead++
-		if !all && numRead == limit {
-			break
 		}
 	}
 
 	if tlt == "object" {
 		return obj, nil
 	}
-	return array, nil
-}
-
-// ReadEntriesToArray reads entries and returns them as a native go array
-func ReadEntriesToArray(reader dsio.EntryReader, all bool, limit int, offset int) ([]interface{}, error) {
-	entries, err := ReadEntries(reader, all, limit, offset)
-	if err != nil {
-		return nil, err
-	}
-
-	array, ok := entries.([]interface{})
-	if !ok {
-		return nil, fmt.Errorf("cannot convert top-level to array")
-	}
-
 	return array, nil
 }
--- a/base/entries.go
+++ b/base/entries.go
@@ -1,0 +1,63 @@
+package base
+
+import (
+	"fmt"
+
+	"github.com/qri-io/dataset/dsio"
+)
+
+// ReadEntries reads entries and returns them as a native go array or map
+func ReadEntries(reader dsio.EntryReader, all bool, limit int, offset int) (interface{}, error) {
+	obj := make(map[string]interface{})
+	array := make([]interface{}, 0)
+	numRead := 0
+
+	tlt, err := dsio.GetTopLevelType(reader.Structure())
+	if err != nil {
+		return nil, err
+	}
+
+	for i := 0;; i++ {
+		val, err := reader.ReadEntry()
+		if err != nil {
+			if err.Error() == "EOF" {
+				break
+			}
+			return nil, err
+		}
+		if !all && i < offset {
+			continue
+		}
+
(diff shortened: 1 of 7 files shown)


cmd/export.go
```diff
@@ -28,9 +27,6 @@ new dataset, use --blank.`,
 		Example: `  # export dataset
   qri export me/annual_pop
 
-  # export without the body of the dataset
```
b5: nice catch on the docs side

base/entries_test.go
```diff
@@ -0,0 +1,134 @@
...
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/qri-io/dataset"
+	"github.com/qri-io/dataset/dsio"
+)
+
+func TestReadEntriesArray(t *testing.T) {
```
b5: I've added some happy-path testing here. Knowing you're not a fan of table tests, I thought we'd try adding a "description" to the anon struct to help some of our issues with table tests.

Many of our table tests tend to take on this shape:
```go
cases := []struct{
}{
   // test this type of thing
  {},
  {},

  // test that type of thing
  {},
  {},
}
```
Not knowing _what a test case is checking_ is causing us problems. I think all new table tests should be required to include a description field that annotates what that case is testing, and we should use it to get around the problematic output of `case 15: ...`.

cc @ramfox

base/entries_test.go
```diff
@@ -0,0 +1,134 @@
...
+	if expLen != len(got) {
+		t.Errorf("entry length mismatch. expected: %d, got: %d", expLen, len(got))
+	}
+
+	if _, err := ReadEntriesToArray(newTestJSONObjectReader(1), false, 1, 0); err == nil {
+		t.Errorf("expected reading object to error")
+	}
+}
+
+// newTestJSONArrayReader creates a dsio.EntryReader with a number of entries that matches entryCount
```
b5: ideally these helper functions can be rolled into some of the work you're doing on overhauling our testing setup. feel free to move them in subsequent PR's

base/entries.go
```diff
@@ -0,0 +1,63 @@
+package base
+
+import (
+	"fmt"
+
+	"github.com/qri-io/dataset/dsio"
+)
+
+// ReadEntries reads entries and returns them as a native go array or map
```
b5: Huge fan of moving this into one spot. Lovely de-duplication. In the long run we may want to put this directly in the dsio package. This is a great start.

base/entries.go
```diff
@@ -0,0 +1,63 @@
...
+	array := make([]interface{}, 0)
+	numRead := 0
+
+	tlt, err := dsio.GetTopLevelType(reader.Structure())
+	if err != nil {
+		return nil, err
+	}
+
+	for i := 0;; i++ {
+		val, err := reader.ReadEntry()
```
b5: think we can get away with using a `dsio.PagedReader` here?

base/entries.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	if tlt == "object" {
+		return obj, nil
+	}
+	return array, nil
+}
+
+// ReadEntriesToArray reads entries and returns them as a native go array
+func ReadEntriesToArray(reader dsio.EntryReader, all bool, limit int, offset int) ([]interface{}, error) {
```
b5: On the other hand I'm not a fan of this. I think this is prone to misuse, or leads the reader to falsely assume we don't/rarely return objects. 

Since this is only being used in testing, I think we should change this to a `Must` function, because it seems the main purpose for using this is within the context of tests:
```go
// MustSliceEntries asserts the returned entries are a slice of interface{}
// it panics if any other value or a non-nil error is returned
// this is mainly intended for use in tests
func MustSliceEntries(ent interface{}, err error) []interface{} {
  if err != nil {
    panic(err)
  }
 entSlice, ok := ent.([]interface{})
  if !ok {
    panic("entries is not of type []interface{}")
  } 
  return entSlice
}
```
use it like this: 
```go
  ent := MustSliceEntries(ReadEntries(rdr, all, limit, offset))
```

lib/export.go
```diff
@@ -41,22 +41,40 @@ func NewExportRequests(node *p2p.QriNode, cli *rpc.Client) *ExportRequests {
 
 // ExportParams defines parameters for the export method
 type ExportParams struct {
-	Ref     repo.DatasetRef
-	RootDir string
-	PeerDir bool
-	Format  string
+	Ref       string
```
b5: 👌 we should chat at some point about normalizing the name of this field across all lib input structs

base/entries_test.go
dustmop: I like using a description much better. It makes it possible to Ctrl+F for the failing test case, which `case 15` doesn't allow for.

base/entries.go
dustmop: Even better, I think I can just move the PagedReader to base/render, the only place that needs to pass (all, limit, offset) to ReadEntries. It ends up making ReadEntries much simpler, though the test cases now become extremely trivial.

base/entries.go
dustmop: Good call. Moved the function to lib/datasets_test (where it is needed) and called it `mustByArray`.

lib/export.go
dustmop: 👍

PR #693, opened 2019-02-13
fix(export): Export --zip flag

dustmop:
Reenable the --zip flag. If the --zip flag is not given, the default format is "json", but if it is given then the default format is "zip". If --zip is combined with a format aside from the "zip", then that format is used but the single-file dataset is inside of a zip file.


lib/export.go
```diff
@@ -102,11 +110,12 @@ func (r *ExportRequests) Export(p *ExportParams, fileWritten *string) (err error
...
-
-		if p.Format == "" {
+		// If format was not supplied as a flag, and we're not outputting a zip, derive format
+		// from file extension.
+		if p.Format == "" && !p.Zipped {
 			format = ext
 		}
-
+		// Make sure the format doesn't contradict the file extension.
 		if ext != format {
```
b5: v cool.

dustmop:
Added api functionality and also fixed the cmd_test integration test while I was at it.

PR #695, opened 2019-02-20
fix(get): Get returns 404 if dataset is not found

dustmop:


dustmop:
Depends upon https://github.com/qri-io/registry/pull/27. Fixes 2/4 parts of https://github.com/qri-io/frontend/issues/464.

api/registry.go
```diff
@@ -131,6 +132,11 @@ func (h *RegistryHandlers) registryDatasetHandler(w http.ResponseWriter, r *http
 
 	err = h.RegistryRequests.GetDataset(&ref, res)
 	if err != nil {
+		// If error was that the dataset wasn't found, send back 404 Not Found.
+		if strings.HasPrefix(err.Error(), "error 404") {
```
b5: we should make this a more robust thing later, like, you know, returning proper status codes :/

PR #697, opened 2019-02-25
fix(list): Flag --num-versions shows number of versions of each dataset

dustmop:
Depends upon https://github.com/qri-io/dataset/pull/177.


cmd/list.go
```diff
@@ -59,11 +60,12 @@ must have ` + "`qri connect`" + ` running in a separate terminal window.`,
...
-	Limit     int
-	Offset    int
-	Peername  string
-	Published bool
+	Format      string
+	Limit       int
+	Offset      int
+	Peername    string
+	Published   bool
+	NumVersions bool
```
b5: I'd prefer this be `ShowNumVersions`, to me `NumVersions` sounds like it should be an integer value for "(max) number of versions to list".

dustmop: Sounds good, I was originally going with `ShowVersions`, but switched to `NumVersions`, though `ShowNumVersions` seems best. I still think it's right to keep the flag as `--num-versions` and `-n`, since that is more in line with existing program flags (I'm thinking of `grep -n`, for example).

PR #698, opened 2019-02-27
feat: overhauled diff command

b5:
Phew, ok this one was, well hard. But we now haz a real differ

This PR doesn't cover cbor, yaml, or xlsx. All of which require further thought, but you can diff all sorts of other stuff.

Some things to think about:
* The output of diff isn't really super readable / helpful yet. We need to have a big conversation about how to represent this kind of diff. I'd like to loop others into this chat
* Large diffs currently just flood the screen, it'd be great to have some heuristic for detecting "too much diff", and a flag to say "yes plz I want too much diff"
* `--format json` is there to help understand the raw diff. Piping to `jq` is good. can help understand what's going on eg: `qri diff file.json me/dataset --format json | jq`
* I'd welcome suggestions for alternatives to the word "elements", "values"? Dunno, let's discuss
* ~~We should also think about warning people that this is experimental, and we haven't verified the output of qri diff in lots of places yet.~~ _added_

Lots more, but this is a good start.

** **

TODO:
* [x] update app documentation
* [x] API endpoint
* [x] API endpoint tests
* [x] update existing tests
* [x] add cmd tests
* [x] add lib tests
* [x] describe feature properly in this PR so we can ship documentation elsewhere


lib/diff.go
```diff
@@ -0,0 +1,191 @@
...
+	if *left == "" && *right == "" {
+		refs := []repo.DatasetRef{}
+		if err = DefaultSelectedRefs(node.Repo, &refs); err != nil {
+			return
+		}
+		switch len(refs) {
+		case 0:
+			// do nothing
+		case 1:
+			// if one ref is speficied, compare with previous version, so we'll put it in right
```
ramfox: 'if one ref is speficied' => 'if one ref is specified'

PR #699, opened 2019-02-28
feat(fill_struct): Json and yaml deserialization rewrite

dustmop:
Add a custom routine FillStruct for deserialization. This will used for parsing dataset json and yaml files when saving. This solves 4 problems:

* Field names are now case insensitive
* Unknown field names (such as misspellings) are now an error
* That is, unless the target struct is able to store arbitrary fields (Meta)
* Allow for saving individual components, by checking the "qri" field

### DIFF
--- a/base/fill_struct.go
+++ b/base/fill_struct.go
@@ -22,6 +22,7 @@ func FillStruct(fields map[string]interface{}, output interface{}) error {
 type KeyValSetter interface {
 	SetKeyVal(string, interface{}) error
 }
+
 // TODO (dlong): Implement this interface for dataset.Meta. It currently has the similar method
 // `Set`, which does more than needed, since it assigns to any field, not just the private map.
 
@@ -76,6 +77,11 @@ func putFieldsToTargetStruct(fields map[string]interface{}, target reflect.Value
 			num, ok := val.(int)
 			if ok {
 				field.SetInt(int64(num))
+				continue
+			}
+			numFloat, ok := val.(float64)
+			if ok {
+				field.SetInt(int64(numFloat))
 				continue
 			}
 			errs = append(errs, fmt.Sprintf("field %s type int, value %s", fieldName, val))
--- a/base/fill_struct.go
+++ b/base/fill_struct.go
@@ -1,0 +1,195 @@
+package base
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"time"
+)
+
+// FillStruct fills in the values of an arbitrary structure using an already deserialized
+// map of nested data. Fields names are case-insensitive. Unknown fields are treated as an
+// error, *unless* the output structure implementes the KeyValSetter interface.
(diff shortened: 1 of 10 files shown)


lib/file.go
```diff
@@ -122,6 +127,39 @@ func ReadDatasetFile(path string) (ds *dataset.Dataset, err error) {
 	return
 }
 
+func fillDatasetOrComponent(fields map[string]interface{}, path string, ds *dataset.Dataset) error {
+	switch fields["qri"] {
```
b5: This really needs to be documented better, but switching on the qri field ("component kinds"), should be examining only the first two characters of kind: https://github.com/qri-io/dataset/blob/f2ced13e41b51fe4f2549d8ed6d585a871659333/kind.go#L13

First, just to formalize the spec, the `qri` field, if present MUST have a value in the form:
```
component_acronym ':' version_number
```
where `component_acronym` MUST be a unique two character identifier for the type of component, and `version_number` is an unsigned integer.

The version number might change, and maintaining code is irritating. So how about:

```go
func fillDatasetOrComponent(fields map[string]interface{}, path string, ds *dataset.Dataset) (err error) {
	var fill interface{}
	fill = ds

	if kindStr, ok := fields["qri"].(string); ok && len(kindStr) > 3 {
		switch kindStr[:2] {
		case "md":
			ds.Meta = &dataset.Meta{}
			fill = ds.Meta
		case "cm":
			ds.Commit = &dataset.Commit{}
			fill = ds.Commit
		case "st":
			ds.Structure = &dataset.Structure{}
			fill = ds.Structure
		}
	}

	if err = base.FillStruct(fields, fill); err != nil {
		return err
	}
	absDatasetPaths(path, ds)
	return nil
}
```

I'm not a massive fan of the `absDatasetPaths(path,ds)` at the end of this func, but after reading, it's clear this is the best place for it currently, so let's keep it.

base/fill_struct.go
```diff
@@ -0,0 +1,243 @@
+package base
```
b5: Fill struct is really good. So good that I think it needs to be it's own package. What do you think of moving the fill_struct code to `qri-io/dataset/fill`? The only exported function could be renamed from `FillStruct` to `Struct`, (package consumers would call `fill.Struct`.

This pattern has legs, and I think it's already forced a number of important conversations about adjacent interface decisions like paths vs. keys; case sensitivity; the role of fill struct in the serialization chain; etc. All of this could be documented better if `fill_struct` had it's own package.

dustmop: Discussed offline, I feel like fill_struct has some UI level concerns with how it handles case-sensitivity and unknown fields, that make it a bit more specific to qri the app than just dataset. Also, it could be used for config.Config, or other structs that are separate from dataset. It could be its own package, or a subpackage of `base`, which @b5 will probably do soon.

PR #704, opened 2019-03-12
feat(remote): Beginning of "remote" mode implementation

dustmop:
On the remote side, the following is being added:
* The flag --remote-mode which enables remote mode
* An api /dataset that can be posted to, only in remote mode
* Receive method that calculates size of dag, and fails if too large

On the client side:
* The `qri publish` command has a "--remote" to control where to publish
* Publishing to a remote sends a dag.Info instead of dataset head

Missing:
* Remote mode should act like read-only
* dag.Info is a stub for now
* `qri publish --remote ...` is ignoring name of remote
* dSync when a push to a remote succeeds


api/remote.go
```diff
@@ -0,0 +1,52 @@
...
+// NewRemoteHandlers allocates a RemoteHandlers pointer
+func NewRemoteHandlers(node *p2p.QriNode) *RemoteHandlers {
+	req := lib.NewRemoteRequests(node, nil)
+	return &RemoteHandlers{*req}
+}
+
+// ReceiveHandler is the endpoint for remotes to receive daginfo
+func (h *RemoteHandlers) ReceiveHandler(w http.ResponseWriter, r *http.Request) {
+	switch r.Method {
+	case "POST":
```
b5: most other handlers include an `OPTIONS` select case for CORS requests that just sends back an empty ok handler, I think it's actually the right decision to _not_ have that here, just wanted to point out why it exists elsewhere in case we end up doing publish-to-remote from a brower.

lib/remote.go
```diff
@@ -0,0 +1,131 @@
+package lib
```
b5: where my test at? Even a basic outline would be great.

api/remote.go
```diff
@@ -0,0 +1,52 @@
+package api
```
b5: some very light testing here would be nice

dustmop:
Added tests

PR #705, opened 2019-03-12
feat(daginfo): add `qri daginfo` command that returns a dag.Info of a dataset

ramfox:
1) adds DAGInfo.go file to cmd
2) adds base NewDAGInfo, that you give  a context, store, nodegetter, and id, and it returns a labeled DAGInfo with the Manifest of the dataset, the sizes of each node. The labels correspond to the different dataset components: body, meta, structure, viz, transform, commit.
3) adds action and lib functions to wire the `qri daginfo` command together

Expanded command to include:
`qri daginfo get [label] <datasetRef>` to get the sub Info at a specific label.

And added a default view that returns the ref, total size, block count, & labels with their sizes
relies on qri-io/dag/pull/7 and will fail until it is merged

Edit:
Refactored, based on discussion, to merge `qri manifest` and `qri daginfo` command into single `qri dag` command:

`qri dag manifest get`
`qri dag manifest missing`
`qri dag info`


b5:
you might just need a rebase to make CI happy. codecov isn't going to like this 😄

Makefile
```diff
@@ -66,6 +66,7 @@ update-qri-deps: require-gopath
 	cd $$GOPATH/src/github.com/qri-io/startf && git checkout master && git pull
 	cd $$GOPATH/src/github.com/qri-io/starlib && git checkout master && git pull
 	cd $$GOPATH/src/github.com/qri-io/dag && git checkout master && git pull
+	cd $$GOPATH/src/github.com/qri-io/ioes && git checkout master && git pull
```
b5: nice catch

cmd/DAGInfo.go
```diff
@@ -0,0 +1,159 @@
+package cmd
+
+import (
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	// "io/ioutil"
```
b5: drop these commented bits

cmd/DAGInfo.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/ghodss/yaml"
+	"github.com/qri-io/dag"
+	"github.com/qri-io/ioes"
+	"github.com/qri-io/qri/lib"
+	"github.com/spf13/cobra"
+)
+
+// NewDAGInfoCommand creates a new `qri daginfo` command that generates a daginfo for a given
+// dataset reference. Referenced dataset must be stored in local CAFS
+func NewDAGInfoCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
```
b5: I was going to ask for tests on this, but I think we should first consider merging `qri manifest` and `qri daginfo` into one master `qri dag` command. so:

`qri dag info`
`qri dag manifest`

So I think we should merge this & follow up with a quick PR that does that, and test _that_ command instead

base/manifest.go
```diff
@@ -18,3 +20,69 @@ func NewManifest(ctx context.Context, ng ipld.NodeGetter, path string) (*dag.Man
 
 	return dag.NewManifest(ctx, ng, id)
 }
+
+// NewDAGInfo generates a DAGInfo for a given node
+func NewDAGInfo(ctx context.Context, store cafs.Filestore, ng ipld.NodeGetter, path string) (*dag.Info, error) {
```
b5: def gonna need at least a light test on this one

lib/datasets.go
```diff
@@ -625,3 +625,54 @@ func (r *DatasetRequests) ManifestMissing(a, b *dag.Manifest) (err error) {
 	*b = *mf
 	return
 }
+
+// DAGInfo generates a dag.Info for a dataset path
+func (r *DatasetRequests) DAGInfo(refstr *string, i *dag.Info) (err error) {
```
b5: also need a test for this

lib/datasets.go
```diff
@@ -625,3 +625,54 @@ func (r *DatasetRequests) ManifestMissing(a, b *dag.Manifest) (err error) {
...
+	return
+}
+
+// SubDAGParams defines parameters for the SubDAGInfo method
+type SubDAGParams struct {
+	RefStr, Label string
+}
+
+// SubDAGInfo generates a dag.Info for a sub DAG at a label in the original dag.Info
+func (r *DatasetRequests) SubDAGInfo(s *SubDAGParams, i *dag.Info) error {
```
b5: hrmmm... possible to have the main `DAGInfo` command accept what is currently `SubDAGParams` and if the "Label" field is empty just give the main `DAGInfo`? it would remove the need for this method entirely

ramfox: yea I thought about this, and actually started doing it that way at first. Then I figure it would be easier to condense it once I got it working separately. Was hoping for this exact feedback.

cmd/DAGInfo.go
ramfox: I really like this!

lib/datasets.go
b5: totally. I find myself writing separate implementations & re-combining them after I understand both all the time. I think it's a healthy pattern especially given that if it's not caught on the first pass, subsequent refactoring usually uncovers this same conclusion.

cmd/dag.go
```diff
@@ -172,3 +203,66 @@ func (o *ManifestOptions) Missing() error {
...
+			if !o.Pretty {
+				buffer, err = json.Marshal(info)
+			} else {
+				buffer, err = json.MarshalIndent(info, "", " ")
+			}
+		case "yaml":
+			buffer, err = yaml.Marshal(info)
+		// case "cbor":
+		// 	buffer, err = info.MarshalCBOR()
+		default:
```
dustmop: Could you change this to `case "":`, and add a default case that returns an error that does `fmt.Errorf("unknown format %s", o.InfoFormat)`. This way we'll catch spelling mistakes like `jsonn`, treating them as errors instead of the default format.

b5:
... or at your earliest convenience

PR #706, opened 2019-03-13
fix(save): Better error message when saving with wrong file

dustmop:


b5:
This is delightful. I'm just gonna merge it

PR #707, opened 2019-03-13
feat(remote): Parse config using FillStruct. Remotes field.

dustmop:
Parse the config using FillStruct. Fix numerous bugs found while supporting the config struct. Add a Remotes component to the config, which uses ArbitrarySetter to support arbitrary remote names.


base/fill_struct.go
```diff
@@ -23,9 +23,9 @@ type ArbitrarySetter interface {
 	SetArbitrary(string, interface{}) error
 }
 
-// timeObj and ifaceObj are used for reflect.TypeOf
+// timeObj and strObj are used for reflect.TypeOf
 var timeObj time.Time
-var ifaceObj interface{}
+var strObj string
```
b5: I  prefer multiple var declarations grouped by parenthesis. pedantic, but just to keep it consistent:
```go
var (
  // timeObj and strObj are use for reflect.TypeOf
  timeObj time.Time
  strObj string
)
```

config/config.go
```diff
@@ -88,20 +89,6 @@ func (cfg Config) SummaryString() (summary string) {
 	return summary
 }
 
-// ReadFromFile reads a YAML configuration file from path
-func ReadFromFile(path string) (cfg *Config, err error) {
```
b5: I'm assuming this is removed to require users to use fill_struct, it'd be great to know how we _would_ read from a file using fill struct, even if just as a comment on this PR. Totally fine to see this deleted

config/remotes.go
```diff
@@ -0,0 +1,29 @@
+package config
+
+import (
+	"fmt"
+)
+
+// Remotes encapsulates configuration options for remotes
+type Remotes struct {
+	nameMap map[string]string
```
b5: because nameMap is unexported this field won't be visible to encoding packages like yaml & json, which will mean this data isn't written to disk (it's stored in yaml). We could refactor to just be:
```go
type Remotes map[string]string
```
and adjust `SetArbitrary` to alter `r *Remotes` values instead of `r.nameMap`.

lib/config.go
```diff
@@ -28,21 +30,29 @@ var SaveConfig = func() error {
 
 // LoadConfig loads the global default configuration
 func LoadConfig(streams ioes.IOStreams, path string) (err error) {
-	var cfg *config.Config
-	cfg, err = config.ReadFromFile(path)
+	var data []byte
```
b5: maybe the changes here could be moved into `config.ReadFromFile` instead of throwing it out, that way we have  a clean package API that properly calls fill_struct.

but I think that would create a circular dependency between `base` & `config`, so we'd need to make fill struct it's own package for that to work. If that's right, no need to do this now.

dustmop: My first approach was doing this in config.ReadFromFile, but yeah it is a circular dependency, because config/ is actually below base/. This implies that fill_struct may need to move to its own package, but I agree we should do it later.

Issue #690, opened 2019-02-11
add flags to `qri save` for setting individual components of a save


b5:
I'm having a semi-frustrating time always having to create a `dataset.yaml` or `dataset.json` just to update a transform script. What I'd like to be able to do from the CLI is one of two things. The first would be providing json/yaml snippits for each component. In each of these cases the `_file` argument would be a file that _only defines that component_, not an entire dataset:
```
$ qri save --meta meta_file.json me/dataset
$ qri save --structure structure_file.json me/dataset

$ qri save --viz viz_file.yaml me/dataset
$ qri save --transform transform_file.json me/dataset

# we already support this:
$ qri save --body file.csv me/dataset
```

The second would be providing a raw script as a file as a shortcut for "default settings, but with this file":
`qri save --transform tf_file.star me/dataset`
ditto for viz scripts:
`qri save --viz template_file.html me/dataset`
This would just do the right thing, populating a component with defaults (like setting `syntax` to `html` or `starlark`) and setting the `scriptPath` to the supplied path argument

In a later PR it'd be great to leverage the `qri` values we embed into datasets to intelligently accept a dataset.json definition, but if passed to a specific component argument, we would traverse to and only use that component of the supplied dataset. For example this:
```
qri save --meta other_dataset.json me/dataset
```
would be the same as saying "I've exported some other dataset as `other_dataset.json`, please copy it's `meta` component onto `me/dataset`. This is future work. For now it'd be great if we errored if you passed a file that had `"qri": "ds"` in the top level to any component argument

dustmop:
I'm not a huge of having a flag for each component, leads to too many flags that are similar things but are also mutually exclusive. I would prefer `--component <component_name>`. Both --file and --body are somewhat special in how they work, for example, either can be used to create a new dataset, while --component couldn't be.

b5:
In the alternative you're outlining I would do `qri save --component structure --file st_file.json`? Just making sure.

Agreed on the many/too many flags issue. I want to make sure I understand the alternative, but for I can see this adding up to overloading the `--file` flag a whole bunch instead of flags for each component:
```
# do what I was mentioning above, inferring this is a default transform
$ qri save --file transform.star me/dataset

# skip the whole --component thing, instead let users specify components with "qri" : "st" top level values
$ qri save --file component_file.json me/dataset
```

This would have a few implications:
* data, script, and viz file extensions would need to be mutually exclusive
[...]

PR #718, opened 2019-03-25
 feat(save): Save a dataset with multiple file arguments

dustmop:


dustmop:
Fixes https://github.com/qri-io/qri/issues/690.

lib/file.go
```diff
@@ -58,51 +58,77 @@ func pathKind(path string) string {
...
 }
 
-// ReadDatasetFile decodes a dataset document into a Dataset
-func ReadDatasetFile(path string) (ds *dataset.Dataset, err error) {
-	var (
-		resp *http.Response
-		f    *os.File
-		data []byte
-	)
+// ReadDatasetFiles decodes a dataset document into a Dataset
```
b5: Outdated comment. It'd be great if this function comment broke down some of the nuances of reading multiple filepaths into a dataset, keeping it as high-level as possible. Something like this (adjusting for accuracy):

```go
// ReadDatasetFiles combines resources at zero or more filepaths, deserializing them into a
// *dataset.Dataset. Files can separately specify different components of a dataset
// but two provided components cannot overlap (modify the same component)
```

lib/file.go
```diff
@@ -58,51 +58,77 @@ func pathKind(path string) string {
...
+func ReadDatasetFiles(pathList []string) (*dataset.Dataset, error) {
```
b5: I'd prefer 
```go
func ReadDatasetFiles(pathList ... string) (*dataset.Dataset, error)
```
To me ReadDatasetFiles would be very useful when using qri-as-a-library by hiding the `[]string` from the caller:
```go
//  variadic arguments allows calls for easier single-file loading:
ds, err := lib.ReadDatasetFiles("./path.json")

//  and chaining that doesn't require []string allocation:
ds, err := lib.ReadDatasetFiles("./structure.json", "./meta.yaml")
```

cmd/save.go
```diff
@@ -57,7 +57,7 @@ commit message and title to the save.`,
 		},
 	}
 
-	cmd.Flags().StringVarP(&o.FilePath, "file", "f", "", "dataset data file (yaml or json)")
+	cmd.Flags().StringSliceVarP(&o.FilePaths, "file", "f", nil, "dataset or component file (yaml or json)")
```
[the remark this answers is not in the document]
dustmop: Added, along with an all-in-one combination of meta + transform + viz.

lib/file.go
dustmop: Good idea, this is much better.

lib/file.go
dustmop: Done.

PR #719, opened 2019-03-26
feat(remote): Send actual dag.Info to remote, perform dsync

dustmop:


api/api.go
```diff
@@ -230,6 +235,11 @@ func NewServerRoutes(s *Server) *http.ServeMux {
 	if s.cfg.API.RemoteMode {
```
b5: I'd love to get a `log.Info` that notifies a user this server is operating with remote mode enabled? Should probably be in `api.Server.Serve`

api/api.go
```diff
@@ -265,3 +275,31 @@ func NewServerRoutes(s *Server) *http.ServeMux {
...
+	if err != nil {
+		return nil, err
+	}
+	bapi := capi.Block()
+
+	return dsync.NewReceivers(context.Background(), ng, bapi), nil
+}
+
+// newIPFSCoreAPI constructs a CoreAPI (part of ipfs) from a qri p2p node
+// copied from qri/actions/dag.go
```
b5: opened #720 to deal with this. Let's get this merged then refactor away the need for this func

api/remote_test.go
```diff
@@ -12,6 +14,13 @@ func TestRemoteHandlers(t *testing.T) {
 		{"POST", "/", mustFile(t, "testdata/postRemoteRequest.json")},
 	}
 
+	// Reject all dag.Info's
+	lib.Config.API.RemoteAlwaysAccept = false
```
b5: Hrm, I'm wondering if we can adjust `remoteAlwaysAccept` to instead be `remoteAcceptSizeMax`, and only allow either 0 (`remoteAlwaysAccept == false`) or -1 (`remoteAlwaysAccept == true`). We could later use that to replace the `allowedDagSize` const, but let's do that as a follow up.

cmd/connect.go
```diff
@@ -54,6 +54,7 @@ peers & swapping data.`,
 	cmd.Flags().BoolVarP(&o.Setup, "setup", "", false, "run setup if necessary, reading options from environment variables")
 	cmd.Flags().BoolVarP(&o.ReadOnly, "read-only", "", false, "run qri in read-only mode, limits the api endpoints")
 	cmd.Flags().BoolVarP(&o.RemoteMode, "remote-mode", "", false, "run qri in remote mode")
+	cmd.Flags().BoolVarP(&o.RemoteAlwaysAccept, "remote-always-accept", "", false, "when running as  a remote, accept all datasets sent")
```
b5: Hrm, I'm wondering if we can adjust `remoteAlwaysAccept` to instead be `remoteMaxSizeAccept`, and only allow either 0 (`remoteAlwaysAccept == false`) or -1 (`remoteAlwaysAccept == true`). We could later use that to replace the `allowedDagSize` const, but let's do that as a follow up.

We're planning to follow up in the future with additional configuration details that will invalidate assumption suggested by "always accept" (because additional things will be able to reject a `dsync`). I'd love to avoid users complaining "but I set `remoteAlwaysAccept` to `true`, shouldn't it _always accept_?"

lib/remote.go
```diff
@@ -76,58 +83,105 @@ func (r *RemoteRequests) PushToRemote(p *PushParams, out *bool) error {
...
 	if err != nil {
 		return err
 	}
 
-	fmt.Printf("Received dag.Info:\n")
-	fmt.Printf(p.Body)
-	fmt.Printf("\n\n")
+	// TODO(dlong): Customization for how to decide to accept the dataset.
+	if !Config.API.RemoteAlwaysAccept {
+		*reason = "not accepting any datasets"
```
b5: If you're ok with the max size adjustment I think this message should stay the exact same if `remoteMaxSizeAccept == 0`

cmd/connect.go
dustmop: I was considering this flag to be temporary, as we're going to have better configuration in the future, and this whole entire feature is in flux for now. But I also agree that `remoteMaxSizeAccept` is better and more accurately reflects what the code is doing now. I'll switch to that.

PR #722, opened 2019-04-01
feat(remote): Retrieve sessionID & diff first, pass both to dsync

dustmop:


lib/remote.go
```diff
@@ -87,7 +91,27 @@ func (r *RemoteRequests) PushToRemote(p *PushParams, out *bool) error {
 		return fmt.Errorf("error code %d: %v", res.StatusCode, rejectionReason(res.Body))
 	}
 
-	// TODO(dlong): Get a dsync session id from the remote, use it to perform dsync.
+	bodyBytes, err := ioutil.ReadAll(res.Body)
```
b5: Hrm, I think this can be refactored to be a little more concise using an anonymous struct, and leveraging the fact that this is a JSON API:

```go
env := struct{ Data ReceiveResult }{}
if err := json.NewDecoder(res.Body).Decode(&env); err != nil {
  return err
}
res.Body.Close()
```

I know I know you made all those changes to fill.Struct. Those changes are still super great and I'm delighted they're in.

b5:
🍡 🍡

Issue #716, opened 2019-03-23
Problem with starlib/xlsx and dataset/dsio - Manjaro with Go 1.12


wblondel:
Hi,

I can't build qri on Manjaro with Go 1.12 (go version go1.12 linux/amd64).

First, when I do `go get github.com/qri-io/qri`, I have these errors:
```
# github.com/qri-io/starlib/xlsx
qri-io/starlib/xlsx/xlsx.go:96:8: assignment mismatch: 1 variable but f.xlsx.GetRows returns 2 values
# github.com/qri-io/dataset/dsio
qri-io/dataset/dsio/xslx.go:70:23: too many arguments in call to r.decode
	have ([]string, error)
	want ([]string)
```

Then, when I `make build`, I have these errors:
```
\n4/5 install gx dep-packages:\n
go get github.com/qri-io/registry/regclient github.com/qri-io/dag github.com/qri-io/qfs github.com/qri-io/startf
# github.com/qri-io/dataset/dsio
../dataset/dsio/xslx.go:70:23: too many arguments in call to r.decode
	have ([]string, error)
	want ([]string)
# github.com/qri-io/starlib/xlsx
../starlib/xlsx/xlsx.go:96:8: assignment mismatch: 1 variable but f.xlsx.GetRows returns 2 values
make: *** [Makefile:48: build] Error 2
```

How can I solve this? I don't know what to do.

Thanks!

b5:
Hi @wblondel, thanks so much for filing! 
`make build` is definitely the way to go (regular `go install` isn't expected to work).

It looks like one of our downstream dependencies changed their API. I'm working on a fix, and will update here.

b5:
ooook, just merged two commits that should address the issue. Would you mind running two commands for me? `make update-qri-deps` followed by `make build`

wblondel:
Success! It works :+1: Great job :)

b5:
Great! Thanks for the heads up, I'll close this

PR #724, opened 2019-04-02
feat(render): overhauled qri render, implement RFC0011

b5:
OOooOOok, this PR implements [RFC0011](https://github.com/qri-io/rfcs/blob/master/text/0011-html_viz.md).

There are a few other changes worth pointing out:

### Qri viz template syntax has changed
This PR completely breaks the existing viz template api, replacing it with the one the RFC describes. I plan to fully document that API & post it up in the next few days. The majority of it is outlined in the RFC.

### `limit`, `offset` and `all` params are removed from render
Good riddance. Instead templates now bear the full responsibility of determining how much of the dataset body to read. This is a necessary change now that `save` and `update` both execute render by default. Speaking of which:

### `qri save` and `qri update` both generate visualizations by default
Qri now stores an executed version of the template at `index.html`. Get the old behaviour back with `--no-render`.


b5:
@ramfox your issue with xlsx is related to #716, and further proof that we need go modules real bad. I've opened #726 to get us going on a better fix for the problem you ran into

PR #728, opened 2019-04-04
refactor(viz): Move code for adding "default viz" out of InferValues.

dustmop:
Rationale: base's functionality should remain simple and single-purpose, without optional flags to enable or disable code paths. The only place where the `inferViz` value was being used was in the higher-level actions/dataset, which implies that it actually wants to exist in a separate function. Another hint is that the description of `InferValues` states that it "populates any missing fields that must exist to create a snapshot". The viz is not something that *must exist*, as evidenced by the fact that it has an optional flag to turn it off.


b5:
can we also get a v0.7.4-dev tag? first PR since release

PR #732, opened 2019-04-05
refactor(p2p): Give node a CoreAPI accessor to simplify dependencies

### LINKED ISSUE #720: refactor p2p.Node.IPFSNode() to return an ipfs.CoreAPI
Now that CoreAPI is here to stay, we have no reason to pass around an actual IPFS node pointer, let's pass around a CoreAPI instance instead. This'll cut down on a bunch of copy-pasting in packages that import p2p.

dustmop:
Instead of passing around IPFSNode pointers everywhere, add an accessor that directly returns a CoreAPI pointer, and an IPFSNamesys pointer. Simplifies a bunch of call sites. Fixes #720.

### DIFF
--- a/p2p/node.go
+++ b/p2p/node.go
@@ -262,7 +262,7 @@ func (n *QriNode) GetIPFSNamesys() (namesys.NameSystem, error) {
 	return ipfsn.Namesys, nil
 }
 
-// IPFSCoreAPI returns a pointer to the core IPFS API
+// IPFSCoreAPI returns a IPFS API interface instance
 func (n *QriNode) IPFSCoreAPI() (coreiface.CoreAPI, error) {
 	ipfsfs, ok := n.Repo.Store().(*ipfs_filestore.Filestore)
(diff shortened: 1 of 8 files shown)


dustmop:
Depends on https://github.com/qri-io/dag/pull/11.

p2p/node.go
```diff
@@ -242,14 +245,33 @@ func (n *QriNode) echoMessages() {
...
+// GetIPFSNamesys returns a namesystem from IPFS
+func (n *QriNode) GetIPFSNamesys() (namesys.NameSystem, error) {
+	ipfsn, err := n.ipfsNode()
+	if err != nil {
+		return nil, err
+	}
+	return ipfsn.Namesys, nil
+}
+
+// IPFSCoreAPI returns a pointer to the core IPFS API
```
b5:
```suggestion
// IPFSCoreAPI returns a IPFS API interface instance
```

PR #733, opened 2019-04-05
fix(ref): CanonicalizeProfile just handles renames all the time

dustmop:
Handle peername changes all the time, by matching profileID, instead of only doing so if a third parameter was being supplied to CanonicalizeProfile.

### DIFF
--- a/repo/ref_test.go
+++ b/repo/ref_test.go
@@ -1,7 +1,7 @@
 package repo
 
 import (
-	"reflect"
+	//"reflect"
 	"testing"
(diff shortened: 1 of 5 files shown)


repo/ref_test.go
```diff
@@ -1,7 +1,7 @@
 package repo
 
 import (
-	"reflect"
+	//"reflect"
```
b5: destroy

b5:
too good, need merge meow

PR #737, opened 2019-04-10
Lib constructor

b5:


api/api.go
```diff
@@ -42,14 +42,16 @@ func init() {
 type Server struct {
 	// configuration options
 	cfg     *config.Config
+	cfgPath string
```
dustmop: How about adding the path to the Config object, such that it doesn't serialize.

lib/datasets.go
```diff
@@ -20,33 +20,66 @@ import (
...
 	}
+}
 
-	return &DatasetRequests{
-		node: node,
-		cli:  cli,
-	}
+// datasetMethods encapsulates business logic for working with Datasets on Qri
+type datasetMethods struct {
+	cli  *rpc.Client
```
dustmop: Nearly every *Methods struct is going to be wrapping the same state: an rpc.Client, a p2p.Node, and a Config. I think it's worth wrapping them together into a single object that we can pass around a pointer to. Call it MethodState or OperationState or something. Then, these *Methods structs are thin encapsulations over this state, along with a bit of info for method dispatch.

Not a huge fan of Capital Interface with lower-case implementation. Makes it hard to speak aloud about. Maybe DatasetMethods for interface, and datasetOperations / datasetOps for internal struct?

lib/datasets.go
```diff
@@ -412,7 +445,7 @@ func (r *DatasetRequests) SetPublishStatus(p *SetPublishStatusParams, publishedR
 
 	if p.UpdateRegistry && r.node.Repo.Registry() != nil {
 		var done bool
-		rr := NewRegistryRequests(r.node, nil)
+		rr := registryMethods{node: r.node}
```
dustmop: Stream of conscious here...

If these *Methods are wrapping a single object (as mentioned above), this will be easier to do. Can we use the interface returning function instead of the internal one? Or would it be doable if we switched to this single object. I'm picturing this:

```
state := NewMethodState(node, rpc.client, config)
m := NewDatasetMethods(state)
m.SetPublishStatus(...)
...
func (m *datasetOperations) SetPublishStatus(...) {
  ...
  r := NewRegistryMethods(m.state)
}
```

I guess the problem is that we want NewDatasetMethods and NewRegistryMethods to be methods on the top-level Lib object, right? In that case I guess the question becomes whether `state` can implement the Instance interface.

lib/datasets.go
b5: Deep agree here. So far I've been calling that encapsulated state an `Instance`. This PR adjusts method constructor func signatures to accept only an Instance. eg: 
```go
func NewDatasetMethods(inst Instance) DatasetMethods
```

So far I'm liking the way the phrasing shakes out: "derive the methods you need from an instance".

As for the exported-interface unexported-implementation, I will admit to being a huge fan. I've seen it in other codebases (examples escape me at the moment, I think it might have been a stats package...). The way I verbalize it is there's the _Interface_, and a _default implementation_. IMHO the reader can clearly see the intent of the private struct: implement the interface.

b5:
This was fun. fleshing out thoughts is fun. Sleep now, sweet PR. ya done good

PR #740, opened 2019-04-15
feat: lib.NewInstance constructor to create a qri instance

b5:
The big refactor here is removing all global configuration state from lib. Lots of changes here, let's take the review process slow. I'll edit this description with details soon


api/api.go
```diff
@@ -40,16 +40,18 @@ func init() {
 // Server wraps a qri p2p node, providing traditional access via http
 // Create one with New, start it up with Serve
 type Server struct {
+	lib.Instance
```
dustmop: With the understanding that this refactoring is large and only partially present in this PR, I think it would be good to get rid of `cfg` and `qriNode` fields on this struct, and just access them directly on the instance.

lib/config.go
```diff
@@ -5,81 +5,47 @@ import (
...
-	}
-
-	if migrated {
-		return SaveConfig()
-	}
-
-	return nil
+// ConfigMethods encapsulates changes to a qri configuration
+type ConfigMethods struct {
+	Instance
```
dustmop: In my humble opinion, the biggest win we want to accomplish with this change is consolidating all of the state that qri uses into a single struct. Moving the Config from a global object into part of this struct is part of accomplishing this goal. All of our private code (inside and under lib/) should be using this struct and passing it around to other parts of the code. However, I don't think it needs to be an interface that hides the details, since it's only for internal use. With that in mind, I think the right approach looks like this:

```
type instance struct {
  cfg *Config
  node *p2p.QriNode
  ctx Context
}

type ConfigMethods struct {
  instance
}

func (c *ConfigMethods) SomeFunc() {
  c.node.Connect() // or whatever
}
```

When we make a public facing interface for uses of lib/, they can get something that hides these details and doesn't give direct access to these internal fields.

api/api.go
```diff
@@ -239,7 +248,7 @@ func NewServerRoutes(s *Server) *http.ServeMux {
 			panic(err)
 		}
 
-		remh := NewRemoteHandlers(s.qriNode, receivers)
+		remh := NewRemoteHandlers(s.qriNode, s.cfg, receivers)
```
dustmop: I assume that in the future this will change to be just:

```
remh := NewRemoteHandlers(s.inst, receivers)
```

It's probably too big of a change to do all of that now, but how about a TODO that mentions this plan?

api/api_test.go
```diff
@@ -197,7 +205,9 @@ func TestServerReadOnlyRoutes(t *testing.T) {
 	if err != nil {
 		t.Fatal(err.Error())
 	}
-	s := New(node, cfg)
+	// TODO (b5) - hack until tests have better instance-generation primitives
+	inst := lib.NewInstanceFromConfigAndNode(cfg, node)
```
dustmop: Instead, once we switch to using the private `instance` struct for internal code, just change this to:

```
inst := &instance{config: cfg, node: node}
```

api/profile_test.go
```diff
@@ -12,12 +12,20 @@ import (
...
 	"github.com/beme/abide"
+	"github.com/qri-io/qri/config"
+	"github.com/qri-io/qri/lib"
 )
 
 func TestProfileHandler(t *testing.T) {
 	node, teardown := newTestNode(t)
 	defer teardown()
 
+	cfg := config.DefaultConfigForTesting()
```
dustmop: Create a helper method that does this, and use it in these functions in order to call out the existence of the pattern:

```
  inst := newTestInstanceWithProfileFromNode(node)
```

```
func newTestInstanceWithProfileFromNode(node *p2p.QriNode) *instance {
  cfg := config.DefaultConfigForTesting()
  pro, _ := node.Repo.Profile()
  cfg.Profile, _ = pro.Encode()
  return &instance{cfg: cfg, node: node}
}
```

cmd/qri.go
```diff
@@ -220,93 +152,106 @@ func (o *QriOptions) CryptoGenerator() gen.CryptoGenerator {
...
+	return o.inst.Node(), nil
 }
 
 // DatasetRequests generates a lib.DatasetRequests from internal state
 func (o *QriOptions) DatasetRequests() (*lib.DatasetRequests, error) {
 	if err := o.Init(); err != nil {
 		return nil, err
 	}
-	return lib.NewDatasetRequests(o.node, o.rpc), nil
+	return lib.NewDatasetRequests(o.inst.Node(), o.inst.RPC()), nil
```
dustmop: Please add a TODO about switches these implementations to passing in an `instance` without pulling it apart.

cmd/render_test.go
```diff
@@ -103,7 +103,7 @@ func TestRenderRun(t *testing.T) {
 		t.Errorf("error setting default template in config: %s", err)
 		return
 	}
-	lib.Config = cfg
+	// lib.Config = cfg
```
dustmop: Delete this line.

cmd/save_test.go
```diff
@@ -118,7 +117,7 @@ func TestSaveRun(t *testing.T) {
 		return
 	}
 
-	lib.Config = config.DefaultConfigForTesting()
+	// lib.Config = config.DefaultConfigForTesting()
```
dustmop: Delete me, too

config/config.go
```diff
@@ -98,14 +101,24 @@ func ReadFromFile(path string) (*Config, error) {
...
+	cfg.path = path
+}
+
+// Path gives the unexported filepath for a config
+func (cfg Config) Path() string {
+	return cfg.path
+}
+
 // WriteToFile encodes a configration to YAML and writes it to path
 func (cfg Config) WriteToFile(path string) error {
```
dustmop: Can this change to using cfg.path instead of a parameter?

lib/lib.go
```diff
@@ -32,16 +45,417 @@ func init() {
 
 // Receivers returns a slice of CoreRequests that defines the full local
 // API of lib methods
-func Receivers(node *p2p.QriNode) []Requests {
+func Receivers(inst Instance) []Requests {
+	node := inst.Node()
+	r := inst.Repo()
```
dustmop: Would prefer not using this temp `r`, and instead just using node.Repo where it's needed.

lib/lib.go
```diff
@@ -32,16 +45,417 @@ func init() {
...
+	// Config returns the current configuration for this
+	Config() *config.Config
+	Node() *p2p.QriNode
+	Repo() repo.Repo
+	RPC() *rpc.Client
+}
+
+// WritableInstance is an instance that can be modified
+// instances that allow config changes should implement this interface
+type WritableInstance interface {
```
dustmop: Discussed offline: I don't think it works as well having this property expressed as an interface. Would prefer to have explicit logic in the config code to control whether the config may be changed. Thinking about it even more, using the name `WritableInstance` also implies that nothing *about* the instance is writable, like it would make me think that the repo and node were also could not be written to if they didn't implement this interface.

lib/lib.go
```diff
@@ -32,16 +45,417 @@ func init() {
...
+	// SetConfig modifies configuration details
+	SetConfig(*config.Config) error
+}
+
+// ErrNotWritable is a canonical error for try to edit an instance
+// that isn't editable
+var ErrNotWritable = fmt.Errorf("this instance isn't writable")
+
+// Requests defines a set of library methods
+type Requests interface {
```
dustmop: With the rename away from Requests, how about calling this `Methods` instead, with the method named `MethodsName` instead? Ok to add a TODO for now.

lib/lib.go
```diff
@@ -32,16 +45,417 @@ func init() {
...
+// Config provides methods for manipulating Qri configuration
+func (inst *instance) Config() *config.Config {
+	return inst.cfg
+}
+
+// SetConfig implements the ConfigSetter interface
+func (inst *instance) SetConfig(cfg *config.Config) (err error) {
+
+	if path := inst.cfg.Path(); path != "" {
+		if err = cfg.WriteToFile(path); err != nil {
```
dustmop: Follow up from my earlier comment on WriteToFile, this could be:

```
cfg.path = path
err = cfg.WriteToFile()
```

lib/lib.go
```diff
@@ -32,16 +45,417 @@ func init() {
...
+		}
+	}
+
+	inst.cfg = cfg
+	return nil
+}
+
+// Node accesses the instance qri node if one exists
+func (inst *instance) Node() *p2p.QriNode {
+	return inst.node
```
dustmop: These bodies to me are the strongest evidence that we should use a non-interface `instance` struct in all of our internal code. There's no real alternative for how an interface could work aside from just returning these private fields. We should just be passing around a simple `instance` struct everywhere.

lib/lib_test.go
```diff
@@ -58,9 +59,13 @@ func TestReceivers(t *testing.T) {
 	}
 
 	node := n.(*p2p.QriNode)
-	reqs := Receivers(node)
-	if len(reqs) != 9 {
-		t.Errorf("unexpected number of receivers returned. expected: %d. got: %d\nhave you added/removed a receiver?", 9, len(reqs))
+	cfg := config.DefaultConfigForTesting()
+	// TODO (b5) - hack until tests have better instance-generation primitives
+	inst := &instance{node: node, cfg: cfg}
```
dustmop: From my perspective, I don't think this is a hack, I think this is the proper use.

lib/config.go
b5: @dustmop & I have been discussing this at length offline. I'm very much won over by this approach, especially when combined with some of the other feedback in this review (like a convenience function for building test `instance`s).

I'm wondering if we should take it one step further: make `instance` an exported struct with no exported fields:

```go
type Instance struct {
  cfg *Config
  node *p2p.QriNode
  ctx Context
}
```

This PR currently has `lib` exporting an `Instance` interface:
```go
// Instance is the interface that bundles the foundational values of a
// qri instance. Instance provides the basis for creating Method constructors,
// which actually do qri things
// think of instance as the "core" of the qri ecosystem
type Instance interface {
	// Context returns the base context this instance is using. Any resources
	// built from this instance should inherit from this context and obey
	// calls from the ctx.Done(), releasing any & all resources
	Context() context.Context
	// Teardown closes the instance by closing the base context
	Teardown()
	// Config returns the current configuration for this
	Config() *config.Config
	Node() *p2p.QriNode
	Repo() repo.Repo
	RPC() *rpc.Client
}
```
The initial thinking is defaulting to being defensive: by passing an interface we're making fewer promises to userland. I've spent some time asking the question: _when would we want multiple implementations of this interface?_ And can't find any reasonable situations where this would be the case.

Combining that with @dustmop's feedback, if `interface` were exported we'd have two nice properties. First, code like this now looks far more ok in `lib` tests because Instance is exported, and lib can access the unexported fields:
```go
inst := &Instance{ node: node }
```

Secondly, we can take Dustmop's suggestion one step further now that Instance is a concrete type, having "Method Types" build directly on an instance if they require no additional fields:
```go
type ConfigMethods Instance

func (c *ConfigMethods) SomeFunc() {
  c.node.Connect() // or whatever
}
```
This is only possible if Instance isn't an interface type.

config/config.go
b5: what if I want to write config somewhere else? Do I have to `SetPath` then `WriteToFile`? I'm ok(ish) with that, just want to be sure that's the API we're after

api/api_test.go
b5: because `api` imports `lib` this won't be possible because lib doesn't export `instance` or any of it's fields. I think in this case the right thing to do would be to use `lib.NewInstance`. `api` is a consumer of `lib`, it's `lib`'s job to make writing these tests straightforward.

cmd/qri.go
b5: sure thing, I added TODOs at all struct definition sites in the lib package

lib/lib.go
b5: ok, let's debate & cleanup in subsequent PRs

lib/lib.go
```diff
@@ -386,19 +381,30 @@ func newFilesystem(cfg *config.Config, store cafs.Filestore) (qfs.Filesystem, er
...
 }
 
-// instance implements the (exported) Instance interface
-// create an instance one with NewInstance
-type instance struct {
+// Instance bundles the foundational values qri relies on, including a qri
+// configuration, p2p node, and base context.
+// An instance wraps required state for for "Method" constructors, which
+// contain qri business logic. Think of instance as the "core" of the qri
+// ecosystem create one with NewInstance
```
dustmop: separate these clauses, period or comma after "ecosystem"

PR #741, opened 2019-04-17
feat(daemonize): Command to daemonize qri

dustmop:


b5:
I'ma close this for now, not because it's bad, because it's going to be a _long_ time before we can entertain doing this. We need to be able to spin up and down IPFS node in a long running process with at least a weak guarantee that no resources will leak while that process is at rest.

For the record, patterns @dustmop developed here are the foundation of `qri update start --daemonize`

PR #744, opened 2019-04-22
refactor(lib.ConfigMethods): removes `lib.ConfigMethods` interface

ramfox:
We've talked a bit about this internally, but we have a new guideline for when something should be an interface: if we can think of more then one implementation... it should be an interface!

Otherwise, let's just make it a concrete struct! 

Our first move over to this philosophy is the `lib.ConfigMethods`. Because we need to serve Qri over RPC, we need to follow some rules about the signatures of the methods we are registering. One of the stipulations for registering a method over rpc, is the struct on which the method operates must be exported. Because we implemented a private `configMethods` struct, that satisfies the `ConfigMethods` interface, we would need to jump through a few hoops to get the `configMethods` struct to be seen as exported when we try registering its methods over RPC. Instead, let's just remove the `ConfigMethods` interface in favor of exporting a `ConfigMethods` struct.

Other things this PR addresses:
- we aren't catching the error from getting a`DatasetRequest` in `cmd/Get`, this causes a panic. So let's catch the error
- refactoring `lib.NewInstance` that makes the RPC dialing/connecting process more clear. Basically, the first time we run `Init`, dialing to RPC should fail. If it doesn't fail, then we already have RPC connection running, so we should connect to that RPC connection and return. Otherwise, we have to do a full set up of a qri instance.

### DIFF
--- a/lib/config.go
+++ b/lib/config.go
@@ -15,8 +15,8 @@ type ConfigMethods struct {
 
 
 // NewConfigMethods creates a configuration handle from an instance
-func NewConfigMethods(inst *Instance) ConfigMethods {
-	return ConfigMethods{inst: inst}
+func NewConfigMethods(inst *Instance) *ConfigMethods {
+	return &ConfigMethods{inst: inst}
 }
 
 // CoreRequestsName specifies this is a configuration handle
@@ -33,7 +33,7 @@ type GetConfigParams struct {
 
 // GetConfig returns the Config, or one of the specified fields of the Config,
 // as a slice of bytes the bytes can be formatted as json, concise json, or yaml
-func (m ConfigMethods) GetConfig(p *GetConfigParams, res *[]byte) (err error) {
+func (m *ConfigMethods) GetConfig(p *GetConfigParams, res *[]byte) (err error) {
 	if m.inst.rpc != nil {
 		return m.inst.rpc.Call("ConfigMethods.GetConfig", p, res)
 	}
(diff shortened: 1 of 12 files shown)


lib/config.go
```diff
@@ -9,29 +9,18 @@ import (
...
+
 // NewConfigMethods creates a configuration handle from an instance
 func NewConfigMethods(inst *Instance) ConfigMethods {
-	return configMethods{inst: inst}
-}
-
-// configMethods is the private implementation of ConfigMethods
-type configMethods struct {
-	inst *Instance
+	return ConfigMethods{inst: inst}
```
b5: One new-ish convention is now exposed: the returned `ConfigMethods` is a value, not a pointer, because ConfigMethods can't change any of it's internal values.

One concern I have with exposing this: we won't be able to get away with returning values for all "Methods" structs, because some structs need additional state to do their work (`RemoteRequests` is a good example). 

This begs a choice: 
Should we favour consistency and  return `*ConfigMethods`, even though all Methods are not defined on a pointer? Or should we favour accuracy, and return the struct value?

ramfox: After doing a quick audit, only `ConfigMethods` and `ProfileMethods` don't return pointers. I think we should just be consistent, as long as we aren't losing anything.

b5: dope. let's do it!

b5:
ok ok we know this'll work now that #746 is in. @ramfox feel free to merge when you get a second!

PR #745, opened 2019-04-22
feat(base/fill): Overhaul error handling

dustmop:
Primary change: the path that an error occurs at is now prepended to the error message. Now they look like "at Meta.Title: need string, got int". Introduce FieldError, for when a field cannot be converted from its given type to the target type. Introduce ErrorCollector, which collects errors and merges them.
Raise test coverage by a significant amount.

### DIFF
--- a/base/fill/error_collector.go
+++ b/base/fill/error_collector.go
@@ -1,0 +1,57 @@
+package fill
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ErrorCollector collects errors, merging them into one, while noting the path where they happen
+type ErrorCollector struct {
+	errs   []error
+	fields []string
+}
+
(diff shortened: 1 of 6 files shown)


base/fill/error_collector.go
```diff
@@ -0,0 +1,57 @@
+package fill
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ErrorCollector collects errors, merging them into one, while noting the path where they happen
+type ErrorCollector struct {
```
b5: Fun! It's worth calling out that a similar pattern would be applicable to our jsonschema package, which returns a slice of `ValError`: https://github.com/qri-io/jsonschema/blob/master/schema.go#L185

PR #750, opened 2019-04-25
fix(linux): Ignore errors from setting rlimit, needs root on linux

dustmop:


dustmop:
The next step is to add documentation about building, yes. I think https://github.com/qri-io/qri/#build is the right spot, just need to expand it.

PR #752, opened 2019-04-29
fix(pagination): add `page` and `pageSize` to api and `limit` and `offset` to cmd

ramfox:
Title basically covers it!

The only other change is outlined in commit 2e623db:
> While working on offset and limit for `qri peers list`, I realized that paginating through the list of peers doesn't really make sense, because to generate the peers list we range through a map, not a slice, we get a different order each time.

> We can talk about whether we want to make this list ordered, but for now, I thought it might be a good place to try to stdout to `less`, rather then return the output to the terminal.

> So, if you look for your cached peers, qri will send you to `less` that displays your cached peers.

This PR also adds pagination tests at the `lib` level:
  `DatasetRequests.List`
  `RegistryRequests.List`
  `SearchRequests.Search`
  `LogRequests.Log`

depends on qri-io/registry#28 before tests will pass


cmd/get.go
```diff
@@ -89,20 +89,20 @@ func (o *GetOptions) Complete(f Factory, args []string) (err error) {
 	}
 
 	if o.Selector == "body" {
-		// if we have a limit, but not offset, assume an offset of 0
-		if o.Limit != -1 && o.Offset == -1 {
-			o.Offset = 0
+		// if we have a PageSize, but not Page, assume an Page of 1
+		if o.PageSize != -1 && o.Page == -1 {
+			o.Page = 1
```
dustmop: Do pages start being indexed at 1 instead of 0?

cmd/get.go
```diff
@@ -122,13 +122,16 @@ func (o *GetOptions) Run() (err error) {
 		}
 	}
 
+	// convert Page and PageSize to Limit and Offset
+	listParams := lib.NewListParams("", o.Page, o.PageSize)
```
dustmop: I don't think it makes sense to construct this just to break it back apart a few lines later. Just use o.Page and o.PageSize to assign to Offset and Limit below.

cmd/list.go
```diff
@@ -87,12 +87,15 @@ func (o *ListOptions) Complete(f Factory, args []string) (err error) {
 // Run executes the list command
 func (o *ListOptions) Run() (err error) {
 
+	// convert Page and PageSize to Limit and Offset
+	listParams := lib.NewListParams("", o.Page, o.PageSize)
```
dustmop: Same comment here.

cmd/log.go
```diff
@@ -79,12 +79,15 @@ func (o *LogOptions) Run() error {
 		return err
 	}
 
+	// convert Page and PageSize to Limit and Offset
+	listParams := lib.NewListParams("", o.Page, o.PageSize)
```
dustmop: Same comment here.

cmd/peers.go
```diff
@@ -208,37 +211,61 @@ func (o *PeersOptions) Info() (err error) {
...
+			Limit:  listParams.Limit,
+			Offset: listParams.Offset,
 			Cached: o.Cached,
 		}
 		res := []*config.ProfilePod{}
 		if err = o.PeerRequests.List(p, &res); err != nil {
 			return err
 		}
 
+		if o.Cached {
```
dustmop: So I mentioned `less` in another thread, but after some digging, it looks like the best thing to do is to check the environment variable "PAGER", and use that, falling back to `less` otherwise: https://en.wikibooks.org/wiki/Guide_to_Unix/Environment_Variables#PAGER and https://unix.stackexchange.com/questions/213367/where-do-editor-pager-browser-environment-variables-come-from.

However, this also is POSIX specific, so it won't work on Windows. It's okay for now, but maybe add a TODO to fix this in the future so we know it's here.

Finally, do we always want to page this (outputting to less or whatever) even if there's only a small number of results? Maybe instead, check the length and just output to the terminal if it's small.

cmd/search.go
```diff
@@ -74,16 +75,18 @@ func (o *SearchOptions) Validate() error {
...
-	// TODO - spinner is failing tests
-	// o.StartSpinner()
-	// defer o.StopSpinner()
+	o.StartSpinner()
+	defer o.StopSpinner()
 
 	// TODO: add reindex option back in
 
+	// convert Page and PageSize to Limit and Offset
+	listParams := lib.NewListParams("", o.Page, o.PageSize)
```
dustmop: Same comment about just assigning o.Page and o.PageSize directly.

base/dataset.go
```diff
@@ -110,6 +110,15 @@ func ListDatasets(r repo.Repo, term string, limit, offset int, RPC, publishedOnl
 		}
 		res = pub[:i]
 	}
+	// if offset is too high, return empty list
+	if offset >= len(res) {
+		return res[:0], nil
```
dustmop: This reads a bit strange - using the name `res` but essentially ignoring its value. Would suggest changing it to: `		return []repo.DatasetRef{}, nil`

cmd/get.go
ramfox: yes!

cmd/peers.go
ramfox: dope this is great.

now we only print to a pager if we are looking at more than 25 items, we check the env variable PAGER, if it's not set we default to less.

cmd/peers.go
```diff
@@ -208,37 +213,67 @@ func (o *PeersOptions) Info() (err error) {
...
 		res := []*config.ProfilePod{}
 		if err = o.PeerRequests.List(p, &res); err != nil {
 			return err
 		}
 
+		// If the length of the peers list is greater then 25
+		// send the output to a terminal pager
+		if len(res) > 25 {
+			// TODO (ramfox): This is POSIX specific, need to expand!
+			envPager := os.Getenv("PAGER")
```
b5: WooOooOOOooah

cmd/peers.go
```diff
@@ -208,37 +213,67 @@ func (o *PeersOptions) Info() (err error) {
...
 			Cached: o.Cached,
 		}
 		res := []*config.ProfilePod{}
 		if err = o.PeerRequests.List(p, &res); err != nil {
 			return err
 		}
 
+		// If the length of the peers list is greater then 25
+		// send the output to a terminal pager
+		if len(res) > 25 {
```
b5: This `less` stuff is _very_ cool. Glad to see you've only introduced it in a small place as an experiment. I think we should merge this & follow up with another PR that unifies all pagniated printing into a single function detects & uses `PAGER`.

b5: but while we're here, I don't think we need the arbitrary 25 distinction. let's just always print with this new codepath.

cmd/peers.go
```diff
@@ -208,37 +213,67 @@ func (o *PeersOptions) Info() (err error) {
...
+			// TODO (ramfox): This is POSIX specific, need to expand!
+			envPager := os.Getenv("PAGER")
+			if envPager == "" {
+				envPager = "less"
+			}
+
+			buf := bytes.Buffer{}
```
b5: allocate a pointer to avoid the three pointer allocations below
```suggestion
			buf := &bytes.Buffer{}
```

cmd/peers.go
```diff
@@ -208,37 +213,67 @@ func (o *PeersOptions) Info() (err error) {
...
+			pager := exec.Command(envPager)
+			pager.Stdin = &buf
+			pager.Stdout = o.Out
+
+			fmt.Fprintln(&buf, "")
+			for i, peer := range res {
+				printPeerInfoNoColor(&buf, i, peer)
+			}
+
+			if err := pager.Run(); err != nil {
```
b5: drop the `if clause`, returning the result of `pager.Run()` directly

lib/search_test.go
```diff
@@ -28,23 +28,30 @@ func TestSearch(t *testing.T) {
...
-	numResults := 3
-	errString := ""
+	cases := []struct {
+		description string
+		params      *SearchParams
+		numResults  int
+		err         string
+	}{
+		{"search datasets - 'cities'", &SearchParams{"cities", 0, 100}, 3, ""},
+		// {"search datasets - 'cities'", &SearchParams{"cities", 0, 2}, 2, ""},
```
b5: either remove or enable, I'm totally fine with just removing for now. This PR is already pretty solid 😄

repo/test/test_repo.go
```diff
@@ -97,15 +97,49 @@ func NewTestRepo(rc *regclient.Client) (mr *repo.MemRepo, err error) {
...
+		if _, err := createDataset(mr, tc); err != nil {
 			return nil, fmt.Errorf("%s error creating dataset: %s", k, err.Error())
 		}
 	}
 
 	return
 }
 
+// NewTestRepoForLog generates a repository with a dataset that has a history, usable for testing purposes
+func NewTestRepoForLog(rc *regclient.Client) (mr *repo.MemRepo, refs []repo.DatasetRef, err error) {
```
b5: I think we need to have a larger conversation about making datasets with long histories, and repos that have datasets with long histories. That time isn't now. For now, can we rename this to `NewTestRepoWithHistory`?

cmd/get.go
dustmop: Spoke about this offline, didn't realize that o.Page and o.PageSize aren't the same values as Limit and Offset, oops!

repo/test/test_repo.go
ramfox: jesus that name is so much better

and yes this is definitely not the optimal way to do this

PR #755, opened 2019-05-03
test(transform): Detect when transforms don't produce any changes

dustmop:
Depends upon https://github.com/qri-io/dsdiff/pull/16 and https://github.com/qri-io/qri/pull/754. Currently `save`ing with a transform that produces no changes is broken, it always creates a commit because the differ see a new transform.ScriptPath. This fixes that.

### DIFF
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -625,6 +625,38 @@ func TestSaveConflictingComponents(t *testing.T) {
 	}
 }
 
+// Test that running a transform without any changes will not make a new commit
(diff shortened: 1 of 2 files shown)


dustmop:
Had to change the base (from `cron` to `master`); now that https://github.com/qri-io/qri/pull/757 was merged to fix the build, no need to base this off of cron. In addition, the description is slightly inaccurate, this functionality was fixed by https://github.com/qri-io/dsdiff/pull/16, this is just adding a test to verify that `transform` is doing the right thing.

PR #758, opened 2019-05-07
fix(transform): Pass prev dataset to ExecScript

dustmop:
Pass prev dataset as well as next dataset, in order to support new get_body functionality added by https://github.com/qri-io/startf/pull/41.

### DIFF
--- a/actions/transform_test.go
+++ b/actions/transform_test.go
@@ -27,11 +27,13 @@ func TestExecTransform(t *testing.T) {
 		t.Fatal(err.Error())
 	}
 
-	ds := &dataset.Dataset{
+	prev := &dataset.Dataset{
 		Structure: &dataset.Structure{
 			Format: "json",
 			Schema: dataset.BaseSchemaArray,
 		},
+	}
+	next := &dataset.Dataset{
 		Transform: &dataset.Transform{
 			Syntax:  "starlark",
 			Config:  map[string]interface{}{"foo": "config"},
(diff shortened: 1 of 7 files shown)


actions/transform_test.go
```diff
@@ -45,9 +47,9 @@ def transform(ds,ctx):
 	ctx.get_secret("bar")
 	ds.set_body([1,2,3])
 	`)
-	ds.Transform.SetScriptFile(qfs.NewMemfileBytes("transform.star", data))
+	next.Transform.SetScriptFile(qfs.NewMemfileBytes("transform.star", data))
 
-	if err := ExecTransform(node, ds, nil, nil); err != nil {
+	if err := ExecTransform(node, next, prev, nil, nil); err != nil {
```
b5: Interesting to note that we're now starting to follow the "diff" pattern of doing a computation on a "src" & "dest" dataset almost everywhere we're making a new version. We've called these parameters at least three things across codebases:

* `prev` and `next` here
* `a` and `b` ([dsdiff](https://github.com/qri-io/dsdiff/blob/4125cbf6b2a42fe8027c0ea13b6cb98a23eae39e/dsdiff.go#L215))
* `d1` and `d2` ([deepdiff](https://github.com/qri-io/deepdiff/blob/613c16a45974ed3e9a781c8c2ca2eb242f570796/deepdiff.go#L19))

I'd like to tinker with the idea of framing `SaveDataset` as the forward-facing version of generating a diff, and starting to call out this "diff pattern" with uniform varnames. I'm a fan of `src` and `dst`. 

None of this should affect this PR, which should be merged meow.

PR #760, opened 2019-05-07
feat(pager): to make it easier to read, long output on the cli is sent to a pager

ramfox:
While embarking on the previous `page` and `pageSize` journey, we found a fun env variable: `PAGER`, by checking if `PAGER` is set, or by defaulting to `more`, we can send users to a pager for large outputs, typically lists. (If `PAGER` isn't set, and if `more` or `less` isn't valid on the user's system, we default to outputting the text to the terminal)

This refactor also unifies our output, especially for lists. It uses the`fmt.Stringer` interface to leverage one function, `printItems`, that formats lists/numbering items.

Each kind of struct that we wish to output (so far, `config.ProfilePod` and `repo.DatasetRef`) is aliased (to `peerStringer`, `refStringer`, and `logStringer`), and given a new `String` method that formats each struct the way we want for that particular usage. So we can have one `repo.DatasetRef`, but it can be cast as a `refStringer` or `logStringer`, depending on what the user is asking to view.

All the `stringers` live in the `cmd/Stringer.go` files. Each `String` method is tested.

These are the commands that use stringer:
- add
- diff
- get
- list
- log
- peers
- registry
- search

These are the commands that use `printItems` (which prints a slice of stringers to a pager):
- diff
- get
- list
- log
- peers
- search

Just have a few finicky test errors to work out (ci is throwing different errors then I get on my machine), and this is good to go!!

### DIFF
--- a/cmd/stringers_test.go
+++ b/cmd/stringers_test.go
@@ -1,0 +1,158 @@
+package cmd
+
+import (
+	"testing"
+	"time"
+
+	"github.com/qri-io/dataset"
+	"github.com/qri-io/qri/config"
+	"github.com/qri-io/qri/repo"
+)
+
+func TestPeerStringer(t *testing.T) {
+	setNoColor(false)
+	cases := []struct {
+		description string
+		peer        *config.ProfilePod
+		expect      string
+	}{
+		// Online & multiple NetworkAddrs
+		{"Peer Stringer - Online & multiple addresses",
+			&config.ProfilePod{
+				Peername:     "cassie",
+				Online:       true,
+				ID:           "Qm...yay",
+				NetworkAddrs: []string{"address_1", "address_2", "address_3"},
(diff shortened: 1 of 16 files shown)


cmd/stringers_test.go
```diff
@@ -0,0 +1,190 @@
...
+						Length:   1,
+						Entries:  1,
+						ErrCount: 1,
+					},
+					NumVersions: 1,
+					Meta: &dataset.Meta{
+						Title: "Dataset Title",
+					},
+				},
+			}, "\u001b[32;1mpeer/ds_name\u001b[0m\nDataset Title\n\u001b[2m/network/hash\u001b[0m\n1 byte, 1 entry, 1 error, 1 version\n\n",
```
b5: long term it might make sense to just test the `noColor` output

PR #762, opened 2019-05-10
OS X daemonized update service

b5:
Adds in ideas developed by @dustmop in #741, bringing a new flag to the update service. Now:
`qri update service start --daemonize` will create a long-running daemon on os x. All other platforms will see an error when run with the `--daemonize` flag. Stop the daemon with `qri update service stop`, see the status of the service with `qri update service status`. 

We should definitely add `qri update service restart` in the future, as well as `qri update service logs` to show the log output of the update scheduler itself.

I've added this code as a new package called `update`, and move the `cron` package into it.


update/update.go
```diff
@@ -1,17 +1,87 @@
...
 
+var log = golog.Logger("update")
+
+// Start starts the update service
+func Start(ctx context.Context, repoPath string, updateCfg *config.Update, daemonize bool) error {
+	if updateCfg == nil {
+		updateCfg = config.DefaultUpdate()
+	}
+
+	cli := cron.HTTPClient{Addr: updateCfg.Address}
```
dustmop: Can you name this variable `httpClient`? Seeing `cli` makes me think "command-line interface". I know we use this for rpc everywhere, but 1) at least that's a consistent pattern, 2) it's confusing to use the same name for a non-rpc client.

cmd/update.go
```diff
@@ -223,6 +223,9 @@ updating process initiates.`,
 			return o.ServiceStart(ioStreams, f.QriRepoPath())
 		},
 	}
+
+	serviceStartCmd.Flags().BoolVarP(&o.Daemonize, "daemonize", "d", false, "start service as an operating system daemon")
```
dustmop: Rather than the phrase "operating system daemon", how about "long-lived daemon".

lib/update.go
```diff
@@ -220,63 +219,49 @@ type ServiceStatus struct {
...
 	}
-
-	ctx := context.Background()
-	if err := local.Start(ctx); err != nil {
-		return err
+	if p.UpdateCfg == nil && m.inst != nil {
+		p.UpdateCfg = m.inst.Config().Update
 	}
-	return nil
+	return update.Start(p.Ctx, p.RepoPath, p.UpdateCfg, p.Daemonize)
```
dustmop: `started` is never assigned to. Instead, check that `Start` returns an error, and if nil, set `started` to true.

lib/update.go
```diff
@@ -220,63 +219,49 @@ type ServiceStatus struct {
...
 }
 
 // ServiceStop halts the scheduler
 func (m *UpdateMethods) ServiceStop(in, out *bool) error {
-	// TODO (b5):
-	return fmt.Errorf("not finished")
+	return update.StopDaemon()
```
dustmop: Same here, assign true to `out`.

lib/lib.go
```diff
@@ -51,7 +52,7 @@ func init() {
 	// TODO (b5): for now this ensures that `qri update service start`
 	// actually prints something. `qri update service start` should print
 	// some basic details AND obey the config.log.levels.cron value
-	golog.SetLogLevel("cron", "debug")
+	// golog.SetLogLevel("cron", "debug")
```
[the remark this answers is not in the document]
b5: yeah... I'm going to keep it for now (and re-enable to debug log level), but move the switch into a clause that checks for daemonization so we don't write a massive log file when daemonized. Let's get rid of it later with a better solution.

PR #767, opened 2019-05-14
refactor(api/body): add limit and offset params to api /body endpoint

### LINKED ISSUE #766: Exposing `offset` and `limit` as flags/params on api `/body`
### ~Exposing `offset` and `limit` as flags/params on api and cmd~

~We recently unified all the API and cmd endpoints to take page and page-size. We recognize though, that for some use cases using limit and offset are an order of magnitude simpler! Let's add those as advanced flags, that will exist, but will not be the primary focus in our documentation. As per this [code conventions issue](https://github.com/qri-io/qri/issues/761#issuecomment-490535806), an offset of 0 and a limit of -1 will return the entire list.~

~If either or limit or offset are set, they will override the page and page-size flags.~

### Exposing `offset` and `limit` as flags/params on api `/body`
As per below, we are scoping this issue down to include `offset` and `limit` params for only the `/body` endpoint!

An offset of 0 and a limit of -1 will return the entire list.

ramfox:
closes #766

This refactor also separates out the part of the bodyHandler that just converts the request params to GetParams, and puts it into a `makeBodyGetParams` func (plz feedback on name). This allows us to test the bodyHandler easier. Added `TestMakeGetBodyParams` which does just that.

This also fixes a linting bug in update/cron/job

### DIFF
--- a/api/datasets.go
+++ b/api/datasets.go
@@ -563,15 +563,57 @@ func loadFileIfPath(path string) (file *os.File, err error) {
 	return os.Open(path)
 }
 
-// default number of entries to limit to when reading
-// TODO - should move this into lib
-const defaultDataLimit = 100
-
 // DataResponse is the struct used to respond to api requests made to the /data endpoint
 // It is necessary because we need to include the 'path' field in the response
 type DataResponse struct {
 	Path string          `json:"path"`
 	Data json.RawMessage `json:"data"`
+}
+
+func makeGetBodyParams(r *http.Request, readOnly bool, path string) (*lib.GetParams, error) {
+	listParams := lib.ListParamsFromRequest(r)
+	download := r.FormValue("download") == "true"
+	format := "json"
+	if download {
+		format = r.FormValue("format")
+	}
+	// if download is not set, and format is set, make sure the user knows that
+	// setting format won't do anything
+	if !download && r.FormValue("format") != "" && r.FormValue("format") != "json" {
+		return nil, fmt.Errorf("the format must be json if used without the download parameter")
+	}
+
+	p := &lib.GetParams{
+		Path:     path,
+		Format:   format,
+		Selector: "body",
+		Limit:    listParams.Limit,
+		Offset:   listParams.Offset,
+		All:      r.FormValue("all") == "true" && !readOnly,
+	}
+
+	if !readOnly {
+		offset, offsetErr := util.ReqParamInt("offset", r)
+		limit, limitErr := util.ReqParamInt("limit", r)
+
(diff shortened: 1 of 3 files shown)


api/datasets.go
```diff
@@ -563,32 +563,15 @@ func loadFileIfPath(path string) (file *os.File, err error) {
...
 }
 
-func (h DatasetHandlers) bodyHandler(w http.ResponseWriter, r *http.Request) {
-	d, err := DatasetRefFromPath(r.URL.Path[len("/body"):])
-	if err != nil {
-		util.WriteErrResponse(w, http.StatusBadRequest, err)
-		return
-	}
-
+func makeGetBodyParams(r *http.Request, readOnly bool, path string) (*lib.GetParams, error) {
```
b5: `getParamsFromRequest`? I think it's ok to define the function in generic terms, and add a comment that says something like:
```go
// getParamsFromRequest creates getParams from a request. It's currently only used for paginating dataset bodies
```

api/datasets.go
```diff
@@ -597,24 +580,67 @@ func (h DatasetHandlers) bodyHandler(w http.ResponseWriter, r *http.Request) {
...
 
 	p := &lib.GetParams{
-		Path:     d.String(),
+		Path:     path,
 		Format:   format,
 		Selector: "body",
 		Limit:    listParams.Limit,
 		Offset:   listParams.Offset,
-		All:      r.FormValue("all") == "true" && listParams.Limit == defaultDataLimit && listParams.Offset == 0,
+		All:      r.FormValue("all") == "true" && !readOnly,
```
b5: in a future PR we should deprecate the `All` field from GetParams in favour of directly supplying`Offset: 0, Limit: -1`. This is great in the meantime

api/datasets_test.go
```diff
@@ -0,0 +1,155 @@
...
+func TestMakeGetBodyParams(t *testing.T) {
+	cases := []struct {
+		description                   string
+		download                      bool
+		format                        string
+		page, pageSize, offset, limit int
+		expOffset, expLimit           int
+		all, readOnly                 bool
+		expAll                        bool
+		expectedErr                   string
```
b5: nice! lately I've been splitting table tests into two slices, one for the "good" (cases that shouldn't error), and another for the "bad" (cases that should _only_ produce errors).  Example: https://github.com/qri-io/qri/blob/master/lib/datasets_test.go#L82-L136

mind splitting these? I think you'll like the result. In the meantime I'll document this over on that code conventions issue...

ramfox: yea that is awesome, way less mental gymnastics.

PR #768, opened 2019-05-14
Organize repo

### LINKED ISSUE #765: canonical location for reading & writing qri "repo data"
A bunch of recent work has shed light on just how fuzzy our definition of "qri path" is. There are currently a slew of different ways to get to the "right place" where code can start to make assumptions about where data can be read from and written to. Some parts of our codebase use the location of the `config.yaml` file, others look for the `$QRI_PATH` environment variable and fall back to `$HOME/.qri`, others rely on a value returned by an Instance. 

I'd like to propose we clean all this up with a clear definition of where Qri should look for and store the data it needs to operate. We should define the code that determines that location in _one_ place, and use the result of that value to build assumptions.

I think this boils down to a question of what is the ultimate source of truth for determining where data should be read from & written to. If two sources are in conflict, who wins? If a `config.yaml` file specifies `fs.path : ~/dev/null`, but that config file is loaded from `~/.qri`, where is the fs repo location? 

As far as I can tell, there are two approaches here that make sense as a source of truth, each with benefits & drawbacks:

### config.yaml file is the source of truth
In this world, we focus on loading a configuration file, and whatever paths that repo specifies for each system is the source of truth. Configuration relies on sensible defaults to achieve a common directory layout structure

_Pros:_ like, _soooooooo_ configurable
_Cons:_ No explicit directory structure be relied upon, because the user may have overridden values. Easy to make mistakes about b/c 90% of repos will look the same until someone starts getting fancy with configuration.

### repo root directory is the source of truth
In this approach, "source of truth" for finding Qri data is a _directory_ with a _consistent  structure_ A Qri Instance is loaded from a single repo directory, which requires `config.yaml` to be defined, and at the same path relative to the repo root. configuration may override details about how to load data, but this is considered specific to the subsystem that's being configured.  Initialization focuses on identifying a _repo directory_ and builds all other assumptions from there.

_Pros:_ consistent, easier to reason about, reduces the number of assumptions Qri developers need to defend against
_Cons:_ brittle

#### I vote for "repo path"

I'm _very_ much in favour of the latter "repo path" approach. I think we have enough difficult-to-manage you-can't-assume type code in our codebase, and I don't think the config-level overrides bring any benefits the different "Type" implementations don't already deliver. I haven't seen many people complain about the inflexibility of the `.git` directory layout. If we go with this, a few assumptions can be built up:

* an instance requires a _repo path_ (root directory) to operate, creating a new instance requires a path. 
* qri repo directories _must_ be laid out the same way, configuration is _always_ loaded from `repoPath/config.yaml` unless overridden at runtime
* the default repo path is `$HOME/.qri`. This can be overridden by setting a `$QRI_PATH` env variable. Both of these values can be overridden with a `--repo` flag on the command line.
* qri must be able to read & write data to this directory, systems within can rely on having a place to read & write
* we can move the code that determines the repo root into the `repo` package, and plumb it up into all other systems
* the CLI can get a new `--repo` flag to replace the currently depricated `--config` flag, which alleviates the potentially super confusing `qri --config ~/.other_qri/config.yaml save --file transform.star --config=api_token,$API_TOKEN me/dataset`
* this clarifies the story around how we do multiple qri repos on the same computer, clearing the way for a single `qri update service` that can be configured with a new `RepoPath` to have multiple repos use the same update daemon

By enforcing this as an assumption we can clean up a bunch of code, and reason better about things like where we can expect to find `update` service resources.

There is one shortcoming to the repo path approach: it's possible to configure qri to operate purely from memory. In this setting it would be ideal to. We mainly use this for testing, but it's a little more work to juggle. For now we can create & run tests from a temp directory, and can even check to make sure the directory remains empty for the duration of the test if we're looking for "pure memory" assurances.

b5:
This PR closes #765 by requiring a `repoPath` param to `lib.NewInstance`.

I've also reorganized the structure of the `update` service in a qri repo. all update stuff except `/bin` is moved into an `update` subdir. files are relabelled.

before:
```
.qri
├── bin
│   └── qri
├── config.yaml
├── cron_jobs.qfb
├── cron_logs.qfb
├── ds_refs.json
├── events.json
├── index.bleve
│   ├── index_meta.json
│   └── store
│       └── ...
├── logs
│   ├── 1557256003-test_count.log
│   ├── 1557773521-country_codes.log
│   ├── 1557773581-test.sh.log
│   ├── ...
│   ├── stderr.log
│   └── stdout.log
├── peers.json
├── peers.json.lock
├── root
└── selected_refs.json
```

after:
```
.qri
├── config.yaml
├── ds_refs.json
├── events.json
├── index.bleve
│   ├── index_meta.json
│   └── store
│       └── ...
├── peers.json
├── peers.json.lock
├── selected_refs.json
└── update
    ├── bin
    │   └── qri
    ├── jobs.qfb
    ├── logfiles
    │   ├── 1557256003-test_count.log
    │   ├── 1557423597-country_codes.log
    │   ├── 1557760599-airport_codes.log
    │   └── ...
    ├── logs.qfb
    ├── root
    ├── service.log
    └── stderr.log
```

### DIFF
--- a/lib/lib.go
+++ b/lib/lib.go
@@ -106,15 +106,6 @@ func OptCtx(ctx context.Context) Option {
 	}
 }
 
-// // OptSetQriRepoPath configures the directory to read Qri from, defaulting to
-// // "$HOME/.qri", unless the environment variable QRI_PATH is set
-// func OptSetQriRepoPath(path string) Option {
-// 	return func(o *InstanceOptions) (err error) {
-// 		o.RepoPath, err = repo.Path(path)
-// 		return
-// 	}
-// }
-
 // OptConfig supplies a configuration directly
 func OptConfig(cfg *config.Config) Option {
 	return func(o *InstanceOptions) error {
--- a/lib/lib.go
+++ b/lib/lib.go
@@ -34,9 +34,6 @@ import (
 )
 
 var (
-	// defaultQriLocation is where qri data defaults to storing. The keyword $HOME
-	// (and only $HOME) will be replaced with the current user home directory
-	defaultQriLocation = "$HOME/.qri"
 	// defaultIPFSLocation is where qri data defaults to looking for / setting up
 	// IPFS. The keyword $HOME will be replaced with the current user home
 	// directory. only $HOME is replaced (no other $ env vars).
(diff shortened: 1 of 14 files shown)


lib/lib.go
```diff
@@ -109,31 +106,14 @@ func OptCtx(ctx context.Context) Option {
...
-	return
-}
+// // OptSetQriRepoPath configures the directory to read Qri from, defaulting to
+// // "$HOME/.qri", unless the environment variable QRI_PATH is set
+// func OptSetQriRepoPath(path string) Option {
+// 	return func(o *InstanceOptions) (err error) {
+// 		o.RepoPath, err = repo.Path(path)
+// 		return
+// 	}
+// }
```
ramfox: If this is dead code, plz remove, or let us know why it's still here :)

PR #769, opened 2019-05-16
refactor lib.NewInstance to require context and a repo path. make cmd the owner of default repo path.

b5:
@dustmop put forth the view that default path negotiation should be the concern of cmd. Sound right to me, which implies that lib.NewInstance should require a path, and cmd should do the work of deciding between flags, $QRI_PATH, and the default repo location, passing that off to lib. Thankfully we already to do this with our cmd.PathFactory interface, and have a hot-swappable version for testing. This commit removes the recently-introduced repo.Path method, and does some minor cleanup.

While we're playing with the signature of NewInstance, it makes sense to also require a context.
library consumers should always be providing a context to the instance constructor, even if just
context.Background(). Adding this requirement signals to the user that they should first understand how context works before using lib, and that library authors will need to respect the passed in context throughout packages below lib.

If you don't have a context but need one, feel free to just use `context.Background`:
```go
inst, err := lib.NewInstance(context.Background(), "/path/to/repo")
```

### DIFF
--- a/api/update_test.go
+++ b/api/update_test.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"context"
 	"io/ioutil"
 	"os"
 	"testing"
@@ -20,7 +21,7 @@ func TestUpdateHandlers(t *testing.T) {
 	cfg.Store.Type = "map"
 	cfg.Repo.Type = "mem"
 
-	inst, err := lib.NewInstance(tmpDir,
+	inst, err := lib.NewInstance(context.Background(), tmpDir,
 		lib.OptConfig(cfg),
 	)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


api/update_test.go
```diff
@@ -20,7 +21,7 @@ func TestUpdateHandlers(t *testing.T) {
 	cfg.Store.Type = "map"
 	cfg.Repo.Type = "mem"
 
-	inst, err := lib.NewInstance(tmpDir,
+	inst, err := lib.NewInstance(context.Background(), tmpDir,
```
dustmop: Move onto one line

b5: totally. I've got this fixed in what will be a follow-up PR

PR #772, opened 2019-05-21
feat(fill.Struct): Support interface{} fields, require map string keys

dustmop:


dustmop:
Some background on why this is needed. Json requires that its map like object (called an Object) uses only strings for keys, and the built-in serializer fails if this assumption is violated. Likewise, the json deserializer only returns maps with strings for keys.

The yaml parser however, does not follow this rule (as per the yaml spec), so it returns maps with interface{}s for keys. This ends up making interop rather awkward.

To be specific, there are places (such as ReadDataset) where we parse json or yaml to build a dataset. Nearly all of the fields of a dataset are corrected assigned by the parser and/or fill.Struct, however the Structure.Schema is arbitrary data of type interface{}. If left as is, this can end up with a dataset in memory that has a schema contains map[interface{}]interface{}.

Later code may fail in this situation. For example, compareDatasets assumes that it can serialize fields to json, then compare the serialized bytes. If a dataset is read as yaml, then it will fail at this point, since the schema is unable to be serialized as json.

The fix is to have fill.Struct always convert map[interface{}]interface{} into map[string]interface{}, meaning that datasets can always be safely reserialized as json or yaml.

PR #773, opened 2019-05-23
fix(config): fix api.Copy not copying all fields

b5:
we missed a few fields in `config.API.Copy()` that's causing `qri connect --setup` to not allow serving remote traffic, because `cmd`'s `Connect` uses a copy of the api config ☹️


b5:
merging b/c it's 10pm here

PR #774, opened 2019-05-23
fix(cmd): don't use FgWhite, breaks light-colored termimals

### LINKED ISSUE #771: cmd.printInfo Output is Nearly Invisible on Light Terminals
I tried running `qri version` and went absurdly deep trying to figure out why it was printing a blank line in my terminal:

qri blank line

However, it turned out the line wasn’t blank at all! It’s just white, and my terminal’s background is very light. If I select the text, I can see it:

qri highlighted line

This is happening because the `version` command uses `cmd.printInfo`, which explicitly sets the color to white: https://github.com/qri-io/qri/blob/9ae8d44df443608c487b5269789181b87988496b/cmd/print.go#L43-L45

I think it might be better to use `color.Reset` instead of `color.FgWhite` for this case, though I’m not sure whether that might cause problems in other places where we call this function (e.g. if it might cause problems when nested inside another colored string or something). I’m happy to make a quick PR for this if that’s not likely to be an issue, though.

b5:
closes #771. 

I wanted to solve this problem the "right" way, by making sure we're always writing ANSI codes that we know will print. After some digging, this turns out to be a _way_ bigger problem than I expected, and the easier answer is to avoid using colors like `FgWhite` or `FgBlack`. Thankfully there are only a few places in our codebase that use these, and one of them was dead code. 

Here's some of the places this took me for when we end up back here in six months:

* https://github.com/twilio/twilio-cli/issues/7
* https://github.com/chalk/chalk/issues/102
* https://stackoverflow.com/questions/2507337/is-there-a-way-to-determine-a-terminals-background-color
* https://bugzilla.gnome.org/show_bug.cgi?id=733423

Neovim has a funny way of handling this:

https://github.com/neovim/neovim/blob/66d127957ed3cc27af0344c742bc5acedd347894/src/nvim/tui/input.c#L367-L430

using `ReverseVideo` is another option I tried, it puts gross boxes around everything:
https://stackoverflow.com/questions/50027162/determine-terminal-tty-background-color-at-runtime

TL;DR; don't use `FgWhite` or `FgBlack`.

### DIFF
--- a/cmd/diff.go
+++ b/cmd/diff.go
@@ -1,11 +1,8 @@
 package cmd
 
 import (
-	"bytes"
 	"encoding/json"
 
-	"github.com/fatih/color"
-	"github.com/qri-io/deepdiff"
 	"github.com/qri-io/ioes"
 	"github.com/qri-io/qri/lib"
 	"github.com/spf13/cobra"
@@ -101,16 +98,14 @@ func (o *DiffOptions) Complete(f Factory, args []string) (err error) {
 
 // Run executes the diff command
 func (o *DiffOptions) Run() (err error) {
-	var stats, text string
-
 	p := &lib.DiffParams{
 		LeftPath:  o.Left,
 		RightPath: o.Right,
 		Selector:  o.Selector,
 	}
 
-	res := lib.DiffResponse{}
-	if err = o.DatasetRequests.Diff(p, &res); err != nil {
+	res := &lib.DiffResponse{}
+	if err = o.DatasetRequests.Diff(p, res); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


Mr0grog:
Oh wow, Neovim’s approach is kind of crazy and amazing. o_O

cmd/diff.go
```diff
@@ -119,25 +114,5 @@ func (o *DiffOptions) Run() (err error) {
...
-			text, err = deepdiff.FormatPrettyColor(res.Diff)
-			if err != nil {
-				return err
-			}
-		}
-	}
-	buf := bytes.NewBuffer([]byte(stats + "\n" + text))
-	printToPager(o.Out, buf)
-	return
+	return printDiff(o.Out, res, o.Summary)
```
Mr0grog: Looks like this is just encapsulating the diff printing code in a function with no other changes related to coloring? Just making sure I’m not missing anything.

b5: yep exactly. This just moves the `print` things from 3 files to 2. I'm trying to parse out my commit history better. It's a work in progress 😄

Issue #763, opened 2019-05-13
update service cleanup


b5:
Tracking issue to deal close out the first pass of `qri update` & implement [RFC0024](https://github.com/qri-io/rfcs/blob/master/text/0024-scheduled-updates.md)

Remaning TODOs from #754:
* [ ] `/update` api tests
* [ ] sanitize & test `qri update schedule` input parameters
* [x] os x `qri update service` daemonization (see #762)
* [ ] usability-tested `qri update` output (output will be a little rough for this first pass, refining after we land new `$PAGER` pagination):
  * [ ] need relative time-until and time-since execution for `qri update list` and `qri update logs` respectively
  * [x] qri update logs should output coloured title in red, green, grey for exit status (grey for `error: no changes detected`)
* [x] **multi-profile scheduling** I already need multi-profile scheduling, and we're definitely going to need it for doing bootstrap peer update pooling. I think the job struct should accept a an additional `configPath` param for job execution gives path to configuration file. This would be fed back to dataset updates as the global `--config` param. We'd need a test to make sure global config works.

Small Bugs that need fixing:
* [ ] `qri update schedule me/dataset` over RPC returns error EOF: _update: I've disabled this for now, should file an issue & fix later_
* [x] `qri update run` isn't working: `cannot run job type: `
* [ ] numbered repeats currently don't seem to be decrementing. `qri update schedule test.sh R1/P1W` is running multiple times, when it should only "repeat once"

### Incrementing counters
the unix datestamp prefixes qri logs should use a single incrementing integer prefix that ups for each scheduled run. In the future we could look at per-job-type integers. Or maybe we need to implement it now? I dunno. We could also punt this until later.
*update:* #788 adds incrementing counters!

PR #779, opened 2019-05-29
Cleanup update output

b5:
should address the "usability tested" checklist item in #763

Best to test this one by pulling the branch & building, then playing with update commands. Ideally the output makes things understandable. The less mental gymnastics, the better.

I've switched our human-formatting stuff to a package because we went from needing just one kind of "human" output (file sizes), to a few (relative dates, etc). `github.com/dustin/go-humanize` seems to fit the bill nicely.


b5:
I was hoping to put more work into this, but it's in a good place for now, and I'd like to get it in ahead of 0.8.0

PR #783, opened 2019-05-30
fix(fill.Struct): If present, use field tag as the field name

dustmop:


base/fill/struct.go
```diff
@@ -67,6 +67,15 @@ func putFieldsToTargetStruct(fields map[string]interface{}, target reflect.Value
 		// Lowercase the key in order to make matching case-insensitive.
 		fieldName := target.Type().Field(i).Name
 		lowerName := strings.ToLower(fieldName)
+		fieldTag := target.Type().Field(i).Tag
```
b5: This is _awesome_. Can we get a package-level comment in this file letting users know `fill` is sensitive to the json tag? adding `and obeys the "json" field tag if present`
```

// Package fill matches arbitrary values to struct fields using reflection
// fill is case-insensitive, and obeys the "json" field tag if present
// It's primary use is to support decoding data
// from a number of serialization formats (JSON,YAML,CBOR) into an intermediate
// map[string]interface{} value which can then be used to "fill" arbitrary struct
// values
package fill
```

PR #785, opened 2019-05-31
bring startf inline

b5:
Oooook, gomodules is forcing this issue on our startf separation anxiety. startf relies on qri. When using go modules it's a major antipattern to depend on test files across repositories, which is happening here. We could get away with this when we had GOPATH, not so much now, but I think the ultimate "right" answer is to have common test data in a common repository.

Let's just bite the bullet & move this package inline


b5:
@dustmop you had an outstanding PR to bring [docrun](https://github.com/qri-io/docrun) meta into the startf readme. I've added that here.

Also, does killing the startf repo make up for me creating apiutil? plzzzzz

dustmop:
Looks good to me!

Issue #726, opened 2019-04-03
switch to go modules


b5:
go-ipfs is set to release v0.4.20 this week, which will be the first release after the switch from gx to go modules. If you're interested in the history of the switch for go-ipfs, these issues are the biggies:
* [discussion around the switch](https://github.com/ipfs/go-ipfs/issues/5850)
* [switch implemented on go-ipfs](https://github.com/ipfs/go-ipfs/pull/6038)
* [v0.4.20 release](https://github.com/ipfs/go-ipfs/issues/6163)

Once v0.4.20 is out, I'd like _our_ next release to make the switch. It'll simplify our build story, and bring us back into line as a "stock" go project. We also have a timing window here: **when go 1.13 is released go modules will be enabled by default**. The release isn't until July, but I'd like to be well over to the go module ecosystem by then.

### Reading up on go modules
Start here: https://blog.golang.org/using-go-modules

### Planning the switch
I'd like this switch to cause minimal disruption to our day-to-day development, but _some_ disruption is necessary just to get everyone comfortable with a new way of working. Switching to go modules will require a PR on every repo we manage (everything under github.com/qri-io), so let's use that to our advantage: **everyone on Qri engineering needs to author at least one PR that switches a repo to go modules**.

We'll start at the bottom of the dependency tree and work up, leaving PRs unmerged until one is open & ready on all repos. We'll take a day to merge everything, and cut a patch release that same day. That patch release should include detailed notes on the switch, and instructions on how to transition to our new development process. Things like `make update-deps` and `make install` are all going to change (in implementation at least).

I've added go module support to [jsonschema](https://github.com/qri-io/jsonschema) & [jsonpointer](https://github.com/qri-io/jsonpointer), they're worth checking as a starting point.

### Releases & CHANGELOG.md

go modules are build on semantic versioning, so to do go modules right, we need to improve our story on subpackage releases. Basically this means doing to _all_ our packages what we currently do with `github.com/qri-io/qri`: add `make update-changelog`, generate a proper `CHANGLOG.md`, and issue pull requests that cut versions on a release cycle. As part of this project **we should aim for all projects to have `make update-changelog` and `CHANGELOG.md`**. Our conventional commit adherence is paying dividends here.

If a package has no prior release, let's start at v0.1.0

### gx

Packages that use gx will need some additional steps:
[...]

PR #786, opened 2019-05-31
chore: deprecate gx, switch to gomodules

b5:
this part of the qri-io/qri#726 mission


b5:
I'm going to close #726 with this, even tho we have a bunch of cleanup work to do, let's break it out into separate issues

PR #787, opened 2019-06-05
fix: update cleanup with better CLI output

b5:
fix some UX issues around update, make `qri update run` actually work. These changes can/should be better tested, but like, deadlines.


b5:
makes inroads on #763

PR #788, opened 2019-06-05
feat(cron): add incrementing job number, refactor job names

b5:
builds on #787, let's review & merge that one first.

After playing with cron for a bit myself, the UX of 'qri update log' needs a little work. I've added a RunNumber field to job that increments with each execution, which will allow us to bring cron into line with it's CI/github issue-inspired workflow. I do keep a rough tally of CI run numbers in my head (generally rounded to the second-largest 10's column), and look forward to having this hueristic in Qri update jobs.

I've also taken this last chance before we ship a release to refactor the field names of our job flatbuffer. This is a very permanent API, so the more right we can get things before release, the better. In this case I changed fields like "lastRunStart" to just "runStart", which more accurately depicts what state the field is tracking. The job itself is either unexecuted (and has no RunStart), executing (has a RunStart but no RunStop) or completed (has a RunStart and RunStop).

Finally, I've disambiguated the use of 'LastRunStart', which had different meanings depending on weather the job had executed or not. Now Job has an field called PrevRunStart that keeps, well, the timestamp of the last time a job was started.

BREAKING CHANGE:
Field names of cron.Job have been refactored, which will break repo/update/logs.qfb and repo/update/jobs.qfb files. Delete them to fix.. This will only affect users who have been building from source between releases.

### DIFF
--- a/lib/update_test.go
+++ b/lib/update_test.go
@@ -15,11 +15,14 @@ import (
 )
 
 func TestUpdateMethods(t *testing.T) {
+	prevDci := cron.DefaultCheckInterval
 	tmpDir, err := ioutil.TempDir("", "update_methods")
 	if err != nil {
 		t.Fatal(err)
+		cron.DefaultCheckInterval = prevDci
 	}
 	defer os.RemoveAll(tmpDir)
+	cron.DefaultCheckInterval = time.Millisecond * 500
 
 	cfg := config.DefaultConfigForTesting()
 	cfg.Update = &config.Update{Type: "mem"}
@@ -37,7 +40,7 @@ func TestUpdateMethods(t *testing.T) {
 		// this'll create type ShellScript with the .sh extension
 		Name: "testdata/hello.sh",
 		// run one time after one second
-		Periodicity: "R1/PT1S",
+		Periodicity: "R1/PT10S",
 	}
 	shellRes := &Job{}
 	if err := m.Schedule(shellJob, shellRes); err != nil {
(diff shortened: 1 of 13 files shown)


lib/update_test.go
```diff
@@ -15,11 +15,14 @@ import (
 )
 
 func TestUpdateMethods(t *testing.T) {
+	prevDci := cron.DefaultCheckInterval
 	tmpDir, err := ioutil.TempDir("", "update_methods")
 	if err != nil {
 		t.Fatal(err)
```
dustmop: I could be wrong, but I thought t.Fatal exits. So the line after this wouldn't run, right?

update/cron/cron.go
```diff
@@ -207,6 +213,8 @@ func (c *Cron) Schedule(ctx context.Context, job *Job) error {
 		return err
 	}
 
+	// TODO (b5) - check for proir job & inhert the previous run number
```
dustmop: proir -> prior. inhert -> inherit

update/cron/cron.go
```diff
@@ -152,7 +153,7 @@ func (c *Cron) Start(ctx context.Context) error {
 	}
 
 	// initial call to check
-	go check(ctx)
+	// go check(ctx)
```
dustmop: Can this be deleted?

lib/update_test.go
b5: yep! Desired behaviour. Fatal exits the _test_, not the process. If we can't get a temp dir to write to, we shouldn't run this test

PR #792, opened 2019-06-06
docs(viz): add outline comment for HTML viz templating

b5:
this includes a copy-paste from the lower-level qri-io/dataset/dsviz outline comment. I couldn't find a nice place to put it, so I've added the comment inline, in `qri-io/qri/base.Render`


b5:
I'ma merge this b/c it's already up at https://qri.io/docs

Issue #782, opened 2019-05-30
Installing qri, then running list before setup causes a segfault


dustmop:
List should print an error instead. Need to check whatever is being dereferenced to cause the segfault (I think it's config).

dustmop:
Fixed by #793.

PR #793, opened 2019-06-06
fix(config): lib.NewInstance option funcs must check for nil pointers

dustmop:


dustmop:
Fixes https://github.com/qri-io/qri/issues/782.

PR #797, opened 2019-06-10
feat(startf): pass present dataset to download special function

b5:
I've encountered one-too-many situations where I need access to the present dataset version in the download step to make a proper decision about what network requests are needed, so I'd like make this change now. Normally I'd write a proper RFC for this, but we're extremely crunched for time at the moment trying to produce teaching materials for others. adding a ds argument to download opens up a number of useful techniques and removes at least two steps from the learning curve of the tutorial I'm presently working on.  This change modifies our starlark `download` special function signature from this:
```python
def download(ctx):
```

to this:
```python
def download(ds, ctx):
```

Where the new ds arg is the current dataset version.

By adding ds to the download step, doing append-only fetching is possible, which cuts down on convoluted logic of overfetching-then-matching in the transform step. On top of this, it's possible to drop the transform step entirely by calling ds.set_body in the download step, pushing the entire conversation about context into a more advanced lesson.

Our primary concern for providing a ds argument in the download step has always been security. In a future world where datasets come with an access control layer, we don't want anyone to be able to write a script that publishes private data to a remote location over HTTP in the download step. Our recent work on storing separate read & write dataset pointers and advances in our understanding of what we can and can't do with starlark have me far less worried about this situation. We have all the tools necessary today to pick up when a user is attempting to access dataset data in the download step, and can interrupt that read attempt based on whatever criteria we see fit.

BREAKING CHANGE:
starlark transform download function has changed from 'download(ctx)` to 'download(ds,ctx)`, wehre the ds argument is the dataset being transformed

### DIFF
--- a/startf/testdata/fetch.star
+++ b/startf/testdata/fetch.star
@@ -1,7 +1,7 @@
 load("http.star", "http")
 load("qri.star", "qri")
 
-def download(ctx):
+def download(ds, ctx):
   res = http.get(test_server_url)
   return res.json()['foo']
 
--- a/startf/transform.go
+++ b/startf/transform.go
@@ -13,10 +13,10 @@ import (
 	"github.com/qri-io/qfs"
 	"github.com/qri-io/qri/p2p"
 	"github.com/qri-io/qri/repo"
-	"github.com/qri-io/starlib"
 	skyctx "github.com/qri-io/qri/startf/context"
 	skyds "github.com/qri-io/qri/startf/ds"
 	skyqri "github.com/qri-io/qri/startf/qri"
+	"github.com/qri-io/starlib"
 	"go.starlark.net/resolve"
 	"go.starlark.net/starlark"
 )
@@ -164,13 +164,16 @@ func ExecScript(next, prev *dataset.Dataset, opts ...func(o *ExecOpts)) error {
 		return err
 	}
 
+	ds := skyds.NewDataset(t.prev, t.checkFunc)
+	ds.SetMutable(t.next)
+
 	funcs, err := t.specialFuncs()
 	if err != nil {
 		return err
 	}
 
 	for name, fn := range funcs {
-		val, err := fn(t, thread, ctx)
+		val, err := fn(t, thread, ds, ctx)
 
 		if err != nil {
 			if evalErr, ok := err.(*starlark.EvalError); ok {
@@ -182,7 +185,7 @@ func ExecScript(next, prev *dataset.Dataset, opts ...func(o *ExecOpts)) error {
 		ctx.SetResult(name, val)
 	}
 
-	err = callTransformFunc(t, thread, ctx)
+	err = callTransformFunc(t, thread, ds, ctx)
 	if evalErr, ok := err.(*starlark.EvalError); ok {
 		return fmt.Errorf(evalErr.Backtrace())
 	}
@@ -247,9 +250,9 @@ func (t *transform) specialFuncs() (defined map[string]specialFunc, err error) {
 	return
 }
 
-type specialFunc func(t *transform, thread *starlark.Thread, ctx *skyctx.Context) (result starlark.Value, err error)
+type specialFunc func(t *transform, thread *starlark.Thread, ds *skyds.Dataset, ctx *skyctx.Context) (result starlark.Value, err error)
 
-func callDownloadFunc(t *transform, thread *starlark.Thread, ctx *skyctx.Context) (result starlark.Value, err error) {
+func callDownloadFunc(t *transform, thread *starlark.Thread, ds *skyds.Dataset, ctx *skyctx.Context) (result starlark.Value, err error) {
 	httpGuard.EnableNtwk()
 	defer httpGuard.DisableNtwk()
 	t.print("📡 running download...\n")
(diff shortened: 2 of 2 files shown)


b5:
Example transform script that would make use of this to build a bibliography:

```python
load("http.star", "http")
load("bsoup.star", "bsoup")

urls = ["https://arxiv.org/abs/1906.02702"]

# download is a special function that provides HTTP access
# qri will call download when saving a dataset version
def download(ds, ctx):
  # get the previous dataset body if one exists, defaulting to 
  # an array if no body is specified
  body = ds.get_body(default=[])
  # fetch is a list of new urls to grab
  fetch = ctx.get_config("urls")
  # remove any urls that have already been fetched from the list:
  fetch = unfetched_urls(body, fetch)
  print("fetching %d urls\n" % len(fetch))

  # fetch new urls, adding them to the end of the list
  body = body + [{ "article" : extract(url), "annotations": { "notes" : "", "rating" : 0 } } for url in fetch]

  # finally, update the dataset with new data by calling set_body
  ds.set_body(body)

def transform(ds, ctx):
  

"""
helper functions:
"""
def unfetched_urls(body, fetch):
  return [url for url in fetch if missing_url(body, url)]

def missing_url(body, url):
  for citation in body:
    if citation.get('article', {}).get('url', '') == url:
      return False
  return True

def extract(url):
  res = http.get(url)
  if res.status_code != 200:
    error("bad response: %d" % res.status_code)
  soup = bsoup.parseHtml(res.body())
  return extract_arxiv(url, soup)


def extract_url(url, soup):
  return {
    'url' : url,
    'about' : '',
    'author' : '',
    'wordCount' : '',
    'name' : '',
    'description' : '',
  }

def extract_arxiv(url, soup):
  return {
    'url' : url,
    'about' : '',
    'author' : '',
    'name' : remove_prefix(soup.find('h1', { 'class': 'title'}).get_text(), "Title:"),
    'description' : remove_prefix(soup.find('blockquote', { 'class': 'abstract'}).get_text(), "Abstract:"),
  }

def remove_prefix(s, prefix):
    return s[len(prefix):] if s.startswith(prefix) else s
```

dustmop:
I'd push back that we shouldn't take this approach. My fundamental concern is of user expectations, specifically around our assertion that starlark is trustable and always safe to run. Benefits of starlark are not only that it is sandboxed and lacks facilities like filesystem access, but also that it is non-Turing complete is can't run forever or use infinite memory. In addition, there are situations where a user has datasets that are not necessarily "private", but are still thought of as being "offline" until the user pushes them to the dweb by running `qri connect`. Due to situations like these, I believe it shouldn't be possible for starlark to leak information about datasets to foreign servers via query parameters in URLs. This is why, so far, we've disallowed the `download` function from having any access at all the `ds` parameter. Basically, "download" cannot "upload".

Of course, there's use cases where myself, using qri, actually wants to use extract something such as urls from my own dataset in order to fetch them. In situations like these, we should have some facility for feeding data into the `download` function. However this works, it should be only allowed on an opt-in basis, and should never be the default behavior, in order to respect user's expectations about the safety of running unknown starlark code. Furthermore, this opt-in should not simply be a yes/no prompt (which *everyone* ignores) but rather a phrase or flag that the user needs to specifically type out.

My proposal is a new special function called `expose`, which runs before `download`, which is able to extract information from the dataset that `download` can then use for itself. The function `expose` is only able to run if the `qri` command is run using a flag `--transform-expose-all` that opt-ins to this feature. In the future, `expose` should instead be an instance of whatever path selector syntax we end up landing on. Inside the `download` function, getting the results of expose would be done using `ctx.expose`, the same way we carry on download's return value to `transform` using `ctx.download`.

There's still an additional step to getting data into the `download` function, but I believe this is actually a good thing to have in a how-to explanation. It demonstrates that the `download` function has a specific purposes, and it's own limitations are a consequence of us thinking carefully of how data-flow movement should match user expectations.

Example adapted as described:

```
# Running this transform requires the --transform-expose-all flag
def expose(ds):
  # get the previous dataset body if one exists, defaulting to 
  # an array if no body is specified
  body = ds.get_body(default=[])
  # expose the url fields in each citation
  return [citation.get('article', {}).get('url', '') for citation in body]
  

# download is a special function that provides HTTP access
# qri will call download when saving a dataset version
def download(ctx):
  # fetch is a list of new urls to grab
  urls = ctx.get_config("urls")
  # remove any urls that have already been fetched from the list:
  fetch_urls = list(set(urls) - set(ctx.expose))
  print("fetching %d urls\n" % len(fetch))
  # fetch new urls, adding them to the end of the list
  results = [{ "article" : extract(url), "annotations": { "notes" : "", "rating" : 0 } } for url in fetch]
  return results


def transform(ds, ctx):
  ... ctx.download ...
```

b5:
After thinking it through, I agree with you @dustmop. This hueristic strikes me as an expectation I'd come to Qri with if I'd never used it before:

> Basically, "download" cannot "upload".

I agree we should strive to preserve this, and require deviating from default settings to

To me this represents a chance to stop & think on my own part. In the commit history of this branch I wrote the following:

>  Normally I'd write a proper RFC for this, but we're extremely crunched for time at the moment trying to produce teaching materials for others.

I made a mistake here, and skipping the RFC was it. Let's not skip process because of time pressure. I'll adjust the learning materials to not require this feature. From there I think we should write an RFC for `expose`, taking the `expose` approach.  @dustmop you mind taking the lead?

PR #804, opened 2019-06-12
fix(webapp): must resolve path before passing path to frontend webapp handler

ramfox:
two funky things were happening when trying set up the `FrontendHandler`

1) `s.resolveWebappPath` was running in a go routine. The `webappEntrypointHash` that was being passed to the `/webapp` handler was therefore old! When the frontend webapp went to make certain requests, such as lazy loading parts of the app, it was being redirected to an incorrect endpoint.

2) the new webapp path was not being saved to the config! This means everytime we `qri connect`, the webapp was being 'updated'

This refactor removes the go routine and saves the new webapp path correctly!

It also adds some logging to the FrontendHandler, noting when we are fetching off the distributed web.

This pr also removes all the ephemeral `qri connect` flags

### DIFF
--- a/cmd/connect.go
+++ b/cmd/connect.go
@@ -38,17 +38,7 @@ peers & swapping data.`,
 		},
 	}
 
-	cmd.Flags().IntVarP(&o.DisconnectAfter, "disconnect-after", "", 0, "duration to keep connected in seconds, 0 means run indefinitely")
-
-	cmd.Flags().BoolVarP(&o.DisableAPI, "disable-api", "", false, "disables api, overrides the api-port flag")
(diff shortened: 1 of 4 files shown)


cmd/connect.go
```diff
@@ -38,17 +38,7 @@ peers & swapping data.`,
 		},
 	}
 
-	cmd.Flags().IntVarP(&o.DisconnectAfter, "disconnect-after", "", 0, "duration to keep connected in seconds, 0 means run indefinitely")
```
b5: yep yep yep. I'm glad this is happening. It'll greatly reduce the complexity around the source of truth when it comes to configuration b/c we no longer need to think about "ephemeral config modification"

PR #809, opened 2019-06-17
fix(webapp loading):  update, fetch, and pin webapp on `qri connect` 

ramfox:
When a user runs `qri connect`, we should 1) checking to see if the current webapp is out of data, 2) trying to fetch and pin a new webapp 3) have the frontend handler send back a temporary script if the webapp has not finished updating 4) send the correct script once it has updated.

This commit creates a go routine that checks is the webapp has updated, and tries to fetch and pin that updated website. This go routine communicates on a channel, to update a `updating` bool. The handler uses this `updating` var to determine if it should send back the temporary script, or the actual webapp.

### DIFF
--- a/api/webapp.go
+++ b/api/webapp.go
@@ -63,10 +63,12 @@ func (s Server) FrontendHandler(prefix string) http.Handler {
 	// * and really the data could be owned by Server and initialized by it,
 	//  as there's nothing that necessitates updating the webapp within FrontendHandler.
 	updating := true
-	updates := make(chan bool)
+	var fetchErr error
+	errs := make(chan error)
 	go func() {
 		if err := s.resolveWebappPath(); err != nil {
-			log.Errorf("error resolving webapp path: %s", err)
+			errs <- err
+			return
 		}
 		path := s.Config().Webapp.EntrypointHash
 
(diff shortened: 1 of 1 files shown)


api/webapp.go
```diff
@@ -59,15 +61,57 @@ func (s Server) FrontendHandler(prefix string) http.Handler {
 	// TODO -
 	// * there's no error handling,
```
b5: I think we need to handle errors here. We can do it by making `updates` a channel of error instead of boolean values. starting on line 65:
```go
     updating := true
     var fetchError error
     errs := make(chan error)
	go func() {
		if err := s.resolveWebappPath(); err != nil {
	          errs <- err
                  return
		}
		path := s.Config().Webapp.EntrypointHash

		// TODO (ramfox): frontend handler should accept a context
		log.Info("fetching webapp off the distributed web...")
		fetcher, ok := s.Repo().Store().(cafs.Fetcher)
		if !ok {
			errs <- fmt.Error("this store cannot fetch from remote sources")
			return
		}
		if _, err := fetcher.Fetch(cafs.SourceAny, path+"/main.js"); err != nil {
			errs <- fmt.Errorf("error fetching file: %s", err.Error())
			return
		}
		log.Info("pinning webapp...")
		pinner, ok := s.Repo().Store().(cafs.Pinner)
		if !ok {
			errs <- fmt.Error("unable to pin webapp")
			return
		}
		if err := pinner.Pin(path, true); err != nil {
			errs <- fmt.Errorf("error pinning webapp: %s", err.Error())
			return
		}
		log.Info("done pinning webapp!")
		errs <- nil
	}()

	go func() {
	   fetchErr = <-errs
           if fetchErr != nil {
               log.Errorf("error fetching webapp:  %s", fetchErr)
           }
          updating = false
	}()
```

Also, I think we should avoid references to 'your repository'. It might be the user's repository, but we can't always assume as much (example: when tests are running, it's not really "your" repo).

api/webapp.go
```diff
@@ -59,15 +61,57 @@ func (s Server) FrontendHandler(prefix string) http.Handler {
...
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if updating == true {
+			// if the app is still pinning
+			// return a temporary script that tells the user what's going on
+			w.Write([]byte(loadingScript))
+			return
+		}
 		path := fmt.Sprintf("%s%s", s.Config().Webapp.EntrypointHash, strings.TrimPrefix(r.URL.Path, prefix))
-		log.Info("fetching webapp off the distributed web...")
 		s.fetchCAFSPath(path, w, r)
```
b5: With the errors channel change from above in place, we'll now need to check if the value of `fetchErr` is non-ni here (after confirming `updating` is false, before `fetchCAFSPath`), and print an error for the user if so.

PR #818, opened 2019-07-02
fix(use): Move `use` up to cmd/ from lib/. Delete select from repo/

dustmop:


lib/lib.go
```diff
@@ -70,7 +70,6 @@ func Receivers(inst *Instance) []Methods {
 		NewConfigMethods(inst),
```
b5: can I haz version bump to `0.8.3-dev` in this file? kthxbye

cmd/get.go
```diff
@@ -92,7 +92,15 @@ func (o *GetOptions) Complete(f Factory, args []string) (err error) {
 			args = args[1:]
 		}
 	}
-	o.Refs = args
+	o.Refs = make([]string, 0, len(args))
```
b5: this has a funny side effect if no args are provided, which, for `qri use`, is kinda the point 😄

repo/ref.go
```diff
@@ -31,17 +31,6 @@ type Refstore interface {
 	RefCount() (int, error)
 }
 
-// ErrRefSelectionNotSupported is the expected error for when RefSelector interface is *not* implemented
```
b5: Hrm... I want to stop & think before we remove this. I'm a _huge_ fan of the changes that are happening in lib & cmd, but think the `RefSelector` interface should stay, and that cmd should use this interface to persist instead of talking directly to the filesystem.

I think subtle bugs will show up around _use_ when we start to undermine the assumption that the qri repo is stored directly on the filesystem. I'm thinking of situations where we're talking to a remote _repo_ that supports selection. Far off? yes. But The intent of this interface is to maintain an abstract "lock" around assets in the qri repo that is owned by the repo interface.

I'd love to see `cmd` use this interface, and nothing else.

cmd/get.go
dustmop: Yeah, this is a bug, GetDatasetRefString is never called in that case. I'll fix.

repo/ref.go
dustmop: I hadn't considered that approach, I'll take a look at implementing it. The problem is that `cmd/` doesn't always have a reference to a `repo` object, as the factory doesn't construct one if `qri connect` is running, and since it's above `lib/`, this code path doesn't have the ability to route a call through RPC. But I guess it could just implement the interface some other way.

b5: you should be able to do `lib.Instance.Repo()`, it'll require access to the `lib.Instance` anywhere that wants to do qri use, but we're trying to go in that direction anyway, getting away from the factory pattern.

PR #823, opened 2019-07-09
 feat(qri-ref): Commands to link a working directory to a dataset

dustmop:


cmd/init.go
```diff
@@ -0,0 +1,187 @@
...
+		// TODO(dlong): Instead, import the body.json file for the new dataset
+		return fmt.Errorf("cannot initialize new dataset, body.json exists")
+	}
+
+	// Suggestion for the dataset name defaults to the name of the current directory.
+	pwd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+	suggestDataset := filepath.Base(pwd)
```
b5: let's run this through [varname](https://github.com/qri-io/varName/)

cmd/init.go
```diff
@@ -0,0 +1,187 @@
...
+		Selector: "",
+	}
+	res := lib.GetResult{}
+	if err = o.DatasetRequests.Get(&p, &res); err == nil {
+		// TODO(dlong): Tell user to use `checkout` if the dataset already exists in their repo?
+		return fmt.Errorf("a dataset with the name %s already exists in your repo", ref)
+	}
+
+	// Validate dataset format
+	if dsFormat != "csv" && dsFormat != "json" {
```
b5: we're so close to having a nice method for checking these strings in dataset: https://godoc.org/github.com/qri-io/dataset#SupportedDataFormats I'll PR in a new package function for checking if a string is a supported Data format

cmd/use.go
```diff
@@ -16,6 +16,9 @@ import (
 // FileSelectedRefs stores selection, is copied from github.com/qri-io/qri/repo/fs/files.go
 const FileSelectedRefs = "/selected_refs.json"
 
+// QriRefFilename links the current working folder to a dataset by containing a ref to it.
+const QriRefFilename = ".qri-ref"
```
b5: ugh I hate to bikeshed, but I think this should be `.qri_ref` to keep consistent to our file naming conventions from `.qri` repo, which we inherited from go language conventions. fight me

cmd/use.go
```diff
@@ -144,6 +154,16 @@ func GetDatasetRefString(f Factory, args []string, index int) (string, error) {
 	return refs[0], nil
 }
 
+// GetLinkedFilesysRef returns whether the current directory is linked to a dataset in your repo,
+// and the reference to that dataset.
+func GetLinkedFilesysRef() (bool, string) {
```
b5: flip the return values to follow the go "comma, ok" pattern

cmd/use.go
```diff
@@ -144,6 +154,16 @@ func GetDatasetRefString(f Factory, args []string, index int) (string, error) {
...
+	data, err := ioutil.ReadFile(QriRefFilename)
+	if err == nil {
+		return true, string(data)
```
b5: need to `return strings.TrimSpace(string(data), true`. Newlines in that file are hurting things. Text editors often auto-add newline endings to files these days 🙍‍♂

cmd/use.go
dustmop: If we're going to change it, the rfc should be updated to match. It is currently using ".qri-ref", and I feel like the rfc should be the place to bikeshed about that. It is still in review.

Arguments in favor of ".qri-ref":
1) Though our source code is written in go, the external interface that users see on their filesystem is not, itself, the same as go. In other words, a separate reimplementation of our binary in another programming language would not follow go language conventions.
2) Following git for inspiration: in the .git repo, dash (-) is used for separating words, such as "packed-refs", "hooks/pre-commit", *unless* the words are entirely uppercase, like "ORIG_HEAD" and "COMMIT_EDITMSG".

Arguments in favor of ".qri_ref":
3) We should be consistent with the convention established in "~/.qri/". This alone seems reason enough to dismiss (1) and (2), if you want to go forward with this change.

dustmop: I should, not that it's very important, but my personal preference is dash (-) to separate words over (_), only because it doesn't require the shift key to type on my keyboard. It's what lisp uses and I've always liked it better. Fwiw.

b5: thx for comments on process, totally agreed we should debate this stuff over in RFC land. I'll move the discussion over there. Great points made here tho 😃

cmd/use.go
dustmop: Oops, correct, fixing...

PR #827, opened 2019-07-17
feat(status): Status UI. RefSelect to simplify handling references

dustmop:
Simplify how references are handled when implied by .qri-ref or use. Necessary for improving the UI of the status command.


cmd/ref_select.go
```diff
@@ -0,0 +1,116 @@
...
+// GetCurrentRefSelect returns the current reference selection
+func GetCurrentRefSelect(f Factory, args []string, allowed int) (*RefSelect, error) {
+	// TODO(dlong): Respect `allowed`, number of refs the command uses. -1 means any.
+	// TODO(dlong): For example, `get` allows -1, `diff` allows 2, `save` allows 1
+	// If reference is specified by the user provide command-line arguments, use that reference.
+	if len(args) > 0 {
+		return NewExplicitRefSelect(args[0]), nil
+	}
+	// If in a working directory that is linked to a dataset, use that link's reference.
+	dir, err := os.Getwd()
```
b5: This looks a lot like `GetLinkedRefSelect` from below, can we use that here instead?

cmd/status.go
```diff
@@ -37,48 +35,47 @@ func NewStatusCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
...
-	if len(res) == 0 {
-		printSuccess(o.ErrOut, "working directory clean!")
-		return nil
-	}
-
+	clean := true
+	valid := true
 	for _, si := range res {
 		if si.Type != "unmodified" {
+			// TODO(dlong): Why write to stderr?
```
b5: yeah.... should be stdout here. These prints are the "subject" of the command

dustmop:
Okay, pushed a commit that fixes your comments, and performs a few minor cleanups as well.

b5:
so good. want merge.

PR #829, opened 2019-07-19
feat(replace-save,removed-status): report removed components in qri status, replace-save support

b5:
First, we now check status mapping in _both_ directions, which means we can report removed components as well as added ones. There are some idiosyncrasies here that I think we need to introduce terminology to describe. Yay Spec(ish) time!

### 'Bound' Components
There are components of a dataset who's existence or lack thereof is "bound" to some other factor. Our document model enforces these restrictions to keep datasets working properly. Immediate examples: 

* the `structure` component is bound to the existence of a body. If a body exists, structure MUST exist. if a body doesn't exist, structure MAY exist.
* `commit` is bound to a dataset version existing. if a dataset version exists, it MUST have a commit component.
* the `dataset` itself is bound to any component existing it's not possible to not have a "dataset" component. Getting a little existential, there is no dataset component if no dataset components exist.

This matters when reporting the status of a dataset, because the user has no control over these components, and so we need to treat the edge cases properly. For example, if a user deletes their "structure.json" file, but still has a structure, we shouldn't report it as "removed", because it can't be. We need to either report the fact that structure will be inferred on save, or we should consider this an "implementation detail" and report nothing at all. Personally I'm a fan of the former.

There is a connection between _bound components_ and  _derived values_. To distinguish the two:
* if a bound component doesn't exist, but must, we will _infer_ one for the user based.
* all derived values will ALWAYS be derived, and cannot be wrong

Inference is different from derivation because it can be "wrong" (where "wrong" is _not what the user intended_). We can infer the wrong schema. We can't derive the wrong checksum.

I should point out that "bound" is a descriptive term. We don't need to explicitly write this stuff in code the way we do with, say, derived values.

### Replace Save
`actions.SaveDataset` has been refactored to accept a struct of switches. That struct has a new switch `Replace` that instructs qri to only use input data when creating a dataset commit. This has the effect of "replacing" the dataset instead of applying the input dataset as modifications to the prior version.

We need this for FSI, which operates on a file-system-reflects-current-state paradigm.

### DIFF
--- a/actions/dataset.go
+++ b/actions/dataset.go
@@ -16,13 +16,26 @@ import (
 	"github.com/qri-io/qri/repo/profile"
 )
 
+// SaveDatasetSwitches provides togglable flags to SaveDataset that control 
+// save behaviour
+// dryRun, pin, convertFormatToPrev, force, shouldRender bool
+type SaveDatasetSwitches struct {
+	Replace bool // 
+	DryRun bool
+	Pin bool
+	ConvertFormatToPrev bool
+	Force bool
+	ShouldRender bool
+}
+
 // SaveDataset initializes a dataset from a dataset pointer and data file
-func SaveDataset(node *p2p.QriNode, changes *dataset.Dataset, secrets map[string]string, scriptOut io.Writer, dryRun, pin, convertFormatToPrev, force, shouldRender bool) (ref repo.DatasetRef, err error) {
+func SaveDataset(node *p2p.QriNode, changes *dataset.Dataset, secrets map[string]string, scriptOut io.Writer, sw SaveDatasetSwitches) (ref repo.DatasetRef, err error) {
 	var (
 		prevPath string
 		pro      *profile.Profile
 		r        = node.Repo
 	)
+
 
 	prev, mutable, prevPath, err := base.PrepareDatasetSave(r, changes.Peername, changes.Name)
 	if err != nil {
@@ -33,7 +46,7 @@ func SaveDataset(node *p2p.QriNode, changes *dataset.Dataset, secrets map[string
 		return
 	}
 
-	if dryRun {
+	if sw.DryRun {
 		node.LocalStreams.PrintErr("🏃🏽‍♀️ dry run\n")
 		// dry-runs store to an in-memory repo
 		r, err = repo.NewMemRepo(pro, cafs.NewMapstore(), node.Repo.Filesystem(), profile.NewMemStore(), nil)
@@ -61,7 +74,7 @@ func SaveDataset(node *p2p.QriNode, changes *dataset.Dataset, secrets map[string
 	}
 
 	if changes.BodyFile() != nil && prev.Structure != nil && changes.Structure != nil && prev.Structure.Format != changes.Structure.Format {
-		if convertFormatToPrev {
+		if sw.ConvertFormatToPrev {
 			var f qfs.File
 			f, err = base.ConvertBodyFormat(changes.BodyFile(), changes.Structure, prev.Structure)
 			if err != nil {
@@ -77,9 +90,11 @@ func SaveDataset(node *p2p.QriNode, changes *dataset.Dataset, secrets map[string
 		}
 	}
 
-	// apply the changes to the previous dataset.
-	mutable.Assign(changes)
-	changes = mutable
+	if !sw.Replace {
+		// apply the changes to the previous dataset.
+		mutable.Assign(changes)
+		changes = mutable
+	}
 
 	// infer missing values
 	if err = base.InferValues(pro, changes); err != nil {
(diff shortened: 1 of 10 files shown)


b5:
cc @chriswhong & @ramfox for the description of bound components and a weigh in on how status should report

lib/datasets.go
```diff
@@ -208,6 +209,9 @@ type SaveParams struct {
 	FilePaths []string
 	// secrets for transform execution
 	Secrets map[string]string
+	// Replace writes the entire given dataset as a new snapshot instead of 
+	// applying save params as agumentations to the existing history
```
dustmop: agumentations -> augmentations

lib/lib_test.go
```diff
@@ -162,7 +162,8 @@ func addCitiesDataset(t *testing.T, node *p2p.QriNode) repo.DatasetRef {
 	ds.Name = tc.Name
 	ds.BodyBytes = tc.Body
 
-	ref, err := actions.SaveDataset(node, ds, nil, nil, false, true, false, false, true)
+	// dryRun, pin, convertFormatToPrev, force, shouldRender bool
```
dustmop: delete this line

actions/dataset.go
```diff
@@ -16,14 +16,27 @@ import (
 	"github.com/qri-io/qri/repo/profile"
 )
 
+// SaveDatasetSwitches provides togglable flags to SaveDataset that control
```
dustmop: togglable  -> toggleable

actions/dataset.go
```diff
@@ -16,14 +16,27 @@ import (
 	"github.com/qri-io/qri/repo/profile"
 )
 
+// SaveDatasetSwitches provides togglable flags to SaveDataset that control 
+// save behaviour
+// dryRun, pin, convertFormatToPrev, force, shouldRender bool
```
dustmop: delete this line

actions/dataset.go
```diff
@@ -77,24 +90,26 @@ func SaveDataset(node *p2p.QriNode, changes *dataset.Dataset, secrets map[string
 		}
 	}
 
-	// apply the changes to the previous dataset.
-	mutable.Assign(changes)
-	changes = mutable
+	if !sw.Replace {
+		// apply the changes to the previous dataset.
```
dustmop: I like using the language "patch" to describe the non-replace behavior. Would put that here for future reference, by changing the comment to "Treat the changes as a set of patches applied to the previous dataset"

cmd/status.go
```diff
@@ -64,8 +65,14 @@ func (o *StatusOptions) Run() (err error) {
...
 	for _, si := range res {
-		if si.Type != "unmodified" {
-			printErr(o.Out, fmt.Errorf("%s: %s (source: %s)", si.Type, si.Path, filepath.Base(si.SourceFile)))
+		switch si.Type {
+		case fsi.STRemoved:
+			printErr(o.Out, fmt.Errorf("  %s:  %s", si.Type, si.Component))
+			clean = false
+		case fsi.STUnmodified:
+			// noop
+		default:
```
dustmop: Instead of using `default`, make this explicit:

```
case fsi.STAdd, fsi.STChange:
```

fsi/status.go
```diff
@@ -4,28 +4,52 @@ import (
...
 var (
 	// STUnmodified is "no status"
 	STUnmodified = "unmodified"
 	// STAdd is an added component
 	STAdd = "add"
 	// STChange is a modified component
 	STChange = "modified"
-	// STRemove is a removed component, currently not really supported?
-	STRemove = "remove"
+	// STRemoved is a removed component, currently not really supported?
```
dustmop: Update this comment, we are handling removed now, right?

fsi/status.go
```diff
@@ -73,23 +97,39 @@ func (fsi *FSI) Status(dir string) (changes []StatusItem, err error) {
...
 	// if err = validate.Dataset(ds); err != nil {
 	// 	return nil, fmt.Errorf("dataset is invalid: %s" , err)
 	// }
 
+	storedComponents := dsComponents(stored)
+
+	for cmpName := range storedComponents {
+		// when reporting deletes, ignore "bound" components that must/must-not
+		// exist based on external conditions
+		if cmpName != componentNameDataset && cmpName != componentNameStructure && cmpName != componentNameCommit {
```
dustmop: Does this mean that if a user deletes the structure but changes nothing else, then `status` won't report any changes? Is that intentional, or are we only trying to handle the case of deleting both body and structure leading to a single message?

b5: yeah, it's true that deleting structure in this case won't produce a change. I think we should handle that scenario in a yet-to-be-written subroutine that checks for values that will be inferred & creates status items for those.

b5:
okok @dustmop said this is good to go on slack

PR #851, opened 2019-07-29
refactor(repo): move fsi link storage to refstore, use flatbuffers in fs.Refstore

### LINKED ISSUE #844: [8] Refactor FSI link store into repo.Refstore
The origin of our need for separate `/fsi` endpoints comes from not storing FSI details in the refstore, let's move 'em there, which will set the stage for removing the `/fsi` prefix.

b5:
closes #844

In an ideal world this refactor wouldn't include transitioning `ds_refs.json` to flatbuffer storage, but I wanted to get that upgrade in as a contribution from our FSI work.

This change does break an assumption repo.Refstore used to be able to make. In the past it was true that _all stored references had paths_. This is no longer the case. All stored

I haven't put enough thought into deleting references. I'd like to follow up on that topic with subsequent PR(s).


cmd/cmd_test.go
```diff
@@ -925,7 +924,9 @@ func (r *TestRepoRoot) CreateCommandRunner(ctx context.Context) *cobra.Command {
 	r.streams = ioes.NewIOStreams(in, out, out)
 	setNoColor(true)
 
-	return NewQriCommand(ctx, r.pathFactory, r.testCrypto, r.streams)
+	cmd := NewQriCommand(ctx, r.pathFactory, r.testCrypto, r.streams)
+	cmd.SetOutput(out)
```
dustmop: This fixes most of the problems with writing to stdout, but we should also change cmd/export.go to write to o.Out instead of fmt.Fprintf on [line 111](https://github.com/qri-io/qri/blob/master/cmd/export.go#L111). For some reason that line still prints out if I run the tests with "-v". Also, this commit is unrelated to the rest of this PR, could we move to a different PR?

b5: agreed, I'll submit a separate PR for this

repo/fs/refstore.go
```diff
@@ -97,106 +102,92 @@ func (n Refstore) GetRef(get repo.DatasetRef) (repo.DatasetRef, error) {
...
 		return 0, err
 	}
-	return len(names), nil
+	return len(refs), nil
 }
 
-func (n *Refstore) names() ([]repo.DatasetRef, error) {
-	data, err := ioutil.ReadFile(n.filepath(n.file))
+func (rs *Refstore) refs() (repo.Refs, error) {
+	data, err := ioutil.ReadFile(rs.filepath(rs.file))
```
dustmop: We should consider cacheing the deserialized refs, then only reread the ref file is it has a greater mtime than the last time we read it.

fsi/fsi_test.go
```diff
@@ -219,40 +222,40 @@ func TestUpdateLink(t *testing.T) {
 	}
 }
 
-func TestRefLink(t *testing.T) {
-	paths := NewTmpPaths()
-	defer paths.Close()
+// func TestRefLink(t *testing.T) {
```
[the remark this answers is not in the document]
b5: A subsequent PR removes this test entirely: #855. I'd like to leave it for now because I'm lazy

repo/fs/refstore.go
b5: Great idea. Filed #857

PR #860, opened 2019-08-01
fix(status): Parse errors should also be shown for schema

dustmop:


cmd/fsi_test.go
```diff
@@ -520,6 +520,74 @@ fix these problems before saving this dataset
...
+
+	// Save a dataset containing a body.json and meta component
+	cmdR := r.CreateCommandRunner(ctx)
+	err = executeCommand(cmdR, "qri save --body=testdata/movies/body_ten.csv me/ten_movies")
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
+
+	// Change to a temporary directory.
+	os.Chdir(rootPath)
```
b5: I'm foreseeing us doing a _lot_ of "isolated directory tests" for FSI integration checks. we should think about adding something to `TestRepoRoot` that includes a "dataset directory" & methods to change to / from it.

I think this is a great start, this suggestion would make a nice follow-up PR

PR #865, opened 2019-08-07
chore(ci): Fix circleci by forcing correct timestamp tags

dustmop:


dustmop:
See https://github.com/golangci/golangci-lint/issues/595.

dustmop:
Absolutely.

First off, we're building with go modules, and our Circle CI needs to use GOPROXY="https://proxy.golang.org" in order to work around an unrelated issue with badger (see https://github.com/dgraph-io/badger/issues/886 and especially https://github.com/dgraph-io/badger/issues/904). At some point in the future we should upgrade our dependencies, mostly ipfs, so that we are on a newer badger, but that will take some time to get right.

The failures we started seeing this week on CircleCI were related to the goproxy (proxy.golang.org) returning 410 for a number of modules. Most other issues concerning this error recommended turning off GOPROXY in order to fix it, but due to the problem above, we are unable to do so. As far as I can tell, both badger and the golangci projects breakages were due to modifying commits or tags for an already existing name, which invalidates certain assumptions that the go tool makes. This may be a recurring problem in the future, especially as gomodule adoption increases.

With @b5's help, we were able to rule out our docker image as being the problem, "circleci/golang:latest". Running our project's HEAD with an old docker image still hit the same failures. Figuring that the problem must be proxy related, I dug into the server responses to see that a timestamp mismatch was causing the problem. Luckily, circleci respects the replace commands in go.mod, so even though our codebase itself doesn't use any of the golangci projects, this commit fixes the paths for whatever circleci is retrieving.

PR #869, opened 2019-08-08
fix(api): body requests honor full references

b5:
desktop needs this to like, work.


b5:
Merging b/c this should get in before next week

PR #871, opened 2019-08-13
Handle null datasets, those with no history. Cleanup FSI semantics of status API.

dustmop:


dustmop:
High level description: fsi=true should only be sent to the API when it knows it's asking for data about the working directory that must be merged with the most recent dataset version. Essentially, this is equivalent to any interaction with the left-hand "status" pane, but nothing else.

The changes within should handle all cases where a dataset has no history. If FSI is true, it doesn't matter that there's no history: the working directory will be merged in. If FSI is false, that should either produce a 422 error from the API, or send "repo: no history" to the command-line.

lib/fsi.go
```diff
@@ -101,27 +102,34 @@ func (m *FSIMethods) Status(dir *string, res *[]StatusItem) (err error) {
...
+	}
+	*res, err = fsint.Status(dir)
 	return err
 }
 
-// StoredStatus returns a status-like report of a dataset reference
-func (m *FSIMethods) StoredStatus(ref *string, res *[]StatusItem) (err error) {
+// StatusAtVersion gets changes that happened at a particular version in the history of the given
+// dataset reference. Not used for FSI.
+func (m *FSIMethods) StatusAtVersion(ref *string, res *[]StatusItem) (err error) {
```
b5: Muuuuuuuch better method name

api/datasets.go
```diff
@@ -306,6 +306,10 @@ func (h *DatasetHandlers) getHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	res := lib.GetResult{}
 	err := h.Get(&p, &res)
+	if err == repo.ErrNoHistory {
```
b5: These no history errors should be handled inside the `if err != nil` check imo. Considering how verbose error handling is in go, I think it's worth keeping all error-returns within the `!= nil` branch

api/datasets.go
```diff
@@ -615,7 +619,12 @@ func (h DatasetHandlers) bodyHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	result := &lib.GetResult{}
-	if err := h.Get(p, result); err != nil {
+	err = h.Get(p, result)
+	if err == repo.ErrNoHistory {
```
b5: ditto here, handle in the `!= nil` statement

api/fsi.go
```diff
@@ -47,22 +48,30 @@ func (h *FSIHandlers) StatusHandler(routePrefix string) http.HandlerFunc {
...
+		useFSI := r.FormValue("fsi") == "true"
 		ref, err := DatasetRefFromPath(r.URL.Path[len(routePrefix):])
 		if err != nil {
 			util.WriteErrResponse(w, http.StatusBadRequest, fmt.Errorf("bad reference: %s", err.Error()))
 			return
 		}
 
 		res := []lib.StatusItem{}
-		if ref.Path != "" {
+		if !useFSI {
```
b5: let's switch the useFSI test to check for truth instead of falsy, call StatusForAlias and early return. Outdent the else clause & make StatusAtVersion the default.

api/fsi_test.go
```diff
@@ -53,3 +59,131 @@ func TestFSIHandlers(t *testing.T) {
...
+
+	dsHandler := NewDatasetHandlers(node, false)
+
+	// Dataset with no history
+	actualStatusCode, actualBody := APICall("/peer/test_ds", dsHandler.GetHandler)
+	if actualStatusCode != 422 {
+		t.Errorf("expected status code 422, got %d", actualStatusCode)
+	}
+	expectBody := `{ "meta": { "code": 422, "error": "no history" }, "data": null }`
+	if expectBody != actualBody {
```
b5: these expect vs actual comparisons would benefit from that new go-cmp package 'Diff' method for easier test maintenance. Not a vital change tho

api/fsi_test.go
```diff
@@ -53,3 +59,131 @@ func TestFSIHandlers(t *testing.T) {
...
+		t.Errorf("expected status code 422, got %d", actualStatusCode)
+	}
+	expectBody = `{ "meta": { "code": 422, "error": "no history" }, "data": null }`
+	if expectBody != actualBody {
+		t.Errorf("expected body %s, got %s", expectBody, actualBody)
+	}
+}
+
+// APICall calls the api and returns the status code and body
+func APICall(url string, hf http.HandlerFunc) (int, string) {
```
b5: This is very pretty. we should use it in more places

api/log.go
```diff
@@ -55,7 +55,12 @@ func (h *LogHandlers) logHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	res := []repo.DatasetRef{}
-	if err := h.Log(params, &res); err != nil {
+	err = h.Log(params, &res)
+	if err == repo.ErrNoHistory {
```
b5: ditto with the err handling within `err != nil` clause

PR #889, opened 2019-08-29
feat(status): FSI status include file mtime. Requires flag for cmdline.

dustmop:


api/fsi_test.go
```diff
@@ -158,7 +167,8 @@ func TestNoHistory(t *testing.T) {
 	}
 	// Handle tempoary directory by replacing the temp part with a shorter string.
 	resultBody = strings.Replace(actualBody, initDir, initSubdir, -1)
-	expectBody = `{"data":[{"sourceFile":"fsi_init_dir/meta.json","component":"meta","type":"add","message":""},{"sourceFile":"fsi_init_dir/schema.json","component":"schema","type":"add","message":""},{"sourceFile":"body.csv","component":"body","type":"add","message":""}],"meta":{"code":200}}`
+	templateBody := `{"data":[{"sourceFile":"fsi_init_dir/meta.json","component":"meta","type":"add","message":"","mtime":"%s"},{"sourceFile":"fsi_init_dir/schema.json","component":"schema","type":"add","message":"","mtime":"%s"},{"sourceFile":"body.csv","component":"body","type":"add","message":"","mtime":"%s"}],"meta":{"code":200}}`
+	expectBody = fmt.Sprintf(templateBody, metaMtime, schemaMtime, bodyMtime)
 	if expectBody != resultBody {
```
b5: tests in CI are sad. can we use go-cmp here instead?

```go
if diff := cmp.Diff(expectBody, resultBody); diff != "" {
  // ...
```

dustmop: Added it for the larger compares, tests are passing now.

PR #897, opened 2019-09-03
feat(cmd): add fsi plumbing subcommand

b5:


cmd/fsi.go
```diff
@@ -0,0 +1,88 @@
...
+		return
+	}
+
+	o.FSIMethods, err = f.FSIMethods()
+	return err
+}
+
+// Link creates a FSI link
+func (o *FSIOptions) Link() (err error) {
+	return fmt.Errorf("not finsihed: link")
```
dustmop: finsihed -> finished

cmd/fsi.go
```diff
@@ -0,0 +1,88 @@
...
+type FSIOptions struct {
+	ioes.IOStreams
+
+	Refs       *RefSelect
+	FSIMethods *lib.FSIMethods
+}
+
+// Complete adds any missing configuration that can only be added just before calling Run
+func (o *FSIOptions) Complete(f Factory, args []string) (err error) {
+	if o.Refs, err = GetCurrentRefSelect(f, args, -1); err != nil {
```
dustmop: This is okay as a starting point, but we also want to add flags to explicitly set the directory and dataset-ref. Since the fsi command is "plumbing", we may actually want to require them, under the assumption that this is only for advanced use cases, and the UI doesn't need to be as friendly. In addition, a common use case is going to be a user deleted their dataset folder, and doesn't have access to the .qri-ref file any more, so they want to remove the entry from the .fbs, but don't have any access to the linked ref. In that case, relying on the implicit ref doesn't work.

cmd/fsi.go
```diff
@@ -0,0 +1,88 @@
...
+		Use:   "unlink",
+		Short: "unlink a .qri-ref",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := o.Complete(f, args); err != nil {
+				return err
+			}
+			return o.Unlink()
+		},
+	}
+
```
dustmop: In the future, would also be nice to have "cleanup", which goes through the .fbs and removes any folders that it can't find anymore. Also "relink", which takes a directory and dsref and moves which directory the .fbs points to for that dsref, removing the old directory's .qri-ref. Or perhaps "link" could just do that.

b5: totally. I'll file an issue.

PR #899, opened 2019-09-03
fix(config): enable setting config over RPC

### LINKED ISSUE #898: cannot set the config from the command line when a qri process is running
When you try to `qri config set` from the  CLI while a qri process is running you get this error:
```
error validating config: /middleware: type should be array
```

The `Repo` object in config has field `Middleware`, which is an array of strings. This is currently an empty array. When sending the config object over rpc, rpc sets that empty array to null. Then, when we go to validate the config, we error, because `Repo.Middleware` should be an array, not null. 

This behavior was confirmed by inspecting the `update` param in `SetConfig` before and after being sent over RPC. Also, if you edit your config.yaml file to include dummy Middleware (eg, `Repo.Middleware = ['test']`), you will be able to edit the config over the command line while the qri process is running.

I'm not sure how to force RPC to send the `Config` struct the way we expect.

b5:
closes #898.

Also makes `https://registry.qri.cloud` the new default.

### DIFF
--- a/config/registry.go
+++ b/config/registry.go
@@ -12,7 +12,7 @@ type Registry struct {
 // DefaultRegistry generates a new default registry instance
 func DefaultRegistry() *Registry {
 	r := &Registry{
-		Location: "https://registry.qri.io",
+		Location: "https://registry.qri.cloud",
 	}
 	return r
 }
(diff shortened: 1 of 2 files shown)


ramfox:
LGTM

PR #901, opened 2019-09-03
fix(regclient.Profile): Profile struct has field formatting details when marshalling to json

### LINKED ISSUE #900: registry profile responses come back with capitalized fields
We are expecting lower case on the desktop app

ramfox:
closes #900


ramfox:
LGTM!

ramfox:
Tests that should be written:
- running a node that has datasets & and ID & calling to /registry/profile/new with a new peername
- while node is running:
  - config.profile.peername should be the new peername
  - datasets should return with new peername as the datasets' peernames

PR #909, opened 2019-09-05
fix(add): restore add

b5:
turns out the bug was actually in dsync, which was fixed with this PR: https://github.com/qri-io/dag/pull/21

This also bumps our version number to 0.9.0-dev


api/meta.json
```diff
@@ -0,0 +1 @@
+{"title": "hello"}
```
dustmop: Remove this. Something in the tests is creating this file, I think it's my fault.

Issue #839, opened 2019-07-29
User can send a a filepath to init, init will copy the file to the working directory


chriswhong:
Prerequisite for https://github.com/qri-io/desktop/issues/29

`qri init` should take a flag so the user can specify a file path to a csv that they would like to turn into a qri dataset.

Qri should:
- copy the file to the working directory
- rename it to `body.[extension]`
- validate it (try to read the csv/json, if it's not in a good enough state to be committed as-is, throw an error)

b5:
closed by #910

PR #910, opened 2019-09-05
feat(fsi): Init may take a --source-body-path to create body file

dustmop:


dustmop:
Fixes https://github.com/qri-io/qri/issues/839.

Issue #907, opened 2019-09-05
In FSI status response, body does not return `"type":"parse error"`


ramfox:
Meta and schema will return parse error correctly, body will only return `modified`, `add`, and `delete`

b5:
closed by #915

Issue #911, opened 2019-09-05
bug: inconsistency in status response btw body and other components


ramfox:
```
0: {sourceFile: "/Users/ramfox/Work/datasets/test_fsi_error/meta.json", component: "meta", type: "add",…}
1: {sourceFile: "/Users/ramfox/Work/datasets/test_fsi_error/schema.json", component: "schema",…}
2: {sourceFile: "body.json", component: "body", type: "add", message: "",…}
meta: {code: 200}
```

`sourceFile` for schema and meta come back with full paths
`sourceFile` for body comes back with just the filename

Not a huge deal, but it is messing with expectations.

b5:
closed by #915

PR #915, opened 2019-09-06
fix(fsi): Add relative directory to fsi body path. Parse body for status.

dustmop:


dustmop:
Fixes https://github.com/qri-io/qri/issues/907 and https://github.com/qri-io/qri/issues/911.

PR #922, opened 2019-09-10
feat(remove): Add FSI options to remove

### LINKED ISSUE #878: delete --all from an FSI-linked directory doesn't clean up my local folder
Not totally sure what we should do here, but I think users will run into trouble with one-way `.qri-ref` files. We should come up with a way to deal with that. My gut says `qui delete --all` is the right thing to do, where "all" implies "the entire namespace"

b5:
closes #878, unblocks qri-io/desktop#118

* removing all versions of a dataset always unlinks a directory
* adds a boolean flag to `qri remove` called `--files`. when true (in any situation), it'll remove all version-tracked files an fsi-linked directory
* To remove a link without deleting versions from the qri DB, run `qri fsi unlink peer/dataset_name`


cmd/fsi.go
```diff
@@ -68,7 +69,22 @@ func (o *FSIOptions) Complete(f Factory, args []string) (err error) {
 
 // Link creates a FSI link
 func (o *FSIOptions) Link() (err error) {
-	return fmt.Errorf("not finished: link command")
+	pwd, err := os.Getwd()
```
dustmop: It's not a show stopper for me, but consider allowing that the directory be another command-line argument, and that it always be required. So linking the current directory would need to be done like `qri fsi link me/my_dataset .` with the "." at the end.

Also, there's also nothing to be gained by allowed an implied dataset ref to come from ".qri-ref", since we're creating a link. Even in the case that we want to re-link the current directory, we're never going to want to use the current value within ".qri-ref", since that would be a no-op.

fsi/fsi.go
```diff
@@ -133,6 +133,9 @@ func (fsi *FSI) UpdateLink(dirPath, refStr string) (string, error) {
 }
 
 // Unlink breaks the connection between a directory and a dataset
+// TODO (b5) - should this even need the refStr, since we consider the .qri-ref
+// file the canonical source of truth, shouldn't this look for that file,
+// and use the reference stored within to talk to the repo?
```
dustmop: What if .qri-ref is deleted, corrupted, or otherwise made unreadable (permissions, network filesystem issues), but the ref link still existed in the fbs? We should still handle that case, ignoring the .qri-ref file's error but still updating the repo.

fsi/fsi.go
```diff
@@ -152,6 +155,10 @@ func (fsi *FSI) Unlink(dirPath, refStr string) error {
 		return err
 	}
 
+	// attempt to remove the directory, this will error if the directory is
+	// not empty. We intentionally ignore this error
+	os.Remove(dirPath)
```
dustmop: Suggestion: be explicit about which errors that we ignore, like "directory not empty", but still acknowledge other errors such as "permission denied".

fsi/mapping.go
```diff
@@ -385,6 +385,47 @@ func WriteComponents(ds *dataset.Dataset, dirPath string) error {
 	return nil
 }
 
+// DeleteDatasetFiles removes mapped files from a directory. if the result of
+// moving all files leaves the directory empty
```
dustmop: This sentence is unfinished

fsi/mapping.go
```diff
@@ -385,6 +385,47 @@ func WriteComponents(ds *dataset.Dataset, dirPath string) error {
...
+		}
+		removed[component] = FileStat{
+			Path:  path,
+			Mtime: stat.Mtime,
+		}
+	}
+
+	// attempt to remove the directory, this will error if the directory is
+	// not empty. We intentionally ignore this error
+	os.Remove(dirPath)
```
dustmop: Same comment as before

fsi/fsi.go
b5: Well, that's a great set of reasons not to rely solely on `.qri-ref` 😄

cmd/fsi.go
b5: ah yes, I'd forgotten about this. I'll add it for `qri fsi link`, we should consider doing the same for `unlink` at some point in the future.

PR #924, opened 2019-09-11
refactor(api init): calls to init should return the dataset

### LINKED ISSUE #923: api init & add dataset request: return the dataset?
When init-ing a dataset via the api, can the response have the same structure as if we had just requested `peername/datasetname?fsi=true`

When adding a dataset from the network via the api, can the response have the same structure as if we had requested `peername/datasetname?fsi=true`,if we have requested the files be checked out to a folder, and `peername/datasetname?fsi=false` if we have not. Checking out to a folder may be a future functionality, tbh not exactly sure where we have settled currently.

Following this convention would help the flow of init-ing and adding a dataset a lot!

ramfox:
closes #923


b5:
👍

b5:
👍

b5:
💯

b5:
🎅

PR #926, opened 2019-09-11
refactor(fsi): New core behavior for FSI, implemented using Component interface

dustmop:


fsi/component/kinds.go
```diff
@@ -0,0 +1,588 @@
...
+	if ds != nil {
+		ds.Commit = cc.Value
+	}
+	return nil
+}
+
+// BodyComponent represents a body component
+type BodyComponent struct {
+	BaseComponent
+	Resolver       qfs.PathResolver
```
b5: This question has shown up a bunch in our work on filters. Do we embed a pointer to the resolver, or require it as an argument all over the place?

I'm pretty sure it's the latter, and like `context.Context`, it's an anti pattern to embed a resolver into a struct that isn't `lib.Instance`. Again, let's keep it for now & return to this question later.

fsi/component/kinds.go
```diff
@@ -0,0 +1,588 @@
...
+func (bc *BodyComponent) LoadAndFill(ds *dataset.Dataset) error {
+	if bc.Value != nil {
+		return nil
+	}
+
+	var err error
+	var entries dsio.EntryReader
+
+	// TODO(dlong): Move this condition into a utility function in dataset.dsio.
+	// TODO(dlong): Should we pipe ctx into this function, instead of using context.Background?
```
b5: yes. Adding a context here is one thing I'd _love_ to see this go-round

fsi/component/list_dir.go
```diff
@@ -0,0 +1,223 @@
...
+		comp := assignField(filesysComponent, "structure", dsComponent)
+		if comp != nil {
+			structure := comp.(*StructureComponent)
+			structure.Value = ds.Structure
+			structure.IsLoaded = true
+			bodyStructure = ds.Structure
+		}
+	}
+	// TODO: viz
+	// TODO: transform
```
b5: yes. let's do these ASAP (next PR?)

fsi/mapping.go
```diff
@@ -1,436 +1,122 @@
...
-
-		componentNameBody: nil,
-	}
-
-	extensions := map[string]decoderFactory{
-		".json": newJSONDecoder,
-		".yaml": newYAMLDecoder,
-		".yml":  newYAMLDecoder,
+// ReadDir reads the component files in the directory, and returns a dataset
+func ReadDir(dir string) (*dataset.Dataset, error) {
```
b5: lit. I'd very much like to see this accept a context. We've encountered some situations where reading from FSI slowness has choked holding an os-level file descriptor. Accepting a context here sets the stage for adding a deadline later

fsi/mapping.go
```diff
@@ -1,436 +1,122 @@
...
-			}
-		} else {
-			// Else, record a problem.
-			if problems == nil {
-				problems = make(map[string]FileStat)
-			}
-			problems[componentNameBody] = FileStat{Path: bodyFilepath, Mtime: mtime}
-		}
+// GetProblems returns the problem messages on a component collection
+func GetProblems(comp component.Component) string {
```
b5: In the future I think this should return a slice of `Problems`

fsi/mapping.go
```diff
@@ -1,436 +1,122 @@
...
+		}
 		// ignore not found errors. multiple components can be specified in the
 		// same dataset file, creating multiple remove attempts to the same path
-		if err := os.Remove(path); err != nil && !os.IsNotExist(err) {
-			return nil, err
-		}
-		removed[component] = FileStat{
-			Path:  path,
-			Mtime: stat.Mtime,
+		if err := os.Remove(subc.Base().SourceFile); err != nil && !os.IsNotExist(err) {
```
b5: it should be a long term goal of this package to _only_ interact with a qfs.Filesystem, pushing all `os` package interaction into qfs land. We're definitely not there today.

lib/datasets.go
```diff
@@ -128,7 +128,7 @@ func (r *DatasetRequests) Get(p *GetParams, res *GetResult) (err error) {
 		if ref.FSIPath == "" {
 			return fsi.ErrNoLink
 		}
-		if ds, _, _, err = fsi.ReadDir(ref.FSIPath); err != nil {
+		if ds, err = fsi.ReadDir(ref.FSIPath); err != nil {
```
b5: much nicer to see these unused return values go away

PR #927, opened 2019-09-12
fix(remove): removing a dataset with no history should work

b5:
need to be able to remove a dataset reference that has no history

### DIFF
--- a/cmd/remove.go
+++ b/cmd/remove.go
@@ -120,14 +120,14 @@ func (o *RemoveOptions) Run() (err error) {
 		}
 		if res.NumDeleted == rev.AllGenerations {
 			printSuccess(o.Out, "removed entire dataset '%s'", res.Ref)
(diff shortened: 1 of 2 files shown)


dustmop:
Looking back at this, I think it's putting too much work into `qri fsi unlink`. That command should *only* be removing the link file (.qri-ref) and removing the FSIPath from the repository. Removing the directory should be the purview of `qri remove` only. I also think the remove functionality (that is, calling os.Remove in Unlink) has been broken because the Dir is always an empty string, but the lack of tests means it wasn't caught. I'll make a follow-up PR that removes the call to os.Remove, and also puts more tests into place that validate the fact that the directory remains after `qri fsi unlink`.

Arqu:
@dustmop I've actually addressed this as part of PR #1203, we can continue the conversation there.

PR #933, opened 2019-09-17
refactor(api): send status code 422 for "dataset is not linked to the filesystem" error

### LINKED ISSUE #932: more specific error code for `dataset is not linked to the filesystem` API error
Right now it returns a very generally 500 error.

We went through this on the `history` side of things: when a dataset exists when fsi=true, but does not have a history in qri, we return error code 422

I would like to surface this same error on the `dataset get` side of things, when fsi=true.
If a dataset does not exist on the filesystem, but it does exist on qri, we are currently returning a 500 error. 404 doesn't make sense because this is what we return when the dataset cannot be found on qri at all. Since it is the mirror of the above history situation, I propose we use the same error code 422. To quote the restAPI tutorial resources on 422 responses:

> the server understands the content type of the request entity, and the syntax of the request entity is correct, but was unable to process the contained instructions

ramfox:
adds lib error `ErrNotLinkedToFilesystem`
adds api func `NoFsiPathErrResponse` that returns a response with the correct error and status code, used by the `getHandler` and `rootHandler`

This pr also fixes a `getHandler` test bug that was requesting data from the wrong path, this fixes the paths and updates the tests

closes #932


api/datasets.go
```diff
@@ -279,6 +279,10 @@ func (h *DatasetHandlers) getHandler(w http.ResponseWriter, r *http.Request) {
 			NoHistoryErrResponse(w)
 			return
 		}
+		if err == lib.ErrNotLinkedToFilesystem {
+			NoFsiPathErrResponse(w)
```
b5: I don't think we need a custom function for this. How about this instead:
```go
if err == fsi.ErrNoLink {
  util.WriteErrResponse(w, http.StatusUnprocessableEntity, err)
  return
}
```

lib/lib.go
```diff
@@ -44,6 +44,10 @@ var (
 	// directory. only $HOME is replaced (no other $ env vars).
 	defaultIPFSLocation = "$HOME/.ipfs"
 
+	// ErrNotLinkedToFilesystem is the err implementers should return when we
+	// are expecting the dataset to have a file system link, but fsiPath is empty
+	ErrNotLinkedToFilesystem = fmt.Errorf("dataset is not linked to the filesystem")
```
b5: Let's move this into the fsi package so fsi can return it as well. It'll require you to import the FSI package within `api/datatasets.go`, I think that's ok.

fsi/fsi.go:
```go
 // ErrNotLinkedToFilesystem is the err implementers should return when we
 // are expecting the dataset to have a file system link, but fsiPath is empty
 ErrNoLink = fmt.Errorf("dataset is not linked to the filesystem")
```

lib/datasets.go
```diff
@@ -126,7 +126,7 @@ func (r *DatasetRequests) Get(p *GetParams, res *GetResult) (err error) {
 	var ds *dataset.Dataset
 	if p.UseFSI {
 		if ref.FSIPath == "" {
-			return fmt.Errorf("%s is not linked to the filesystem", ref.AliasString())
+			return ErrNotLinkedToFilesystem
```
b5: this'll become `fsi.ErrNoLink`

lib/datasets.go
```diff
@@ -314,7 +314,7 @@ func (r *DatasetRequests) Save(p *SaveParams, res *repo.DatasetRef) (err error)
 			return err
 		}
 		if ref.FSIPath == "" {
-			return fmt.Errorf("%s is not linked to the filesystem", ref.AliasString())
+			return ErrNotLinkedToFilesystem
```
b5: ditto, `fsi.ErrNoLink`

api/datasets.go
ramfox: Dope.

Should I also refactor the NoHistoryErrResponse func? It's where I took the pattern

b5: that'd be lovely!

PR #941, opened 2019-09-25
logbook records dataset histories

b5:
logbook is an operation-based conflict-free replicated data type of single-writer append-only logs.

This feature is a long time coming, and forms the basis for collaboration in qri. I want to roll this out in a series of PRs that focus on first moving the logs of an author's dataset around in a meaningful way. Next steps after this PR:

* add a logbook to a qri instance.
* log version, create, remove, publish, and unpublish events
* use logs to in `qri log`
* add methods to a remote for syncing logs
* push logs to remotes on publish
* pull logs from remotes on add
* add a new `qri fetch` and `/fetch` command to sync replicated logs independent of dataset fetches

This will accomplish the first step of showing histories independent from data possession.


b5:
I'm going to merge this without review because we're on a time crunch. We'll do a full rundown on this package in a follow up PR down the road.

PR #942, opened 2019-09-25
feat(add): add link flag to lib.Add, CLI, API

### LINKED ISSUE #940: Update /add to add fsi linked datasets
`/init` currently takes query params `dir` (working directory) and `mkdir`(name of folder to create to store fsi files)

`/add` should take the same, allowing for adding datasets and linking them with one api call.  This will power the Add Dataset modal in Qri Desktop

b5:
closes #940
This allows adding a dataset & linking it to a directory in the same command. you can now pass `dir` as a query param to `POST /add/[name]/dataset` and provide a filepath

### DIFF
--- a/api/datasets.go
+++ b/api/datasets.go
@@ -364,13 +364,19 @@ func (h *DatasetHandlers) addHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// TODO (b5) - move this into lib.Add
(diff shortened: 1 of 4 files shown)


b5:
@chriswhong made a note that this method differs from `/init`, which accepts a `mkdir` param. We should unify these two methods in the future

Issue #943, opened 2019-09-25
Display issue in WSL


bmcguirk:
Probably a weirdness of WSL, but when I run `qri search "<dataset>"` in WSL running Ubuntu, I get this: 

![qri-error](https://user-images.githubusercontent.com/7449705/65640342-25a41e80-dfb8-11e9-8e88-d9715a608f93.png)

`lsb_release -a` output:

```shell
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.3 LTS
Release:        18.04
Codename:       bionic
```

`uname -a` output:

```shell
Linux [machine_name] 4.4.0-17134-Microsoft #706-Microsoft Mon Apr 01 18:13:00 PST 2019 x86_64 x86_64 x86_64 GNU/Linux
```

`go version`: 

```shell
go version go1.13 linux/amd64
```

`qri version`: 

```shell
0.9.0-dev
```

Sorry if this is a stupid issue to file. New to Go/WSL, etc. But very excited about qri!

b5:
Thanks so much for filing @bmcguirk! And definitely not a stupid issue to file. No such thing 😄.

This is indeed a bug. We've been actively refactoring search in the last little while, and I suspect part of the issue is result reporting, but there may be more than one issue at play here. I'll get an initial change up and report back here to see if it improves your experience

bmcguirk:
@b5 - Thanks! Glad to know I (probably) didn't miss something obvious. I'll stay tuned for the next release or releases. :)

b5:
Hi @bmcguirk we just cut a new release today that merges a (potential) fix for this. I'd love to know if you can confirm it works. on my machine if I search for "world" I get something that looks like this: 
<img width="1368" alt="Screen Shot 2019-09-26 at 8 40 11 PM" src="https://user-images.githubusercontent.com/1154390/65733714-e4863a00-e09d-11e9-9d6d-82054dedc6a8.png">

b5:
I'll assume your +1 means things are working, and will close this. Feel free to re-open if we haven't fully-addressed the issue @bmcguirk

Issue #908, opened 2019-09-05
invalid character 'd' looking for beginning of value


benmazzotta:
Hi, I tried running the CLI quickstart on Linux Ubuntu 18.04. Every search command returns the same error code. 

`invalid character 'd' looking for beginning of value`

Any thoughts? Thanks in advance.

b5:
Hi @benmazzotta! 

Apologies, we're currently in the process of transitioning our "registry" code, which powers search. It's going to be a minute before search is properly restored (we're aiming for Monday of next week)

b5:
Hey Ben, We've shipped a bunch of fixes around search, and this should be much smoother if you run the following command:

```
qri config set registry.location https://registry.qri.cloud
```

after running that configuration update, search should work! You'll need to be sure `qri version` reports at least `v0.9.0`

Thanks!

PR #945, opened 2019-09-26
fix(search): update to new registry api search result format

b5:
putting in some work on #943, #908

Search is kinda gorked right now and should get a fix before we cut a release. This cleans up responses & brings us in line with the new search API response structure. For now we've moved to standard dataset responses only as search results. This defines a new search result stringer and makes room for a registry URL link.


b5:
@chriswhong mind pulling this & running search from the CLI as a review?

PR #946, opened 2019-09-26
refactor(fsi): export to structure.json instead of schema.json by default

b5:
as desktop starts to gain its footing, we're realizing that `structure.json` will work better as a default component file. It'll allow us to cache structure details we've been proxying at the time of computing status, and allow us to build an editor that affects formatConfig without creating an additional file.

This PR switches the default output of FSI to use `body`, `meta.json`, and `structure.json`, and cleans up a few bugs in the process. structure now by default contains a `schema` property that carries the FSI-linked schema definition.

_It's still possible to create a file named schema.json and use that for schema definition_.

We should test this branch a bunch with desktop before merging.


dustmop:
Discussed offline; I put in some additional changes. We're now totally removing the ability to use schema.json as it's own component, the schema will always live in structure.json.

PR #957, opened 2019-10-11
refactor(dsref): rename rev package to dsref, a home for all ref string parsing

b5:
we need a place for housing both parsing & parsed dataset references. Other code needs this refactoring to sort out import cycles

### DIFF
--- a/dsref/info.go
+++ b/dsref/info.go
@@ -1,0 +1,23 @@
+package dsref
+
+import (
+	"time"
+)
+
+// Info describes a version of a datsaet
+type Info struct {
+	// Refer
+	Ref Ref `json:"ref,omitempty"`
+
+	// FSIPath is this dataset's link to the local filesystem if one exists
+	FSIPath string `json:"fsiPath,omitempty"`
+	// Published indicates whether this reference is listed as an available dataset
+	Published bool `json:"published"`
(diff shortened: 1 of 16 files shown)


dsref/info.go
```diff
@@ -0,0 +1,23 @@
...
+type Info struct {
+	// Refer
+	Ref Ref `json:"ref,omitempty"`
+
+	// FSIPath is this dataset's link to the local filesystem if one exists
+	FSIPath string `json:"fsiPath,omitempty"`
+	// Published indicates whether this reference is listed as an available dataset
+	Published bool `json:"published"`
+	// If true, this version doesn't exist locally
+	Foreign bool `json:"foreign,omitempty"`
```
dustmop: Are all combinations of these two booleans possible? Put another way, does Foreign==true imply Published? If so, they're not independent, and maybe should be an enum instead. Or, is it possible to have Foreign==true and Published==false due to p2p stuff?

b5: These are independent. Someone can hand you a reflog which will be `foreign=true` and from the perspective of the reflog receiver `reflog=false`. In the future we're going to expand reflog to be a slice of strings indicating _where_ something is published. I expect `foreign` to become the longer-living value

PR #960, opened 2019-10-14
refactor(fsi): schema should be generated and added to structure on init

ramfox:
However, all calls to fsi datasets should reflect the content in fsi, not attempt to generate any fields

Branch to help me work through some structure changes. Because the `mapping` function is largely changed with dustin's pr, this is just up as a reference for what the desktop is expected from the backend.

What desktop expects:
- on init, with a body file, we should generate a schema
- if we delete the schema or structure, and fsi === true, do not send desktop any generated fields EXCEPT FOR FORMAT, which we rely on
- on save, structure file is generated with a detected schema


ramfox:
2 fixes before this can merge:

1) we should be able to remove the structure.json file and still be able to read the body
2) if we remove the structure.json file, then write to it (for example, a user adds a formatConfig option from the desktop), but does not add a "format" field or a "schema" field, on `save` the backend must add those fields. (currently `save` will error)

For future prs: goal would be to, on init, have structure return the formatConfig based on `detect`, currently it is only returning the `schema`, not the formatConfig options

ramfox:
both @dustmop and i think this looks good! going to merge!

PR #966, opened 2019-10-17
add logbook hooks to remote

b5:
**Rename `log` to `oplog`**
using this package confuses me with logging packages. oplog disambiguates

**Adding Hooks**
This PR adds hooks for logsync operations. the remote package does the necessary work to export the same `Hook` function signature that dataset uses. This should simplify the API for the outside world.

Hooks having the same signature simplifies the interface, but we still need a way to get `oplog` data in a hook. This PR adds a new function `OplogFromContext` for getting oplog data within a hook. I've followed go idioms for embedding values in a context, but would like to add another:

> _If a package embeds values in a context, place context interaction code in a file called context.go_

That way we can see at a glance what packages jam stuff into a context.

**Adding Tests To the Remote package**
This PR adds a missing test suite for `remote`, bringing coverage up from an embarrassing `0%`.

One thing I'd like to see improve oner time is the story behind test setup for the remote package. The remote package lives very high in our import tree, and requires a _lot_ of setup to get into a testable state. Dsync relies on a proper IPFS node to do it's work, so we have a whole test file copied out of `qri-io/qri/dag/dsync` for that.

I'd love to see this story improve generally. We might need a pattern for composing test runners.

**Next steps for the remote package**
Remotes are going to be the primary site for utilizing access control once we get it up & running. Knowing access control checks should be happening in the remote package, we can start to design around it.

For starters: logsync and dsync currently have different data structures & properties for representing users/peers. The first thing we should do is unify this with an interface that supports working with PKI.


logbook/logbook.go
```diff
@@ -402,46 +402,65 @@ func (book *Book) WriteCronJobRan(ctx context.Context, number int64, ref dsref.R
...
+// only the username and name components of a dataset.
+// the passed in oplog must refer unambiguously to a dataset or branch.
+// book.Log() returns exact log references
+func DsrefAliasForLog(log *oplog.Log) (dsref.Ref, error) {
+	ref := dsref.Ref{}
+	if log == nil {
+		return ref, fmt.Errorf("logbook: log is required")
+	}
+	if log.Model() != nameModel {
+		return ref, fmt.Errorf("logbook: log doesn't descibe a dataset")
```
dustmop: descibe -> describe

logbook/logsync/logsync.go
```diff
@@ -9,36 +9,55 @@ import (
...
-	// ReceiveCheck is called before accepting a log, returning an error from this
-	// check will cancel receiving
-	ReceiveCheck Hook
-	// DidReceive is called after a log has been merged into the logbook
-	DidReceive Hook
-
 	// to send & push over libp2p connections, provide a libp2p host
 	Libp2pHost host.Host
+
+	// called before accepting a log, returning an error cancel receiving
```
dustmop: cancel -> cancels

logbook/logsync/logsync_test.go
```diff
@@ -108,6 +110,183 @@ func Example() {
...
+	worldBankRef, err := writeWorldBankLogs(tr.Ctx, tr.B)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	hooksCalled := []string{}
+	callCheck := func(s string) Hook {
+		return func(ctx context.Context, a Author, ref dsref.Ref, l *oplog.Log) error {
+			hooksCalled = append(hooksCalled, s)
+			return fmt.Errorf("boooooooooo")
```
dustmop: More sensible string here, please. Like "hook failed" or something.

dustmop:
LGTM!

Issue #968, opened 2019-10-18
can't set `acceptSizeMax` via CLI


b5:
looks like an issue with fillstruct not accepting `int64` types:

```
$ qri config set remote.acceptsizemax 100000
at "remote.acceptsizemax": unknown kind: int64
```

PR #969, opened 2019-10-18
fix(fill): Support setting int64

dustmop:


dustmop:
Fixes https://github.com/qri-io/qri/issues/968.

PR #970, opened 2019-10-19
Refactor actions

### LINKED ISSUE #962: refactoring away elements of the actions package
It's time to remove the actions package, there are some quick pickups 

* [x] **actions/transform.go -> startf** 
now that startf is within the qri repo, this is an easy pickup. Let's be sure to move the tests as well.
* [x] **remove actions/dataset_head**
DatasetHead tries to fetch a dataset from the network if it doesn't exist. This is a bit _too_ optimistic. Let's remove it, replacing calls with `CanonicalizeDatasetRef` and `base.ReadDataset`
* [x] **actions/dag.go -> p2p**
Long term I think these manifest methods belong in _repo_, but that's a different project. For now let's get this into p2p, where it makes the most natural sense to live while we work on a broader "manifests for qfs" undertaking.
* [x]  **actions/peers.go -> p2p**
The listPeers function can move down into the p2p package, probs just as a method on p2p.QriNode. `ConnectedQriProfiles` can be removed entirely, it already just proxies to p2p.
* [x] **Validate -> base**
once `DatasetHead` is removed, validate no longer relies on a `QriNode`, and can be moved into base.
* [x] **Diff -> base**
once `DatasetHead` is removed, diff no longer relies on a `QriNode`, and can be moved into base.
* [x] **Recall -> base**
`Recall` only relies on a repo, move it into base.
* [x] **actions/body.go -> base**
`GetBody` isn't using the passed-in `p2p.QriNode` at all anymore. This can be moved down.
* [x] **remove Log**
we've removed the p2p log fetch method. `actions.DatasetLog` can become just base.DatasetLog.

The implication of these changes: _reposition p2p within our import hierarchy_. Instead we'll be more deliberate, and build back up toward a dynamic system that fetches stuff on the fly later on. Remotes and peer selection from log histories mean we need to re-approach the topic of fetching anyway.

b5:
closes #962.


b5:
superseded by #971

PR #973, opened 2019-10-23
fix(remote): `DeleteDataset` func will now delete an entire dataset's history

ramfox:
This pr renames `RemoveDatasetRef` to `RemoveDataset`. It calls on `base.RemoveNVersionsFromStore` to remove the blocks from the filestore, and uses `repo.DeleteRef` to remove the dataset reference from the repo

`RemoveDataset` calls the `datasetRemovePreCheck` function, if it exists, before it removes the dataset and its history.

`RemoveDataset` calls the `datasetRemoved` function, if it exists, after it removes the datasets and its history

This pr also moves all signature logic to a new `signature.go` file.

It adds a `VerifySigParams` that takes a publicKey, a `params` `map[string]string`, and returns a bool, error. Bool is true if the signature is verified. Basically the inverse of the `sigParams` func.

When remotes have their own keystore, we should make `VerifySigParams` private.

### DIFF
--- a/remote/remote.go
+++ b/remote/remote.go
@@ -158,6 +158,11 @@ func (r *Remote) ResolveHeadRef(ctx context.Context, peername, name string) (*re
 }
 
 // RemoveDataset handles requests to remove a dataset
+// currently removes all versions of a dataset
+// TODO (ramfox): add `gen` params that indicates how many versions of the dataset, starting
+// with the most recent version, we should remove. This should remove the latest version of
+// the dataset ref from the refstore and add the (n + 1)th to the refstore
+// gen = -1 should indicate that we remove all the dataset versions
 func (r *Remote) RemoveDataset(ctx context.Context, params map[string]string) error {
 	pid, ref, err := r.pidAndRefFromMeta(params)
 	if err != nil {
--- a/remote/remote.go
+++ b/remote/remote.go
@@ -11,6 +11,7 @@ import (
 	golog "github.com/ipfs/go-log"
 	"github.com/qri-io/dag"
 	"github.com/qri-io/dag/dsync"
+	"github.com/qri-io/qri/base"
 	"github.com/qri-io/qri/config"
 	"github.com/qri-io/qri/dsref"
 	"github.com/qri-io/qri/logbook/logsync"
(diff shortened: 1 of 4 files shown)


remote/signature_test.go
```diff
@@ -0,0 +1,41 @@
...
+		panic(err)
+	}
+
+	verified, err := VerifySigParams(peerInfo.PubKey, sigParams)
+	if err != nil {
+		t.Errorf("case 'should verify', expected no error, got '%s'", err)
+	}
+	if verified == false {
+		t.Errorf("case 'should verify', expected verification to be true, was false")
+	}
```
b5: We should also have a test here that uses mismatched keys, and expects `err` to not be nil.

remote/remote.go
```diff
@@ -156,14 +157,21 @@ func (r *Remote) ResolveHeadRef(ctx context.Context, peername, name string) (*re
 	return ref, err
 }
 
-// RemoveDatasetRef handles requests to remove a dataset
-func (r *Remote) RemoveDatasetRef(ctx context.Context, params map[string]string) error {
+// RemoveDataset handles requests to remove a dataset
+func (r *Remote) RemoveDataset(ctx context.Context, params map[string]string) error {
```
b5: This shift to removing both persisted versions and a ref kinda breaks the original intention of the `/refs` endpoint of the remote api. It's a little easier to see from the api routes definition: https://github.com/qri-io/qri/blob/902927181e03c9e68510528d85e842b6b0b91d05/api/api.go#L248-L255

The purpose of the `/refs` endpoint predates our logbook work, used to both resolve alias references with HEAD paths, and remove a ref on an unpublish request. There's some work sewing in support for deleting single dags in dsync.

That being said, there's nothing in dsync that supports the idea of removing multiple generations of a history at once, which we need _asap_ So let's merge this & start refactoring past it.

We have a few options for long-term solutions:
1. Multi-rooted DAG manifests, and have `dsync` interpret remove requests as a request to remove all root hashes recursively.
2. Create a higher-order route for deleting logs, references, and persisted versions in a single endpoint. We can use logsets in that context to determine which dags to remove.

PR #974, opened 2019-10-24
feat(readme): Readme component, prepare for removing viz

dustmop:
Support the readme component for save and FSI and components. Modify fill.Struct to handle binary blobs / byte slices. Remove the unused base.Diff. Put some TODOs in place to eventually get rid of viz. Move regex for checking dataset field names into component package.


base/revisions.go
```diff
@@ -102,6 +102,13 @@ func sel(r *dsref.Rev, ds, res *dataset.Dataset) bool {
 				res.Viz = ds.Viz
 			}
 		}
+	case "rm":
```
b5: we should also add this to the `dsref` package

cmd/dag.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/qri-io/dag"
 	"github.com/qri-io/ioes"
+	"github.com/qri-io/qri/fsi/component"
```
b5: weird to see `fsi` in this part of the codebase. We should think about moving `component` into `base` in the near future

fsi/component/kinds.go
```diff
@@ -5,12 +5,13 @@ import (
...
 	"encoding/json"
 	"fmt"
-	"gopkg.in/yaml.v2"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 
+	"gopkg.in/yaml.v2"
+
```
b5: we don't normally use spaces between import domains. IMHO should be:

```
import (
  std_lib_stuff

  everything_else
)
```

code review comments don't have much opinion on the grouping beyond this.
https://github.com/golang/go/wiki/CodeReviewComments#imports

One thing I've seen emerge in other projects is to group imports owned by the org

cmd/dag.go
dustmop: Agreed, will move that soon, it's not strictly a part of fsi.

PR #983, opened 2019-10-25
Refactor dataset ref string

nikon72ru:
Now datasetRef.String() returns standardized answer in format peername/name@/path
New function datasetRef.Absolute() which return answer in format peername/name@ProfileID/path

The problem I encountered: changes that affect on /api tests can broke your terminal encoding. It's happen when the abide package (github.com/beme/abide) try to show differences in your result and stored result. It happen when endpoint exports as a zip file, which means the contents of the response are encoded. Abide does the comparison and can see that something changed, but when it displays those changes, your terminal is getting confused because it is displaying weirdly encoded characters and something isn't getting escaped properly. To fix it you need to commenting out the entire exportCases function. Then take a look at the output from abide. If things have changed in expected ways, abide gives you instructions on how to update the snapshots. Uncomment out the export file, and update those snaphots using abide's instructions. Then, run the tests again to make sure everything updated correctly (thank @ramfox for help)


b5:
Looks like #988 also affected API output, and you'll need to rebase to get this merged. Try the following:

```sh
# you'll need to pull from the qri-io/master repo:
$ git checkout master
$ git pull
$ git checkout refactor_datasetRef_string
$ git rebase master

# you should see a conflict, which can be run by re-upping the API tests:
$ cd ~/qri/api
$ go test -v -- -u
$ git add -A
$ git rebase --continue
$ git push -f YOUR_REPO_BRANCH
```

Hopefully that helps. This PR's looking great!

nikon72ru:
@b5 i'm not sure why my merge change  go.mod file. Is it correct?

b5:
Yep! Looks great. The `go.mod` changes are in fact correct, as these dependencies were accidentally introduced by another dev, and somewhere along the way your go tooling ran `go mod tidy`, which cleaned this up. This looks great! Let's merge it!

b5:
Thanks so much @nikon72ru! Congrats on your first Open Source contribution!

PR #985, opened 2019-10-25
chore(mod): update ipfs & libp2p deps

b5:
let's try to be a little speedier with these in the future :)

This moves us over to some random commit of IPFS (we can sync to the latest patch once it comes out). but more importantly updates our libp2p dependencies to use [go-libp2p-core](https://github.com/libp2p/go-libp2p-core)


cmd/fetch_test.go
```diff
@@ -47,7 +47,8 @@ func TestFetchCommand(t *testing.T) {
 		}
 	}()
 
-	time.Sleep(time.Second * 2)
+	// TODO (b5) - this is horrible. we should block on a channel receive for connectedness
```
dustmop: 😢

PR #986, opened 2019-10-28
fix(fsi): Don't output schema for default csv, many related fixes

dustmop:
Move component from fsi/ to base/. Use component for `qri init`. Only generate schema for `qri init` when the --source-body-path flag is provided.


go.mod
```diff
@@ -37,6 +37,8 @@ require (
 	github.com/libp2p/go-libp2p-peerstore v0.1.3
 	github.com/libp2p/go-libp2p-swarm v0.2.2
 	github.com/mitchellh/go-homedir v1.1.0
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
+	github.com/modern-go/reflect2 v1.0.1 // indirect
```
b5: mind running `go mod tidy`? I'd like see if this gets removed. If this is intentional, where is this dep being used?

fsi/init.go
```diff
@@ -84,65 +80,55 @@ func (fsi *FSI) InitDataset(p InitParams) (name string, err error) {
...
+		// Create structure by detecting it from the body.
+		file, err := os.Open(p.SourceBodyPath)
+		if err != nil {
+			return "", err
 		}
-	} else {
-		// Create body file by reading the sourcefile.
-		if bodyBytes, err = ioutil.ReadFile(p.SourceBodyPath); err != nil {
+		defer file.Close()
+		// TODO(dlong): This should move into `dsio` package.
```
b5: 👍

go.mod
dustmop: Success, `go mod tidy` removed these

PR #995, opened 2019-11-04
refactor(logbook): working identifers into logbook

### LINKED ISSUE #997: renaming a dataset must write log operations
currently we aren't properly tracking renaming properly in logbook, which means we won't be able to use logbook as a canonical source of naming events. We're also _very_ short on rename tests in general.

* Write a test that proves rename presently doesn't work on the CLI.
* Write a test that proves logbook operations aren't being written on renaming
* Add calls to logbook to write an operation to the dataset namespace on rename
* make tests pass
* party

b5:
closes #997

This PR lands a few key changes to logbooks:
* arranges logs into a cleaner hierarchy of user >> dataset >> branch, removing the top level arrange-by-model variation
* disambiguates what is written to each branch. User logs only contain operations that affect the user model, Dataset logs only contain operations that affect entire datasets, and Branch logs only contain operations for a line of commit activity
* introduces methods for fetching by _ head reference_ using names, and ID, using a log identifier
* renames "name" model to "dataset", so we now have `user`, `dataset`, and `branch` models
* renames `WriteNameInit` methods to match dataset business logic conventions. Now called: `WriteDatasetInit`, `WriteDatasetRename`, and `WriteDatasetDelete`
* adds logic to track oplog `Remove` operations. Fetching a log by head reference that has been removed will now return `ErrNotFound`. Deleted logs can still be fetched by `ID` reference

### DIFF
--- a/fsi/init.go
+++ b/fsi/init.go
@@ -131,7 +131,7 @@ func (fsi *FSI) InitDataset(p InitParams) (name string, err error) {
 		}
 	}
 
-	if err = fsi.repo.Logbook().WriteDatasetInit(context.TODO(), name); err != nil {
+	if err = fsi.repo.Logbook().WriteDatasetInit(context.TODO(), ref.Name); err != nil {
 		if err == logbook.ErrNoLogbook {
 			err = nil
 			return name, nil
--- a/fsi/init.go
+++ b/fsi/init.go
@@ -131,7 +131,7 @@ func (fsi *FSI) InitDataset(p InitParams) (name string, err error) {
 		}
 	}
 
-	if err = fsi.repo.Logbook().WriteNameInit(context.TODO(), name); err != nil {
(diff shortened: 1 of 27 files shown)


fsi/init.go
```diff
@@ -131,7 +131,7 @@ func (fsi *FSI) InitDataset(p InitParams) (name string, err error) {
 		}
 	}
 
-	if err = fsi.repo.Logbook().WriteDatasetInit(context.TODO(), name); err != nil {
+	if err = fsi.repo.Logbook().WriteDatasetInit(context.TODO(), ref.Name); err != nil {
```
b5: @dustmop case-and-point when dealing with strings. This used to write a `name` string, which was passing in `peername/name` here, and causing silent creation of a _second_ logbook reference for `peername/name`. Given that logbook doesn't do any validation in relation to naming, we should either add some sort of logbook-level validation, or find some way to ensure the data structures passed to logbook are uniform and ideally always valid.

Issue #993, opened 2019-11-04
logbook is handling deleted datasets incorrectly


dustmop:
If I have a dataset such as "me/my_ds" that has a few versions, then delete it, then create a new dataset with the same name, the logbook incorrectly shows all the events from the old dataset.

* Deleting a dataset entirely should finish off the logbook entirely for that dataset.
* Using a name for a dataset that doesn't currently exist shouldn't depend upon the history of how that name was used before.
* We should not be looking up logbooks using pretty dataset names, we should be mapping them to stable identifiers. This will be important for renames as well.
* All of this behavior needs documentation and tests.

dustmop:
Fixed by #1003.

PR #1003, opened 2019-11-05
fix(log): Deleting a dataset ends the log for that dataset

dustmop:


dustmop:
Closes https://github.com/qri-io/qri/issues/993

PR #1008, opened 2019-11-08
fix(ls): Don't crash when `list` runs over RPC

dustmop:


b5:
@dustmop just mentioned that some of this originates from FSI, which allows `nil` pointers for fields like structure

PR #1009, opened 2019-11-08
feat: stats

ramfox:
To give users greater insights into their data quicker and with less friction, we have created a way to get some statistical metadata for each column/field in a dataset!

Using the `qri stats <dataset_ref>` command or `/stats/<dataset_ref` api endpoint, you can pull stats on a specific dataset. This is great for our Desktop App, we can use these stats to give the user more details about their dataset in a more digestible way.

**boolean fields**:
  - count: total count of entries in this column
  - trueCount: count of all true entries
  - falseCount: count of all false entries

**string fields**
  - count: total count of entries in this column
  - minLength: shortest length of a string
  - maxLength: longest length of a string
  - unique: the number of unique entries
  - frequencies: a map of entries that appear more than once and the frequency in which they occur

**numeric fields**
  - count: total count of entries in this column
  - min: smallest number
  - max: largest number
  - unique: the number of unique entries
  - frequencies: a map of entries that appear more than once and the frequency in which they occur

**null fields**
  - count: total count of entries in this column


lib/datasets.go
```diff
@@ -831,3 +831,42 @@ func (r *DatasetRequests) DAGInfo(s *DAGInfoParams, i *dag.Info) (err error) {
...
+	ref, err := base.ToDatasetRef(p.Ref, r.node.Repo, false)
+	if err != nil {
+		return err
+	}
+	ref.Dataset, err = dsfs.LoadDataset(ctx, r.node.Repo.Store(), ref.Path)
+	if err != nil {
+		return fmt.Errorf("loading dataset: %s", err)
+	}
+
+	// stats.JSON assumes we have a bodyFile. base.OpenDataset will attach the
```
b5: Don't feel like this comment is necessary

lib/lib.go
```diff
@@ -222,6 +224,14 @@ func OptRegistryClient(cli *regclient.Client) Option {
 	}
 }
 
+// OptStatsCache overrides the configured stats cache
+func OptStatsCache(stats *stats.Stats) Option {
```
b5: argument here should be a `stats.Cache`, not a `*stats.Stats`

stats/cache.go
```diff
@@ -0,0 +1,82 @@
...
+
+	return fmt.Errorf("not finished")
+}
+
+// JSON gets cached byte data for a path
+func (c osCache) JSON(ctx context.Context, path string) (r io.Reader, err error) {
+	return nil, ErrCacheMiss
+}
+
+// Hash uses lower-case base32 encoding for id bytes for a few reasons:
```
b5: stale comment. pretty sure this is my fault

lib/lib_test.go
```diff
@@ -113,6 +113,9 @@ func TestNewDefaultInstance(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	if _, err = os.Stat(filepath.Join(tempDir, "stats")); os.IsNotExist(err) {
```
b5: Nice addition!

config/stats.go
```diff
@@ -0,0 +1,61 @@
+package config
+
+import (
+	"github.com/qri-io/jsonschema"
+)
+
+// Stats configures a qri stats
+type Stats struct {
```
b5: There's some confusion here between **stats** configuration and **stats cache** configuration, partially because we didn't talk about this at great length 😄. 

We need a place to configure how stats are calculated. As an example, a user may be working with massive datasets on a slow machine, and want to lower `stats.StopFreqCountThreshold` to `1` so that frequency calculation is skipped entirely, regardless of where it's being cached.

So we need to be able to configure stats, and move the conversation about cache configuration into it's own structure. This is a great use case for a struct literal:

```go
// Stats configures qri statistical metadata calculation
type Stats struct {
  Cache struct {
    Type string `sson:"type"`
    MaxSize  uint64 `json:"maxsize"`
    Path  string `json:"path,omitempty"`
  }
  // For later addition:
  // StopFreqCountThreshold int
}
```

Issue #1000, opened 2019-11-04
qri remove has many fsi bugs


dustmop:
First off, running `qri remove` with no dataset deference (outside or inside of FSI) does nothing but also prints no output.

Running `qri remove` in FSI also ignores the dataset reference in .qri-ref. So even running `qri remove --revisions=2` inside FSI will do nothing.

In FSI, running `qri remove --revisions=2 me/test_ds` will remove 2 versions from the repo, but won't update the working directory. So `qri status` will show dirty files and the user has to use `qri restore` to get to a clean directory.

*However*, run the previously mentioned command *will* remove the link from the repo's fbs file, but *not* modify the `.qri-ref` in the working directory. Running future commands will return the error "dataset is not linked to the filesystem". We should be treating the .qri-ref file as authoritative and fixing up the fbs instead.

In addition, .qri-ref should be updated so that it contains the dataset ref path after a `remove` command.

b5:
very fitting that issue #1000 is just like "hey, there's a bunch of bugs here".

dustmop:
Fixed by https://github.com/qri-io/qri/pull/1012.

PR #1012, opened 2019-11-09
fix(remove): Fix multiple problems with remove

dustmop:
1) remove now works with FSI, by reading the ref in .qri-ref
2) remove with no reference is an error, before it was silently ignored
3) removing a single version no longer unlinks the dsref in FSI
4) change --files to --keep-files, which has the opposite meaning
5) in FSI, removing a dataset version will update files so the wd is clean
6) but removing a dataset is not allowed if the wd is dirty
7) unless the --keep-files flag is enabled, which won't modify the wd

Fixes all of issue #1000


remote/remote.go
```diff
@@ -187,7 +187,7 @@ func (r *Remote) RemoveDataset(ctx context.Context, params map[string]string) er
 	}
 
 	// remove all the versions of this dataset from the store
-	if err := base.RemoveNVersionsFromStore(ctx, r.node.Repo, &ref, -1); err != nil {
+	if _, err := base.RemoveNVersionsFromStore(ctx, r.node.Repo, &ref, -1); err != nil {
```
b5: This is an interesting case. Looks like we need to think more about the interaction between remotes & a restore.

lib/datasets.go
```diff
@@ -625,43 +624,55 @@ func (r *DatasetRequests) Remove(p *RemoveParams, res *RemoveResponse) error {
...
+			ds.Peername = head.Peername
+			if err = base.OpenDataset(ctx, r.inst.Repo().Filesystem(), ds); err != nil {
+				return err
+			}
 
-	// If deleting more revisions then exist, delete the entire dataset.
-	if p.Revision.Gen >= len(log) {
-		return removeEntireDataset()
-	}
+			// TODO(dlong): Add a method to FSI called ProjectOntoDirectory, use it here
```
b5: Would love to see an issue made out of this

cmd/remove_integration_test.go
```diff
@@ -0,0 +1,538 @@
...
+}
+
+// ExecCommand executes the given command string
+func (run *RemoveTestRunner) ExecCommand(cmdText string) error {
+	run.CmdR = run.RepoRoot.CreateCommandRunner(run.Context)
+	return executeCommand(run.CmdR, cmdText)
+}
+
+// MustExec runs a command, returning standard output, failing the test if there's an error
+func (run *RemoveTestRunner) MustExec(cmdText string) string {
```
b5: not a huge fan of the way this `Must` pattern swallows the `t *testing.T` pointer as state. It makes this stuff really hard to use in benchmarks (which take a `*testing.B`) and subtests (which create new testing.T pointers). To me the test runner pattern shines when it's concise without being magical. I'd prefer:

```go
run, delete := newRemoveTestRunner(t, "test_peer", "example")
defer delete()

run.MustExec(t, "qri save --body=testdata/movies/body_two.json me/remove_test")
```

base/dataset.go
```diff
@@ -278,24 +278,24 @@ func UnpinDataset(ctx context.Context, r repo.Repo, ref repo.DatasetRef) error {
 // the most recent version
 // when n == -1, remove all versions
 // does not remove the dataset reference
-func RemoveNVersionsFromStore(ctx context.Context, r repo.Repo, ref *repo.DatasetRef, n int) error {
+func RemoveNVersionsFromStore(ctx context.Context, r repo.Repo, ref *repo.DatasetRef, n int) (*repo.DatasetRef, error) {
```
b5: Lovely. Returning the new head here provides a great separation between block persistence and it's affect on refs.

base/dataset_test.go
```diff
@@ -147,7 +147,7 @@ func TestRemoveNVersionsFromStore(t *testing.T) {
 	}
 
 	for _, c := range bad {
-		err := RemoveNVersionsFromStore(ctx, c.store, c.ref, c.n)
+		_, err := RemoveNVersionsFromStore(ctx, c.store, c.ref, c.n)
```
b5: Looking forward to leveraging the return value of this func in tests 😄

PR #1019, opened 2019-11-14
feat(stats): add histogram, mean, median to numeric stat

b5:
This PR adds `mean ` (average), `median` (middle value) and `histogram` properties to the numeric stat, removes `frequencies` and `unique`.

Histograms bucket frequency counts of a variable across evenly distributed "bins" of value ranges.  I'm using a default, hardcoded 10 bins for histogram calculation for now.

JSON output of a `numeric` stat now looks like this:
```
{
  "count": 21159,
  "histogram": {
    "bins": [
      0,
      24500000.1,
      49000000.2,
      73500000.30000001,
      98000000.4,
      122500000.5,
      147000000.60000002,
      171500000.70000002,
      196000000.8,
      220500000.9,
      245000001
    ],
    "frequencies": [
      21141,
      14,
      1,
      1,
      1,
      0,
      0,
      0,
      0,
      1
    ]
  },
  "max": 245000000,
  "mean": 327137.47497518786,
  "median": 50000,
  "min": 0,
  "type": "numeric"
}
```

The length of the `bins` array will always be 1 element longer than the frequencies array, so the first frequency is the number of values that fall in the range `bin[0]` to `bin[1]`


stats/stats.go
```diff
@@ -383,41 +392,43 @@ func (acc *numericAcc) Map() map[string]interface{} {
...
-			}
-		}
-		if len(acc.frequencies) == 0 {
-			acc.frequencies = nil
-		}
+	// finalize avg
+	acc.mean = acc.mean / float64(acc.count)
+
+	if len(acc.histogram) > 0 {
+		acc.median = acc.histogram[len(acc.histogram)/2]
```
ramfox: This is grabbing the middle entry before the histogram is sorted. Also, there is no catch in place for if the histogram list is even, in which we should be averaging the two middle entries,

b5:
DOPE FIXED MERGING

PR #1020, opened 2019-11-14
fix(stats): gets stats with no structure

ramfox:
Hook the `lib.Get` (which fetches `Stats` if `Selection="stats"`, into the `statsHandler`. `lib.Get` correctly fetches fsi datasets as well as versioned datasets.
If we don't have a structure, such as a case where we are using `fsi` and the user has removed their `structure.json`, stats should detect a structure and use that detected structure.schema to calculate stats.

### DIFF
--- a/lib/datasets.go
+++ b/lib/datasets.go
@@ -8,10 +8,12 @@ import (
 	"io/ioutil"
 	"net/rpc"
 	"os"
+	"path/filepath"
 
 	"github.com/ghodss/yaml"
 	"github.com/qri-io/dag"
 	"github.com/qri-io/dataset"
+	"github.com/qri-io/dataset/detect"
 	"github.com/qri-io/jsonschema"
 	"github.com/qri-io/qfs"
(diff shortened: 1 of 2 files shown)


lib/datasets.go
```diff
@@ -901,6 +903,20 @@ func (r *DatasetRequests) Stats(p *StatsParams, res *StatsResponse) (err error)
...
+		p.Dataset.Structure.Schema, _, err = detect.Schema(p.Dataset.Structure, p.Dataset.BodyFile())
+		if err != nil {
+			return err
+		}
+		// TODO (ramfox): this feels gross, but since we consume the reader when
+		// detecting the schema, we need to open up the file again, since we don't
+		// have the option to seek back to the front
+		if err = p.Dataset.OpenBodyFile(ctx, r.node.Repo.Filesystem()); err != nil {
+			return err
+		}
```
b5: agreed it's a little gross, we'll survive for now.

PR #1022, opened 2019-11-15
fix(remove): Can remove a dataset if its working directory is missing

dustmop:


lib/datasets.go
```diff
@@ -597,7 +598,15 @@ func (r *DatasetRequests) Remove(p *RemoveParams, res *RemoveResponse) error {
 				if wdErr == fsi.ErrWorkingDirectoryDirty {
 					return fmt.Errorf("cannot remove from dataset while working directory is dirty")
 				}
-				return wdErr
+				if strings.Contains(wdErr.Error(), "not a linked directory") {
+					// If the working directory has been removed (or renamed), could not get the
+					// status. However, don't let this stop the remove operation, since the files
+					// are already gone, and therefore won't be removed.
+					log.Debug("could not get stauts for %s, maybe removed or renamed", ref.FSIPath)
```
b5: status -> status

`log.Debug` should be `log.Debugf`. go vet won't catch this because the log package isn't directly invoking `fmt`

cmd/remove_integration_test.go
```diff
@@ -512,3 +513,30 @@ func TestRemoveAllAndKeepFiles(t *testing.T) {
...
+	os.Chdir(parentDir)
+
+	// Remove the working directory
+	err := os.RemoveAll(workDir)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Remove all should still work, even though the working directory is gone.
+	run.MustExec(t, "qri remove --revisions=all me/remove_no_wd")
```
b5: I'd prefer the "check" part of a test not use `Must` functions, even if just for stylistic reasons. It's not a huge thing, but a quick `if err != nil` helps the readability of what's being tested

PR #1023, opened 2019-11-15
fix(rename): Allow rename for dataset with no history. Update .qri-ref

dustmop:


base/ref.go
```diff
@@ -80,8 +80,10 @@ func ModifyDatasetRef(ctx context.Context, r repo.Repo, current, new *repo.Datas
 		return err
 	}
 	if err := repo.CanonicalizeDatasetRef(r, current); err != nil {
-		log.Debug(err.Error())
-		return fmt.Errorf("error with existing reference: %s", err.Error())
+		if err != repo.ErrNoHistory {
```
b5: I'm a fan of combining the if statement:

```go
if err := repo.CanonicalizeDatasetref(r, current); err != nil && err != repo.ErrNoHistory {
  // ...
}
```

but, might be worth pushing back & keeping it as written. It calls attention to the `ErrNoHistory` check

lib/fsi.go
```diff
@@ -61,14 +62,29 @@ func (m *FSIMethods) CreateLink(p *LinkParams, res *string) (err error) {
...
-	return err
+	if p.ToModify == "dir" {
+		*res = true
+		return m.inst.fsi.ModifyLinkDirectory(p.Dir, p.Ref)
+	} else if p.ToModify == "ref" {
+		*res = true
+		return m.inst.fsi.ModifyLinkReference(p.Dir, p.Ref)
+	} else {
+		return fmt.Errorf("ToModify has unknown value %q", p.ToModify)
+	}
```
b5: drop the else in favour of always returning the error

base/ref.go
dustmop: I'm fine with combining them, done.

PR #1031, opened 2019-11-25
feat(commit): Friendly commit descriptions based on deepdiff

dustmop:
Instead of the old dsdiff package, switch to using deepdiff to generate commit titles and messages. This took a few steps:

* Add more tests for the old functionality
* Refactor generateCommitMsg a little bit
* Add new base/friendly package that wraps deepdiff
* Update all of the tests

There's still a bit more work to do on this:

* Although friendly has the ability to make nice messages out of body changes, due to how base/dataset:PrepareDataset works, the body is not inlined so it doesn't get passed into deepdiff. It's not clear to me the best way to handle this.
* Certain fields, like the structure derived fields, should be ignored when making descriptions. Maybe base/component should be integrated to do this.
* There's a lot of inefficient string formatting in base/friendly. It should probably use strings.Builder instead.


base/dsfs/dataset_test.go
```diff
@@ -385,74 +387,292 @@ func TestWriteDataset(t *testing.T) {
...
-		ds, prev *dataset.Dataset
-		force    bool
-		expected string
-		errMsg   string
+		description string
+		prev, ds    *dataset.Dataset
+		force       bool
+		expectShort string
+		expectLong  string
+		errMsg      string
```
b5: please split table tests into two groups, one set of `good` cases and one set of `bad` cases:

```go
good := []struct{
  description string
  prev, ds *dataset.Dataset
  force bool
  expectShort string
  expectLong string
}{ }

bad := []struct{
  prev, ds *dataset.Dataset
  force bool
  errMsg string
}{ }
```

bad cases get to drop the description. But these tests are great!

dustmop:
Fixed all relevant comments, tests are passing everywhere

PR #1032, opened 2019-11-26
test(p2p): TestMissingManifest should not hang when hashes change.

dustmop:


p2p/dag_test.go
```diff
@@ -46,27 +47,25 @@ func TestMissingManifest(t *testing.T) {
...
-		},
-	}
+	// Select some blocks from the saved dataset. Don't hardcode block ids, because if they
+	// ever change this test will hang.
+	capi, _ := node.IPFSCoreAPI()
+	blocks := p2ptest.GetSomeBlocks(capi, ref, 2)
+	in := &dag.Manifest{Nodes: blocks}
 
+	// TODO(dlong): This function seems to not work correctly. If any blocks are missing, it
+	// doesn't return them, instead it hangs forever.
```
b5: Ah yes, I'm aware of this, and will ship a proper fix in the future.

PR #1036, opened 2019-11-29
feat(refs): `qri list --raw` will display raw reference data

dustmop:


dustmop:
Going to merge this, seems good to have and there's no giant questions about it's impact.

Issue #1037, opened 2019-12-02
Save API needs a method of creating a name from a source-body-path, telling it to desktop


dustmop:
For an easier import experience, the `/save/` api call should take a new query parameter, let's call it `generatename`, which creates a unique dataset name from the source-body-path filename. This is already mostly happening, but we also need to avoid conflicting with existing dataset names. Once this name is created, it can be passed back to the API caller, which can then parse it out and use it in its UI.

dustmop:
Of note: running the command `qri save --body=my_bodyfile.json` used to infer the dataset name "my_bodyfilejson" via the call to `varName.CreateVarNameFromString` in base/dataset_prepare.go. It seems this broke at some point, probably due to FSI / ref_select stuff, and there was no cmd test so this regression wasn't caught.

dustmop:
Ran into an interesting edge case...

If a user first runs `qri save --body=my_bodyfile.json`, the name "my_bodyfilejson" will be inferred. If the user again runs `qri save --body=my_bodyfile.json` (perhaps the same file with modifications, or an entirely different file with the same name) it's unclear what they want. Do they want to create a new version of the same dataset (relying on our inference), or do they want to create an entirely new dataset (relying on the command's earlier behavior of making a new dataset version). It seems like the first possibility is dangerous, as the name is "magically" inferred (and we should be allowed to change the inference algorithm at will). We should assume that this inference is a sign that user doesn't quite care what the dataset is named for now.

With this assumption, doing `qri save --body=my_bodyfile.json` and ending up with a name that is already in use should lead to an error. We need some way of letting the user be explicit about what they want.

Long ago, there was a `qri new` command. This was the same as our current `qri save` command, except it could only be used to create new datasets, and was also the only way to create new datasets. I'm thinking we should bring back something similar, in the form of a `--new` flag for `save`, with the following semantics:

* `qri save --body=my_bodyfile.json --new` will infer a name from the bodyfile, and add extra information onto the end in order to guarantee uniqueness (my_bodyfile_1, my_bodyfile_2...)
* `qri save --body=my_bodyfile.json` will infer a name from the bodyfile, if that name is already in use, it's an error. The error can recommend the user either add the `--new` flag or explicitly add the dataset reference
* `qri save --body=my_bodyfile.json me/my_ds` will create a new dataset or add a version, as it already does
* `qri save --body=my_bodyfile.son me/my_ds --new` will create a new dataset only, raising an error if the dataset already exists

[...]

PR #1039, opened 2019-12-03
feat(save): Flag --new for save ensures that we're saving a new dataset

dustmop:
When running `save` without a dataset ref, --new will ensure that a unique dataset name is generated. When running `save` with a ref, --new will only succeed if the dataset name is not in use. Add API support so that desktop can easily import files without having to ask for a dataset name first. Fix up some inconsistent or incomplete logic in base around saving datasets.


dustmop:
Fixes https://github.com/qri-io/qri/issues/1037.

PR #1040, opened 2019-12-06
test(add): Mock remote client, enables tests for add command

dustmop:
Add the ability to mock the remoteClient, done in tests by setting a specific value on the context. This makes it possible to construct test runners that will create commands that create lib.Instances which use the mockRemoteClient. Using this, test the behavior of add.


dustmop:
Putting this in place in order to fix and add tests for our deletion bugs.

Issue #1035, opened 2019-11-29
removing a foreign dataset breaks if there's no logbook


dustmop:
Discovered by @chriswhong in Desktop, but this is specifically a backend problem.

Removing a dataset uses `Remove` in lib.datasets.go, which does the following:
* check fsi status, if there's a working directory
* call `base.DatasetLog` to get revisions to delete
* if `--all` was used:
* call `base.RemoveNVersionsFromStore`
* get logbook and call `WriteDatasetDelete`
* call `repo.DeleteRef`
* cleanup fsi directory

When removing one of your own datasets, if there's no logbook for the dataset, both `base.DatasetLog` and `base.RemoveNVersionsFromStore` will create an oplog for that dataset. However, if the dataset is foreign (`add`ed from a peer), neither of these will create that oplog, which causes the call to `WriteDatasetDelete` to fail.

1. The function `Remove` shouldn't fail if `WriteDatasetDelete` returns a "not found" error. This is a case of "there's nothing to remove", so the function should log.Debug and keep going.
2. We need a test for this situation, but right now there's no good utility for testing foreign datasets. That's a huge weakspot in our tests and needs to be improved.

dustmop:
Fixed by #1041.

PR #1041, opened 2019-12-06
fix(remove): Remove works on foreign datasets, even if logbook missing

dustmop:
When removing a dataset, ignore logbook "not found" errors, since we're removing the dataset anyway, it doesn't matter if something is missing. Fmt the mock_client added in the previous PR.


dustmop:
This will fix https://github.com/qri-io/qri/issues/1035.

Issue #1038, opened 2019-12-03
a test is creating a file cmd/repo/logbook.qfb


dustmop:
Some test, probably in cmd/repo and related to logbook, is creating the file "cmd/repo/logbook.qfb". This should be fixed, tests shouldn't be writing to the source directory.

dustmop:
Was able to narrow down the specific test. I used the following:

```
grep "func Test" cmd/*.go | awk '{print $2}' > ../commands.sh
```

then used a keyboard macro to make the a script like this:

```
go test ./cmd/. -run TestAddNoParam
ls cmd/repo/logbook.qfb
go test ./cmd/. -run TestAddAndListRefs
ls cmd/repo/logbook.qfb
go test ./cmd/. -run TestAddWithCheckout
ls cmd/repo/logbook.qfb
...
```

By running that, I found that `TestSearchRun` is creating this file.

dustmop:
Fixed by https://github.com/qri-io/qri/pull/1048.

PR #1048, opened 2019-12-09
fix(search): Run TestSearchRun in a tmp directory

dustmop:
Run this test in a temporary directory. This prevents repo/logbook.qfb from being created when the test runs.


dustmop:
Fixes https://github.com/qri-io/qri/issues/1038.

Issue #1055, opened 2019-12-13
bug(remove): remove hanging when removing a dataset not in my namespace


ramfox:
This bug only occurs when the Qri daemon is running.

The remove command hangs when the api is on and you try to remove a dataset from another user's namespace.

dustmop:
This is happening because base.RemoveNVersionsFromStore fails to look up dataset versions that don't exist locally, it just hangs trying to fetch them from IPFS. When not connected, these lookups fail fast instead.

Fix incoming.

Relatedly, when we `qri add` someone's dataset, currently logbooks are not being fetched:
```
> qri add b5/world_bank_population
⠒ 16:28:09.264 ERROR        lib: pulling logs: decoding public key: proto: required field "Type" not set datasets.go:831
```

dustmop:
Fixed by #1059.

PR #1059, opened 2019-12-13
fix(remove): Remove foreign datasets while connected, don't hang

dustmop:


dustmop:
Fixes https://github.com/qri-io/qri/issues/1055.

PR #1061, opened 2019-12-18
feat(websocket): Open websocket server, watch filesystem for events.

dustmop:
Open a websocket server that clients can connect to. Watch filesystem using fsnotify. The paths to watch are all FSI working directories. If files are created, modified, or deleted in those directories then an event is sent to all websocket clients.


dustmop:
Talked to @b5 in person, good to merge this.

Issue #1060, opened 2019-12-18
qri list breaks if repo has a dataset with no history and no working directory


dustmop:
Discovered by @b5. Reproduction steps:

```
mkdir my_ds
cd my_ds
qri init --name my_ds --format csv
cd ..
rm -rf my_ds
qri list
```

This ends up with a ref that has no Path and no FSIPath (due to https://github.com/qri-io/qri/pull/1049), which causes repo.PutRef to reject it.  `qri list` does not handle this error properly, bailing out instead of skipping that entry. Ideally, the ref should be removed entirely, since it now refers to nothing at all.

dustmop:
Fixed by https://github.com/qri-io/qri/pull/1062.

PR #1062, opened 2019-12-18
fix(fsi): Remove references that have no path and no working directory.

dustmop:


dustmop:
Fixes https://github.com/qri-io/qri/issues/1060

PR #1070, opened 2020-01-10
fix(ref): CanonicalizeDatasetRef sets fsipath even if ref is full.

dustmop:


dustmop:
Fixes https://github.com/qri-io/desktop/issues/379.

PR #1071, opened 2020-01-13
fix(validate): Improve validate so it works with FSI

dustmop:
FSI broke validate because we switched from schema.json back to structure.json. The old functionality was looking specifically for schema.json (in the cmd package, no less) and did not survive this change. Instead, we should use the FSI package which correctly finds relevant component files.

Cleanup the lower levels of how validate works. Before, it was converting things back and forth, between cmd -> lib -> base -> dataset -> jsonschema. No need to do all of that unnecessary work.

### DIFF
--- a/base/validate.go
+++ b/base/validate.go
@@ -1,116 +1,46 @@
 package base
 
 import (
-	"bytes"
 	"context"
-	"encoding/json"
 	"fmt"
 	"io/ioutil"
 
 	"github.com/qri-io/dataset"
-	"github.com/qri-io/dataset/detect"
-	"github.com/qri-io/dataset/dsio"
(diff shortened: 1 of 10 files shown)


ramfox:
After an in room discussion w/ @dustmop, the qri command line command `qri validate` takes a `body` filepath and a `schema` filepath. However, this means that any of a user's fsi datasets can not be easily passed into the `qri validate` command. Since we have switched to using `structure.json` files in fsi to represent the structure, you cannot easily use the body of one dataset with the `structure` of another, since the command is expecting a `schema`.

We decided that there should be both a `schema` and `structure` flag, and it should be invalid to use both.

PR #1072, opened 2020-01-17
fix(change username): successful registry signup that changes a username must not break datasets made with prior peer name

b5:
as mentioned before, changing your peer name is precarious. This makes it one degree less-so.

Before we can merge this, we need a high-level test that does the following:

1. creates a new qri repo
2. creates a dataset
3. Adds a commit to that dataset
4. signs up with a fake registry, changing their peername in the process
5. run `qri ls`, must work
6. rename that dataset
7. run qri ls again, must work.


b5:
Looks great! I think we should merge

PR #1073, opened 2020-01-17
fix(log): Add --log-all flag to enable logging everywhere

dustmop:
The --log-all flag can be used by Desktop to log more information to the default file we create at startup. Add logging to lib.Get errors and fsi.Checkout so we can check errors that occur in the wild.

### DIFF
--- a/lib/lib.go
+++ b/lib/lib.go
@@ -106,6 +106,7 @@ type InstanceOptions struct {
 	regclient  *regclient.Client
 	statsCache *stats.Cache
 	logbook    *logbook.Book
+	logAll     bool
 
 	remoteMockClient bool
 	// use OptRemoteOptions to set this
@@ -189,6 +190,14 @@ func OptCheckConfigMigrations(cfgPath string) Option {
 		if migrated {
 			return o.Cfg.WriteToFile(cfgPath)
 		}
+		return nil
+	}
+}
+
+// OptSetLogAll sets the logAll value so that logging is enabled for all packages
+func OptSetLogAll(logAll bool) Option {
+	return func(o *InstanceOptions) error {
+		o.logAll = logAll
 		return nil
 	}
 }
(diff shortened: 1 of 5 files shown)


lib/lib.go
```diff
@@ -306,6 +315,16 @@ func NewInstance(ctx context.Context, repoPath string, opts ...Option) (qri *Ins
 		}
 	}
 
+	// if logAll is enabled, turn on logging for all packages. Packages need to be explicitly
+	// enumerated here
+	if o.logAll {
+		allPackages := []string{"qriapi", "qrip2p", "base", "cmd", "config", "dsref", "fsi", "lib", "logbook", "repo"}
```
b5: This is going to spit out an _insane_ amount of logs, particularly `qrip2p`, which IIRC spits out a debug message _on every peer connection attempt_. Have you run this & looked at just how verbose the output gets?

If the output is in fact too much, I think the first answer is just to curb the amount of debug messaging in `qrip2p`.

lib/lib.go
```diff
@@ -193,6 +194,14 @@ func OptCheckConfigMigrations(cfgPath string) Option {
 	}
 }
 
+// OptSetLogAll sets the logAll value so that logging is enabled for all packages
```
b5: I'd love a note that mentions this turns on _debug level logging_ for all _qri_ packages. go-log is also connected to IPFS.

lib/lib.go
dustmop: It's not ideal, but I don't think it's *too* bad. Right now, `/status` requests are worse than p2p connections, but those will soon get fixed by websockets. I just ran the app for about 10 minutes, and the resulting logfile is 95kb. This means that an hour will get to about 640kb, less than even a Meg. Running for a full day will make a 15mb file. We're going to have to post-filter logs anyway (to get rid of the color bytes), so post-filtering to get rid of p2p connection information won't be too hard to do, and the rest of the log should remain entirely readable. The important thing is that p2p connection attempts only output debug messages every few seconds, so it's not like a sequence of important steps (within the handling of a single api request) will get broken up by a bunch fo p2p messages.

lib/lib.go
dustmop: Done.

PR #1075, opened 2020-01-23
docs(cmd): document registry signup and prove commands

b5:
we had someone ask in the discord how to connect to an existing qri.cloud account, and I realized we haven't documented these commands at all.

### DIFF
--- a/cmd/registry.go
+++ b/cmd/registry.go
@@ -9,15 +9,17 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// NewRegistryCommand creates a `qri registry` subcommand for working with configured registries
-// TODO - registry command is currently removed in favor of the newer "qri publish" command
+// NewRegistryCommand creates a `qri registry` subcommand for working with the
+// configured registry
+// TODO (b5) - registry publish commands are currently removed in favor of the
+// newer "qri publish" command.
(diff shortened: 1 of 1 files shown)


cmd/registry.go
```diff
@@ -117,12 +129,25 @@ This dataset will no longer show up in search results.`,
 		},
 	}
 
-	signup.Flags().StringVar(&o.Username, "username", "", "registry username to prove")
-	signup.Flags().StringVar(&o.Password, "password", "", "registry password")
-	signup.Flags().StringVar(&o.Email, "email", "", "email")
+	signup.Flags().StringVar(&o.Username, "username", "", "desired username")
+	signup.Flags().StringVar(&o.Password, "password", "", "a new password for online login")
```
dustmop: So, this isn't urgent, and it's already the status quo, but I would really like at some point in the future to move away from requiring users to enter passwords as command-line flags. It's bad for multiple reasons. My preference would be to always use something like https://godoc.org/golang.org/x/crypto/ssh/terminal#Terminal.ReadPassword which won't echo back to the terminal, and provides just a bit more protection for sensitive info.

b5: love it, mind filing an issue?

PR #1078, opened 2020-01-24
test(cmd): refactor TestRepoRoot down into repo/test package

b5:
Now that we have the TestRunner struct in cmd it's possible to refactor a bunch of cmd-specific bits up into that struct, and push the repo mocking bits down so other packages can use it. This kills a long-standing TODO in the process.

I've also merged the lib/test package into the repo/test package. The central magic of a pre-zipped IPFS repo comes from there, and is only being used by cmd tests.


cmd/test_runner.go
```diff
@@ -3,59 +3,68 @@ package cmd
...
 	Context     context.Context
 	ContextDone func()
 	TsFunc      func() time.Time
 	CmdR        *cobra.Command
 	Teardown    func()
+
+	pathFactory PathFactory
 }
 
 // NewTestRunner constructs a new TsetRunner
```
dustmop: Could you fix this typo as well, please

repo/test/test_repo.go
```diff
@@ -351,3 +356,150 @@ var BadStructureFile = qfs.NewMemfileBytes("badStructure.csv", []byte(`
...
+	QriPath := filepath.Join(RootPath, "qri")
+	err = os.MkdirAll(QriPath, os.ModePerm)
+	if err != nil {
+		return r, err
+	}
+	// Create empty config.yaml into the test repo.
+	cfg := config.DefaultConfigForTesting().Copy()
+	cfg.Profile.Peername = peername
+
+	// PathFactory returns the paths for qri and ipfs roots.
```
dustmop: Remove these two lines

repo/test/test_repo.go
```diff
@@ -351,3 +356,150 @@ var BadStructureFile = qfs.NewMemfileBytes("badStructure.csv", []byte(`
...
+	// If dataset doesn't exist, return an empty string for the path.
+	if len(refs) == 0 {
+		return "", err
+	}
+
+	return refs[index].Path, nil
+}
+
+// ReadBodyFromIPFS reads the body of the dataset at the given keyPath stored in CAFS.
+// TODO (b5): reprecate this rediculous function
```
dustmop: Though this function shouldn't be used in production code, I think it's pretty acceptable both for (1) test code, and also (2) to document quickly how our stack is using IPFS to store data. I would rather replace the TODO with a notation like "only for use in test code".

repo/test/test_repo.go
```diff
@@ -351,3 +356,150 @@ var BadStructureFile = qfs.NewMemfileBytes("badStructure.csv", []byte(`
...
+func (r *MockRepo) ReadBodyFromIPFS(keyPath string) (string, error) {
+	ctx := context.Background()
+	// TODO: Perhaps there is an existing cafs primitive that does this work instead?
```
dustmop: I wrote this comment originally (I'm pretty sure?) and the answer seems to be "no", so please remove it.

PR #1080, opened 2020-01-27
feat: add initial "browse" lib methods, utilities for testing registry interaction, much refactoring

b5:
Ok, this PR does too many things at once. Apologies. Here's the highlights:

* Adds initial support for `browse`, library handlers & actions for surfing the qri network (sourcing from either p2p or the registry)
* renames the new `repotest.MockRepo` into `repotest.TempRepo` to reflect the _mocks vs. temp pattern_ (more on that below) 
* moves a bunch of repo constructor code from `lib.NewInstance` down into a new package: `buildrepo`
* organize repo sub packages more, creating a new `spec` subpackage specifically for testing if a repo implementation conforms to the repo spec.
* Re-establishes a clean set of HTTP endpoints for a registry, delegating a great deal of work to remote endpoints
* creates `temp_registry_server`, a separate binary for mocking a registry outside of go programming environments

### Temp Pattern
I think we need a way to distinguish between _mocks_ and this thing where we create & tear down full stuff and _temporary resources_ that a test may create. I'm proposing we formally separate the two words:

A *mock* stubs out some resource without actually being that resource. They're lightweight, carry little business logic, and are great for unit tests. mocks have the word `Mock` somewhere in their name. 

A *temp* is an ephemeral instance of a resource that fully implements it's business logic. Temp have the word `Temp` somewhere in their name, are great for integration tests. Temp resources need to be _cleaned up_ after their done. A `Temp` _must_ return a cleanup function that removes the resources consumed when called.


cmd/test_runner.go
```diff
@@ -40,28 +44,45 @@ func NewTestRunner(t *testing.T, peerName, testName string) *TestRunner {
...
+	if err != nil {
+		t.Fatalf("creating registry: %s", err)
+	}
+
+	// TODO (b5) - wouldn't it be nice if we could pass the client as an instance configuration
+	// option? that'd require re-thinking the way we do NewQriCommand
+	_, server := regserver.NewMockServerRegistry(*reg)
+
+	runner := newTestRunnerFromRoot(&root)
+	runner.Teardown = func() {
```
dustmop: This works now, but could potentially hide a bug, if there already exists a Teardown function in some future codebase. Should assign `prevTeardown := runner.Teardown`, then called `prevTeardown()` inside the new cleanup function.

registry/temp_registry_server/server.go
```diff
@@ -3,41 +3,38 @@
...
 		port = "3000"
 	}
 
-	pro := handlers.NewBAProtector("username", adminKey)
-	ps := registry.NewMemProfiles()
-	reg := registry.Registry{
-		Profiles: ps,
+	inst, reg, cleanup, err := NewTempRepoRegistry(ctx)
+	if err != nil {
+		log.Fatalf("creating mock registry: %s", err)
```
dustmop: this is a temp, not mock, right?

dustmop:
Also, please bump version/version.go to 0.9.5-dev

PR #1081, opened 2020-01-27
feat(dscache): Dscache holds info about datasets based upon logbook

dustmop:
Define dscache as a flatbuffer. Add a function that converts a (logbook + peerList + dsRef) triple into a dscache. Add a flag to `list` that will build this dscache and list datasets from the result. Cleanup some naming in logbook for the PlainLogs that are used to implement `logbook --raw`.


dustmop:
No tests yet. Wanted to get the design out first, will work on tests all of tomorrow.

PR #1082, opened 2020-01-28
test(TwoActorRegistryIntegration): lib-level integration test showing logsync & clone behaviour

b5:
wanted to get this test in place to both prove things are working properly in go land, and lay the foundation for writing this same test as an e2e test on desktop.


b5:
merging w/out review because we're shorthanded today.

PR #1083, opened 2020-01-28
test(dscache): Tests for dscache, especially conversion. TODOs for more.

dustmop:


dscache/capture_log.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	golog "github.com/ipfs/go-log"
+)
+
+// captureLog can replace this package's logger so that it counts calls instead of actually logging.
+// To be used only from tests
```
b5: If this is only for tests rename this file to `capture_log_test.go` to prevent anything in here from making it into the compiled binary

dscache/convert_test.go
```diff
@@ -0,0 +1,327 @@
...
+			FSIPath: "/path/to/second_workspace",
+		},
+	}
+
+	dsInfoList, err := convertLogbookAndRefs(ctx, book, dsrefs)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	dsBytes, err := json.MarshalIndent(dsInfoList, "", " ")
```
b5: why convert this to JSON? Any reason not to just hand-construct the expected `dsInfoList` & feed that to `cmp.Diff`?

dscache/convert_test.go
```diff
@@ -0,0 +1,327 @@
...
+		t.Errorf("expected to get 1 call to log.Error, got %d", gotNum)
+	}
+	gotText := replaceLog.ErrorLast()
+	expectText := "could not find dataset second_name in dsrefs"
+	if gotText != expectText {
+		t.Errorf("expected to get %q, got %q", expectText, gotText)
+	}
+
+	// Still successfully got a list of dataset references.
+	dsBytes, err := json.MarshalIndent(dsInfoList, "", " ")
```
b5: ditto on the JSON

dscache/capture_log.go
```diff
@@ -0,0 +1,96 @@
...
+type captureLog struct {
```
b5: I'm not a huge fan of using logs for tests. Why do you think this is the right approach in this context?

dscache/capture_log.go
dustmop: Done.

dscache/convert_test.go
dustmop: I had thought it would be easier to take the json approach, but trying it just now using literals, I was clearly mistaken.

dscache/capture_log.go
dustmop: So, I am unsure about this. The real goal of `TestConvertLogbookAndRefsMissingDsref` is to make sure that the dscache conversion will work if the logbook knows about a ref that is not found in the dsrefs. But this probably will never happen, while the opposite *will* happen since there are still cases were users (us developers) may have entries in our dsrefs that are not in logbook. Asserting that the log.Error in findFSIPath gets hit isn't really important to the test itself, but it helps ensure that expectations about the code are being met.

Could you elaborate on the problem you raised with using this mock log? Wondering if it's worth using it like this, or if the downside outweighs the (minor) benefit mentioned above.

b5: My concern is this change makes logging take on an additional meaning beyond just spitting out data about a running program. Adding or removing logging information should never break a program (tests are a program), and that'll happen here if a developer alters logging output.

We haven't yet made the jump from just logging stuff to gathering observability metrics which is the thing we _should_ be overloading logging with, but when we do I expect our approach to logging to change a bunch. I'd heavily prefer changes to logging _never_ affect our test suite.

If we need to test something that feels like it can't be tested, let's change code to make it testable.

dustmop: Okay, removed. While exploring ahead I realized this function `findFSIPath` is going to change anyway, so it's also not worth adding here.

PR #1085, opened 2020-01-29
feat(remote): ephemeral log fetching with remoteClient.FetchLogs

b5:
we need a way to fetch logs & _not_ persist them. This does that.

I've also added a --logs-only flag to add which brings back the old fetch behaviour of persisting logs on fetch. Long term I'd like to merge all log pulling into "qri log" (for ephemeral fetching) and "qri clone" (for persisted fetching)


b5:
merging w/out review because we're shorthanded today & time is a precious thing

PR #1087, opened 2020-01-30
feat(registry): add preview support to registry fetching

b5:
Initial support for registry-only preview fetching for desktop


base/preview.go
```diff
@@ -0,0 +1,47 @@
...
+		log.Errorf("CreatePreview opening body file: %s", err.Error())
+		return nil, err
+	}
+
+	st := &dataset.Structure{
+		Format: "json",
+		Schema: ds.Structure.Schema,
+	}
+
+	data, err := ConvertBodyFile(ds.BodyFile(), ds.Structure, st, 100, 0, false)
```
dustmop: I would much prefer if this 100 was a well-named constant, like `defaultNumBodyRowsInPreview`, in order to call attention to it, and make it clear that it was a deliberate choice. As opposed to the couple of places in our code where we use something like 100000 to essentially mean "infinite" (see calls to `References`).

registry/regserver/handlers/handlers.go
```diff
@@ -83,3 +93,70 @@ func HealthCheckHandler(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusOK)
 	w.Write([]byte(`{"meta":{"code": 200,"status":"ok"},"data":null}`))
 }
+
+// HomeFeedHandler provides access to the home feed
+func HomeFeedHandler(r repo.Repo) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		refs, err := base.ListDatasets(req.Context(), r, "", 30, 0, false, true, false)
```
dustmop: Same comment here about using a constant instead of 30.

PR #1088, opened 2020-01-30
fix(dscache): Add refs in dsref that logbook is missing. More tests.

dustmop:
Our repos may contain datasets that are in dsref but not in logbook. This can happen from datasets created before logbook was implemented, or due to deleting logbook (which was necessary to work around bugs for a little while), or from the registry not syncing logs yet. For now, such dsrefs should still be added to dscache, but this behavior will not be needed in the future once we move over to dscache all the time.

Also add some more tests for dscache building, which eliminates a few TODOs.

### DIFF
--- a/dscache/convert_test.go
+++ b/dscache/convert_test.go
@@ -130,7 +130,7 @@ func (run *DscacheTestRunner) Delete() {
 }
 
 // Test the convertLogbookAndRefs function, which turns a logbook and dsrefs into a list of dsInfo
-func TestConvertLogbookAndRefs(t *testing.T) {
+func TestConvertLogbookAndRefsBasic(t *testing.T) {
 	run := NewDscacheTestRunner()
 	defer run.Delete()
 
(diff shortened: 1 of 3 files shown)


dscache/convert_test.go
```diff
@@ -233,12 +236,152 @@ func TestConvertLogbookAndRefsMissingDsref(t *testing.T) {
...
+	defer run.Delete()
+
+	ctx := context.Background()
+
+	peerInfo := testPeers.GetTestPeerInfo(0)
+	book := makeFakeLogbook(ctx, t, "test_user", peerInfo.PrivKey)
+
+	// Dsrefs has a third reference that is not in logbook.
+	dsrefs := []repo.DatasetRef{
+		repo.DatasetRef{
```
b5: For future reference, the go compiler can infer the types of elements of a slice in declaration like this, you can use just curly braces.

PR #1095, opened 2020-02-05
feat(dscache): Dscache will get updated by save, if it exists in the repo

dustmop:
Add a logbook.Action that describes a change to logbook when a save happens.

Add dscache.Update which receives an action, and modifies the dscache (by copying, since flatbuffers don't support mutation), and saves the result to the repo.

Change dscache constructor so that a non-null dscache is always returned, but it'll just be "empty" if there's nothing in the repo. That allows us to "save" since a filename within the repo should always exist.

Rename --via-dscache flag to --use-dscache.

Make an integration test that demonstrates using `list` to make a dscache, then `save` to update it.

### DIFF
--- a/base/save.go
+++ b/base/save.go
@@ -200,8 +200,17 @@ func CreateDataset(ctx context.Context, r repo.Repo, streams ioes.IOStreams, ds,
 	ds.ProfileID = pro.ID.String()
 	ds.Peername = pro.Peername
 	ds.Path = path
-	if err = r.Logbook().WriteVersionSave(ctx, ds); err != nil && err != logbook.ErrNoLogbook {
+	action, err := r.Logbook().WriteVersionSave(ctx, ds)
+	if err != nil && err != logbook.ErrNoLogbook {
 		return
+	}
+	dscache := r.Dscache()
+	if dscache != nil && !dscache.IsEmpty() {
+		log.Info("dscache: update and save new version")
+		err = dscache.Update(action)
+		if err != nil {
+			log.Error(err)
+		}
 	}
 
 	if err = ReadDataset(ctx, r, &ref); err != nil {
--- a/cmd/save_test.go
+++ b/cmd/save_test.go
@@ -7,8 +7,11 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/qri-io/ioes"
+	"github.com/qri-io/qfs/localfs"
 	"github.com/qri-io/qri/base/dsfs"
+	"github.com/qri-io/qri/dscache"
 	"github.com/qri-io/qri/lib"
+	"github.com/qri-io/qri/logbook"
 	"github.com/qri-io/qri/startf"
 )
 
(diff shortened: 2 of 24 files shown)


dscache/dscache.go
```diff
@@ -18,28 +22,48 @@ var (
...
+		// Ignore error, as dscache loading is optional
+		return &Dscache{Filename: filename}
 	}
 	root := dscachefb.GetRootAsDscache(buffer, 0)
-	return &Dscache{Root: root, Buffer: buffer}, nil
+	return &Dscache{Filename: filename, Root: root, Buffer: buffer}
+}
+
+// IsEmpty returns whether the dscache has any constructed data in it
+func (d *Dscache) IsEmpty() bool {
```
b5: can we make `Dscache` nil-callable? We could do this in a follow-up PR, but given that we're modifying the repo interface to include one, I'd love to be able to wire up a nil cache & not have everything explode. This test spells it out:

https://github.com/qri-io/qri/blob/e3086cfd1944cc782f960356cb4bf6bdca618bfb/logbook/logbook_test.go#L183-L232

Many pointer methods are calling `IsEmpty` as a guard, which is already halfway there. Could also add:
```go
func (d *Dscache) IsEmpty() bool {
  if d == nil {
    return true
  }
  return d.Root == nil
}
```

for all non-error-return funcs

dscache/dscache.go
```diff
@@ -18,28 +22,48 @@ var (
...
 	ProfileIDToUsername map[string]string
 }
 
-// LoadDscacheFromFile will load a dscache from the given filename
-func LoadDscacheFromFile(filename string) (*Dscache, error) {
-	buffer, err := ioutil.ReadFile(filename)
+// NewDscache will construct a dscache from the given file, or will construct an empty dscache
+// that will save to the given filename
+func NewDscache(fsys qfs.Filesystem, filename string) *Dscache {
+	ctx := context.TODO()
```
b5: Looks like this constructor needs to take a context arg. I know, irritating.

repo/repo.go
```diff
@@ -69,6 +70,9 @@ type Repo interface {
 	// Repos have a logbook for recording & storing operation logs
 	Logbook() *logbook.Book
 
+	// Dscache is a cache of datasets that have been built according to logbook
+	Dscache() *dscache.Dscache
```
b5: Can we move this closer to refstore, and add a note to refstore mentioning that we're transitioning toward using dscache?

dscache/dscache.go
```diff
@@ -18,28 +22,48 @@ var (
...
 	Buffer              []byte
 	ProfileIDToUsername map[string]string
 }
 
-// LoadDscacheFromFile will load a dscache from the given filename
-func LoadDscacheFromFile(filename string) (*Dscache, error) {
-	buffer, err := ioutil.ReadFile(filename)
+// NewDscache will construct a dscache from the given file, or will construct an empty dscache
+// that will save to the given filename
+func NewDscache(fsys qfs.Filesystem, filename string) *Dscache {
```
b5: Can we get a comment mentioning that passing the empty string as `filename` will skip loading and saving entirely? It's a great convention, I'd like consumers to know it's both available & intentional.

dscache/mutator.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	dscachefb.DscacheStartUsersVector(builder, len(userList))
+	for i := len(userList) - 1; i >= 0; i-- {
+		u := userList[i]
+		builder.PrependUOffsetT(u)
+	}
+	return builder.EndVector(len(userList))
+}
+
+func (d *Dscache) copyReferenceListWithReplacement(
```
b5: This is incredibly fancy. Like, jazz-hands-level fancy. fun.

base/save.go
```diff
@@ -200,9 +200,18 @@ func CreateDataset(ctx context.Context, r repo.Repo, streams ioes.IOStreams, ds,
...
 		return
 	}
+	dscache := r.Dscache()
+	if dscache != nil && !dscache.IsEmpty() {
+		log.Info("dscache: update and save new version")
+		err = dscache.Update(action)
+		if err != nil {
+			log.Error(err)
+		}
+	}
```
b5: These two things must always be called together. We should keep an eye out for other places where two subsystems have required interactions like this. Lots of this PR is feeling like another step toward an event bus

cmd/save_test.go
```diff
@@ -235,6 +238,78 @@ func TestSaveInferName(t *testing.T) {
 	}
 }
 
+func TestSaveDscache(t *testing.T) {
```
b5: Solid story test. Ty!

cmd/test_runner_test.go
```diff
@@ -168,6 +169,15 @@ func (run *TestRunner) DatasetMarshalJSON(t *testing.T, ref string) (data string
...
 }
 
+// Dscache returns the dscache from the repo
+func (run *TestRunner) Dscache(t *testing.T) *dscache.Dscache {
+	repo, err := run.RepoRoot.Repo()
+	if err != nil {
+		t.Fatal(err)
+	}
+	return repo.Dscache()
+}
```
b5: This is only used in one place, and its main point of failure is on Repo not existing. Do we really need it?

logbook/action.go
```diff
@@ -0,0 +1,22 @@
...
+// ActionType is the type of action that a logbook just completed
+type ActionType byte
+
+const (
+	// ActionMoveCursor is an action that moves the cursor for a dataset
+	ActionMoveCursor ActionType = iota
+)
+
+// Action represents the result of an action that logbook just completed
+type Action struct {
```
b5: This can stay here for now, but I think in the _very_ near future everything in this file should move to a new package: `event`. 

`Action` is a concrete type that currently bridges between `dscache` and `logbook`. To that end, I don't think this belongs in logbook, as it's not really logbook's primary concern

logbook/logbook_test.go
```diff
@@ -73,7 +73,7 @@ func Example() {
 
 	// create a log record of the version of a dataset. In practice this'll be
 	// part of the overall save routine that created the above ds variable
-	if err := book.WriteVersionSave(ctx, ds); err != nil {
+	if _, err := book.WriteVersionSave(ctx, ds); err != nil {
```
b5: The fact that we're ignoring this response so often (even in tests) to me is an indication that we should have some sort of internal wiring where _any_ call to WriteVersionSave is publishing an `Event` to a channel that others can subscribe to, instead of returning actions.

dscache/dscache.go
dustmop: Easy enough, done.

base/save.go
dustmop: I'm going to keep this as in for now, but I see the problem. I'll change this over to using a channel in the next dscache PR.

PR #1097, opened 2020-02-06
Fix broken link to registry in readme

andrew:
It wasn't exactly clear if it should be `https://github.com/qri-io/registry` or `https://github.com/qri-io/qri/tree/master/registry`, happy to update if I picked the wrong one.


b5:
👋 @andrew! Nice catch. We have indeed moved the registry definition into this repo, so the correct link would be `https://github.com/qri-io/qri/tree/master/registry`. If you wouldn't mind amending your commits to follow our [commit guidelines](https://github.com/qri-io/qri/blob/master/DEVELOPERS.md#-git-commit-guidelines) it'll avoid my name getting added to your commits. Something like

```
docs(readme): Fix broken link to registry
```

would be perfect. Thanks!

andrew:
Updated, might be an idea to archive the [old repo](https://github.com/qri-io/registry) or add a note to it's readme too.

b5:
Great idea. Just archived the old repo. Thanks!

PR #1102, opened 2020-02-06
refactor(dsref): Rename DetailedRef to VersionInfo

dustmop:
Add a few more fields for commit info. Use VersionInfo everywhere, instead of using one-off structs all of the place. A bunch of conversion code goes away. A couple of TODOs get eliminated.


api/log.go
```diff
@@ -54,7 +55,7 @@ func (h *LogHandlers) logHandler(w http.ResponseWriter, r *http.Request) {
 		ListParams: lp,
 	}
 
-	res := []lib.DatasetLogItem{}
```
b5: ❤️

PR #1104, opened 2020-02-07
fix(regclient): update to new search endpoint, fix response ds.Name

b5:


registry/regserver/handlers/handlers.go
```diff
@@ -52,7 +52,7 @@ func NewRoutes(reg registry.Registry, opts ...func(o *RouteOptions)) *http.Serve
 
 	pro := o.Protector
 	mux := http.NewServeMux()
-	mux.HandleFunc("/", HealthCheckHandler)
```
b5: This was covering up many bad things.

base/dataset.go
```diff
@@ -159,6 +159,8 @@ func ListDatasets(ctx context.Context, r repo.Repo, term string, limit, offset i
 				}
 				return nil, fmt.Errorf("error loading ref: %s, err: %s", ref.String(), err.Error())
 			}
+			ds.Peername = res[i].Peername
+			ds.Name = res[i].Name
```
dustmop: This through me for a loop, but I understand it now. Refs have a Peername and Name, and so do Datasets (as transient values), so we're copying them from the ref to the dataset before adding that to the result ref. The long term solution is getting rid of this duplication by switching off of reporef.Dataset, and instead using dsref.Ref and dataset.Dataset only.

registry/regserver/handlers/handlers.go
dustmop: Excellent fix.

PR #1106, opened 2020-02-07
fix(publish): Use Refselect for publish, so FSI works. Use dsref, add tests

dustmop:


repo/ref/dataset_ref.go
```diff
@@ -122,3 +121,13 @@ func (r DatasetRef) IsPeerRef() bool {
 func (r DatasetRef) IsEmpty() bool {
 	return r.Equal(DatasetRef{})
 }
+
+// SimpleRef converts a DatasetRef to a dsref.Ref
+func (r DatasetRef) SimpleRef() dsref.Ref {
```
b5: I'd prefer we use `repo.ConvertToDsref`, which includes a note not to use it. Any chance we can also move `repo.ConvertToDsref` into this package with this PR? It's extra work, but I suspect that function not living in this package is why you felt the need to create `SimpleRef` in the first place

dustmop: Done.

PR #1109, opened 2020-02-10
test(testdata): Locate testdata directory without GOPATH.

dustmop:


repo/test/test_repo.go
```diff
@@ -54,7 +54,10 @@ func init() {
 }
 
 func testdataPath(path string) string {
-	return filepath.Join(os.Getenv("GOPATH"), "/src/github.com/qri-io/qri/repo/test/testdata", path)
+	// Get the testdata directory relative to this source file.
+	_, currfile, _, _ := runtime.Caller(0)
+	testdataPath := filepath.Join(filepath.Dir(currfile), "testdata")
+	return filepath.Join(testdataPath, path)
```
b5: This is lovely! Any chance we can fix the other places this occurs within our codebase with the same thing? Other Places we rely on `GOPATH` for tests:

* https://github.com/qri-io/qri/blob/27a9009621f1e3dc12341dcb46cf3958c4133fbf/base/base_test.go#L50-L52
* https://github.com/qri-io/qri/blob/27a9009621f1e3dc12341dcb46cf3958c4133fbf/base/dsfs/dsutil/dsutil_test.go#L128-L130
* https://github.com/qri-io/qri/blob/27a9009621f1e3dc12341dcb46cf3958c4133fbf/lib/lib_test.go#L160-L162
* https://github.com/qri-io/qri/blob/27a9009621f1e3dc12341dcb46cf3958c4133fbf/repo/test/test_repo.go#L56-L186

That last one (test_repo.go) is really gross

dustmop: Sure, good call. I'll push a commit and then merge.

PR #1115, opened 2020-02-11
refactor(remote): move feed and preview interfaces into remote

b5:


dsref/version_info_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/qri-io/dataset"
+)
+
+func TestConvertToVersionInfo(t *testing.T) {
+	ds := &dataset.Dataset{
+		Peername: "a",
+		Name:     "b",
+		Path:     "c",
+		Commit: &dataset.Commit{
+			Timestamp: time.Time{},
```
dustmop: Would be nicer to use a non-zero value here, to verify that the test is actually copying the field.

PR #1121, opened 2020-02-14
fix(profileID): Disambiguate profileID construction to avoid subtle bugs

dustmop:
There's a potential for bugs related to how we construct profileIDs. I've run into this bug a few times, and the issue is getting worse during the switch to dscache. This change will help to solve the problem going forward.

We use a few different types around profileIDs:

1) ipfs's peerID which is a string of bytes derived from cryptographic information

2) repo/profile.ID which is an alias of peerID, moved into our own namespace

3) a string of a base58 encoded profileID that starts with "Qm..."

The issue originates from the fact that all three are basically strings, but with different semantics. (1) and (2) are byte sequences with the same contents, but occassionally need conversion due to having different go types. (3) is a base58 encoded string, meaning it exists as different bytes in memory, and needs decoding work done to convert it to (1) or (2). Due to the fact that all three are actually sequences of bytes, the go type system fails to catch a certain class of errors.

Many places in our code construct a profileID using a call to profile.ID(). These breakdown into three cases:

1) profile.ID(peerID)

   This is always valid, as a peerID is an alias for profile.ID. It's just a type cast.

2) profile.ID(raw bytes)

   This is mostly used in tests, as a way to construct an invalid profileID. For example profile.ID("bad"). This works because it's a simple type conversion from bytes to strings.

3) profile.ID(base58EncodedProfileID)

   This is a bug. The string should be decoded (using IDB58Decode) but here it is simply being reinterpretted.

This change proposes that we ban calls to "profile.ID" completely in our code (except in the one file that needs it, repo/profile/id.go). It's not clear from looking at the code quickly whether the call is a simple conversion between aliases or an inappropriate cast when decoding was needed.

Instead, we should have separate functions that call out these three use cases explicitly:

1) profile.IDFromPeerID(peer.ID)

   Takes a peer.ID as a parameter, so the compiler guarantees type safety

2) profile.IDRawBytes(raw bytes)

   Should only be used in tests. It will be immediately obvious if this is used with anything aside from raw strings

3) profile.IDB58Decode

   The current decoding function. We also have IDB58DecodeOrEmpty for convenience.

Finding all callsites of profile.ID:
```
grep -n -r "profile\.ID(" * 
```

### DIFF
--- a/repo/profile/id.go
+++ b/repo/profile/id.go
@@ -45,10 +45,29 @@ func (id *ID) UnmarshalYAML(unmarshal func(interface{}) error) (err error) {
 	return
 }
 
+// IDRawBytes constructs an ID from raw bytes. No decoding happens. Should only be used in tests
+func IDRawBytes(data string) ID {
+	return ID(data)
+}
+
+// IDFromPeerID type casts a peer.ID from ipfs into an ID
+func IDFromPeerID(pid peer.ID) ID {
+	return ID(pid)
+}
+
 // IDB58Decode proxies a lower level API b/c I'm lazy & don't like
 func IDB58Decode(proid string) (ID, error) {
 	pid, err := peer.IDB58Decode(proid)
 	return ID(pid), err
+}
+
+// IDB58DecodeOrEmpty decodes an ID, or returns an empty ID if decoding fails
+func IDB58DecodeOrEmpty(proid string) ID {
+	pid, err := peer.IDB58Decode(proid)
+	if err != nil {
+		pid = ""
+	}
+	return ID(pid)
 }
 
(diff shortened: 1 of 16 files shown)


repo/profile/id.go
```diff
@@ -45,12 +45,31 @@ func (id *ID) UnmarshalYAML(unmarshal func(interface{}) error) (err error) {
 	return
 }
 
+// IDRawBytes constructs an ID from raw bytes. No decoding happens. Should only be used in tests
+func IDRawBytes(data string) ID {
```
b5: function name `RawBytes` that accepts a `string` type feels smelly to me. Can we rename to `IDRawByteString`?

dustmop: Will do, good idea.

PR #1122, opened 2020-02-14
fix(dsref): Parse dsref from strings. Use for rename and init.

dustmop:
Parser for dsref. Use for the rename command, and for validating dataset names for init. Refactor base.ModifyDatasetRef so it is more obvious how it works.


dustmop:
Include one of the changes from https://github.com/qri-io/qri/pull/1121 (repo/profile/id.go). Tests are in that PR.

dsref/parse.go
```diff
@@ -0,0 +1,162 @@
...
+//
+// Parse will parse a human-friendly reference, or only a concrete reference, or a full reference.
+//
+// ParseHumanFriendly will only successfully parse a human-friendly reference, and nothing else.
+//
+// The grammar is here:
+//
+//  <dsref> = <humanFriendlyRef> [ <concreteRef> ] | <concreteRef>
+//  <humanFriendlyRef> = <username> '/' <datasetname>
+//  <concreteRef> = '@' [ <profileID> ] '/' <network> '/' <commitHash>
```
b5: This is great. I think we should merge as-is because it's a better, tested implementation of the way we currently think about ref parsing. 

I'd like to change a few things about the grammar for reference strings in the future, and would like to structure these as follow-up PRs, using this one as a baseline

* we need consistent naming for the components of a reference string. in some places we call the left side an `Alias`, others the `Name`, and here we're now adding `human-friendly reference`. 
* We need to introduce grammar for reference strings that use logbook identifiers.
* profileID should move within the grammar, and be a variant of a human friendly name. 

Lots to think about, just wanted to surface here so we can revisit this PR later.

b5:
oh yes, let's def merge #1121 first

PR #1123, opened 2020-02-14
refactor(dscache): Rename dsInfo to entryInfo. Align with RefEntryInfo in fbs

dustmop:


dustmop:
The conversion code internal to dscache was using its own struct called dsInfo. There's no need for this. Instead, wrap VersionInfo, adding the the unique index fields, and call the new struct entryInfo. Rename the fbs struct RefCache to be RefEntryInfo so it matches this internal entryInfo struct.

Issue #1129, opened 2020-02-19
bug(get): different responses for components w/ scriptPaths in linked vs non linked dataset


ramfox:
Currently working on the feature to allow users to edit datasets directly in the Desktop app.

When requesting a Dataset from the backend (`/peername/datasetname`) of a linked dataset, the dataset comes back with the `Readme.ScriptBytes` field hydrated.

However, when requesting a non-linked dataset at the same endpoint, the `Readme.ScriptPath` field is hydrated, not the `ScriptBytes` field.

Requesting that Datasets return with the same expected fields hydrated for both fsi `true` & `false`.

I would prefer it if `Readme.ScriptBytes` were the chosen field, but would understand if we need to switch to using a `Selectors` param in the get request in order to get the `ScriptBytes`.

b5:
closed by #1135

PR #1135, opened 2020-02-21
fix(api): inline script bytes for get requests

b5:
Turns out all we need to do is consume the files opened by `lib.DatasetRequests.Get`'s `OpenDataset` invocation.

I'm being a terrible testing citizen. `dsutil.InlineScriptsToBytes` needs a test real bad, but I'm in a jam. Will file an issue.


b5:
This closes #1129

PR #1140, opened 2020-02-27
feat(dscache): Fill dscache when saving, using func pointer in logbook.

dustmop:
Logbook has a callback for init and save actions. Dscache registers to receive these actions. When saving a dataset for the first commit, the flag --use-dscache will opt-in to dscache by creating a new instance.


b5:
Lemmie PR a version bump first

PR #1143, opened 2020-03-01
chore(circleci): add performance benchmarks on master branch push

### LINKED ISSUE #1112: Qri takes ages when loading a large file as a body
I tried adding a dataset to Qri, but the dataset is modestly large (~7GB or so), and Qri seems to have serious trouble with it. It seems like it is parsing the whole dataset, or trying to do something with the whole dataset, while setting up my project - because it takes forever.

The file I'm adding is in CSV format, and is well-formed, so there is no need to parse it except to infer structure, which should be an optional feature anyway. There are a handful of columns (certainly less than 10), all with simple types (strings, numbers).

Is this a bug or expected behavior?

b5:
closes #1112. This test will run & publish an update to [qri_eng/qri_cli_user_benchmarks](https://qri.cloud/qri_eng/qri_cli_user_benchmarks) whenever we push to master. Now that we can measure performance, we'll know when it's better 😄

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,5 @@
-version: '2'
+version: '2.1'
+
 jobs:
   build:
     working_directory: /go/src/github.com/qri-io/qri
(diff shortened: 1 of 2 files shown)


b5:
closes #1112

PR #1148, opened 2020-03-03
fix(base/dsfs/dataset): avoid redundant validation on metadata updates

### LINKED ISSUE #1145: saving with only a meta change should be fast
When a user updates their meta for an existing dataset, using either Desktop or `qri save --file=meta.json ...`, it should be very fast to save because so little is changing. However, `prepareDataset` in `base/dsfs/dataset.go` does a lot of work even in this case. It will assign `bfPrev` the bodyfile from the previous commit, then a few lines later sees that `bf` is nil, and assigns `bf = bfPrev`. Then some goroutines kick off to calculate the `errCount`, `depth`, `entryCount`, `checksum` etc by reading the entire body file.

However, this step is unnecessary. The code path here knows that it is reusing the body from the previous version, so there's no way the structure can vary if nothing has changed. In this situation, prepare should just copy the structure component and skip all of this goroutine creation.

Hackerpilot:
Skip some validation steps in the case that the body file is not being
changed in order to speed up saving.

Closes #1145

I'm very new to this codebase and the Go langauge, so I'm not sure if this is the right way to solve the issue. Review would be greatly appreciated.

### DIFF
--- a/base/dsfs/dataset.go
+++ b/base/dsfs/dataset.go
@@ -262,52 +262,57 @@ func prepareDataset(store cafs.Filestore, ds, dsPrev *dataset.Dataset, privKey c
 		return fmt.Errorf("bodyfile or previous bodyfile needed")
 	}
 
+	// If bf is nil, this is a metadata update and we will assume that the
+	// previous version of the file had passed validation.
 	if bf == nil {
 		bf = bfPrev
-	}
+	} else {
+		errR, errW := io.Pipe()
+		entryR, entryW := io.Pipe()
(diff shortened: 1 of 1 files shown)


Hackerpilot:
Yeah, it looks like I messed up in a few ways here. I tried running the tests, and saw a message in the output about `listen tcp :80: bind: permission denied update.go:86`, and I jumped to the conclusion that the failures were caused by not running the tests in Docker or something.

Thanks for the feedback. I'll look for a better first issue.

PR #1149, opened 2020-03-04
feat(diff): update to new deepdiff output, show context in diffs

b5:


base/friendly/friendly.go
```diff
@@ -107,38 +115,48 @@ func DiffDescriptions(deltas []*deepdiff.Delta, stats *deepdiff.Stats) (string,
 	return shortTitle, longMessage
 }
 
-// preprocess makes delta lists easier to work with, by combining operations when possible
-func preprocess(deltas []*deepdiff.Delta) []*deepdiff.Delta {
+const dtReplace = deepdiff.Operation("replace")
+
+
```
dustmop: Remove these extra blank lines

base/friendly/friendly.go
```diff
@@ -107,38 +115,48 @@ func DiffDescriptions(deltas []*deepdiff.Delta, stats *deepdiff.Stats) (string,
...
+					last.Type = dtReplace
 					continue
 				}
 			}
 		}
+
+		p := joinPath(path, d.Path.String())
+		// TODO (b5) - We need this because it's possible to write a transform script
+		// that changes nothing other than the script itself, and we currently reject
+		// that change. Should we? I think I'm missing something.
```
dustmop: Discussed offline, we should definitely *not* be rejecting a change that consists of a transform modification that produces no body diff. There are very good use cases for that sort of thing.

base/friendly/friendly_test.go
```diff
@@ -9,22 +9,21 @@ import (
 )
 
 func TestFriendlyDiffDescriptions(t *testing.T) {
+	// TODO (b5) - we should update this test to use deepdiff to generate these diffs
```
dustmop: Disagree. I think it's correct to have unit tests that only exercise the code in friendly, in isolation from how deepdiff works. It makes it easier to see the expectations that the code makes, and how it behaves in isolated cases, helping users to understand the intentions of the code. We can also have additional tests higher up that verify the two in combination, as integration tests.

cmd/fsi_integration_test.go
```diff
@@ -1029,17 +1029,26 @@ run ` + "`qri save`" + ` to commit this dataset
...
+   1: 
+     0: "four"
+     1: "five"
+    -2: 6
+    +2: 321
+ meta: 
+   qri: "md:0"
+  +title: "hello"
+ qri: "ds:0"
+ structure: {"format":"csv","formatConfig":{"lazyQuotes":true},"qri":"st:0","schema":{"items":{"items":[{"title":"field_1","type":"string"},{"title":"field_2","type":"string"},{"title":"field_3","type":"integer"}],"type":"array"},"type":"array"}}
```
dustmop: Probably can't happen in time for this PR to merge, but I don't think printing how all of structure here is very useful. Makes it really hard to read just what changed in the diff. Having nearby context is necessary in the diff output, but this is pretty far away from the other changes.

cmd/print.go
```diff
@@ -167,26 +167,17 @@ func doesCommandExist(cmdName string) bool {
...
-		stats = deepdiff.FormatPrettyStatsColor(res.Stat)
-		if !summaryOnly {
-			text, err = deepdiff.FormatPrettyColor(res.Diff)
-			if err != nil {
-				return err
-			}
+	// should use the IsATTY package from mattn
+	deepdiff.FormatPrettyStats(buf, res.Stat, !color.NoColor)
+	if !summaryOnly {
+		buf.Write([]byte("\n"))
```
dustmop:
```suggestion
		buf.WriteByte('\n'))
```

lib/diff.go
```diff
@@ -232,9 +235,44 @@ func (r *DatasetRequests) Diff(p *DiffParams, res *DiffResponse) (err error) {
...
+		if err != nil {
+			return nil, nil, err
+		}
+
+		return dd.StatDiff(ctx, left, right)
+	}
+	return dd.StatDiff(ctx, left.InferredSchema, right.InferredSchema)
+}
+
+// TODO - holy shit dis so bad. fix
```
dustmop: Can you elaborate why this is needed, what the problem is that it's temporarily working around, and what needs to be done instead to fix it?

cmd/cmd_test.go
```diff
@@ -441,21 +441,39 @@ func TestDiffRevisions(t *testing.T) {
...
+ 9: ["Harry Potter and the Half-Blood Prince ",153]
+ 10: ["Batman v Superman: Dawn of Justice ",183]
+ 11: ["Superman Returns ",169]
+ 12: ["Quantum of Solace ",106]
+ 13: ["Pirates of the Caribbean: Dead Man's Chest ",151]
+ 14: ["The Lone Ranger ",150]
+ 15: ["Man of Steel ",143]
+ 16: ["The Chronicles of Narnia: Prince Caspian ",150]
+ 17: ["The Avengers ",173]
++18: ["Dragonfly ",104]
```
dustmop: What do you think about adding a TODO to limit the amount of context shown? I feel 4 or 5 lines is plenty to understand the position where the file is changing, which is similar to what Unix Diff does. Any more than that and it gets hard to read.

b5: Agreed. I think this would be best filed as a feature request issue on deepdiff, as that's where it needs to originate from. I'd like to be able to arbitrarily set the amount of context, but for that to work we need to flesh out the spec for "jump points" that skip rows in the patch.

cmd/fsi_integration_test.go
b5: Agreed, but it's _very_ hard to come up with a solid heuristic for that sort of thing.

base/friendly/friendly_test.go
b5: I agree with your disagreement. Removing.

PR #1151, opened 2020-03-05
fix(cmd/list): show username with no datasets to list

advanderveer:
This is a very superficial fix for the username not displaying on a call to `qri list` without providing specific peer as an argument. 

It is not perfect because `repo/store.go#CanonicalizeProfile` shows that it is also possible to refer to one's own datasets by using `me` as a magic peername.  So running `qri list --peer me` now still gives a weird (meme worthy) message: `me has no datasets`. If that also needs to be taken into account the rabbit hole starts to become pretty deep for a simple feedback message. 

I would say that it is not worth complexity and just return the message "found no datasets".

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -128,10 +128,15 @@ func (o *ListOptions) Run() (err error) {
 	}
 
 	if len(infos) == 0 {
+		pn := fmt.Sprintf("%s has", o.Peername)
+		if o.Peername == "" {
+			pn = "you have"
+		}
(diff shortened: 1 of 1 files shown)


dustmop:
Seems good! Passing `--peer me` is not an expected user action, so it's okay for now if it behaves a bit strange. We're working on some longer term name resolution changes, so that will probably fix that issue eventually. Merging this change now, thanks for the fix!

PR #1170, opened 2020-03-11
Document CLI args for fsi commands

Mr0grog:
I was poking around at the fsi commands this morning while misusing Qri Desktop and noticed that the arguments felt a little under-documented. This just adds positional argument info to the usage string for `qri fsi link` and `qri fsi unlink` to make them slightly clearer.

I thought about adding a longer description with more info, but since these are low-level commands that don’t normally show up when you run `qri --help`, that didn’t really seem worthwhile.

### DIFF
--- a/cmd/fsi.go
+++ b/cmd/fsi.go
@@ -20,8 +20,8 @@ func NewFSICommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	}
 
 	link := &cobra.Command{
-		Use:   "link",
-		Short: "link a .qri-ref",
+		Use:   "link DATASET PATH",
+		Short: "link a dataset to a directory on disk",
 		Example: `link a dataset to the current working directory:
   $ qri fsi link peername/dataset .`,
 		Args: cobra.MinimumNArgs(2),
@@ -34,8 +34,8 @@ func NewFSICommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	}
 
 	unlink := &cobra.Command{
-		Use:   "unlink",
-		Short: "unlink a .qri-ref",
+		Use:   "unlink [DATASET]",
+		Short: "unlink a dataset from a directory on disk",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := o.Complete(f, args); err != nil {
 				return err
--- a/cmd/fsi.go
+++ b/cmd/fsi.go
@@ -24,7 +24,7 @@ func NewFSICommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 		Short: "link a dataset to a directory on disk",
 		Example: `link a dataset to the current working directory:
   $ qri fsi link peername/dataset .`,
-		Args: cobra.MinimumNArgs(2),
+		Args: cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := o.Complete(f, args); err != nil {
 				return err
@@ -36,6 +36,7 @@ func NewFSICommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	unlink := &cobra.Command{
 		Use:   "unlink [DATASET]",
 		Short: "unlink a dataset from a directory on disk",
+		Args:  cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := o.Complete(f, args); err != nil {
 				return err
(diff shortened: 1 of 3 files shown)


cmd/fsi.go
```diff
@@ -20,8 +20,8 @@ func NewFSICommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	}
 
 	link := &cobra.Command{
-		Use:   "link",
-		Short: "link a .qri-ref",
+		Use:   "link DATASET PATH",
+		Short: "link a dataset to a directory on disk",
 		Example: `link a dataset to the current working directory:
   $ qri fsi link peername/dataset .`,
 		Args: cobra.MinimumNArgs(2),
```
Mr0grog: In practice, it doesn’t seem like providing multiple refs makes sense here, so maybe this should be `ExactArgs(2)`. (Also, maybe it should set the `allowed` arg to `1` instead of `-1` when it calls `GetCurrentRefSelect()`?) Wasn’t sure, though, so didn’t change it.

cmd/fsi.go
```diff
@@ -34,8 +34,8 @@ func NewFSICommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	}
 
 	unlink := &cobra.Command{
-		Use:   "unlink",
-		Short: "unlink a .qri-ref",
+		Use:   "unlink [DATASET]",
+		Short: "unlink a dataset from a directory on disk",
```
Mr0grog: While trying to make sure I was documenting correctly, I realized this only takes one arg in practice (even though it parses a second arg, it never uses it). Wondering if we should be setting `Args: cobra.MaximumNArgs(1)` here.

cmd/fsi.go
```diff
@@ -20,8 +20,8 @@ func NewFSICommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	}
 
 	link := &cobra.Command{
-		Use:   "link",
-		Short: "link a .qri-ref",
+		Use:   "link DATASET PATH",
```
b5: Love it. Reviewing this also had me go lookup the fact that `Use` can be more than one-word long. TIL.

b5:
If you don't feel like making any of the above changes @Mr0grog I'd be happy to just rebase and apply the ones you've made. Thanks so much for filing!

Mr0grog:
> I'd love it if we could also make the ExactArgs and GetCurrentRefSelect adjustments you'd mentioned.

Can do! Would you prefer that as a separate PR, since it changes functionality?

> we'd need your commit message to match our commit guidelines

D’oh, totally forgot. Will amend as soon as I get a chance.

b5:
> Can do! Would you prefer that as a separate PR, since it changes functionality?

Nah. In an ideal world split it out into a separate commit
```
fix(cmd): require fsi link provide exactly two arguments
```

that'll pick up as a change just fine.

cmd/fsi.go
```diff
@@ -34,8 +34,9 @@ func NewFSICommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	}
 
 	unlink := &cobra.Command{
-		Use:   "unlink",
-		Short: "unlink a .qri-ref",
+		Use:   "unlink [DATASET]",
+		Short: "unlink a dataset from a directory on disk",
+		Args:  cobra.MaximumNArgs(1),
```
Mr0grog: Well now I have more questions. I used `MaximumNArgs` instead of `ExactArgs` here because I was thinking this would operate on the current directory or the `use` dataset if none was specified, but somehow totally missed that the first thing we do when parsing the arg values is to ensure there is at least one arg: https://github.com/qri-io/qri/blob/924157c253bf571ec9df92b0fd63e316e67bc6ad/cmd/fsi.go#L62-L65

…I left it as `MaximumNArgs` so that code can print a nicer message than `ExactArgs` would. BUT now I’m wondering if this is as expected (you must always specify a dataset name, in which case I need to fix the usage string) or if it *should* be falling back to the current directory or whatever was specified with `use` (in which case our parsing needs to change). Any thoughts here?

b5: hmmm, What do you think?

I'm of the opinion that the FSI commands should _all_ be sensitive to `pwd`. After all, that's kinda the point of FSI. It'd be nice to state "all FSI commands default to the current linked directory if one exists".

cc @dustmop, who is generally the maintainer of FSI related things these days

Mr0grog: If I were just going on what I expected/thought seemed reasonable, I’d say the first arg to `qri fsi unlink` should be optional and it should default to the working directory, like other qri commands. But then the same probably goes for `qri fsi link` — the second arg (the directory to link to) should be optional and defaults to the current working dir. But now this change has gotten pretty complex. 😜

b5: Good point. I vote we merge this & follow up with a PR that does a more comprehensive audit of default args 😄.

PR #1171, opened 2020-03-12
docs(developer.md): add note about ulimit workaround

b5:
came up a few times in discussion on discord, might as well have a linkable header we can point to

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -3,6 +3,7 @@
 * [Development Setup](#setup)
 * [Coding Rules](#rules)
 * [Commit Message Guidelines](#commits)
+* [Troubleshooting](#troubleshooting)
 
 ## <a name="setup"></a> Development Setup
 
@@ -18,8 +19,6 @@ machine:
   Installing Git][git-setup] is a good source of information.
 
 * [The Go Programming Language](https://golang.org): see golang.org to get started
-
-* [gx](https://github.com/whyrusleeping/gx/): gx is a distributed package management tool needed to build IPFS.
 
 * [golint](https://github.com/golang/lint): Golint is a linter for Go source code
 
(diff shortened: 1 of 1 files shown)


DEVELOPERS.md
```diff
@@ -166,4 +165,18 @@ A detailed explanation can be found in this [document][commit-message-format].
 [yarn-install]: https://yarnpkg.com/en/docs/install
 
 
+## <a name="troubleshooting"> Troubleshooting
+
+#### "Too Many Open Files" on OS X
```
Mr0grog: Pedantic fix (not that anybody is confused by this, though):

```suggestion
#### "Too Many Open Files" on MacOS
```

DEVELOPERS.md
```diff
@@ -166,4 +165,18 @@ A detailed explanation can be found in this [document][commit-message-format].
...
+
+If while running the test suite you get failing tests with "too many open files", you may need to adjust the number of files your shell process is willing to open. This is particularly common with the `p2p` and `cmd` pacakges. Both rely heavily on interacting temporary directories on the file system to run tests.
```
Mr0grog:
```suggestion
If, while running the test suite, you get failing tests with "too many open files", you may need to adjust the number of files your shell process is willing to open. This is particularly common with the `p2p` and `cmd` pacakges. Both rely heavily on interacting temporary directories on the file system to run tests.
```

DEVELOPERS.md
```diff
@@ -166,4 +165,18 @@ A detailed explanation can be found in this [document][commit-message-format].
...
+
+Often this is caused by having too low a `ulimit` value for the shell. Try the following command:
```
Mr0grog:
```suggestion
Often this is caused by having too low a file limit for your shell. You can use the [`ulimit` command](https://ss64.com/osx/ulimit.html) to check or change the limits. Try the following command to set the limit to 1000 open files:
```

DEVELOPERS.md
```diff
@@ -166,4 +165,18 @@ A detailed explanation can be found in this [document][commit-message-format].
...
+
+```
+ulimit -S -n 1000
+```
+
+We recommend a ulimit value of at least 1000, feel free to go higher if you need. Some other software require values as high as 10000 to avoid max open file issue.
```
Mr0grog:
```suggestion
We recommend a ulimit value of at least 1000; feel free to go higher if you need. Some other software require values as high as 10000 to avoid the max open file issue.
```

b5:
Thanks so much for the detailed review @Mr0grog. I've incorporated all of your fixes, learned to refer to it as `MacOS`, and rebased to incorporate fixes from #1174.

PR #1173, opened 2020-03-12
docs(cmd): DATASET is required in `fsi unlink`

Mr0grog:
This is a follow-on to #1170. We merged that one a bit too hastily and the usage string implied that DATASET was optional, when it actually isn't.

The last bit of conversation (https://github.com/qri-io/qri/pull/1170#discussion_r391354964) implied we might want to change the functionality to make this optional, though, so feel free to throw out this PR if you just want to move straight ahead with that instead.


Mr0grog:
👍 Sounds good.

PR #1175, opened 2020-03-12
docs(cmd): improve CLI docs for `dag` commands

Mr0grog:
Make the usage docs and argument parsing a little clearer for the `qri dag` family of commands:
- Only hide the top level command, so `qri dag --help` actually lists the subcommands.
- List optional and required positional arguments in usage string.
- Mark required flags as required.
- Only parse a "label" style argument when it's actually applicable.

I'm not deeply familiar with any of these commands, so I didn't actually try to spruce up the description strings. I mainly just followed the logic to understand what arguments were required (and what those arguments should generally represent).

### DIFF
--- a/cmd/dag.go
+++ b/cmd/dag.go
@@ -28,16 +28,16 @@ func NewDAGCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	}
 
 	manifest := &cobra.Command{
-		Use:    "manifest",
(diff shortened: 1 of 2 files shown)


Mr0grog:
Whoops, that was a really bad typo in the commit message. Fixed.

b5:
Bahahahah BAD Typo!? Gotta ❤️ Dat!

PR #1177, opened 2020-03-12
feat(cmd): remove `--blank` option from `export`

Mr0grog:
The `qri export --blank` command is now redundant — the more modern path to accomplish a similar need is with `qri init`. (Per this discussion on Discord with @b5 and @dustmop: https://discordapp.com/channels/497052800778502152/571427701475377152/687690467394977822) This removes it.

### DIFF
--- a/cmd/export.go
+++ b/cmd/export.go
@@ -2,8 +2,6 @@ package cmd
 
 import (
 	"fmt"
-	"io/ioutil"
-	"os"
 
 	"github.com/qri-io/ioes"
 	"github.com/qri-io/qri/lib"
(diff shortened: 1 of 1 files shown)


dustmop:
Just for clarification, `qri init` doesn't quite do the same thing that `qri export --blank` used to do. The main purpose of `qri init` is to initialize a new "working directory" with a hidden ".qri-ref" file that links that directory to a dataset in the user's repository. Changes to those files act as a staging area for new commits, and can be added to the linked database by using a bare `qri save` command with no explicit reference.

Perhaps generating scaffolding files can also be folded into its behavior, but that's a pretty orthogonal use case, so if we do it we need to explain it in a way that makes sense.

Mr0grog:
👍 I think it might be reasonable for `qri init` to scaffold, and it wouldn’t be too confusing (but not sure whether the default should be to do it or not).

FWIW, the way I think of the use case here (and why I was thinking of them as redundant) was more “get me started making a new dataset” than “set up the right files and content.”

dustmop:
Yes, I think that framing makes sense.

Anyway, I forgot I have to merge this. Doing that now!

PR #1178, opened 2020-03-12
test(body): update expected message for missing host in url

Arqu:
Small fix to the body fetch tests as the responses are not consistent with the expected value.

### DIFF
--- a/base/body_test.go
+++ b/base/body_test.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
 	"github.com/qri-io/dataset"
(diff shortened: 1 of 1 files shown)


base/body_test.go
```diff
@@ -59,7 +59,7 @@ func TestDatasetBodyFile(t *testing.T) {
 		{&dataset.Dataset{Structure: &dataset.Structure{Format: "csv"}, BodyBytes: []byte("a,b,c\n1,2,3")}, "body.csv", 11, ""},
 
 		// urlz
-		{&dataset.Dataset{BodyPath: "http://"}, "", 0, "fetching body url: Get \"http:\": http: no Host in request URL"},
+		{&dataset.Dataset{BodyPath: "http://"}, "", 0, "fetching body url: Get http:: http: no Host in request URL"},
```
b5: After a slack chat, we've come to realize this change is actually a question of what versions of go we'd like to support. If it's 1.14 this change needs to be reverted, if it's 1.13, we need to generalize this test to accommodate both versions.

Our docs say 1.13 is minimum, we voted to generalize.

Arqu:
As discussed, generalized the case to look for a match based on the core substring. Didn't want to go overboard by supporting multiple messages and juggling those across versions.

Issue #1159, opened 2020-03-06
TestUpdateMethods is flaky, fails on CI


dustmop:
CI parallelization causes this test to fail since qri connect takes up an operating system resource that causes a global conflict.

```
=== RUN   TestUpdateMethods
    TestUpdateMethods: update_test.go:47: apologies, we currently can't schedule updates while `qri connect` is running
--- FAIL: TestUpdateMethods (0.00s)
```

```
=== RUN   TestUpdateMethods
    TestUpdateMethods: update_test.go:65: it looks like a qri update service is already running on port 2503, please close before running tests
--- SKIP: TestUpdateMethods (0.00s)
```

PR #1182, opened 2020-03-12
docs(cmd): add argument names to CLI help text

Mr0grog:
Add positional argument information to all commands and clean up the descriptions and examples a bit.

~This is only half the commands and *very* much needs a second pass (lots of things I learned or decided partway through), but posting it here for safekeeping and if anyone wants to review early.~

The descriptions and examples now follow some basic styles for consistency:

  - No leading/trailing blank lines. (This is pretty arbitrary; I was mainly trying to make things consistent. We could make other choices about this.)

  - Short descriptions are phrases, not complete sentences — they start with a lower-case letter and do not end with a period. (Also arbitrary; was trying to go with what seemed most common.)

  - Long descriptions are full sentences and paragraphs. They start with a capital unless the first word is the name of a command or special word that should never be upper-case.

  - Examples:
    - Indented with two spaces.
    - Example descriptions start with `#`.
    - Lines that should be typed start with `$`.

I also fixed a few typos and added more detailed descriptions where the obviously didn't fully address what the command does, but otherwise tried to leave descriptive information alone here.

### DIFF
--- a/cmd/checkout.go
+++ b/cmd/checkout.go
@@ -16,13 +16,15 @@ import (
 func NewCheckoutCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	o := &CheckoutOptions{IOStreams: ioStreams}
 	cmd := &cobra.Command{
-		Use:     "checkout",
-		Short:   "checkout creates a linked directory and writes dataset files to that directory",
-		Long:    ``,
-		Example: ``,
+		Use:   "checkout DATASET",
+		Short: "checkout creates a linked directory and writes dataset files to that directory",
+		Long:  ``,
+		Example: `  # Place a copy of me/annual_pop in the ./annual_pop directory:
+  $ qri checkout me/annual_pop`,
 		Annotations: map[string]string{
 			"group": "dataset",
 		},
+		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := o.Complete(f, args); err != nil {
 				return err
--- a/cmd/connect.go
+++ b/cmd/connect.go
@@ -14,12 +14,11 @@ func NewConnectCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	o := ConnectOptions{IOStreams: ioStreams}
 	cmd := &cobra.Command{
 		Use:   "connect",
-		Short: "Connect to the distributed web by spinning up a Qri node",
+		Short: "connect to the distributed web by spinning up a Qri node",
 		Annotations: map[string]string{
 			"group": "network",
 		},
-		Long: `
-While it’s not totally accurate, connect is like starting a server. Running 
+		Long: `While it’s not totally accurate, connect is like starting a server. Running 
 connect will start a process and stay there until you exit the process 
 (ctrl+c from the terminal, or killing the process using tools like activity 
 monitor on the mac, or the aptly-named “kill” command). Connect does three main 
@@ -30,6 +29,7 @@ things:
 
 When you run connect you are connecting to the distributed web, interacting with
 peers & swapping data.`,
+		Args: cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := o.Complete(f, args); err != nil {
 				return err
--- a/cmd/diff.go
+++ b/cmd/diff.go
@@ -13,9 +13,9 @@ import (
 func NewDiffCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	o := DiffOptions{IOStreams: ioStreams}
 	cmd := &cobra.Command{
-		Use:   "diff",
-		Short: "Compare differences between two data sources",
-		Long: `diff is a new & experimental feature, please report bugs here:
+		Use:   "diff ([COMPONENT] [DATASET [DATASET]])|(PATH PATH)",
+		Short: "compare differences between two data sources",
+		Long: `'qri diff' is a new & experimental feature, please report bugs here:
 https://github.com/qri-io/deepdiff
 
 Diff compares two data sources & generates a description of the difference
@@ -29,22 +29,22 @@ qri diff works on structured data. qri diffs are measured in elements
 elements), delete (removed elements), or update (changed values).
 
 Each change has a path that locates it within the document`,
-		Example: `  diff between a latest version & the next one back:
+		Example: `  # Diff between a latest version & the next one back:
   $ qri diff me/annual_pop
 
-  diff current "qri use" selection:
+  # Diff current "qri use" selection:
   $ qri diff
 
-  diff dataset body against it's last version
+  # Diff dataset body against its last version:
   $ qri diff body me/annual_pop
 
-  diff two dataset meta sections:
+  # Diff two dataset meta components:
   $ qri diff meta me/population_2016 me/population_2017
 
-  diff two local json files:
+  # Diff two local json files:
   $ qri diff a.json b.json
 
-  diff a json & csv file
+  # Diff a json & csv file:
   $ qri diff some_table.csv b.json`,
 		Annotations: map[string]string{
 			"group": "dataset",
--- a/cmd/fetch.go
+++ b/cmd/fetch.go
@@ -13,13 +13,14 @@ import (
 func NewFetchCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	o := &FetchOptions{IOStreams: ioStreams}
 	cmd := &cobra.Command{
-		Use:     "fetch",
-		Short:   "fetch logbook info for a dataset reference",
+		Use:     "fetch DATASET [DATASET...]",
+		Short:   "show log of remote dataset commits",
 		Long:    ``,
 		Example: ``,
 		Annotations: map[string]string{
 			"group": "network",
 		},
+		Args: cobra.MinimumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := o.Complete(f, args); err != nil {
 				return err
--- a/cmd/get.go
+++ b/cmd/get.go
@@ -16,32 +16,27 @@ import (
 func NewGetCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	o := &GetOptions{IOStreams: ioStreams}
 	cmd := &cobra.Command{
-		Use:   "get",
-		Short: "Get elements of qri datasets",
-		Long: `Get the qri dataset (except for the body). You can also get portions of 
+		Use:   "get [COMPONENT] [DATASET]",
+		Short: "get components of qri datasets",
+		Long: `Get the qri dataset (except for the body). You can also get components of 
 the dataset: meta, structure, viz, transform, and commit. To narrow down
 further to specific fields in each section, use dot notation. The get 
 command prints to the console in yaml format, by default.
 
-You can get pertinent information on multiple datasets at the same time
-by supplying more than one dataset reference.
-
 Check out https://qri.io/docs/reference/dataset/ to learn about each section of the 
 dataset and its fields.`,
-		Example: `  # print the entire dataset to the console
-  qri get me/annual_pop
+		Example: `  # Print the entire dataset to the console:
+  $ qri get me/annual_pop
 
-  # print the meta to the console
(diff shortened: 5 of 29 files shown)


cmd/remove.go
```diff
@@ -72,7 +73,7 @@ func (o *RemoveOptions) Complete(f Factory, args []string) (err error) {
 		return err
 	}
 	if o.Refs, err = GetCurrentRefSelect(f, args, -1, nil); err != nil {
-		return err
+		return lib.NewError(lib.ErrBadArgs, "please specify a dataset path or name you would like to remove from your qri node")
```
Mr0grog: Near as I could tell, `o.Validate` was never getting called, so I just moved its logic here. Not sure if that’s kosher, though. In another spot with a similar setup, I called `o.Validate()` after `o.Complete()`.

Mr0grog:
*Almost* done with a first pass. Still have `update.go`, `use.go`, `validate.go`, and `version.go` to do. Then a second pass for double checking and cleanup of wording and language.

cmd/status.go
```diff
@@ -15,13 +15,28 @@ import (
...
+		Use:   "status [DATASET]",
+		Short: "Show status of working directory",
+		Long: `List what components and files of the working directory's dataset have been
+added, removed, or changed.
+
+If you specify a DATASET, this will list what components were changed in a
+particular commit (or the latest commit if none is specified). You can specify
+a commit alongside a dataset like:
+
+    me/dataset_name@/ipfs/Qmu...`,
```
Mr0grog: I had no idea you could do this at all. 😯

Mr0grog: (FWIW, seems kinda odd for the `status` command, but a useful feature to have overall.)

dustmop: FYI, I'm going to be trying to remove the ability to use an explicit ref with path for the `status` command. We use this functionality in Desktop in order to see what changed in previous commits. I think it makes more sense as a separate command `showcommit`, so want to extract it out. These changes are fine for now, just answering your curiosity.

Mr0grog: 👍 Was thinking it fit better in log or something similar. Makes sense.

cmd/fetch.go
```diff
@@ -13,13 +13,14 @@ import (
 func NewFetchCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	o := &FetchOptions{IOStreams: ioStreams}
 	cmd := &cobra.Command{
-		Use:     "fetch",
-		Short:   "fetch logbook info for a dataset reference",
+		Use:     "fetch DATASET [DATASET...]",
+		Short:   "fetch logbook info for datasets",
```
Mr0grog: I think this is actually fetching the logs, not the logbook. Is that right and I should change the description?

(Along those lines, is there a meaningful distinction between this and `qri log` other than this operates on remote information and that [only?] operates on the local repo?)

cmd/get.go
```diff
@@ -16,9 +16,9 @@ import (
 func NewGetCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	o := &GetOptions{IOStreams: ioStreams}
 	cmd := &cobra.Command{
-		Use:   "get",
-		Short: "Get elements of qri datasets",
-		Long: `Get the qri dataset (except for the body). You can also get portions of 
+		Use:   "get [COMPONENT] [DATASET...]",
```
Mr0grog: Multiple datasets actually doesn’t work here. There’s a TODO comment about restoring that functionality, but it’s been there for 8 months, so unclear what the status is. Left the documentation describing the currently-broken behavior; not sure what the right thing here is. ¯\_(ツ)_/¯

cmd/remove.go
```diff
@@ -30,15 +29,20 @@ adjust this cap using IPFS, qri will respect it.
...
 		},
+		// Use *max* so we can print a nicer message for no or malformed args
+		Args: cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := o.Complete(f, args); err != nil {
 				return err
 			}
+			if err := o.Validate(); err != nil {
+				return err
+			}
```
Mr0grog: `Validate()` previously wasn’t getting called at all (though it was tested), so I added it.

cmd/remove.go
```diff
@@ -72,7 +76,11 @@ func (o *RemoveOptions) Complete(f Factory, args []string) (err error) {
 		return err
 	}
 	if o.Refs, err = GetCurrentRefSelect(f, args, -1, nil); err != nil {
-		return err
+		// This error will be handled during validation
+		if err != repo.ErrEmptyRef {
+			return err
+		}
+		err = nil
```
Mr0grog: I initially removed the `Validate()` function and put its error here (it previously wasn’t getting called at all), but there is a test for it. I was trying not to make huge code changes here, so I put this together in a way that satisfies the existing tests. That said, I feel like keeping `Validate()` separate makes for more awkward code here: there’s a lot more checking in this spot, and if there is no repo ref (or it’s invalid), we will log that error *last* rather than first (because there are other validation-esque errors that come from `Complete()`).

Not sure whether the right solution is for other errors in `Complete()` to move to `Validate()`, to remove `Validate()` like I originally did, or something else.

cmd/rename.go
```diff
@@ -31,6 +30,9 @@ renames to a minimum.`,
 			if err := o.Complete(f, args); err != nil {
 				return err
 			}
+			if err := o.Validate(); err != nil {
+				return err
+			}
```
Mr0grog: Another instance where `Validate()` was never called before.

Mr0grog:
One other thought I had here is that it might be nice to use [lithammer/dedent](https://pkg.go.dev/github.com/lithammer/dedent?tab=doc) or [Parquery/gocontracts/dedent](https://pkg.go.dev/github.com/Parquery/gocontracts/dedent?tab=doc) to clean up and reformat descriptions and examples so they can be formatted a little nicer in code and be a little more forgiving in terms of formatting.

That could also be done by using Cobra’s `SetUsageFunc()` instead of `SetUsageTemplate()` to do some custom string cleanup before sending to the template for final output: https://github.com/qri-io/qri/blob/a1c3a9935d60f8b6f8cbe05850b569d04bac4a95/cmd/qri.go#L76-L78

Even if you think that’s a good idea, though, I think that’s a little bigger change than is necessarily relevant for this PR.

cmd/peers.go
```diff
@@ -17,9 +17,8 @@ func NewPeersCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	o := &PeersOptions{IOStreams: ioStreams}
 	cmd := &cobra.Command{
 		Use:   "peers",
-		Short: "Commands for working with peers",
-		Long: `
-The ` + "`peers`" + ` commands allow you to interact with other peers on the Qri network.
+		Short: "commands for working with peers",
+		Long: `The ` + "`peers`" + ` commands allow you to interact with other peers on the Qri network.
```
b5: While we're here. `allow you to` is passive voice (not your change, I know). I'd prefer `The ` + "`peers`" + ` commands interact with other peers on the Qri network.`

cmd/remove.go
b5: I like the answer that seems to be emerging from this PR: use cobra's built-ins more. 

This will be a bigger change, but I think the right answer is to call/adjust `Complete` methods into  [cobra pre-run hooks](https://github.com/spf13/cobra#prerun-and-postrun-hooks), and focus on defining all validation that can be done at the cmd level as `cobra.Command.Args` functions.

cmd/fetch.go
b5: Sorta correct. it's fetching logbook data, just not _persisting_ it.

 As we discussed on discord, We'd like to merge `fetch` into the `log` command. Given how long that might take, I think it's still worth updating the doc here

cmd/connect.go
```diff
@@ -14,12 +14,11 @@ func NewConnectCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
...
 	cmd := &cobra.Command{
 		Use:   "connect",
-		Short: "Connect to the distributed web by spinning up a Qri node",
+		Short: "connect to the distributed web by spinning up a Qri node",
 		Annotations: map[string]string{
 			"group": "network",
 		},
-		Long: `
-While it’s not totally accurate, connect is like starting a server. Running 
+		Long: `While it’s not totally accurate, connect is like starting a server. Running
```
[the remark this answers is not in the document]
Mr0grog: Heh, I definitely considered diving into this one, too. Another day. :D

cmd/checkout.go
```diff
@@ -16,10 +16,11 @@ import (
...
 	o := &CheckoutOptions{IOStreams: ioStreams}
 	cmd := &cobra.Command{
-		Use:     "checkout",
-		Short:   "checkout creates a linked directory and writes dataset files to that directory",
-		Long:    ``,
-		Example: ``,
+		Use:   "checkout DATASET",
+		Short: "checkout creates a linked directory and writes dataset files to that directory",
+		Long:  ``,
+		Example: `  # Place a copy of me/annual_pop in the ./annual_pop directory:
```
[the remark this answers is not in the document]
Mr0grog: FWIW, I tried to add `Args` everywhere that `Complete()` or `Run()` didn’t do the same work (and hopefully give a more specific and clear error message). I think I missed this one, though.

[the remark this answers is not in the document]
b5: s'all good. I think it'll make for a great next-pass

cmd/diff.go
```diff
@@ -13,9 +13,9 @@ import (
 func NewDiffCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	o := DiffOptions{IOStreams: ioStreams}
 	cmd := &cobra.Command{
-		Use:   "diff",
-		Short: "Compare differences between two data sources",
-		Long: `diff is a new & experimental feature, please report bugs here:
+		Use:   "diff ([SECTION] [DATASET [DATASET]])|(PATH PATH)",
```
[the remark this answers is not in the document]
Mr0grog: Yeah, I *think* I recall having tried that and it didn’t work, but I could be wrong. It’s also possible I may have accidentally been tripping over a version of #1185, where a path can get parsed as a component field.

On a separate note, reading the code for this totally threw me for a loop when I realized `GetCurrentRefSelect()` doesn’t actually validate that the values it’s picking up are dataset refs or even in the format of dataset refs.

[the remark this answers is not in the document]
Mr0grog: (In the mean time, I updated the language here to say “component” instead of “section.” I missed that in my last pass.)

Mr0grog:
I think I addressed everything you mentioned here, @b5. Give it another look and let me know what you think.

cmd/diff.go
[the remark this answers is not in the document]
Mr0grog: Ah, yeah, combination of DATASET and PATH is explicitly disallowed: https://github.com/qri-io/qri/blob/da62f794bb866bf82bb76fa80874a167aa99fc97/lib/diff.go#L96-L99

Mr0grog:
Hmmm, test failure looks like maybe something flaky? Doesn’t seem like it could be these changes:

```
=== RUN   TestUpdateMethods
    TestUpdateMethods: update_test.go:47: apologies, we currently can't schedule updates while `qri connect` is running
--- FAIL: TestUpdateMethods (0.00s)
```

dustmop:
https://github.com/qri-io/qri/issues/1159 known that TestUpdateMethods is flaky. I'll rerun the tests.

b5:
I didn't address your proposal around dedenting: https://github.com/qri-io/qri/pull/1182#issuecomment-599688362

Super interesting. I'm into it if we can address two concerns:
* we use these docs to generate markdown documentation. We'd just need to do the same thing there, and confirm it works
* zomg we have so many dependencies. Is it worth adding another?

I'll let others weight in 😄

b5:
Looks like tests are passing. @Mr0grog are you able to push the big green merge button? If so we'd be delighted to have you do the merge 😄

Mr0grog:
> are you able to push the big green merge button? If so we'd be delighted to have you do the merge 😄

Nope, I am but a mere community member, a proposer of provocations, a pettish pot-stirrer, a hobbyist of hypotheticals on this repo. No commit rights. ;)

dustmop:
Merged, thanks so much @Mr0grog!

Issue #1155, opened 2020-03-05
`qri init` sets executable bit on the files it creates


feep:
Wrong permissions?

```
❯ qri init
Name of new dataset [scratch]:   
Format of dataset, csv or json [csv]: 
initialized working directory for new dataset feep/scratch

❯ ls 
.rw-r--r-- 12 rusty 03-05 14:25 .qri-ref
.rwxr-xr-x 22 rusty 03-05 14:25 body.csv
.rwxr-xr-x 18 rusty 03-05 14:25 meta.json
.rwxr-xr-x 36 rusty 03-05 14:25 structure.json
```

Hackerpilot:
Looks like this is caused by the `os.ModePerm` constant being used when the files are created. I'm not really sure why this is, as I think this constant is supposed to be a bit-mask used to separate the permission bits from other parts of a file mode. The line that causes this specific issue seems to be https://github.com/qri-io/qri/blob/924157c253bf571ec9df92b0fd63e316e67bc6ad/base/component/kinds.go#L502

dustmop:
@Hackerpilot your diagnosis sounds correct, that should be a relatively easy fix. I think it should be 0644 instead of the current value. It would be great to have a test updated to check for this behavior, for example `TestInitStatusSave` in the file `cmd/fsi_integration_test.go` is listing the directory and checking the filenames; it could also check the permissions on those files.

mecm1993:
This verification was added into `TestInitStatusSave` to validate the permission of each file,

```golang
// Verify the permissions for each generated file.
files := filesDirectory(workDir)
mode := int(0644)
expectPermission := os.FileMode(mode)
for _, file := range files {
        if file.Mode() != expectPermission {
		t.Errorf("%s does not have the correct permission", file.Name())
	}
}
...
func filesDirectory(path string) []os.FileInfo {
	finfos, err := ioutil.ReadDir(path)
	if err != nil {
		return nil
	}
	return finfos
}
```

Moreover, the following modifications can be added to `kinds.go`, so the correct permission can be assigned for the file.

```golang
const Perm = os.FileMode(int(0644))
...
func (bc *BodyComponent) WriteTo(dirPath string) (targetFile string, err error) {
	if bc.Value == nil {
		err = bc.LoadAndFill(nil)
		if err != nil {
			return
		}
	}
	body := bc.Value
	if bc.Structure == nil {
		return "", fmt.Errorf("cannot write body without a structure")
	}
[...]

PR #1183, opened 2020-03-13
fix(base): permission for files generated on init

mecm1993:
This should fix issue #1155 and I also took into consideration the suggestions of @dustmop to add an additional validation on the tests regarding the files' permission.


dustmop:
This is great, thank you @mecm1993! Our continuous integration tests were failing because some of them are flaky, but they're passing now. I'm going to merge your change.

Issue #1120, opened 2020-02-12
Consider the data structures that are used throughout the codebase


dustmop:
A week or two ago, we had `dsref.Ref`, `repo.DatasetRef`, and `dataset.Dataset`. As part of the `dscache` implementation (see https://github.com/qri-io/qri/issues/1089), I started the work of switching over to `dsref.Ref` everywhere possible instead of `repo.DatasetRef`. This is part of the work of moving our dependencies lower, so that more of the codebase can use common, dumb, leaf-level structures wherever possible.

Many places in the code were using `repo.DatasetRef` to talk about the results of a "list" operation, where one or more datasets get retrieved and certain fields are observed, often to create a quick summary of a dataset's meta and structure as an overview. This tended to be the same set of data, so we introduced a `dsref.VersionInfo` which is basically a `dsref.Ref` with a few additional fields, which can replace most of these uses of `repo.DatasetRef`.

Other commands are also being converted over to this new paradigm, such as `log`, and the upcoming `feed` which uses a new "preview" feature. There's danger in stuffing too much into `dsref.VersionInfo`, moving it away from its original intention of being a small summary for listing purposes. Already, `log` added fields for `commitTitle` and `commitMessage` to `VersionInfo`, despite being the only operation that uses these two fields. The danger is that, by adding too much to `VersionInfo`, we lose track of what operations have dependencies on which fields. Instead, the definition of a struct, as a static type, should inform which places it is being used, to make data flow obvious and understandable.

The alternative we should follow is instead to introduce structs for operations that need them. For example, `log` should being using a `LogInfo` type that composes a `VersionInfo` with the additional commit fields that it needs. Dataset previews should similarly create a `PreviewInfo` type that adds "diff stats", and "body preview". By utilizing composition, we maintain clarity about what fields are needed where, without having to have excessive type conversion code throughout the codebase.

dustmop:
Tagging this as a good piece of work for @Arqu to take on

As part of the refactoring work to convert to dsref, I made commit https://github.com/qri-io/qri/commit/4ce3eaa3228aa6f159bb27bbc1dbda7c6ab2c906, which introduced VersionInfo (a dsref plus some additional fields needed for the `list` command, a very cacheable piece of data) and converted the `log` command to return a list of VersionInfo.

[...]

Issue #1144, opened 2020-03-02
restructure dependencies as dscache comes into being


dustmop:

b5:
> We should move fsi under dscache so that it is no longer exposed to lib. This will help simplify our story about handling datasets in repo vs fsi in the same fashion.

Disagree on this. I don't think file system integration is purely the domain of a cache. In #1133, we started to talk about _critical path packages_ organized into an import stack and _event driven packages_ that must be siblings with few/no critical path imports. [Full notes here](https://github.com/qri-io/qri/pull/1133#issuecomment-592135598). A `lib.Instance` should have an internal `fsi` field. 

I totally see the impetus to cut down the complexity of all these subsystems interacting, just not sure that we should solve that by stacking imports.

We might be arriving at a place where subsystems _need_ to use events to communicate between each other. The need for callbacks seems to indicate yes, but then we're stuck back at this "how do you know you're finished" problem.

dustmop:
Ok, I thought some more about this, and I have a recommendation. First, agreed that FSI is not in the domain of a cache, totally correct. However, I think the current relationship with lib and fsi is a problem. There's a lot of code duplication like this:

```
		ref, err := base.ToDatasetRef(p.Ref, r.repo, p.UseFSI)
		if err != nil {
			return err
		}
		if p.UseFSI {
			if ref.FSIPath == "" {
				return fsi.ErrNoLink
			}
			if ds, err = fsi.ReadDir(ref.FSIPath); err != nil {
				return fmt.Errorf("loading linked dataset: %s", err)
			}
		} else {
			ds, err = dsfs.LoadDataset(ctx, r.repo.Store(), ref.Path)
			if err != nil {
				return fmt.Errorf("loading dataset: %s", err)
			}
		}
```

this can be seen in `Get` in lib/datasets.go and `RenderReadme` in lib/render.go and `Validate` in lib/datasets.go. This pattern should be extracted and put into a single location that is responsible for all ref resolution.

Relatedly, while looking at the codepath in `base/save.go` specifically `PrepareDataset` I noticed it is calling CanonicalizeDataset in order to get the previousPath. This is bad for two reasons:
* calling CanonicalizeDataset this low in the stack is a problem, resolution should live higher
* we should switch to initIDs at the base level, instead of passing dsrefs (string or parsed)

Ideally, as we move to dscache, the `base` package wouldn't have a reference to the dscache at all, in order to match the dependency tree I made above.

Here's the proposal:
[...]

PR #1187, opened 2020-03-14
refactor(dsref): separating commit details from VersionInfo

Arqu:
Expanding on #1120. I've isolated most of the reverting changes and made them, however, feels like it's still touching too many things.

Obviously WIP, tests are yet to be touched.


base/log.go
```diff
@@ -11,10 +11,22 @@ import (
 	reporef "github.com/qri-io/qri/repo/ref"
 )
 
+type DatasetLogItem struct {
```
Arqu: I think it's no longer the right place for this struct to be here. Same as the DatasetRef, it's a bit too high up in the chain and forces some other parts of code.

base/log.go
```diff
@@ -11,10 +11,22 @@ import (
...
+	// Decription of a dataset reference
+	VersionInfo dsref.VersionInfo `json:"versionInfo,omitempty"`
+	// Timestamp field from the commit
+	CommitTime time.Time `json:"commitTime,omitempty"`
```
Arqu: The CommitTime is also extracted as it flows into the DataLogItem better than in VersionInfo due to it's dependency on the dataset.

base/log.go
```diff
@@ -31,24 +43,32 @@ func DatasetLog(ctx context.Context, r repo.Repo, ref reporef.DatasetRef, limit,
...
 
 	rlog, err := DatasetLogFromHistory(ctx, r, ref, offset, limit, loadDatasets)
 	if err != nil {
 		return nil, err
 	}
-	items := make([]dsref.VersionInfo, len(rlog))
+	items := make([]DatasetLogItem, len(rlog))
 	for i, vref := range rlog {
-		items[i] = reporef.ConvertToVersionInfo(&vref)
+		items[i].VersionInfo = reporef.ConvertToVersionInfo(&vref)
```
Arqu: I wanted to extract this peice, but since DataLogItem is too high up in base, couldn't drop the login into the repo/ref/converter.

dscache/convert.go
```diff
@@ -100,8 +100,8 @@ func buildDscacheFlatbuffer(userPairList []userProfilePair, entryInfoList []*ent
 		prettyName := builder.CreateString(ce.Name)
 		metaTitle := builder.CreateString(ce.MetaTitle)
 		themeList := builder.CreateString(ce.ThemeList)
-		commitTitle := builder.CreateString(ce.CommitTitle)
-		commitMessage := builder.CreateString(ce.CommitMessage)
+		// commitTitle := builder.CreateString(ce.CommitTitle)
```
Arqu: I don't feel comfortable with this right now as I have a feeling it goes down the wrong path and will change behavior we don't want to change. Haven't thoroughly explored this yet, but all the changes in dscache feel a bit off right now.

lib/datasets.go
```diff
@@ -170,17 +170,31 @@ func (r *DatasetRequests) List(p *ListParams, res *[]dsref.VersionInfo) error {
...
-		infos[i] = reporef.ConvertToVersionInfo(&r)
+		dataLogs[i] = ConvertToDatasetLogItem(&r)
 	}
-	*res = infos
+	*res = dataLogs
 
 	return err
 }
 
+// ConvertToVersionInfo converts an old style DatasetRef to the newly preferred DatasetLogItem
```
Arqu: This is here because I couldn't drop it in the repo/ref/converter due to dependency cycles. Ideally we should put the DataLogItem lower and not have it in base.

logbook/logbook.go
```diff
@@ -789,8 +789,6 @@ func infoFromOp(ref dsref.Ref, op oplog.Op) dsref.VersionInfo {
 		ProfileID:   ref.ProfileID,
 		Name:        ref.Name,
 		Path:        op.Ref,
```
Arqu: Also doesn't feel right, but didn't want to start re-doing this piece without a better understanding of it's impact.

base/log.go
dustmop: I think it's ok here. This struct is specific to the `log` operation, and nothing else should be using it except for higher up functions that also call `base/log`.

dscache/convert.go
dustmop: It's correct to remove `commitTitle` and `commitMessage` from this location. Only log is using those fields, and no other users from dscache care about them. But, `commitTime` should stay, that's used in other callers.

lib/datasets.go
dustmop: Ah, sorry for the confusion, `List` shouldn't be affected by this change. Keep on using `VersionInfo` here, not `DatasetLogItem`.

logbook/logbook.go
dustmop: Correct, no need to change logbook.

base/log.go
Arqu: Moving CommitTime back as even though it flows from the same source, it's more relevant and used through the VersionInfo

Arqu:
I've addressed the inputs with some updated code changes and took care of most of the tests. There's some more bits do be adapted + need to go over things and see what is not covered and get that in order as well. @dustmop think that this should be pretty close right now.

base/log.go
Arqu: Just for future reference, this is me being wrong with some assumptions. Refer to https://github.com/qri-io/qri/issues/1144 for a better overview of the components and it's interactions.

lib/datasets.go
```diff
@@ -759,11 +759,15 @@ func (r *DatasetRequests) Remove(p *RemoveParams, res *RemoveResponse) error {
 	}
 
 	// Get the revisions that will be deleted.
-	history, err := base.DatasetLog(ctx, r.node.Repo, ref, p.Revision.Gen+1, 0, false)
-	if err == nil && p.Revision.Gen >= len(history) {
+	historyDataLog, err := base.DatasetLog(ctx, r.node.Repo, ref, p.Revision.Gen+1, 0, false)
```
dustmop: I think a better (simpler) approach is change `history` to be a `[]base.DatasetLogItem`, then change `base.RemoveEntireDataset` so that it's third parameter is also a `[]base.DatasetLogItem`.

lib/log_test.go
```diff
@@ -27,23 +27,28 @@ func TestHistoryRequestsLog(t *testing.T) {
...
+	items := make([]DatasetLogItem, len(refs))
 	for i, r := range refs {
-		items[i] = reporef.ConvertToVersionInfo(&r)
-		items[i].MetaTitle = ""
-		items[i].BodyRows = 0
-		items[i].NumErrors = 0
-		items[i].BodyFormat = ""
+		ds := r.Dataset
+		items[i].VersionInfo = reporef.ConvertToVersionInfo(&r)
+		items[i].VersionInfo.MetaTitle = ""
```
dustmop: Can get rid of these `VersionInfo` accesses here (after the first one which assigns the whole struct). Just do `items[i].MetaTitle = ""` etc.

api/log.go
```diff
@@ -35,6 +35,8 @@ func (h *LogHandlers) LogHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+type DatasetLogItem = base.DatasetLogItem
```
dustmop: This needs a comment describing it, which begins `// DatasetLogItem ...` in order to pass our `golint` step. For example `// DatasetLogItem aliases the type from base`.

base/log.go
```diff
@@ -11,10 +11,20 @@ import (
...
+
 // DatasetLog fetches the change version history of a dataset
-func DatasetLog(ctx context.Context, r repo.Repo, ref reporef.DatasetRef, limit, offset int, loadDatasets bool) ([]dsref.VersionInfo, error) {
+func DatasetLog(ctx context.Context, r repo.Repo, ref reporef.DatasetRef, limit, offset int, loadDatasets bool) ([]DatasetLogItem, error) {
 	if book := r.Logbook(); book != nil {
 		if versions, err := book.Versions(ctx, reporef.ConvertToDsref(ref), offset, limit); err == nil {
+			items := make([]DatasetLogItem, len(versions))
 			// logs are ok with history not existing. This keeps FSI interaction behaviour consistent
 			// TODO (b5) - we should consider having "empty history" be an ok state, instead of marking as an error
 			if len(versions) == 0 {
```
dustmop: Above the comment a few lines down (starting with "Logbook doesn't store...") to mention the other fields being assigned in this loop.

dscache/fill_info.go
```diff
@@ -35,11 +35,6 @@ func fillInfoForDatasets(ctx context.Context, store cafs.Filestore, filesys qfs.
 			info.BodyFormat = ds.Structure.Format
 			info.NumErrors = ds.Structure.ErrCount
 		}
-		if ds.Commit != nil {
-			info.CommitTime = ds.Commit.Timestamp
```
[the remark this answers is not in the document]
Arqu: Good catch, missed it in my reverts.

PR #1191, opened 2020-03-16
feat(sql): run SQL queries using qri datasets as tables

b5:
This PR adds a new command `qri sql` that executes Structured Query Language queries using qri datasets as sources. Build from this branch and, try this:

```
qri sql "SELECT * FROM me/dataset_refence"
```

### This PR:
* imports [OctoSQL](https://github.com/cube2222/octosql) as a dependency
* Implements a qri OctoSQL [data source](https://github.com/cube2222/octosql/tree/master/storage)
* adds an SQL package, sql lib & cmd methods
* Adds a SQL preprocessor that maps dataset references to legal SQL table names

## Nice To Haves
* Auto Table Aliasing
* Support `DESCRIBE` statements
* Auto-clone when a query runs
* Warn users about query size / downloads if they exceed a predefined threshold

### Error Handling
There are some common gotchas that need clear error messages when they're encountered:

* [ ] invalid dataset reference strings need a clear error message
* [ ] Running a query against a dataset you don't have locally will fail
* [ ] we won't support JSON documents.
* [ ] we don't support a number of statements in the SQL grammar. Need to error with "not supported"
* [ ] nested FROM statements aren't supported yet: parser can't support parenthesis
* [ ] illegal column names.

### Required Tests
* [ ] test a hash-based dataset reference: `@/ipfs/Qm....`
* [ ] test that


sql/preprocess.go
```diff
@@ -0,0 +1,312 @@
...
+	asTok
+	fromTok
+	groupTok
+	havingTok
+	unionTok
+	whereTok
+	keywordEnd // keywordEnd marks the end of keyword tokens in the token enumeration
+)
+
+func (t tokenType) String() string {
```
Arqu: Over time will probably need more like join, on etc

sql/qds/datasource.go
```diff
@@ -0,0 +1,221 @@
...
+				ds:        ds,
+			}, nil
+		},
+		nil,
+		availableFilters,
+		metadata.BoundedFitsInLocalStorage,
+	)
+}
+
+// // NewDataSourceBuilderFactoryFromConfig creates a data source builder factory using the configuration.
```
Arqu: Should probably remove this if it's unused.

sql/qds/datasource.go
```diff
@@ -0,0 +1,221 @@
...
+
+	if err != nil {
+		return nil, errors.Wrap(err, "couldn't read record")
+	}
+
+	aliasedRecord := make(map[octosql.VariableName]octosql.Value)
+	if rec, ok := ent.Value.([]interface{}); ok {
+		for i, x := range rec {
+			switch v := x.(type) {
+			case string:
```
Arqu: Should we also have an explicit case on date or simply rely on strings?

lib/sql.go
```diff
@@ -0,0 +1,52 @@
...
+// ExecParams provides parameters to the execute command
+type SQLExecParams struct {
+	Query        string
+	OutputFormat string
+}
+
+// Exec runs an SQL query
+func (m *SQLMethods) Exec(p *SQLExecParams, results *[]byte) error {
+	if m.inst.rpc != nil {
+		return m.inst.rpc.Call("SQLMethods.Exec", p, results)
```
Arqu: We should wrap this with better error handling for more user friendliness.

Also error handling in general - it's probably going to reach the user very often as we won't really support "all of SQL" out of the box and they will constantly be hitting edge cases.

cmd/sql.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"bytes"
+
+	"github.com/qri-io/ioes"
+	"github.com/qri-io/qri/lib"
+	"github.com/spf13/cobra"
+)
+
+// NewSQLCommand creates a new `qri search` command that searches for datasets
```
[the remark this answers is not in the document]
b5: apologies, this is an old comment from copy-pasting `cmd/search.go`, will update!

base/component/kinds.go
```diff
@@ -535,6 +535,20 @@ func OpenEntryReader(file *os.File, format string) (dsio.EntryReader, error) {
 // TODO(dlong): Move this to dataset.dsio
 func SerializeBody(source interface{}, st *dataset.Structure) ([]byte, error) {
 	buff := bytes.Buffer{}
+
+	// ensure tabular data formats have a schema
+	if st.RequiresTabularSchema() && st.Schema == nil {
```
Arqu: I like how the other PR nicely feeds into this :)

sql/preprocess/preprocess.go
```diff
@@ -0,0 +1,412 @@
...
+	groupTok
+	havingTok
+	limitTok
+	orderTok
+	unionTok
+	whereTok
+	keywordEnd // keywordEnd marks the end of keyword tokens in the token enumeration
+)
+
+func (t tokenType) String() string {
```
Arqu: I think it would be a good idea to have a complete list of "reserved words" here and then just noop or throw on them when parsing, otherwise things like "join" could get caught as a table name instead of an operator. And might be better for us to throw here an "unsuported operation error" than having it explode on the user from deeper in the stack.

sql/preprocess/preprocess.go
```diff
@@ -0,0 +1,412 @@
...
+			default:
+				p.unscan(tok)
+			}
+			return nil
+		}
+	}
+}
+
+func toLegalName(refStr string) string {
+	refStr = strings.Replace(refStr, "@", "_at_", 1)
```
Arqu: Think we still don't handle spaces and numeric names well here.

Arqu: On the other hand we can have the user ensure a well formated query instead of forcing a convention on to them or automagically swapping out too many things.

b5:
Ok, I'm going to merge this, and will follow up with cleanup PRs. Here goes nothing sportfans.

PR #1197, opened 2020-03-16
test(fsi): Written fsi components need to respect umask setting

dustmop:


dustmop:
Test at head is currently broken, because it depends upon the umask of the user. The test should be considering this in order to pass everywhere.

dustmop:
No problem, I think it was my fault

Issue #1057, opened 2019-12-13
should api calls have an "fsi" parameter?


dustmop:
Currently, desktop has to be aware of whether a dataset reference is checked out to a working folder, or whether it is unlinked, in order to make api calls with `fsi` either true or false. It knows if the reference is checked out based upon the results of an earlier `list` operation. The semantics behind the `fsi` parameter are whether the backend should be merging the working directory into the dataset as a "staging area" for an upcoming commit, and it is a bug to use the wrong fsi value (such as fsi=true for a dataset that is not checked out).

An alternative system would be to move all of this logic to the backend. Desktop would just ask for a dataset's body or head etc, and the backend would know if it should merge the working directory based upon whether there's an FSIPath value for that reference. This simplifies desktop, leaves it with one less thing to keep track of, and eliminates a large surface area for potential bugs.

The only downside is that there are some places where the desktop is actually asking for different things when using the `fsi` parameter. For example, `localhost:2503/status/me/my_ds?fsi=true` is asking for the cleanliness of the working directory, while `localhost:2503/status/me/my_ds?fsi=false` is asking about whether changes were made in the stored commit of the head version of that dataset. If we got rid of the `fsi` parameter, we would have to replace this latter call with something like `localhost:2503/history` or `localhost:2503/status_at_version`, but at least in these cases it's clear that desktop is asking for something specific.

dustmop:
Fixed by https://github.com/qri-io/qri/pull/1228.

PR #1198, opened 2020-03-17
feat(showcommit): Split cmd status-at-version into showcommit.

dustmop:
As part of removing the fsi=true, we need to split status into two commands. This is the only command that has semantic different when the parameter is set to true, so it should be split off.


cmd/showcommit.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"fmt"
+
+	"github.com/qri-io/ioes"
+	"github.com/qri-io/qri/lib"
+	"github.com/spf13/cobra"
+)
+
+// NewShowCommitCommand creates a new `qri search` command that searches for datasets
```
b5: :) you copypasta the search command too, eh?

cmd/showcommit.go
```diff
@@ -0,0 +1,79 @@
...
+func NewShowCommitCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
+	o := &ShowCommitOptions{IOStreams: ioStreams}
+	cmd := &cobra.Command{
+		Use:   "showcommit",
+		Short: "Show information about a commit",
```
b5: Ditto from #1182: 

> Short descriptions are phrases, not complete sentences — they start with a lower-case letter and do not end with a period

cmd/showcommit.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/qri-io/ioes"
+	"github.com/qri-io/qri/lib"
+	"github.com/spf13/cobra"
+)
+
+// NewShowCommitCommand creates a new `qri search` command that searches for datasets
+func NewShowCommitCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
+	o := &ShowCommitOptions{IOStreams: ioStreams}
+	cmd := &cobra.Command{
+		Use:   "showcommit",
```
b5: Documentation of this this command should continue the work @Mr0grog landed in #1182. This should be `showcommit DATASET`

cmd/showcommit.go
```diff
@@ -0,0 +1,79 @@
...
+		Short: "Show information about a commit",
+		Long: `
+The command showcommit will show information about a commit, specifically what
+components were added or removed, at a specific version. This is analagous to
+the status command, except only available for dataset versions in history.
+
+  # show information about the head commit
+  qri showcommit me/dataset_name`,
```
b5: plz move this into `Example:`, again following guidelines. Examples should be
* Indented with two spaces.
* Example descriptions start with #.
* Lines that should be typed start with $.

Mr0grog:
> I'd like to explore alternatives to showcommit. This command sorta matches the behavior of git whatchanged How would you feel about whatchanged?

Me, too! But I’d pull for one of:

- `qri get commit me/dataset@/ipfs/version` (or `qri get /ipfs/version` in a linked dir). If thinking in Git analogues, this starts to make `qri get` feel a lot like `git show` (not sure if good or bad).

- `qri log --verbose` or `qri log /ipfs/version --verbose` (or something other than `--verbose`, dunno 🤷‍♂). I’m thinking of how this is thematically similar to the log, and how `git log` has a bunch of options and arguments that let you look at specific commits or commit ranges.

FWIW, the manpage for `git whatchanged` says:

> New users are encouraged to use git-log(1) instead. The whatchanged command is essentially the same as git-log(1) but defaults to show the raw format diff output and to skip merges.
>
> The command is kept primarily for historical reasons; fingers of many people who learned Git long before git log was invented by reading Linux kernel mailing list are trained to type it.

…so maybe that’s another argument in favor of `qri log --something-or-other`.

dustmop:
> While it's totally a plumbing command that we're not going to see much, I'd like to explore alternatives to `showcommit`. This command sorta matches the behavior of `git whatchanged` How would you feel about `whatchanged`?

I'm totally good with `whatchanged`, I even like it more. Agreed, we should expect this command will really only be used by the api, it exists on the command-line just for parity.

cmd/showcommit.go
dustmop: Bleh, I thought I fixed up all of the old names, guess I missed this.

b5:
I'd like to avoid `qri get commit me/dataset@/ipfs/version`. I like the rule that `get` _only returns dataset data derivable from a single version_.

I'm a fan of `qri log --verbose`, but log is getting a lot of attention these days. For now, as @dustmop mentioned, this is just for parity. I'd be a fan of keeping `whatchanged`, with the option to later migrate into `qri log --verbose`. I'd love it if `qri log --verbose` included cached diffing info as well as high level component status changes.

Regarding the `whatchanged` page, I found it helpful to actually look at the _output_ of the command, along with the `man` page. Here's a snippet:

```git
commit b7717f82e846393bc745d216097e37d73aa5d045 (HEAD -> password-no-echo, origin/password-no-echo)
Author: Dustin Long <me@dustmop.io>
Date:   Tue Mar 17 14:21:07 2020 -0400

    test(registry): Fix cmd test that uses registry signup

:100644 100644 b787aef7 72d7d6f5 M      cmd/cmd_test.go
:100644 100644 319b19e3 dab03c5f M      cmd/registry.go
:100644 100644 43db508a cf8ebc3e M      cmd/test_runner_test.go

commit 3906b49f6d7596623507da7f404b4e369e621ba7
Author: Dustin Long <me@dustmop.io>
Date:   Tue Mar 17 14:02:18 2020 -0400

    fix(registry): Prompt for password without echoing it to terminal

:100644 100644 ecf43c8e 319b19e3 M      cmd/registry.go
```

To me that's _very_ close to what we're looking for. Commit details, plus component-level add/modified/deleted

dustmop:
This is fairly distinct from `get` or `log`, it's only used to get the list of "modified" or "unmodified" or "added" etc labels for each component, such that Desktop can show the history of each commit.

I can imagine a future where this functionality is folded into some other command, but the goal for now is to split `status` so that we can remove `fsi=true` from the api. Larger discussion of that: https://github.com/qri-io/qri/issues/1057

Mr0grog:
> the goal for now is to split `status` so that we can remove `fsi=true` from the api. Larger discussion of that: #1057

In this particular case, couldn’t Desktop send the name of the head version, the same way you currently do on the CLI? It *seems* like that would be enough information to imply what the backend needs to do without sending `fsi=true/false` (though I’m possibly being naïve and missing something here).

Otherwise, should this new command be hidden, like `qri fsi`? https://github.com/qri-io/qri/blob/bfbcbfae2783ab4125c1ca23f47af3f508e98271/cmd/fsi.go#L16-L20

I mostly just worry (or as a user, get annoyed by) unnecessary UI churn. If it’s improving the feature with something new, that’s great. But when it’s just moving the feature around repeatedly so users can never find the things they use after updating, that’s a recipe for real anger. (Obviously, the flip-side here is that this feature probably has very few users.)

dustmop:
Moving this functionality to a subcommand of `fsi` feels wrong as this new command `whatchanged` actually has nothing to do with FSI. It's entirely about asking what's in history in the repository.

Mr0grog:
> Moving this functionality to a subcommand of `fsi` feels wrong

Whoa, that’s not what I was suggesting at all! Apologies if I was unclear. I was just suggesting you add `Hidden: true` to this new command.

dustmop:
Oh whoops, I read that too quickly. My mistake, I wasn't familiar with that option at all. Sure, that sounds reasonable to have.

Issue #1163, opened 2020-03-10
qri registry signup should not use a --password command-line flag


dustmop:
It's bad practice to use a command-line flag for setting or sending a password. Onlookers can easily see the password typed into the screen. The password remains visible until scrolled off screen. By default, the password will get saved in bash's history in plain text. 

A better thing to do is to have a non-echoing prompt to get the user's password: https://stackoverflow.com/questions/2137357/getpasswd-functionality-in-go

b5:
Big fan of this. I'd love to see this addressed ASAP

dustmop:
Fixed by https://github.com/qri-io/qri/pull/1199.

PR #1199, opened 2020-03-17
fix(registry): Prompt for password without echoing it to terminal

dustmop:


dustmop:
Fixes https://github.com/qri-io/qri/issues/1163.

PR #1200, opened 2020-03-18
fix(cmd): added the email flag to 'registry prove' as it is required

Arqu:
Needed to re-authenticate/bind myself as I nuked my repo. 
`qri registry prove` requires an email to be supplied on the remote end but had no way to pass it.


b5:
dang, needs a rebase

Issue #1172, opened 2020-03-12
TestDatasetPullPushDeleteFeedsPreviewHTTP is flaky, failed on CI


dustmop:
Saw the following failure:

```
=== RUN   TestDatasetPullPushDeleteFeedsPreviewHTTP
0/6 blocks transferred
1/6 blocks transferred
2/6 blocks transferred
3/6 blocks transferred
4/6 blocks transferred
5/6 blocks transferred
6/6 blocks transferred
done!
    TestDatasetPullPushDeleteFeedsPreviewHTTP: remote_test.go:135: result mismatch (-want +got):
          []string{
          	... // 6 identical elements
          	"DatasetPushPreCheck",
          	"DatasetPushFinalCheck",
        + 	"DatasetPushFinalCheck",
        + 	"DatasetPushFinalCheck",
        + 	"DatasetPushed",
        + 	"DatasetPushed",
          	"DatasetPushed",
          	"LogRemovePreCheck",
          	... // 4 identical elements
          }
```

dustmop:
This is interesting, in the test, these hooks are being run either by `callCheck` or `requireLogAndRefCallCheck`. _Only_ `DatasetPushFinalCheck` and `DatasetPushed` are using `requireLogAndRefCallCheck`, and they're the only hooks that sometimes appear too many times.

dustmop:
Changing those hooks to use `callCheck` doesn't fix the flakiness, false alarm.

dustmop:
Still flaky, causing CI failures. Also observed by @Arqu, and by me just now:

```
=== RUN   TestDatasetPullPushDeleteFeedsPreviewHTTP
0/6 blocks transferred
1/6 blocks transferred
2/6 blocks transferred
3/6 blocks transferred
4/6 blocks transferred
5/6 blocks transferred
6/6 blocks transferred
done!
    TestDatasetPullPushDeleteFeedsPreviewHTTP: remote_test.go:135: result mismatch (-want +got):
          []string{
          	... // 6 identical elements
          	"DatasetPushPreCheck",
          	"DatasetPushFinalCheck",
        + 	"DatasetPushFinalCheck",
        + 	"DatasetPushed",
          	"DatasetPushed",
          	"LogRemovePreCheck",
          	... // 4 identical elements
          }
--- FAIL: TestDatasetPullPushDeleteFeedsPreviewHTTP (0.04s)
```

dustmop:
[...]

PR #1201, opened 2020-03-18
chore(dag): Bump dag commit to get flaky test fixes

dustmop:
This pulls in https://github.com/qri-io/dag/pull/28 which should fix https://github.com/qri-io/qri/issues/1172 and possibly other flaky tests that rely on remotes only calling the finalize hook once.


go.sum
```diff
@@ -869,6 +869,8 @@ github.com/qri-io/compare v0.1.0 h1:A/MRx3uEnJ/iMjfJY1VOqH9CYs9zFSEYaFVeXuGfmis=
 github.com/qri-io/compare v0.1.0/go.mod h1:i/tVuDGRXVxhuZ8ZUieF23u6rQ6wLGJl7KKWpoMRaTE=
 github.com/qri-io/dag v0.2.1-0.20191025201336-254aa177fbd7 h1:hujyzdufAz3Hi6I0I0yKyJTLCM9eoLamKhlpwaqE7ec=
```
b5: mind running `go mod tidy` and amending this commit? It'll drop a bunch of stale imports from `go.sum` like this one

Issue #1190, opened 2020-03-16
`.DS_Store` trips up `qri remove --all --force`


feep:
This isn’t qri’s fault. It is the Finder crapping all over the filesystem

All the same, you are going to have Mac users.

Do we want to special case a directory containing only `.DS_Store` as empty?

```
❯ qri remove --all --force 
for linked dataset [feep/ca_parks_starlark]

remove /Users/rusty/qri/caparks/qri/starlark: directory not empty

❯ ls
.rw-r--r--@ 6.1k rusty 03-16 00:55 -I .DS_Store
```

dustmop:
I think the best bet here would be to get an explicit list of "low-information" files that `remove --force` should be okay with deleting. Definitely `.DS_Store`, but there's also things like `.thumbs.db` and some other Windows specific files such as `desktop.ini`. Would be nice to have a well-curated list that is generally agreed upon, and then we should document that as well.

feep:
A commented blacklist here:

https://github.com/sindresorhus/junk/blob/a7c280daddca13fe5d2f731235b97bdfb5940fda/index.js#L3

More, but that one seems pretty good.

https://github.com/search?q=.ds_store+thumbs.db

Arqu:
Just to bring our slack conversation back into this:
- I wanted to know the intention of --force; Is it just to completely wipe a dataset and anything associated with the directory or purely just qri related files hence the need for an ignore list of common low value files.
It's the latter.

PR #1203, opened 2020-03-19
fix(fsi): qri remove --all --force should not fail on low value files

Arqu:
Addresses issue #1190.

### DIFF
--- a/cmd/fsi_integration_test.go
+++ b/cmd/fsi_integration_test.go
@@ -104,11 +104,11 @@ func (run *FSITestRunner) ChdirToWorkDir(subdir string) string {
 }
 
 // CreateSubDir creates a sub directory from the current working directory
-func (run *FSITestRunner) CreateSubDir(subdir string) string {
+func (run *FSITestRunner) CreateSubDir(t *testing.T, subdir string) string {
 	subDirPath := filepath.Join(run.WorkPath, subdir)
 	err := os.MkdirAll(subDirPath, os.ModePerm)
 	if err != nil {
-		panic(err)
+		t.Fatal(err)
 	}
 	return subDirPath
 }
--- a/cmd/fsi_integration_test.go
+++ b/cmd/fsi_integration_test.go
@@ -103,6 +103,16 @@ func (run *FSITestRunner) ChdirToWorkDir(subdir string) string {
 	return run.WorkPath
 }
 
+// CreateSubDir creates a sub directory from the current working directory
+func (run *FSITestRunner) CreateSubDir(subdir string) string {
+	subDirPath := filepath.Join(run.WorkPath, subdir)
+	err := os.MkdirAll(subDirPath, os.ModePerm)
+	if err != nil {
+		panic(err)
+	}
+	return subDirPath
+}
+
 // CreateAndChangeToWorkDir creates and changes to the working directory
 func (run *FSITestRunner) CreateAndChdirToWorkDir(subdir string) string {
 	run.WorkPath = filepath.Join(run.RootPath, subdir)
--- a/cmd/remove_integration_test.go
+++ b/cmd/remove_integration_test.go
@@ -514,19 +514,9 @@ func TestRemoveAllVersionsWorkingDirectoryLowValueFiles(t *testing.T) {
 
 	}
 
-	lowValueDirs := []string{
-		".Spotlight-V100", // Directory that might appear on external disk
-		"@eaDir",          // Synology Diskstation "hidden" folder where the server stores thumbnails
-	}
-
 	// Add low value files
 	for _, file := range lowValueFiles {
 		run.MustWriteFile(t, file, "\n")
-	}
-
-	// Add low value dirs
-	for _, dir := range lowValueDirs {
-		run.CreateSubDir(t, dir)
 	}
 
 	// Remove all versions
--- a/cmd/remove_integration_test.go
+++ b/cmd/remove_integration_test.go
@@ -451,7 +451,7 @@ func TestRemoveAllVersionsWorkingDirectory(t *testing.T) {
 	}
 
 	// Remove all versions
-	run.MustExec(t, "qri remove --all=1")
+	run.MustExec(t, "qri remove --all")
 
 	// Verify that dsref of HEAD is empty
 	dsPath3 := run.GetPathForDataset(t, 0)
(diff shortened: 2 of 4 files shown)


fsi/fsi.go
```diff
@@ -210,6 +211,11 @@ func (fsi *FSI) Unlink(dirPath, refStr string) error {
 	}
 
 	defer func() {
+		// clean up low value files before running Remove
+		err := removeLowValueFiles(dirPath)
+		if err != nil {
+			log.Errorf("removing directory: %s", err.Error())
+		}
```
dustmop: I think this would be better if it lived here: https://github.com/qri-io/qri/blob/master/lib/datasets.go#L793. For one, this behavior isn't directly needed by `fsi.Unlink`, only by `lib.Remove`. If a user runs `qri fsi unlink` then this function also gets called, via the `Unlink` function in `lib/fsi.go`, but that shouldn't be deleting files. Furthermore, having this functionality here means that it doesn't respect the `--keep-files` flag, which if true shouldn't modify any files at all.

Would also be great to have a test for this in `cmd/`, which creates a directory, runs `qri init`, runs `save`, writes a `.DS_Store` file to that directory, then runs `remove --all` to make sure the whole thing is deleted. Copying `TestRemoveOneRevisionFromWorkingDirectory` then modifying it would be a great start.

b5: Agreed on moving the call site to https://github.com/qri-io/qri/blob/master/lib/datasets.go#L793, but I think the logic of what we mean by `RemoveAll` should live in the `fsi` package. To me the best of both worlds is if `fsi` had an exported function:

```go
package fsi

// RemoveAll is like os.RemoveAll, but it also ignores "low-value files"
func RemoveAll(path string) error {
  // ...
}
```

Then call that function from lib.

fsi/fsi.go
```diff
@@ -253,6 +259,48 @@ func removeLinkFile(dir string) error {
...
+		pattern := strings.Join(filesToBeRemoved, "|")
+		matched, err := regexp.MatchString(pattern, filepath.Base(f.Name()))
+		if err != nil {
+			return err
+		}
+		if matched {
+			os.Remove(path)
+		}
+	}
+	return nil
```
b5: My only other question here: should we also be removing directories if they are empty after removing low-value files?  Personally, I think we should. when I run `qri remove --all --force` I'm trying to blow the directory away.

We could save that for another PR, but if we can get it knocked out now, I'd be all for it

fsi/fsi.go
Arqu: I took a bit of a split approach with `datastore` retaining control but moving the execution logic to `fsi`.

fsi/fsi.go
Arqu: Got covered as part of the update :)

Arqu:
Still pending to finish add the tests.

fsi/fsi.go
```diff
@@ -230,6 +222,36 @@ func (fsi *FSI) Unlink(dirPath, refStr string) error {
...
+}
+
+// RemoveAll unlinks the dataset and removes all associated files
+func (fsi *FSI) RemoveAll(dirPath, refStr string) error {
+	if err := fsi.Unlink(dirPath, refStr); err != nil {
+		return err
+	}
+	if err := os.RemoveAll(dirPath); err != nil {
```
dustmop: I'm wary of ever calling `RemoveAll`, because it's potentially very destructive, especially when taking unknown / arbitrary input as a parameter. The only places we use it in the codebase right now are for tests that construct temporary directories, and one or two places that are part of rarely used commands, like a user deleting their whole repository. Since we're a data storage system, when it comes to deletes, I'd rather err on the side of being cautious over convenient, and whitelist the things we delete, rather than doing a `RemoveAll`.

In fact, this makes me realize, if a user at some point runs `qri init` in their home directory, and then later run `remove --all --force`, this codepath would try and delete their entire home directory. That alone is reason enough why we shouldn't make `--force` mean "remove even unknown files". The current semantics of `--force` are "delete things even if we get errors, including the working directory not being clean", and I think it should stay that way.

lib/datasets.go
```diff
@@ -828,7 +828,15 @@ func (r *DatasetRequests) Remove(p *RemoveParams, res *RemoveResponse) error {
 	if p.Revision.Gen == dsref.AllGenerations {
 		// removing all revisions of a dataset must unlink it
 		if ref.FSIPath != "" {
-			if err := r.inst.fsi.Unlink(ref.FSIPath, ref.AliasString()); err == nil {
+			var err error
+			if p.Force {
+				err = r.inst.fsi.RemoveAll(ref.FSIPath, ref.AliasString())
+			} else if !p.KeepFiles {
+				err = r.inst.fsi.Remove(ref.FSIPath, ref.AliasString())
```
dustmop: This call is happening before `fsi.DeleteComponentFiles` below, which is what removes "body.csv", "structure.json", "meta.json" etc. That means in nearly all cases, the working directory will not be empty, and the call to `os.Remove` within `fsi.Remove` will fail.

There's already a call to `os.Remove` below on the next line (after `fsi.DeleteComponentFiles`). I think the logic of removing low-value files should be happening over there, instead.

fsi/fsi.go
```diff
@@ -253,6 +275,70 @@ func removeLinkFile(dir string) error {
...
+			"^\\.AppleDouble$", // Stores additional file resources
+			"^\\.LSOverride$", // Contains the absolute path to the app to be used
+			"^Icon\\r$", // Custom Finder icon: http://superuser.com/questions/298785/icon-file-on-os-x-desktop
+			"^\\._.*", // Thumbnail
+			"^\\.Spotlight-V100(?:$|\\/)", // Directory that might appear on external disk
+			"\\.Trashes", // File that might appear on external disk
+			"^__MACOSX$", // Resource fork
+
+			// linux specific files
+			"~$", // Backup file
```
dustmop: Not fan of this one pattern. Any file ending with a `~` will be considered low-value, that's too broad and easy to get false positives.

fsi/fsi.go
```diff
@@ -253,6 +275,70 @@ func removeLinkFile(dir string) error {
...
+		}
+		if matched {
+			return true
+		}
+	}
+	return false
+}
+
+func getLowValueFiles(files *[]string) filepath.WalkFunc {
+	return func(path string, f os.FileInfo, err error) error {
```
dustmop: Make sure to run `go fmt ./...` in order to apply the standard golang code style guide.

cmd/fsi_integration_test.go
```diff
@@ -103,6 +103,16 @@ func (run *FSITestRunner) ChdirToWorkDir(subdir string) string {
 	return run.WorkPath
 }
 
+// CreateSubDir creates a sub directory from the current working directory
+func (run *FSITestRunner) CreateSubDir(subdir string) string {
+	subDirPath := filepath.Join(run.WorkPath, subdir)
+	err := os.MkdirAll(subDirPath, os.ModePerm)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
Arqu: Yeah, lot's of these small bits we're just taken out of other examples. I'll look over it and appropriate this + if I find this setup anywhere else.

cmd/remove_integration_test.go
```diff
@@ -465,6 +465,88 @@ func TestRemoveAllVersionsWorkingDirectory(t *testing.T) {
...
+		run.MustWriteFile(t, file, "\n")
+	}
+
+	// Add low value dirs
+	for _, dir := range lowValueDirs {
+		run.CreateSubDir(dir)
+	}
+
+	// Remove all versions
+	run.MustExec(t, "qri remove --all=1")
```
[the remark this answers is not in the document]
Arqu: Similar to the above, seen this in several places. For this PR will fix just this case but will open another to clean up the rest.

fsi/fsi.go
Arqu: Agree on RemoveAll being very agressive and some bad input could wreak havoc.
On the alternative side, doing just RemoveAll and Remove will essentially result in the same behavior, unless Remove no longer removes low value files and it just triggers on `--force` which technically is the purpose of this PR.

Think the above might be a best blend between the two things while remaining cautious.

fsi/fsi.go
Arqu: Happy to remove it, mostly picked it up from some other list of "low value files"

lib/datasets.go
Arqu: 👍 Currently only works due to `os.RemoveAll` from above.

Issue #1169, opened 2020-03-11
Unlink deletes .qri-ref in the current directory, even if that’s not where the specified dataset was


Mr0grog:
Calling `qri fsi unlink me/some_dataset` or `qri fsi unlink me/some_dataset path/to/some_other_dataset` (where `me/some_dataset` is linked to a different directory than `path/to/some_other_dataset`) from a different directory than the specified dataset does not remove the `.qri-ref` file from the dataset’s linked directory and instead attempts to delete it from the working directory (woe unto you if you happen to call this from a different dataset’s linked directory like I did).

Steps to reproduce:

1. Checkout one dataset somewhere.
2. Checkout a different dataset in another directory.
3. `cd` into the second dataset’s directory.
4. Run `qri fsi unlink <name of first dataset>`

Result:

The first dataset is unlinked, but the directory it was linked to still has a `.qri-ref` file. The current directory (for the second dataset, which should still be linked) does *not* have a `.qri-ref` file.

Expected:

The directory for the first dataset should no longer have a `.qri-ref` file and the directory for the second dataset should still have an unaltered `.qri-ref` file.

It looks like this happens because explicitly specifying a dataset ref results in a ref object with no directory path: https://github.com/qri-io/qri/blob/924157c253bf571ec9df92b0fd63e316e67bc6ad/cmd/ref_select.go#L113-L122

and: https://github.com/qri-io/qri/blob/924157c253bf571ec9df92b0fd63e316e67bc6ad/cmd/ref_select.go#L43-L45

Then the call to unlink sets the directory to whatever directory was attached to the ref without checking anything. Since it initializes to an empty string, that winds up implying the current working directory: https://github.com/qri-io/qri/blob/924157c253bf571ec9df92b0fd63e316e67bc6ad/cmd/fsi.go#L108-L113

And the actual unlink logic never sanity checks the directory path it’s given: https://github.com/qri-io/qri/blob/924157c253bf571ec9df92b0fd63e316e67bc6ad/fsi/fsi.go#L201-L210

I’m not sure exactly where in here the right place to do some checking is, but maybe `cmds.fsi.Unlink` should be calling `GetLinkedRefSelect()` of the ref before grabbing the refs directory?

dustmop:
[...]

PR #1205, opened 2020-03-19
fix(cmd): unlink now uses it's respective dir instead of current dir

Arqu:
WIP
Now it propperly fetches the associated directory. One thing to note is that this technically breaks the "explicit" ref assignment even though behavior is unchanged.

All of this is related to issue #1169

### DIFF
--- a/cmd/fsi.go
+++ b/cmd/fsi.go
@@ -7,6 +7,7 @@ import (
 	"github.com/qri-io/ioes"
 	"github.com/qri-io/qri/lib"
 	"github.com/spf13/cobra"
+	reporef "github.com/qri-io/qri/repo/ref"
 )
 
 // NewFSICommand creates a new `qri fsi` command for working with file system
@@ -108,8 +109,14 @@ func (o *FSIOptions) Unlink() error {
 		printRefSelect(o.ErrOut, o.Refs)
 
 		p := &lib.LinkParams{
-			Dir: o.Refs.Dir(),
 			Ref: ref,
+		}
+
+		dsRef := reporef.DatasetRef{}
+		if err := o.FSIMethods.FSIDatasetForRef(&ref, &dsRef); err != nil {
+			p.Dir = o.Refs.Dir()
+		} else {
+			p.Dir = dsRef.FSIPath
 		}
 
 		if err := o.FSIMethods.Unlink(p, &res); err != nil {
--- a/cmd/ref_select.go
+++ b/cmd/ref_select.go
@@ -9,7 +9,6 @@ import (
 	"github.com/qri-io/qri/fsi"
 	"github.com/qri-io/qri/lib"
 	"github.com/qri-io/qri/repo"
-	reporef "github.com/qri-io/qri/repo/ref"
 )
 
 // RefSelect represents zero or more references, either explicitly provided or implied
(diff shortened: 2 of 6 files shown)


cmd/ref_select.go
```diff
@@ -40,7 +41,14 @@ func NewEmptyRefSelect() *RefSelect {
 }
 
 // NewExplicitRefSelect returns a single explicitly provided reference
-func NewExplicitRefSelect(ref string) *RefSelect {
+func NewExplicitRefSelect(ref string, fsi *lib.FSIMethods) *RefSelect {
+	if fsi != nil {
+		dsRef := reporef.DatasetRef{}
+		err := fsi.FSIDatasetForRef(&ref, &dsRef)
+		if err == nil {
+			return &RefSelect{kind: "for linked", refs: []string{ref}, dir: dsRef.Path}
```
Arqu: We do a check on `kind` to define an `explicit` ref and `dri` to check if it is a `linked` ref.

dustmop:
Sorry for the confusion caused by this code being misleading, but I don't think this is the right approach. NewExplicitRefSelect should not be changing to have the linked directory; the intended purpose of that function is to bypass FSI and use an explicit reference.

This should be better documented with an explanation somewhere, but basically, with FSI, the .qri-ref file in the working directory acts an a way to use an implied reference. So instead of needing to type `qri get meta me/my_dataset` in a working directory, you can just type `qri get meta` and qri looks at the .qri-ref file to get the reference `me/my_dataset`. An explicit reference represents `me/my_dataset` when typed by the user, so it shouldn't be acting like an implied reference.

The real bug is in `cmd/fsi.go`, where the Dir is passed into Unlink even though it's blank. What should happen instead is that the ref gets passed to `CanonicalizeDatasetRef` which looks up that ref in the repository and sets the FSIPath if any exists. *Then* the FSIPath can be used to find the linkFile (.qri-ref) to delete. Notice that `Unlink` in fsi/fsi.go is already calling `CanonicalizeDatasetRef`, but it does so after calling `removeLinkFile`, which is incorrect.

Furthermore, I think it's a bug that `Unlink` is calling `os.Remove`, but that's a separate issue. I'll follow up with that elsewhere.

Arqu:
This makes a lot more sense now. I've simply tried to address the initial problem by shoehorning it into `NewExplicitRefSelect`. I'll do a rework that lives higher up in `cmd/fsi.go` and have it only deal with this case instead of the more "global" change.

cmd/fsi.go
```diff
@@ -108,10 +109,16 @@ func (o *FSIOptions) Unlink() error {
 		printRefSelect(o.ErrOut, o.Refs)
 
 		p := &lib.LinkParams{
-			Dir: o.Refs.Dir(),
 			Ref: ref,
 		}
 
+		dsRef := reporef.DatasetRef{}
```
Arqu: Note that we don't need full Canonicalization, we just want the FSIPath.

cmd/fsi.go
```diff
@@ -108,10 +109,16 @@ func (o *FSIOptions) Unlink() error {
...
+		if err := o.FSIMethods.FSIDatasetForRef(&ref, &dsRef); err != nil {
```
dustmop: Note that the implementation of `FSIDatasetForRef` calls `CanonicalizeDatasetRef`, which is how it's getting the `FSIPath`. I'm actually not a fan of `FSIDatasetForRef`, it's not being called anywhere in the codebase, and I'm of the opinion that it should be deleted since it's not particularly doing any work.

cmd/fsi.go
```diff
@@ -108,10 +109,16 @@ func (o *FSIOptions) Unlink() error {
...
+			p.Dir = o.Refs.Dir()
```
dustmop: The most common case where `FSIDatasetForRef` returns an error is when the provided reference is unknown (or misspelled). For example, if I run `qri fsi unlink me/not_a_dataset` then we'll get an error and hit this codepath. Because the reference has to be explicitly provided, it's constructed by `NewExplicitRefSelect`, so we're sure that `Dir` is an empty string. That will lead to the same behavior as the original bug: the `fsi.Unlink` function will remove the `.qri-ref` from the current directory. Instead, hitting this error should do nothing aside from displaying the error.

dustmop:
Looking into this in a bit more detail, the fix is a lot more complicated then it seemed at first glance. I'm going to start a new PR with what I think is the needed approach.

Arqu:
Closing in favor of https://github.com/qri-io/qri/pull/1213

Issue #1153, opened 2020-03-05
`qri init` does not mention the name length max of 144 chars


feep:
You fixed #1116 before I tried to file it. Thanks.

`qri init` does not mention the max length for names, but rejects them anyway.

```
❯ qri init
Name of new dataset [scratch]: a234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890
Format of dataset, csv or json [csv]:   
dataset name must start with a letter, and only contain letters, numbers, and underscore

❯ qri version
0.9.6
```

That name `a234...` passes all the criteria mentioned in the error message, but it is 160 characters long.

Also, it uses a different error message than `qri save`.

```
❯ qri save
for linked dataset [feep/en.wikipedia.org,List_of_highest-grossing_films,Highest-grossing_films]invalid name: error: illegal name 'en.wikipedia.org,List_of_highest-grossing_films,Highest-grossing_films', names must start with a letter and consist of only a-z,0-9, and _. max length 144 characters
```

That’s probably 0.9.5. Can’t reproduce, because you fixed the init bug. =]

Yeah. I know, ridiculous name. I had reasons™.

They don’t call me the human fuzz tester for nothin’.

dustmop:
Fixed the `qri init` issue by https://github.com/qri-io/qri/pull/1208.

Issue #1132, opened 2020-02-20
datasets should use names that only allow lowercase letters


dustmop:
The docs say that dataset names can only use lower case characters (not upper case). Such as here: https://qri.io/docs/getting-started/qri-cli-quickstart

However, `save` lets you create dataset names with upper case characters, and the newly added dsref parser also recognizes there. The problem is when different datasets can have names that are the same except for case differences. We have to remove this problem.

We should discourage upper case characters in dataset names. The first step is to fail to parse all commands that use upper case characters in dataset names, except for rename. If a user has an existing dataset that uses upper case characters in the name, they must rename it before continuing. Over time we can have some auto-migration features for automatically changing these dataset names.

dustmop:
We should fix this sooner rather than later. As long as it's possible to create invalid dataset names, users will keep making them even though it's discouraged.

To clarify, right now a user can do:
```
qri save --body my_rows.csv me/Not_A_Valid_Name
```
we still need to be able to parse these invalid names (because datasets with those names exist in users' repos), but shouldn't let users do anything with them aside from rename. So we need a separate parse step for the left and right parameters for this:
```
qri rename me/Not_A_Valid_name me/totally_okay_name
```

Here's how I picture this work:
1) Add functionality to `dsref/parse.go` so that it returns a standardized error if the dataset name contains an uppercase character. Something like `ErrBadCaseName`.
2) For every command in qri (except rename), if the implementation is already using `dsref` in its `cmd/` handler, check for that error and display a warning like "dataset name should not use upper-case characters, please use rename to change the dataset name". The command should continue to work for now.
3) If the cmd handler is not using dsref, just use it to parse the ref, to check for this warning, even if we throw away the ref after the fact.
4) Don't allow the creation of new datasets using `save` if they have bad-case in their name.
5) Allow rename to parse bad-cased dataset names *for the first parameter*, but not the second. This will allow users to fix datasets that were created before we enforce this restriction.
6) A few versions down the line, turn the warning into an error.
7) Come up with a solution of Desktop, which maybe just down-cases everything automatically? Make sure that the input fields on Desktop are using the same lower-case only regex that `dsref/parse.go` has.

Related bugs:
https://github.com/qri-io/qri/issues/1153
https://github.com/qri-io/qri/issues/1116 (fixed)

dustmop:
- [x] Add functionality to dsref/parse.go so that it returns a standardized error if the dataset name contains an uppercase character. Something like ErrBadCaseName.
[...]

PR #1208, opened 2020-03-19
fix(save): Cannot save new datasets if name contains upper-case characters

dustmop:
Dsref parser returns a particular error if the name is using upper-case characters. For save, refuse to save new datasets if that error is hit. If a dataset already exists with that name, display a warning that the user should rename their dataset. Also display this same warning for `qri get`; other commands should also start showing this warning.

Fix `qri init`'s error message when the dataset name is too long. Remove an unneeded parse operation from cmd.Save now that parsing happens in lib instead. Remove code in cmd that added "me" username since that's unnecessary.

Add a bunch of basic functionality tests for the save cmd, including tests that you can't save a dataset with a different username then your own.


dustmop:
Fixes https://github.com/qri-io/qri/issues/1153, and handles steps 1 and 4 of https://github.com/qri-io/qri/issues/1132.

cmd/save_test.go
```diff
@@ -199,6 +204,129 @@ func TestSaveRun(t *testing.T) {
...
+		//	"qri save --file structure.json me/my_dataset",
+		//	"TODO(dustmop): Should be possible to save a dataset with structure and no body",
+		//},
+		//{
+		//	"structure file explicit ref",
+		//	"qri save --file structure.json test_peer/my_dataset",
+		//	"TODO(dustmop): Should be possible to save a dataset with structure and no body",
+		//},
+	}
+	for _, c := range goodCases {
```
b5: I'd prefer each case run as a subtest. Example:

https://github.com/qri-io/qri/blob/a9ed29555085b123e2cf263e511677a72f651cfe/base/dsfs/dataset_test.go#L653-L669

cmd/cmd.go
```diff
@@ -112,11 +109,3 @@ func parseSecrets(secrets ...string) (map[string]string, error) {
...
 	return s, nil
 }
-
-// parseCmdLineDatasetRef parses DatasetRefs, assuming peer "me" if none given.
-func parseCmdLineDatasetRef(ref string) (reporef.DatasetRef, error) {
-	if !strings.ContainsAny(ref, "@/") {
-		ref = "me/" + ref
-	}
-	return repo.ParseDatasetRef(ref)
-}
```
b5: Delighted to see this go

cmd/save_test.go
```diff
@@ -199,6 +204,129 @@ func TestSaveRun(t *testing.T) {
...
+			"qri save --file dataset.yaml me/my_dataset@/ipfs/QmVersion",
+			"ref can only have username/name",
+		},
+		{
+			"body file other username",
+			"qri save --body body_ten.csv other/my_dataset",
+			"cannot save using a different username than \"test_peer\"",
+		},
+	}
+	for _, c := range badCases {
```
b5: ditto, run as subtest.

cmd/save_test.go
```diff
@@ -199,6 +204,129 @@ func TestSaveRun(t *testing.T) {
...
+		run := NewTestRunner(t, "test_peer", "qri_test_save_basic")
+		defer run.Delete()
```
b5: It's funny to think about what this defer statement is doing. Since [defers call on enclosing function return](https://golang.org/ref/spec#Defer_statements) this is stacking up a bunch of delete calls in a row then calling them all in succession after the test function returns.

No need to change it, but running this in a subtest _would_ alter this

cmd/test_runner_test.go
```diff
@@ -246,3 +249,49 @@ func executeCommandC(root *cobra.Command, args ...string) (err error) {
...
+	ref, err := dsref.ParseHumanFriendly(refStr)
+	if err != nil && err != dsref.ErrBadCaseName {
+		t.Fatal(err)
+	}
+
+	ds.Peername = ref.Username
+	ds.Name = ref.Name
+
+	inst, err := lib.NewInstance(ctx, run.RepoRoot.QriPath)
+	// NOTE(dustmop): There's a bug with TestRepo that I don't understand completely. The commands
```
b5: innnteresting. definitely please file this same comment as an issue so we can get it cleaned up

lib/datasets.go
```diff
@@ -235,6 +235,11 @@ func (r *DatasetRequests) Get(p *GetParams, res *GetResult) (err error) {
 	}
 	ctx := context.TODO()
 
+	// Check if the dataset ref uses bad-case characters, show a warning.
+	if _, err := dsref.Parse(p.Path); err == dsref.ErrBadCaseName {
+		log.Error(dsref.ErrBadCaseShouldRename)
```
b5: works for now, we should have a long conversation about logging & how we actually want it to work. Another time

lib/datasets.go
```diff
@@ -445,36 +449,59 @@ func (r *DatasetRequests) Save(p *SaveParams, res *reporef.DatasetRef) (err erro
...
+	// Validate that username is our own, it's not valid to try to save a dataset with someone
+	// else's username. Without this check, base will replace the username with our own regardless,
+	// it's better to have an error to display, rather than silently ignore it.
+	pro, err := r.node.Repo.Profile()
+	if err != nil {
+		return err
+	}
+	if ref.Username != "" && ref.Username != "me" && ref.Username != pro.Peername {
+		return fmt.Errorf("cannot save using a different username than \"%s\"", pro.Peername)
+	}
```
b5: this seems like valuable logic to put elsewhere at some point. Not totally sure where yet. Off the top of my head it'd be a method on a `Profile`:

```go
profile.OwnsRef(ref dsref.Ref) bool
```

cmd/save_test.go
dustmop: Whoops, the intention was actually to have that latter behavior, it was buggy before. Putting it into subtests fixes that.

lib/datasets.go
dustmop: Another option would be to set a value on the lib.Get result, and have the caller display the warning instead.

cmd/save_test.go
dustmop: Good call, will do.

Issue #1046, opened 2019-12-09
TestFetchCommand is flaky, fails occassionally on CI, and shouldn't run `connect`


dustmop:
Sometimes this test fails with:
```
=== RUN   TestFetchCommand
17:38:33.564 ERROR     qrip2p: error adding peer: open /tmp/qri_test_commands_connect/qri/peers.json.lock: no such file or directory bootstrap.go:30
17:38:33.566 ERROR     qrip2p: error adding peer: open /tmp/qri_test_commands_connect/qri/peers.json.lock: no such file or directory bootstrap.go:30
17:38:33.570 ERROR     qrip2p: error adding peer: open /tmp/qri_test_commands_connect/qri/peers.json.lock: no such file or directory bootstrap.go:30
17:38:33.578 ERROR     qrip2p: error adding peer: open /tmp/qri_test_commands_connect/qri/peers.json.lock: no such file or directory bootstrap.go:30
17:38:38.619 ERROR     flatfs: could not store final value of disk usage to file, future estimates may be inaccurate flatfs.go:883
--- FAIL: TestFetchCommand (5.14s)
    fetch_test.go:68: Get http://localhost:2503/remote/logsync?ref=peer_a/test_movies: dial tcp 127.0.0.1:2503: connect: connection refused
```

Also, since it runs `qri connect`, there's race conditions with other tests, they may change behavior if they think connect is running.

dustmop:
Got a new error in this test:
```
--- FAIL: TestFetchCommand (6.82s)
    fetch_test.go:68: Get http://localhost:2503/remote/logsync?ref=peer_a/test_movies: dial tcp [::1]:2503: connect: connection refused
18:35:23.980 ERROR     flatfs: could not store final value of disk usage to file, future estimates may be inaccurate flatfs.go:883
18:35:24.292 ERROR     flatfs: could not store final value of disk usage to file, future estimates may be inaccurate flatfs.go:883
panic: Fail in goroutine after TestFetchCommand has completed

goroutine 2775 [running]:
testing.(*common).Fail(0xc0000eb100)
	/usr/local/go/src/testing/testing.go:609 +0x151
testing.(*common).FailNow(0xc0000eb100)
	/usr/local/go/src/testing/testing.go:631 +0x2b
testing.(*common).Fatal(0xc0000eb100, 0xc0008f9fa8, 0x1, 0x1)
	/usr/local/go/src/testing/testing.go:710 +0x78
[...]

PR #1211, opened 2020-03-20
test(fetch): Rewrite fetch test. Don't run connect, test expectations.

dustmop:
Rewrite the TestFetchCommand to fix a number of problems. No longer runs "qri connect". Doesn't use the 2503 port. No longer serves RPC and Websockets. Removes the 5 second wait. Actually inspects the output of `log` and `fetch` to validate that they're working, with stable timestamps. Should fix a number of flakey tests and overall run faster.

### DIFF
--- a/cmd/fetch_test.go
+++ b/cmd/fetch_test.go
@@ -2,77 +2,157 @@ package cmd
 
 import (
 	"context"
+	"fmt"
+	"net/http"
+	"regexp"
 	"testing"
 	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/qri-io/qri/api"
+	"github.com/qri-io/qri/config"
+	"github.com/qri-io/qri/lib"
+	"github.com/qri-io/qri/logbook"
 )
 
 func TestFetchCommand(t *testing.T) {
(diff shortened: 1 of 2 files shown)


dustmop:
Should fix https://github.com/qri-io/qri/issues/1046, may also fix https://github.com/qri-io/qri/issues/1159.

cmd/fetch_test.go
```diff
@@ -2,77 +2,157 @@ package cmd
...
 func TestFetchCommand(t *testing.T) {
-	r := NewTestRunner(t, "peer_a", "qri_test_fetch_a")
-	defer r.Delete()
+	a := NewTestRunner(t, "peer_a", "qri_test_fetch_a")
+	defer a.Delete()
 
-	ctx, done := context.WithCancel(context.Background())
-	defer done()
+	// TODO(dustmop): Move most of the below hooks into a common testRunner. Maybe the basic
+	// TestRunner will work?
```
b5: The test runner itself looks like it's worth a little scrutiny, but agree with putting a TODO here & bundling up examining the test runner for another day

PR #1213, opened 2020-03-23
fix(unlink): Fsi unlink handles more edge-cases, acts more sane.

dustmop:
Fsi unlink command will delete the appropriate link file, regardless of how it is called. Better handle edge cases, like inconsistent folder state. Add lots of tests to cover all of the types of behavior fsi unlink has.

### DIFF
--- a/cmd/test_runner_test.go
+++ b/cmd/test_runner_test.go
@@ -182,10 +182,7 @@ func (run *TestRunner) LookupVersionInfo(refStr string) *dsref.VersionInfo {
 	if err != nil {
 		return nil
 	}
-	datasetRef := reporef.DatasetRef{
-		Peername: dr.Username,
-		Name:     dr.Name,
-	}
+	datasetRef := reporef.RefFromDsref(dr)
 	// TODO(dustmop): Work-around for https://github.com/qri-io/qri/issues/1209
 	// Would rather do `run.RepoRoot.Repo()` but that doesn't work.
 	ctx := context.Background()
@@ -210,10 +207,7 @@ func (run *TestRunner) ClearFSIPath(t *testing.T, refStr string) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	datasetRef := reporef.DatasetRef{
-		Peername: dr.Username,
-		Name:     dr.Name,
-	}
+	datasetRef := reporef.RefFromDsref(dr)
 	// TODO(dustmop): Work-around for https://github.com/qri-io/qri/issues/1209
 	// Would rather do `run.RepoRoot.Repo()` but that doesn't work.
 	ctx := context.Background()
--- a/cmd/test_runner_test.go
+++ b/cmd/test_runner_test.go
@@ -18,6 +18,8 @@ import (
 	"github.com/qri-io/qri/lib"
 	"github.com/qri-io/qri/registry"
 	"github.com/qri-io/qri/registry/regserver"
+	"github.com/qri-io/qri/repo"
+	reporef "github.com/qri-io/qri/repo/ref"
 	repotest "github.com/qri-io/qri/repo/test"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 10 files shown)


dustmop:
Fixes https://github.com/qri-io/qri/issues/1169

fsi/fsi.go
```diff
@@ -199,27 +201,32 @@ func (fsi *FSI) ModifyLinkReference(dirPath, refStr string) error {
...
 
-	ref.FSIPath = ""
-	return fsi.repo.PutRef(ref)
+	// The FSIPath is *not* set, which removes it from the refstore
+	datasetRef := reporef.DatasetRef{
+		Peername:  ref.Username,
+		ProfileID: profile.IDB58DecodeOrEmpty(ref.ProfileID),
+		Name:      ref.Name,
+		Path:      ref.Path,
+	}
```
b5: Feels like this should be moved into a utility function so others don't trip on the `ProfileID` decoding. Because `qri-io/qri/repo/ref` imports `qri-io/qri/dsref`, I'd suggest:

```go
package ref

func RefFromDSRef(dsr dsref.Ref) DatasetRef {
  return DatasetRef{
  		Peername:  ref.Username,
  		ProfileID: profile.IDB58DecodeOrEmpty(ref.ProfileID),
  		Name:      ref.Name,
  		Path:      ref.Path,
  	}
}
```

then call that here

dsref/parse.go
```diff
@@ -134,6 +134,15 @@ func ParseHumanFriendly(text string) (Ref, error) {
...
 }
 
+// MustParse parses a dsref from a string, or panics if it fails
+func MustParse(text string) Ref {
+	ref, err := Parse(text)
+	if err != nil {
+		panic(err)
+	}
+	return ref
+}
```
b5: dope.

cmd/test_runner_test.go
```diff
@@ -157,6 +159,73 @@ func (run *TestRunner) CreateCommandRunner(ctx context.Context) *cobra.Command {
...
+		return false
+	}
+	return true
+}
+
+// LookupVersionInfo returns a versionInfo for the ref, or nil if not found
+func (run *TestRunner) LookupVersionInfo(refStr string) *dsref.VersionInfo {
+	// TODO(dustmop): Could directly parse reporef.DatasetRef instead, but we should transition
+	// to dsref's data structures where possible. This will make it easier to switch to dscache
+	// once it exists.
```
b5: that utility function would come in handy here

fsi/fsi.go
```diff
@@ -199,27 +201,32 @@ func (fsi *FSI) ModifyLinkReference(dirPath, refStr string) error {
...
-// Unlink breaks the connection between a directory and a dataset
-func (fsi *FSI) Unlink(dirPath, refStr string) error {
-	ref, err := repo.ParseDatasetRef(refStr)
-	if err != nil {
-		return err
-	}
-
+// Unlink removes the link file (.qri-ref) in the directory, and removes the fsi path
+// from the reference in the refstore
+func (fsi *FSI) Unlink(dirPath string, ref dsref.Ref) error {
```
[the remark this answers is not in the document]
dustmop: Yes, the ref is inside the .qri-ref. The idea is that `Unlink` in lib/fsi.go works exactly as you described, with the assumption that it should have friendlier behavior. `Unlink` in fsi/fsi.go is lower-level, so it acts dumber, essentially performing two separate tasks: deleting the .qri-ref file in the `dir`, and removing the `FSIPath` from the ref in the refstore.

We still want to handle the case where the .qri-ref is deleted from the folder, for which `qri fsi unlink` should still remove the `FSIPath` from the ref. That can't happen if we only pass in the `dir` value. See the test `TestUnlinkMissingLinkFile` which handles this situation.

Arqu:
Sweet! See how the move of Canonicalization actually helps the resolution of the dir now without fumbling for `lib` all the way from `cmd`.
Closing out https://github.com/qri-io/qri/pull/1205 in favor of this.

fsi/fsi.go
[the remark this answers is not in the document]
b5: exactly the explanation I was hoping for. Thanks!

dustmop:
Good call on the utility conversion function. Added, going to merge once the tests complete.

Issue #1158, opened 2020-03-06
TestTwoActorRegistryIntegration is flaky, failed a few different ways on CI


dustmop:
```
    TestTwoActorRegistryIntegration: integration_test.go:40: 0 Peername:  nasim
          ProfileID: QmWYgD49r9HnuXEppQEq1a7SUUryja4QNs9E6XCH2PayCD
          Name:      world_bank_population
          Path:      /ipfs/QmVfExwDm79Eb8KcLxmtMyWESKiztqS8HXmezCLnVYcTMc
          FSIPath:   
          Published: true
        
    TestTwoActorRegistryIntegration: integration_test.go:288: json: cannot unmarshal number into Go value of type struct { Data []*dataset.Dataset; Meta struct { Error string; Status string; Code int } }
--- FAIL: TestTwoActorRegistryIntegration (0.13s)
```

also saw this:

```
=== RUN   TestTwoActorRegistryIntegration
    TestTwoActorRegistryIntegration: integration_test.go:160: registry qri path: /tmp/integration_two_actor_registry_registry515667419/qri
--- FAIL: TestTwoActorRegistryIntegration (0.01s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x68 pc=0x198c82c]

goroutine 4265 [running]:
testing.tRunner.func1.1(0x1d68f60, 0x2e74fb0)
	/usr/local/go/src/testing/testing.go:941 +0x3d0
testing.tRunner.func1(0xc00125a6c0)
	/usr/local/go/src/testing/testing.go:944 +0x3f9
panic(0x1d68f60, 0x2e74fb0)
	/usr/local/go/src/runtime/panic.go:967 +0x15d
github.com/qri-io/qri/remote.NewRemote(0x0, 0xc000453c40, 0x0, 0x0, 0x0, 0x1, 0x1, 0xc0006727e0)
	/go/src/github.com/qri-io/qri/remote/remote.go:133 +0x5dc
github.com/qri-io/qri/lib.(*NetworkIntegrationTestRunner).InitRegistry(0xc0011dded0, 0xc00125a6c0)
	/go/src/github.com/qri-io/qri/lib/integration_test.go:185 +0x4ee
github.com/qri-io/qri/lib.NewNetworkIntegrationTestRunner(...)
[...]

Issue #1047, opened 2019-12-09
TestConnectQriNodes is flaky, fails sometimes on CI


dustmop:
```
--- FAIL: TestConnectQriNodes (0.10s)
    p2ptest_test.go:81: QmbHm5S9hRsvZWE56XWA7dksfXDNABRwK6LaH6KS1z7uEw error upgrading connection to QmPeFTNHcZDr3ZFEfFfepxS5PqHAmfBRGQNPJ389Cwh1as: test qri protocol not supported
    p2ptest_test.go:98: node QmbHm5S9hRsvZWE56XWA7dksfXDNABRwK6LaH6KS1z7uEw does not have a record that node QmPeFTNHcZDr3ZFEfFfepxS5PqHAmfBRGQNPJ389Cwh1as can communicate over the testQri protocol
    p2ptest_test.go:106: node QmbHm5S9hRsvZWE56XWA7dksfXDNABRwK6LaH6KS1z7uEw has no addrs for node QmPeFTNHcZDr3ZFEfFfepxS5PqHAmfBRGQNPJ389Cwh1as
    p2ptest_test.go:115: node QmbHm5S9hRsvZWE56XWA7dksfXDNABRwK6LaH6KS1z7uEw tag value for QmPeFTNHcZDr3ZFEfFfepxS5PqHAmfBRGQNPJ389Cwh1as incorrect. expected: 6, got: 0
FAIL
coverage: 63.7% of statements
FAIL	github.com/qri-io/qri/p2p/test	0.276s
```

dustmop:
Just got a CI failure:

```
=== RUN   TestConnectQriNodes
--- FAIL: TestConnectQriNodes (0.12s)
    p2ptest_test.go:81: QmbHm5S9hRsvZWE56XWA7dksfXDNABRwK6LaH6KS1z7uEw error upgrading connection to QmPeFTNHcZDr3ZFEfFfepxS5PqHAmfBRGQNPJ389Cwh1as: test qri protocol not supported
    p2ptest_test.go:98: node QmbHm5S9hRsvZWE56XWA7dksfXDNABRwK6LaH6KS1z7uEw does not have a record that node QmPeFTNHcZDr3ZFEfFfepxS5PqHAmfBRGQNPJ389Cwh1as can communicate over the testQri protocol
[...]

PR #1076, opened 2020-01-23
feat(cmd): add --no-p2p flag to qri connect to hard-disable p2p connections

b5:
working on this as an ingredient for our `p2p` switch on qri desktop.

PR #1214, opened 2020-03-23
test(connect): Configure connect_test so that it uses its own ports

dustmop:


dustmop:
I believe this will fix https://github.com/qri-io/qri/issues/1158, at least *some* of its failures. Should also finally fix https://github.com/qri-io/qri/issues/1159. May also fix https://github.com/qri-io/qri/issues/1047, though we haven't seen that failure in a while.

cmd/connect_test.go
```diff
@@ -2,51 +2,42 @@ package cmd
...
+	defer run.Delete()
+
+	// Construct a mock registry to pass to the connect command
 	_, registryServer := regmock.NewMockServer()
 
-	path := filepath.Join(os.TempDir(), "qri_test_commands_connect")
-	t.Logf("temp path: %s", path)
+	// Configure ports such that other tests do not conflict with the connection ports
+	// TODO(dustmop): Add websocket.port to config, set that here
+	run.MustExec(t, "qri config set api.port 9871 rpc.port 9872")
```
b5: I'd be awesome if we could set these to `0`, which would act as a sentinel for "choose any available port". If I'm not mistaken feeding the empty string to `net.Listen()` has this behaviour.

cmd/connect_test.go
```diff
@@ -2,51 +2,42 @@ package cmd
...
+	run := NewTestRunner(t, "test_peer", "qri_test_connect")
+	defer run.Delete()
+
+	// Construct a mock registry to pass to the connect command
 	_, registryServer := regmock.NewMockServer()
 
-	path := filepath.Join(os.TempDir(), "qri_test_commands_connect")
-	t.Logf("temp path: %s", path)
+	// Configure ports such that other tests do not conflict with the connection ports
+	// TODO(dustmop): Add websocket.port to config, set that here
```
b5: there's a partial-implementation of this in https://github.com/qri-io/qri/pull/1076, I'd love to pull off the `--no-p2p` part of that PR & revisit it.

dustmop: I'll leave the TODO for now, but will definitely look into that to fix websockets soon.

Issue #1218, opened 2020-03-25
fetch doesn't return the commit message & title as history does


Arqu:
History:
```
[{
  versionInfo: { path: 'qmfoo', ... }
  commitMessage: 'foo',
  commitTitle: 'bar'
}]
```
Fetch:
`[{ 'path': 'qmfoo', ...}]`

dustmop:
Something I missed in the original review: in base/log.go, we have this:

```
type DatasetLogItem struct {
	// Decription of a dataset reference
	dsref.VersionInfo `json:"versionInfo,omitempty"`
    ....
```

If you get rid of the json annotation and change that part to:


```
type DatasetLogItem struct {
	// Decription of a dataset reference
	dsref.VersionInfo
    ....
```

then the api will instead output this json:

```
[{
  path: 'qmfoo',
  ...
  commitMessage: 'foo',
  commitTitle: 'bar'
}]
```

In other words it properly embeds the substructure into itself, instead of nesting it down a level. That makes it much easier to use in Desktop and to convert between DatasetLogItem and VersionInfo. This is a big part of the reason why I was saying we should have tests similar to `TestCheckoutAndRestore` that actually show what the json is expected to look like.

PR #1221, opened 2020-03-26
fix(fetch): fetch should return DatasetLogItems and added tests

Arqu:
WIP, amending PR with comments in the code.

Applies to issue:
https://github.com/qri-io/qri/issues/1218
Related PR:
https://github.com/qri-io/qri/pull/1187
Parent issue:
https://github.com/qri-io/qri/issues/1120

### DIFF
--- a/base/log.go
+++ b/base/log.go
@@ -7,49 +7,40 @@ import (
 	"github.com/qri-io/dataset"
 	"github.com/qri-io/qri/base/dsfs"
 	"github.com/qri-io/qri/dsref"
+	"github.com/qri-io/qri/logbook"
 	"github.com/qri-io/qri/repo"
 	reporef "github.com/qri-io/qri/repo/ref"
 )
 
-// DatasetLogItem is a line item in a dataset response
-type DatasetLogItem struct {
-	// Decription of a dataset reference
-	dsref.VersionInfo
-	// Title field from the commit
-	CommitTitle string `json:"commitTitle,omitempty"`
-	// Message field from the commit
-	CommitMessage string `json:"commitMessage,omitempty"`
-}
+// DatasetLogItem aliases the type from logbook
+type DatasetLogItem = logbook.DatasetLogItem
 
 // DatasetLog fetches the change version history of a dataset
 func DatasetLog(ctx context.Context, r repo.Repo, ref reporef.DatasetRef, limit, offset int, loadDatasets bool) ([]DatasetLogItem, error) {
 	if book := r.Logbook(); book != nil {
-		if versions, notes, err := book.VersionsWithNotes(ctx, reporef.ConvertToDsref(ref), offset, limit); err == nil {
-			items := make([]DatasetLogItem, len(versions))
+		if items, err := book.Items(ctx, reporef.ConvertToDsref(ref), offset, limit); err == nil {
 			// logs are ok with history not existing. This keeps FSI interaction behaviour consistent
 			// TODO (b5) - we should consider having "empty history" be an ok state, instead of marking as an error
-			if len(versions) == 0 {
+			if len(items) == 0 {
 				return nil, repo.ErrNoHistory
 			}
 			// Logbook doesn't store the CommitMessage and CommitTitle
 			// (see infoFromOp in logbook/logbook.go), so we need to load
 			// each dataset, and assign the CommitMessage and CommitTitle field.
-			for i, v := range versions {
-				if v.Path != "" {
-					local, err := r.Store().Has(ctx, v.Path)
+			for i, item := range items {
+				if item.Path != "" {
+					local, err := r.Store().Has(ctx, item.Path)
 					if err != nil {
 						continue
 					}
 					if local {
-						if ds, err := dsfs.LoadDataset(ctx, r.Store(), v.Path); err == nil {
+						if ds, err := dsfs.LoadDataset(ctx, r.Store(), item.Path); err == nil {
 							if ds.Commit != nil {
 								items[i].CommitMessage = ds.Commit.Message
 							}
 						}
 					}
-					versions[i].Foreign = !local
-					items[i].VersionInfo = versions[i]
-					items[i].CommitTitle = notes[i]
+					items[i].Foreign = !local
 				}
 			}
 			return items, nil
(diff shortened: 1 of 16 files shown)


base/log.go
```diff
@@ -44,12 +44,12 @@ func DatasetLog(ctx context.Context, r repo.Repo, ref reporef.DatasetRef, limit,
 						if ds, err := dsfs.LoadDataset(ctx, r.Store(), v.Path); err == nil {
 							if ds.Commit != nil {
 								items[i].CommitMessage = ds.Commit.Message
-								items[i].CommitTitle = ds.Commit.Title
 							}
 						}
 					}
 					versions[i].Foreign = !local
 					items[i].VersionInfo = versions[i]
+					items[i].CommitTitle = notes[i]
```
Arqu: For better fill rate of the `DatasetLogItem` we now rely on the `op.Log` to provide the title so this works well for foreign datasets as well.

Arqu: Also need to update the above comments as they are incorrect.

cmd/fetch.go
```diff
@@ -68,7 +67,7 @@ func (o *FetchOptions) Run() error {
 
 		items := make([]fmt.Stringer, len(res))
 		for i, r := range res {
-			items[i] = versionInfoStringer(r)
+			items[i] = versionInfoStringer(r.VersionInfo)
```
Arqu: I wanted to start a small discussion here @b5 @dustmop. Per the intention of the PR this should also be converted to a `DatasetLogItem` and use the `dsLogItemStringer` however that changes the output of what we tend to show with the versionInfo. Now we have a choice of whether to appropriate the `dsLogItemStringer` to show a more comprehensive output and similar to `versionInfoStringer`, agree to keep the current setup (which is not in line of what we wanted from fetch) or to accept the new output provided by the current implementation of `dsLogItemStringer`.

In either case, with this the difference between `qri log` and `qri fetch` is becoming non-existant as they both come by the same process from the logbook, just two different branches and now also share the same output.

cmd/fetch_test.go
```diff
@@ -146,7 +153,7 @@ func TestFetchCommand(t *testing.T) {
 	}
 
 	// Regex that replaces the timestamp with just static text
-	fixTs := regexp.MustCompile(`"timestamp":"[0-9TZ.:-]*"`)
+	fixTs := regexp.MustCompile(`"timestamp":"[0-9TZ.:+-]*"`)
```
Arqu: @dustmop just wanted to higlight that I've added the `+` here as the test was failing in my case having a `+01:00` at the end probably due to my TZ. Also need to update the tests to use this bellow as well.

logbook/logbook.go
```diff
@@ -805,6 +805,72 @@ func (book Book) Versions(ctx context.Context, ref dsref.Ref, offset, limit int)
 	return Versions(l, ref, offset, limit), nil
 }
 
+// VersionsWithNotes plays a set of operations for a given log, producing a State struct
+// that describes the current state of a dataset
+func (book Book) VersionsWithNotes(ctx context.Context, ref dsref.Ref, offset, limit int) ([]dsref.VersionInfo, []string, error) {
```
Arqu: I don't like these changes, but didn't manage to think of any other elegant way to handle passing the `op.Note's` up as `DatasetLogItem` doesn't belong this low, yet bubbling up the `op.Log` beats the purpose of having `logbook/logbook.go`.
Open to suggestions on this one.

cmd/fetch.go
Arqu: Also wondering whether we should also follow suit for `qri list` as VersionInfo is intended more as an internal structure and convert the output format to `DatasetLogItem` as well. Also `qri list` no longer sports `CommitTitle` and `CommitMessage` either.

logbook/logbook.go
dustmop: Ah, I see the troublesome situation. `VersionInfo` doesn't have the `commitTitle` and `commitMessage`, and we can't just rely on looking them up from the Dataset store, because for foreign data we won't find them. So we *have* to use the `commitTitle` (op.Note) from the logbook.

I think your first instinct is actually correct. We should move `DatasetLogItem` down to here. Looking at the original refactoring https://github.com/qri-io/qri/commit/4ce3eaa3228aa6f159bb27bbc1dbda7c6ab2c906 which switched `Versions()` over to returning `VersionInfo`: this file used to have a struct called a `DatasetInfo` which have the fields "Ref", "Published", "Timestamp", "CommitTitle", and "Size". That's basically everything stored in the logbook. That means to me we should be doing something similar with having a direct representation at this level. Here are the steps:

```
* move DatasetLogItem from base/ to logbook/
* rename infoFromOp -> itemFromOp
* Versions() returns a []DatasetLogItem
  * commitTitle is assigned, commitMessage will be added later
* rename the accumulator in Versions() to be items instead
* update all callers and tests
```

cmd/fetch.go
Arqu: @ramfox Just wanted to get your eye in this and see if it alignes with expectations?
So `qri fetch/log` will return the `DatasetLogItem` which contains the `CommitTitle` and `CommitMessage`. 
`qri list` (while unsure if it had it before) will be losing `CommitTitle` and `CommitMessage`.

cmd/fetch_test.go
```diff
@@ -124,29 +129,35 @@ func TestFetchCommand(t *testing.T) {
...
+	RemoteHost := fmt.Sprintf("http://localhost:%d", RemotePort)
+
 	// Assign peer A as a remote for peer B
-	cfgCmdText := fmt.Sprintf("qri config set remotes.a_node http://localhost:%d", RemotePort)
+	cfgCmdText := fmt.Sprintf("qri config set remotes.a_node %s", RemoteHost)
 	b.MustExec(t, cfgCmdText)
 
 	// Have peer B fetch from peer A, output correlates to the log from peer A earlier
 	actual = b.MustExec(t, "qri fetch peer_a/test_movies --remote a_node")
-	expect = `1   peer_a/test_movies
```
Arqu: @b5 @dustmop Note the change here. Just to make sure the new output is what is expected to face the user.
The dsLogStringer doesn't show `number of entries`, `number of errors` and whether a particular dataset is `foreign` and finally it doesn't show the `dataset name` at the start of each version.

cmd/fetch.go
ramfox: Got it! This works with desktop expectations. Thanks :)

cmd/fetch_test.go
dustmop: Normally, we shouldn't make UI changes in a refactoring such as this, but in this case it's the correct course of action. I went back to the commit before https://github.com/qri-io/qri/commit/4ce3eaa3228aa6f159bb27bbc1dbda7c6ab2c906 and checked what the output used to be, and validated that you're properly restoring it to what it was. My commit (https://github.com/qri-io/qri/commit/4ce3eaa3228aa6f159bb27bbc1dbda7c6ab2c906) broke it, but `cmd/fetch_test.go` didn't have any expectations that were looking at the actual command output, so that fact wasn't caught.

FYI, it sort of *does* show if the dataset is `foreign`, just in a different way. It says `Storage: remote`, which means the same thing.

Issue #1217, opened 2020-03-25
subsequent saves with same transform and meta result in a save instead of 'no changes'


chriswhong:
While writing some automation scripts, I noticed that saving again with the same command and `files` resulted in a new version with modified `body` and an automated commit message of `updated structure and transform`.

I expected a subsequent save to result in `no changes` and no new commit unless the CSV downloaded by the transform script changes.

Command: `qri save --file tmp/popular_baby_names/transform.star --file tmp/popular_baby_names/meta.json me/popular_baby_names`

`meta.json`

```
{
  "qri": "md:0",
  "identifier": "25th-nujf",
  "title": "Popular Baby Names",
  "theme": [
    "Health"
  ],
  "keywords": [
    "popular name",
    "birth name",
    "name",
    "health",
    "male",
    "female",
    "boy",
    "girl",
    "vital statistic",
    "race",
    "ethnicity",
    "2018od4a-video"
  ],
  "description": "Popular Baby Names by Sex and Ethnic Group\r\nData were collected through civil birth registration. Each record represents the ranking of a baby name in the order of frequency. Data can be used to represent the popularity of a name. Caution should be used when assessing the rank of a baby name if the frequency count is close to 10; the ranking may vary year to year.",
  "contributors": [
    {
      "name": "Department of Health and Mental Hygiene (DOHMH)"
    }
  ],
  "accessUrl": "https://data.cityofnewyork.us/api/views/25th-nujf/rows.csv?accessType=DOWNLOAD\n",
  "createdAt": 1382724894,
  "downloadCount": 102644,
  "rowsUpdatedAt": 1560022887,
  "accrualPeriodicity": "Biannually",
  "agency": "Department of Health and Mental Hygiene (DOHMH)"
}
```
`transform.star`:

```
load("http.star", "http")
load("encoding/csv.star", "csv")

# get the NYC Open Data Catalog as json
def download(ctx):
  csvDownloadUrl = "https://data.cityofnewyork.us/api/views/25th-nujf/rows.csv?accessType=DOWNLOAD"
  return http.get(csvDownloadUrl).body()

# set the body
def transform(ds, ctx):
  csv = ctx.download
  ds.set_body(csv, parse_as='csv')
  
```

b5:
closed by #1258

Issue #1215, opened 2020-03-24
Qri Save with new version of transform results in `error saving: no changes`


chriswhong:
Using 0.9.6-dev on CLI.  Ran `qri save --file catalog-transform.star  me/catalog` and successfully ran the transform script, creating a new dataset.  

Added some comments to the transform script file and attempted to save again: `qri save --file catalog-transform.star  me/catalog` and got `error saving: no changes`

I expect to be able to create a new version of the dataset with the same body but with a new version of the transform script.

dustmop:
This should be fixed by https://github.com/qri-io/qri/pull/1222, could you give it a shot when you get a chance, @chriswhong?

b5:
we were tinkering with this today, is this closable?

chriswhong:
Looks like I'm seeing `no changes` as expected.  Close it!

Issue #1150, opened 2020-03-04
qri status should have better performance


dustmop:
`qri status` is currently being calculated in a very inefficient manner, especially for the body component. Due to how `Compare` in base/component/kinds.go is implemented, both the prev body (from ipfs) and the curr body (from fsi) are opened, and read entry by entry until a difference is found. In addition, since incorporating websockets into the backend, Desktop is getting filesys write events and then sending a request to get status, which requires a network round-trip. All these slowdowns are adding up.

All we really need to do is to know when each component is different from what that component is in the current ipfs hash reference. However, this difference has to be checked against the structured data, not on a byte-by-byte basis.

Unfortunately, speeding this up can't take advantage of mtime nor filesize. mtime will always be greater than the initial value once a write event is seen by the filesys watcher. And filesize isn't sufficient due to needing the difference to be structured; adding or removing whitespace leads to different filesizes but no difference in structure.

Some suggestions on how to speed up `qri status`:
* First, we can prevent the round-trip that Desktop needs by sending the status across the Websocket. This alone should make a huge difference
* We don't need to read the whole file off of ipfs. Instead, we could store a checksum of the structured value, and then calculate the same of the fsi body. If they don't match we can be sure that the body is modified
* Taking this further, we could also store a checksum of every 1000 entries, or something like that. The moment the entry reader for the fsi file sees a mismatched checksum, it knows the body is modified. This might take a lot of memory, and won't help in the worst case (end of the body was changed, but the rest is the same)
* Once the file reaches a certain size, it might be pointless to checksum the whole thing. Instead, we could just assume that any write to the size is a modification, and only when a `save` or `restore` command is run should the body be considered clean.
* If we really want to experiment, it might be possible to do some OS specific tricks, and check inode values or read other filesystem specific data. I don't know if this will work at all, but I think that if a large file is modified, it's possible that sectors of the disc may be reused, making it trivial to know modifications have not occured. Unclear if this would actually work.

PR #1222, opened 2020-03-26
feat(save): Friendly commit messages by analyzing head, body, and transform

dustmop:
If a body is small enough, inline it before calling friendly. This is because friendly uses deepdiff, which is not necessarily fast enough for large bodies.

Read transform scripts to see if they changed between commits, letting us have a commit where the transform changes but body stays the same.

Drop derived values from structure before making the friendly message. This is currently less than ideal, since component needs some work to operate on maps.

Diff the head and body separately to get much better estimates for how much of a body changes.

Remove orphaned test private key in the old dataset tests, move it to testPeers.


dustmop:
Fixes https://github.com/qri-io/qri/issues/1217 and https://github.com/qri-io/qri/issues/1215.

dustmop:
Should also mention, this is the sequel to https://github.com/qri-io/qri/pull/1031.

dustmop:
Performance numbers for this feature. Note that the implementation cuts off body diffing at 20M, so we should expect larger save operations to not be affected.

I used the `.circleci/user_benchmarks.sh` script for this. Timing of `save` action, numbers are for "user","system","total", 4 runs each:

Before (master):

```
Small (10,000):
0.477 0.116 0.578
0.495 0.114 0.782
0.498 0.110 0.594
0.489 0.112 0.580

Medium (50,000):
2.167 0.339 1.952
2.135 0.331 2.010
2.117 0.315 1.886
2.107 0.319 1.911

Large (100,000):
5.050 0.736 4.320
4.240 0.595 3.606
4.290 0.608 3.643
4.294 0.603 3.638
```

After (friendly-commit):

```
Small (10,000):
0.689 0.138 0.796
0.684 0.131 0.783
0.686 0.139 0.799
0.676 0.134 0.791

Medium (50,000):
2.129 0.331 1.918
2.119 0.322 1.892
2.171 0.344 1.950
2.141 0.332 1.920

Large (100,000):
4.584 0.665 3.949
4.489 0.678 3.851
4.380 0.640 3.788
4.290 0.628 3.666
```

I also tried increasing the limit to 200M to see how it affects saving, but only ran 1 time not 4:

```
Small (10,000):
0.682 0.143 0.806

Medium (50,000):
3.254 0.447 3.066

Large (100,000):
6.594 0.956 6.115
```

We can see that including body diffing increasing the time to save by about 50%. I would argue that this is acceptable for small bodies (below 20M) as done in the implementation, but probably too slow after that.

base/dsfs/dataset.go
```diff
@@ -21,10 +22,16 @@ import (
...
 	"github.com/qri-io/qfs/cafs"
+	"github.com/qri-io/qfs/localfs"
 	"github.com/qri-io/qri/base/friendly"
 	"github.com/qri-io/qri/base/toqtype"
 )
 
+var (
+	// BodySizeSmallEnoughToDiff sets how small a body must be to generate a message from it
+	BodySizeSmallEnoughToDiff = 20000000 // 20M or less is small
+)
```
b5: skip parens if only one variable

base/dsfs/dataset.go
```diff
@@ -241,17 +248,30 @@ var Timestamp = func() time.Time {
...
+// BodyAction represents the action that should be taken to understand how the body changed
+type BodyAction string
+
+const (
+	// BodyDefault is the default action: compare them to get how much changed
+	BodyDefault = BodyAction("default")
+	// BodySame means that the bodies are the same, no need to compare
+	BodySame = BodyAction("same")
+	// BodyTooBig means the body is too big to compare, and should be assumed to have changed
+	BodyTooBig = BodyAction("too_big")
```
b5: fan of this. Helps me think about FSI & transform script interactions where a file exists on disk that we don't want to use if unchanged. Don't think we can solve that here, but enumerating what actions we can take is a nice technique

base/dsfs/dataset.go
```diff
@@ -470,16 +509,49 @@ func setChecksumAndLength(ds *dataset.Dataset, data qfs.File, buf *bytes.Buffer,
...
+	if prev.Transform != nil && prev.Transform.ScriptPath != "" {
+		err := prev.Transform.OpenScriptFile(ctx, store)
+		if err == nil {
+			tfFile := prev.Transform.ScriptFile()
+			prev.Transform.ScriptBytes, _ = ioutil.ReadAll(tfFile)
+		}
+	}
+	if ds.Transform != nil && ds.Transform.ScriptPath != "" {
+		// TODO(dustmop): The ipfs filestore won't recognize local filepaths, we need to use
+		// local here. Is there some way to have a cafs store that works with both?
```
b5: yes. Thats covered in the interface that's intended to replace `cafs.Filestore`. 

`qfs.Filesystem` has a `Mux` implementation for exactly this: https://github.com/qri-io/qfs/blob/a6a7b98f0fa0969f6eae50fe14c2c6b5b32f4b4f/mux.go#L13-L17

To get there we'd need to do a _lot_ of refactoring.

base/dsfs/dataset.go
```diff
@@ -509,6 +634,37 @@ func generateCommitDescriptions(prev, ds *dataset.Dataset, force bool) (short, l
 	return shortTitle, longMessage, nil
 }
 
+// copied from base/body.go to avoid circular dependency
```
b5: innnnnnteresting. we should think about reversing the flow of this, moving `ReadAllEntries` into dsfs

base/dsfs/dataset_test.go
```diff
@@ -2,34 +2,25 @@ package dsfs
...
...cBels3nm/fWScvjH4dMQXlavwcwPgKhy2NczDhr4X69oEw6Msd4hQiqJrlWd8juUg6vIsrl1wS/JAOCS65fuyJfV3Pw64rWbTPMwO3FOvxj+rFghZFQgjg/i45uHA2UUkM+h504M5Nzs6Arr/rgV7uPGR5e5OBw3lfiS9ZaA7QZiOq7sMy1L0qD49YO1ojqWu3b7UaMaBQx1Dty7b5IVOSYG+Y3U/dLjhTj4Hg1VtCHWRm3nMOE9cVpMJRhRzKhkq6gnZmni8obz2BBDF02X34oQLcHC/Wn8F3E8RiBjZDI66g+iZeCCUXvYz0vxWAQQKBgQDEJu6flyHPvyBPAC4EOxZAw0zh6SF/r8VgjbKO3n/8d+kZJeVmYnbsLodIEEyXQnr35o2CLqhCvR2kstsRSfRz79nMIt6aPWuwYkXNHQGE8rnCxxyJmxV4S63GczLk7SIn4KmqPlCI08AU0TXJS3zwh7O6e6kBljjPt1mnMgvr3QKBgQD6fAkdI0FRZSXwzygx4uSg47Co6X6ESZ9FDf6ph63lvSK5/eue/ugX6p/olMYq5CHXbLpgM4EJYdRfrH6pwqtBwUJhlh1xI6C48nonnw+oh8YPlFCDLxNG4tq6JVo071qH6CFXCIank3ThZeW5a3ZSe5pBZ8h4bUZ9H8pJL4C7yQKBgFb8SN/+/qCJSoOeOcnohhLMSSD56MAeK7KIxAF1jF5isr1TP+rqiYBtldKQX9bIRY3/8QslM7r88NNj+aAuIrjzSausXvkZedMrkXbHgS/7EAPflrkzTA8fyH10AsLgoj/68mKr5bz34nuY13hgAJUOKNbvFeC9RI5g6eIqYH0FAoGAVqFTXZp12rrK1nAvDKHWRLa6wJCQyxvTU8S1UNi2EgDJ492oAgNTLgJdb8kUiH0CH0lhZCgr9py5IKW94OSM6l72oF2UrS6PRafHC7D9b2IV5Al9lwFO/3MyBrMocapeeyaTcVBnkclz4Qim3OwHrhtFjF1ifhP9DwVRpuIg+dECgYANwlHxLe//tr6BM31PUUrOxP5Y/cj+ydxqM/z6papZFkK6Mvi/vMQQNQkh95GH9zqyC5Z/yLxur4ry1eNYty/9FnuZRAkEmlUSZ/DobhU0Pmj8Hep6JsTuMutref6vCk2n02jc9qYmJuD7iXkdXDSawbEG6f5C4MUkJ38z1t1OjA==`)
```
b5: nice pickup

base/dsfs/structure.go
```diff
@@ -8,6 +8,15 @@ import (
 	"github.com/qri-io/qfs/cafs"
 )
 
+// BaseTabularSchema is the base schema for tabular data
```
b5: When we start supporting object-of-equal-length-array tables as a format, this'll explode..

We should add a note here that this shouldn't be used. Instead favor methods in `github.com/qri-io/dataset/tabular`. 

The even better thing to do would be to move this into the `tabular` package (not in this PR, some other time) and rename it to `BaseSchemaArray`, like we do with JSON identity schemas

base/friendly/friendly.go
```diff
@@ -63,19 +72,15 @@ func DiffDescriptions(deltas []*deepdiff.Delta, stats *deepdiff.Stats) (string,
...
 					// it for both the long message and short title.
 					msg = fmt.Sprintf("%s:\n\t%s", compName, changes.Rows[0])
 					shortTitle = fmt.Sprintf("%s %s", compName, changes.Rows[0])
-				} else if len(changes.Rows) == 0 && compName == "transform" {
-					// TODO (b5) - this is a hack to make TestSaveTransformWithoutChanges
-					// in the cmd package pass. We're getting to this stage with 0 rows of
-					// changes, which is making msg & title not-empty, which is in turn allowing
-					// a commit b/c it looks like a change. ideally we don't make it here at all,
-					// but we DEFINITELY need a better hueristic for dsfs.CreateDataset's
-					// change detection check. Maybe use diffstat?
```
b5: yessssss

base/dsfs/dataset.go
dustmop: I agree, will do in a follow-up change.

base/dsfs/structure.go
dustmop: I used to be able to use dataset's BaseSchemaArray, but some tests were failing that "items" needed to exist at the top-level. I think some lower codepath has started relying on arrays always being 2-dimensional?

dustmop:
Something I forgot to mention, that I added as a TODO and am also copying here. Limiting the body size for diffs causes an undesirable side-effect: if a user has a dataset body larger than this size, then any `save` operation will create a commit, even if no changes have been made, since our change detection will see BodyTooBig and assume something changed. Some solutions to this:
* above 20M start using a less-accurate non-structured checksum instead of deepdiff
* calculate a structured checksum and compare that to the last commit's checksum
* immitate "redo" (https://apenwarr.ca/log/20181113), store file attributes, see if any change
* at a large enough size, like 4G, just assume the file is always changed. don't be expensive
We should make this algorithm agree with how `status` works. See issue: https://github.com/qri-io/qri/issues/1150

b5:
hrm... I didn't catch the proviso for files over `BigFile`. sounds like a job for a rolling hash function.

PR #1223, opened 2020-03-27
fix(save): Infered name must start with a letter

dustmop:


dustmop:
Fixes https://github.com/qri-io/desktop/issues/500.

PR #1228, opened 2020-03-31
refactor(fsi): Remove fsi=true parameter, use FSIPath everywhere

dustmop:
Change semantics of lib operations so that, instead of looking for a UseFSI parameter, they see if a reference has an FSIPath. If so, they follow fsi semantics, otherwise refstore. Removes the need to have operations return ErrNoLink if UseFSI is requested but no FSIPath exists.

Most instances of checking for ErrNoHistory now disappear (except for `/history`) since only working directories that have been freshly `init`d will not have history, and now they'll just follow their FSIPath.

It is now illegal to use `status` to check the historical commits of a dataset. Callers should use `whatchanged` instead.

The "fsi=true" api parameter is now non-functional. It now not illegal to send requests with it, it just doesn't do anything.


dustmop:
Shouldn't merge this one until Desktop is updated, to avoid breaking functionality. @ramfox, the calls to `/status` on the "history" pane should call `/whatchanged` instead. For the FSI-powered "status" pane, you should still use `/status`.

ramfox:
So pumped we are leaving fsi-true-false land!

Can you confirm some of my assumptions:

1) If the dataset is not fsi linked, and you ask for HEAD, it will return the latest commit
2) If the dataset IS fsi linked, and you ask for HEAD, it will return the contents of the linked folder
3) Regardless of whether or not the dataset is fsi linked, if you ask for the dataset at a specific path you will get the contents of that commit.

Trying to understand what logic we can drop on frontend!

You basically said this but want to confirm: Does `status` now only makes sense in the context of fsi? Is this also true for any dataset at HEAD? If I have an non-linked dataset and I ask for status at `peername/name`, is that an error? Again, trying to figure out where we still need to keep track of if a dataset is fsi true or not.

base/ref.go
```diff
@@ -32,9 +32,9 @@ func SetPublishStatus(r repo.Repo, ref *reporef.DatasetRef, published bool) erro
 	return r.PutRef(*ref)
 }
 
-// ToDatasetRef parses the dataset ref and returns it, allowing datasets with no history only
-// if FSI is enabled.
-func ToDatasetRef(path string, r repo.Repo, allowFSI bool) (*reporef.DatasetRef, error) {
+// ToDatasetRef parses the dataset ref and looks it up in the refstore, allows refs with no history
+// TODO(dustmop): In a future change, remove the third parameter from this function
+func ToDatasetRef(path string, r repo.Repo, _ bool) (*reporef.DatasetRef, error) {
```
b5: My only complaint here: we're increasing our reliance on `reporef.DatasetRef`, when we should be transitioning towards `dsref.Ref`. That struct doesn't have a `FSIPath` field, and I'd like to keep it that way. Instead I think we should use this function to process the filepath, and store it in the `Path` field with a prefix that indicates it's an FSI path. Off the top of my head, we could use `file://` for this. 

The long term solution for this whole problem is to use a muxed `qfs.Filesystem` interface that knows how to deal with both local & IPFS paths, further removing the need for `if` statements that redirect to FSI, but that's a _long_ ways off.

This PR is still an invaluable intermediate step that it's not worth delaying on, We should follow up in a future PR that switches the return value.

dustmop:
> So pumped we are leaving fsi-true-false land!
> 
> Can you confirm some of my assumptions:

Excellent questions! This is a good level of detail we should be operating out in order to hash out how this stuff works. Here we good.

> 
> 1. If the dataset is not fsi linked, and you ask for HEAD, it will return the latest commit

True.

> 2. If the dataset IS fsi linked, and you ask for HEAD, it will return the contents of the linked folder

True.

> 3. Regardless of whether or not the dataset is fsi linked, if you ask for the dataset at a specific path you will get the contents of that commit.

I should double-check that this works as described, I'm not 100% sure that asking for a specific path will work correctly if the reference is fsi linked. It *should* work that way, right?

> Trying to understand what logic we can drop on frontend!
> 
> You basically said this but want to confirm: Does `status` now only makes sense in the context of fsi?

Yes.

> Is this also true for any dataset at HEAD?

Not sure I completely follow, but maybe this is the answer: you can only ask for `status` at HEAD for a given reference, asking for `status` with a path is an error.

If you want to ask what changed at an older commit, which used to be `status` with a path, now use `/whatchanged`.

Rational: we can't rely on the existence of a "path" to determine if we're asking for `status` or `whatchanged`, because using "HEAD" should behave equivalently to being given the explicit path for "HEAD" (in order for CanonicalizeDatasetRef to have rational semantics). We originally created the `fsi=true` parameter to determine this behavior, but really they should just be separate commands.

> If I have an non-linked dataset and I ask for status at `peername/name`, is that an error? 

Yes, correct.

> Again, trying to figure out where we still need to keep track of if a dataset is fsi true or not.

Hopefully we can remove as much of that logic as possible. Ideally, fsi-ness should only affect UI, not which api calls need to be made.

base/ref.go
dustmop: The true fix I think is to switch to dscache, and get rid of base.ToDatasetRef entirely. I have a PR waiting in the wings that starts that change. For now, this PR doesn't increase the usage of base.ToDatasetRef, only remove its third parameter.

Having a muxed qfs.Filesystem would be good, as noted in https://github.com/qri-io/qri/pull/1222.

ramfox:
issue tracking desktop refactor to match this API: https://github.com/qri-io/desktop/issues/511

dustmop:
> Regardless of whether or not the dataset is fsi linked, if you ask for the dataset at a specific path you will get the contents of that commit.

Pushed a new commit, this is fixed now, made a new test that verifies this behavior.

Issue #1227, opened 2020-03-30
RPC/EOF errors should be handled better and respond with human readable messages


Arqu:
We leverage RPC to interact from the command line with a running qri node. However, due to `net/rpc` having poor error handling, typically on most error scenarios it responds with non-serializable responses or simply returns `EOF`. One of the common sources is that some commands either fail when `qri connect` is running but shouldn't or are not meant to be run against a live node. In either case there should be an appropriate response so users can troubleshoot.

Issue #813, opened 2019-06-24
dry-run save transform while connected produces error


dustmop:
Discovered by myself and @ramfox late last week, together we tracked down the problem to being connected and then running:
`qri save --file=code.star me/my_dataset --dry-run`

Gives the error:
`reading body EOF`

Running the same while not connected works okay.

Arqu:
This still hangs, issue was introduced here: https://github.com/qri-io/dataset/pull/169
In `dataset.go` the struct `Dataset` has  a field `bodyFile qfs.File` which isn't serializable. Quite possibly `Body interface{}` is problematic as well as `net/rpc` doesn't generally handle vague interfaces well.

A similar issue was tracked here: https://github.com/qri-io/qri/issues/1193 and it's respective PR https://github.com/qri-io/qri/pull/1226

Arqu:
Similar issue here: https://github.com/qri-io/qri/blob/master/lib/update.go#L73

dustmop:
Removing `qfs.File` is not an option, it's pretty ingrained in our code. Instead, we should see if there's a way to make this work by using `gob.Register`, like in the `init` function in `lib/lib.go`.

Arqu:
Yes, and the other alternative being going to our `Pod (Plain Old Data)` objects.

b5:
yet another alternative: use the JSON API in these cases, which has none of these problems.

Arqu:
I generally tried to avoid having a converter from RPC to API calls, or at least do it separately. 
But yeah that would solve this.

Arqu:
Turns out, qfs.File is fine, it was the `json.RawMessage` from the `p.ReturnBody` as that actually abstracts the response from the raw `DatasetRef` and we don't have to deal with qfs at all.

PR #1229, opened 2020-03-31
fix(rpc): better error messages on RPC EOF

Arqu:
https://github.com/qri-io/qri/issues/1227

### DIFF
--- a/lib/lib.go
+++ b/lib/lib.go
@@ -748,8 +748,11 @@ func (inst *Instance) Teardown() {
 // checkRPCError validates RPC errors and in case of EOF returns a
 // more user friendly message
 func checkRPCError(err error) error {
+	if err == nil {
+		return nil
+	}
 	if strings.Contains(err.Error(), "EOF") {
-		msg := `Qri wasn't able to execute your command. 
+		msg := `Qri couldn't parse the response and is unsure if it was successful. 
 It is possible you have a Qri node running or the Desktop app is open.
 Try closing them and running the command again.
 Check our issue tracker for RPC issues & feature requests:
--- a/lib/lib.go
+++ b/lib/lib.go
@@ -23,6 +23,7 @@ import (
 	"github.com/qri-io/qri/config"
 	"github.com/qri-io/qri/config/migrate"
 	"github.com/qri-io/qri/dscache"
+	qrierr "github.com/qri-io/qri/errors"
 	"github.com/qri-io/qri/event"
 	"github.com/qri-io/qri/fsi"
 	"github.com/qri-io/qri/logbook"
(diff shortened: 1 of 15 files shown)


lib/update.go
```diff
@@ -70,7 +70,7 @@ func (m *UpdateMethods) Schedule(in *ScheduleParams, out *cron.Job) (err error)
 		return fmt.Errorf("apologies, we currently can't schedule updates while `qri connect` is running")
 		// TODO (b5) - this returns error EOF when run. need to figure out why. I suspect it's b/c
 		// job contains an interface field: Options
-		// return m.inst.rpc.Call("UpdateMethods.Schedule", in, out)
+		// return checkRPCError(m.inst.rpc.Call("UpdateMethods.Schedule", in, out))
```
Arqu: Will investigate this as part of https://github.com/qri-io/qri/issues/813

lib/lib.go
```diff
@@ -743,3 +744,20 @@ func (inst *Instance) RemoteClient() remote.Client {
 func (inst *Instance) Teardown() {
 	inst.teardown()
 }
+
+// checkRPCError validates RPC errors and in case of EOF returns a
+// more user friendly message
+func checkRPCError(err error) error {
+	if strings.Contains(err.Error(), "EOF") {
```
b5: err will often be `nil` here when it works, and will panic when the `Error()` method is called on a nil value. Adding this above will fix:

```go
if err == nil {
  return nil
}
```

lib/lib.go
```diff
@@ -743,3 +744,20 @@ func (inst *Instance) RemoteClient() remote.Client {
...
+		msg := `Qri wasn't able to execute your command. 
+It is possible you have a Qri node running or the Desktop app is open.
+Try closing them and running the command again.
+Check our issue tracker for RPC issues & feature requests:
+  https://github.com/qri-io/qri/issues?q=is:issue+label:RPC
+
+Error:
+%s`
```
b5: Language of this looks great. 

I just kinda hacked together this convention in the SQL pr, does it work for you? More info we could include?

Arqu: Yeah I saw that back in the PR and thought it was a bit ellaborate but very much user friendly. 
Just a couple bits on the language itself:
Qri actually does execute the command but couldn't parse the response (and we don't know if it was successful or not)! maybe should change to this.
Other than that, I'd just like more "specific" error handling and not just a blanket RPC error. Though that can come over time.

b5: totally. At least this gives clear feedback. We can workshop the friendly error format as time goes on 😄

lib/lib.go
b5: very sad to see this didn't crash the test suite anywhere.

Arqu: Yeah, thing is I thought adding a whole test suite for the RPCs against a running qri node. But thats quite a substantial bit of work and not really sure if it's worth the time if we're going to move away from it anyways.

b5: completely agreed. I'd _much_ rather see that energy go into refactoring our api package tests

Issue #1179, opened 2020-03-12
`qri peers list` does not work offline


Mr0grog:
The help text for `qri peers list` claims it does not work offline, but the code tries to function offline anyhow, and it prints a message implying that should be OK:

```sh
$ qri peers list
qri not connected, listing cached peers
error: not connected, run `qri connect` in another window
```

Per @b5 ([on Discord](https://discordapp.com/channels/497052800778502152/571427701475377152/687705661001302039)), it *should* work offline. We also need to change the help text to indicate you don’t have to be online for it to work.

PR #1234, opened 2020-03-31
fix(cmd): pass node when not making online request for peers

Arqu:
Fixes https://github.com/qri-io/qri/issues/1179

The issue was when running 'offline' we will neither trigger an RPC call nor do we pass the instance/Node along which always drops the request due to 'not being connected'. In reality we want to check against the instance created by the `qri` command if we're not going to query against a live `qri connect` session and execute an RPC call if there is. RPC always takes precedence vs command local execution.

### DIFF
--- a/lib/peers.go
+++ b/lib/peers.go
@@ -30,9 +30,6 @@ func (d PeerRequests) CoreRequestsName() string { return "peers" }
 // NewPeerRequests creates a PeerRequests pointer from either a
 // qri Node or an rpc.Client
 func NewPeerRequests(node *p2p.QriNode, cli *rpc.Client) *PeerRequests {
-	if node != nil && cli != nil {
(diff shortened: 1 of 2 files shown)


lib/peers.go
```diff
@@ -30,10 +30,7 @@ func (d PeerRequests) CoreRequestsName() string { return "peers" }
 // NewPeerRequests creates a PeerRequests pointer from either a
 // qri Node or an rpc.Client
 func NewPeerRequests(node *p2p.QriNode, cli *rpc.Client) *PeerRequests {
-	if node != nil && cli != nil {
```
Arqu: This check is not needed as it will never conflict. In `PeerRequests.List` we always check if `cli` is present in which case an RPC request is fired and thus result returned and function terminated.

PR #1238, opened 2020-04-01
refactor(test): test requires stable output

Arqu:
This addresses these:
```
--- FAIL: TestGetPreviousVersionExplicitPath (0.33s)
panic: expected output to start with "saved:", got "" [recovered]
        panic: expected output to start with "saved:", got ""

goroutine 5175 [running]:
testing.tRunner.func1.1(0x542e2a0, 0xc0513bbef0)
        /usr/local/Cellar/go/1.14/libexec/src/testing/testing.go:941 +0x3d0
testing.tRunner.func1(0xc050648900)
        /usr/local/Cellar/go/1.14/libexec/src/testing/testing.go:944 +0x3f9
panic(0x542e2a0, 0xc0513bbef0)
        /usr/local/Cellar/go/1.14/libexec/src/runtime/panic.go:967 +0x15d
github.com/qri-io/qri/cmd.parseRefFromSave(0x0, 0x0, 0x5704723, 0x3d)
        /Users/arqu/dev/qri/qri/cmd/fsi_integration_test.go:1656 +0x142
github.com/qri-io/qri/cmd.TestGetPreviousVersionExplicitPath(0xc050648900)
        /Users/arqu/dev/qri/qri/cmd/fsi_integration_test.go:766 +0x12c
testing.tRunner(0xc050648900, 0x586d840)
        /usr/local/Cellar/go/1.14/libexec/src/testing/testing.go:992 +0xdc
created by testing.(*T).Run
        /usr/local/Cellar/go/1.14/libexec/src/testing/testing.go:1043 +0x357
```

### DIFF
--- a/cmd/fsi_integration_test.go
+++ b/cmd/fsi_integration_test.go
@@ -758,19 +758,19 @@ func TestGetPreviousVersionExplicitPath(t *testing.T) {
 	defer run.Delete()
 
 	// First version has only a body
-	output := run.MustExec(t, "qri save --body=testdata/movies/body_two.json me/get_ver")
+	output := run.MustExecCombinedOutErr(t, "qri save --body=testdata/movies/body_two.json me/get_ver")
 	ref1 := parseRefFromSave(output)
 
 	// Add a meta
-	output = run.MustExec(t, "qri save --file=testdata/movies/meta_override.yaml me/get_ver")
+	output = run.MustExecCombinedOutErr(t, "qri save --file=testdata/movies/meta_override.yaml me/get_ver")
 	_ = parseRefFromSave(output)
 
 	// Modify the body
-	output = run.MustExec(t, "qri save --body=testdata/movies/body_four.json me/get_ver")
+	output = run.MustExecCombinedOutErr(t, "qri save --body=testdata/movies/body_four.json me/get_ver")
 	ref3 := parseRefFromSave(output)
 
 	// Change the meta
-	output = run.MustExec(t, "qri save --file=testdata/movies/meta_another.yaml me/get_ver")
+	output = run.MustExecCombinedOutErr(t, "qri save --file=testdata/movies/meta_another.yaml me/get_ver")
 	_ = parseRefFromSave(output)
 
 	run.ChdirToRoot()


Arqu:
Another note is that I get a flaky test since the big refactor (seems to double execute some steps?) on:
```
--- FAIL: TestDatasetPullPushDeleteFeedsPreviewHTTP (0.05s)
    remote_test.go:135: result mismatch (-want +got):
          []string{
              ... // 6 identical elements
              "DatasetPushPreCheck",
              "DatasetPushFinalCheck",
        +      "DatasetPushFinalCheck",
        +      "DatasetPushed",
              "DatasetPushed",
              "LogRemovePreCheck",
              ... // 4 identical elements
          }
```

cmd/fsi_integration_test.go
```diff
@@ -758,19 +758,19 @@ func TestGetPreviousVersionExplicitPath(t *testing.T) {
 	defer run.Delete()
 
 	// First version has only a body
-	output := run.MustExec(t, "qri save --body=testdata/movies/body_two.json me/get_ver")
+	output := run.MustExecCombinedOutErr(t, "qri save --body=testdata/movies/body_two.json me/get_ver")
```
Arqu, line 761: Also not even sure this is the correct approach, but got me over the fence pretty quickly.

b5: yep I think this is correct. I had to do the same in my branch

dustmop:
Whoops! Looks like https://github.com/qri-io/qri/pull/1228 and https://github.com/qri-io/qri/pull/1237 didn't commute properly. The former added a test that the latter caused to fail.

Seeing `TestDatasetPullPushDeleteFeedsPreviewHTTP` fail is worrying, that was supposed to be fixed by https://github.com/qri-io/dag/pull/28. I don't understand how that is still broken, unless you're using an old version of the `dag` module?

Arqu:
Looking at `go.sum` I get `github.com/qri-io/dag v0.2.1-0.20191025201336-254aa177fbd7 h1:hujyzdufAz3Hi6I0I0yKyJTLCM9eoLamKhlpwaqE7ec=` which should be fine. 
I was just guessing on the issue, but seems like `run.MustExec` doesn't properly return the output (or there's something else wrong and things get output on the `Err` stream instead of `Out`)

dustmop:
`run.MustExec` used to return the combined stdout and stderr, but now returns only stdout. There was some hacky black magic happening in TestRunner that caused this to happen (by accident) while also piping only stderr to another variable on the TestRunner.RepoRoot. I cleaned that up with https://github.com/qri-io/qri/pull/1237, so now tests that require combined stdout and stderr need to be explicit about it. Your fix for `TestGetPreviousVersionExplicitPath` was correct; it was added by a different PR that was merged simultaneously. For some reason `qri save` writes its "saved" message to stderr, which seems like a bug to me. This was quite a hassle to untangle, but is better off for it now.

That flakey test is weird, though. There must be a second race condition that was missed on the first attempt.

Arqu:
For what it's worth, a common "abuse" of the stdErr stream is to separate what gets piped to other commands while keeping the rest invisible.

dustmop:
Agreed, we do this for FSI when printing `for linked dataset [me/my_ds]` so that things like `get` will correctly pipe while this "info" message goes to stderr. But that's why I think `save` should go to stdout instead of stderr, it should operate like get, as it doesn't output anything *except* the full path of the dataset that got written.

Arqu:
Yeah. This is the offending line:
https://github.com/qri-io/qri/blob/master/cmd/save.go#L191
Should I just update it to `o.Out` or should I also change the message to output just the `DatasetRef.String()`?

dustmop:
Hold off on it for now. Want to look into it a bit, discuss, and make sure we're not missing something.

PR #1239, opened 2020-04-01
fix(starlark): various cleanups to starlark runtime

b5:
**fix(dry-run): dry-runs must never add to the refstore**
fixed this by unwrapping an extraneous call to load a newly saved dataset, shedding a
canonicalization call. I've Also added a number of guards within lib.DatasetRequests.Save to prevent dry-run from messing everything up.


cmd/fsi_integration_test.go
```diff
@@ -758,19 +758,19 @@ func TestGetPreviousVersionExplicitPath(t *testing.T) {
 	defer run.Delete()
 
 	// First version has only a body
-	output := run.MustExec(t, "qri save --body=testdata/movies/body_two.json me/get_ver")
+	output := run.MustExecCombinedOutErr(t, "qri save --body=testdata/movies/body_two.json me/get_ver")
```
dustmop, line 761: These were fixed by https://github.com/qri-io/qri/pull/1238. Rebase against master.

base/save.go
```diff
@@ -199,32 +199,31 @@ func CreateDataset(ctx context.Context, r repo.Repo, streams ioes.IOStreams, ds,
 		Path:      path,
 	}
```
dustmop, line 201: Just making sure this is correct. If dry-run is true, we still call dsfs.CreateDataset, right? So the blocks are added to ipfs, but we never make a reference to them?

b5: yes. Earlier in the codepath we pass down a `MemStore` instead of the IPFS store, which is the source of the (bad for storing) `/map/` prefix in the CID. So we are calling CreateDataset, but on a throwaway repo

dustmop: Oh of course. 👍 great!

Issue #1236, opened 2020-03-31
Autocomplete on CLI would be very handy


Arqu:
Navigating the CLI can be tough sometimes, especially on dataset names.

Arqu:
Since this issue is not complete, bellow is a list of items that are done and things that are still todo.

No dependencies:
- [x] qri connect
- [x] qri init (intentionally left manual)
- [x] qri registry (intentionally left manual)

Depends on qri search:
- [x] qri fetch
- [x] qri add
- [x] qri search

Depends on qri list:
- [x] qri render
- [x] qri stats
- [x] qri logbook
- [x] qri validate
- [x] qri save
- [x] qri export
- [x] qri use
- [x] qri rename
- [x] qri remove
- [x] qri publish
- [x] qri log
- [x] qri checkout
- [x] qri get

Todo (currently blocking):
- [ ] qri config (config args)
- [ ] qri get (structure args)
- [ ] qri fetch (—remote)
- [ ] qri publish (—remote)
- [ ] qri diff (combined completion components, datasets and files)
- [ ] qri restore? (versions, structure args)
- [ ] qri list (peers)
- [ ] qri dag (hidden)
- [ ] qri fsi (hidden) 
- [ ] qri (--repo)
- [ ] qri sql (completely different type of autocomplete)
- [ ] qri update (complex)
- [ ] qri peers (complex)
- [ ] qri status (status currently does not behave as expected, additionally it depends on full dataset+location completions)

feep:
Very nice.

One vote for fish support.

https://fishshell.com/docs/current/index.html#completion-own

Arqu:
Bad news, fish isn't yet supported with cobra (which is what we use underneeth the CLI)
The GREAT news is that this is happening: https://github.com/spf13/cobra/pull/1048
Once that's in, I'll follow up on this and try to incorporate fish as well.

feep:
It is (spf13/cobra#1048) merged.

Let me know when the package hits HEAD, I’ll let you know how it works in fish.

Arqu:
Sweet, I'll try to get around to it this week. Will ping back here once it's in!

feep:
Busy right now (and, there’s absolutely no hurry). I’ll make a note to test.

Eh, building, cause I know I’ll forget.

Thanks.

feep:
I presume this help message `$ source <(qri completion [bash|zsh|fish])` is from upstream?

feep:
Works for the subset that you say works.

[...]

PR #1240, opened 2020-04-01
feat(cmd): added command to generate basic cmd autocompletion scripts

Arqu:
Relevant to https://github.com/qri-io/qri/issues/1236

Took a mix of inspiration between `cobra` docs and `kubectl` handling of `zsh` support.
Currently it only handles registered commands and flags, no support for dataset name autocomplete just yet. 
The convoluted approach for `zsh` is due to `cobra` not supporting "advanced/custom" autocomplete functions and thus would not work for the future dataset name autocompletion (or similar scenarios like peers etc).

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -1,0 +1,219 @@
+package cmd
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/qri-io/ioes"
+	"github.com/spf13/cobra"
+)
+
+// NewAutocompleteCommand creates a new `qri complete` cobra command that prints autocomplete scripts
+func NewAutocompleteCommand(_ Factory, ioStreams ioes.IOStreams) *cobra.Command {
+	o := &AutocompleteOptions{IOStreams: ioStreams}
+	cmd := &cobra.Command{
+		Use:   "completion [bash|zsh]",
+		Short: "Generates shell completion scripts",
+		Long: `To load completion run
+
+. <(qri completion [bash|zsh])
+
+To configure your bash/zsh shell to load completions for each session add to your bashrc/zshrc
+
+# ~/.bashrc or ~/.zshrc
+. <(qri completion [bash|zsh])
+`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return o.Run(cmd, args)
+		},
+		ValidArgs: []string{"bash", "zsh"},
+	}
+
+	return cmd
+}
+
+// AutocompleteOptions encapsulates completion options
+type AutocompleteOptions struct {
+	ioes.IOStreams
+}
+
+// Run executes the completion command
+func (o *AutocompleteOptions) Run(cmd *cobra.Command, args []string) (err error) {
+	if len(args) == 0 {
+		return fmt.Errorf("shell not specified")
+	}
+	if len(args) > 1 {
+		return fmt.Errorf("too many arguments, expected only the shell type")
+	}
+	if args[0] == "bash" {
+		cmd.Parent().GenBashCompletion(o.Out)
+	}
+	if args[0] == "zsh" {
+		zshHead := `# reference kubectl completion zsh
+
+__qri_bash_source() {
+	alias shopt=':'
+	alias _expand=_bash_expand
+	alias _complete=_bash_comp
+	emulate -L sh
+	setopt kshglob noshglob braceexpand
+
+	source "$@"
+}
+
+__qri_type() {
+	# -t is not supported by zsh
+	if [ "$1" == "-t" ]; then
+		shift
+
+		# fake Bash 4 to disable "complete -o nospace". Instead
+		# "compopt +-o nospace" is used in the code to toggle trailing
+		# spaces. We don't support that, but leave trailing spaces on
+		# all the time
+		if [ "$1" = "__qri_compopt" ]; then
+			echo builtin
+			return 0
+		fi
+	fi
+	type "$@"
+}
+
+__qri_compgen() {
+	local completions w
+	completions=( $(compgen "$@") ) || return $?
+
+	# filter by given word as prefix
+	while [[ "$1" = -* && "$1" != -- ]]; do
+		shift
+		shift
+	done
+	if [[ "$1" == -- ]]; then
+		shift
+	fi
+	for w in "${completions[@]}"; do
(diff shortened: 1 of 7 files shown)


Arqu:
Depends on https://github.com/qri-io/qri/pull/1242 and will also require updates/new flags for some commands to have easily digestable output for shell scripts.

Arqu:
Depends on https://github.com/qri-io/qri/pull/1247

cmd/completion.go
```diff
@@ -0,0 +1,253 @@
...
+
+		o.Out.Write([]byte(zshHead))
+		o.Out.Write(zshBody.Bytes())
+		o.Out.Write([]byte(zshTail))
+	}
+	return nil
+}
+
+const (
+        bash_completion_func = `
```
dustmop, line 225: from golint:
```
don't use underscores in Go names; const bash_completion_func should be bashCompletionFunc
```

cmd/completion.go
```diff
@@ -0,0 +1,253 @@
...
+	-e "s/${LWORD}_get_comp_words_by_ref${RWORD}/__qri_get_comp_words_by_ref/g" \
+	-e "s/${LWORD}__ltrim_colon_completions${RWORD}/__qri_ltrim_colon_completions/g" \
+	-e "s/${LWORD}compgen${RWORD}/__qri_compgen/g" \
+	-e "s/${LWORD}compopt${RWORD}/__qri_compopt/g" \
+	-e "s/${LWORD}declare${RWORD}/builtin declare/g" \
+	-e "s/\\\$(type${RWORD}/\$(__qri_type/g" \
+	<<'BASH_COMPLETION_EOF'
+`
+
+		zshBody := new(bytes.Buffer)
```
dustmop, line 207: We tend to use `bytes.Buffer{}` in our codebase.

cmd/completion.go
```diff
@@ -0,0 +1,253 @@
...
+		return fmt.Errorf("shell not specified")
+	}
+	if len(args) > 1 {
+		return fmt.Errorf("too many arguments, expected only the shell type")
+	}
+	if args[0] == "bash" {
+		cmd.Parent().GenBashCompletion(o.Out)
+	}
+	if args[0] == "zsh" {
+		zshHead := `# reference kubectl completion zsh
```
dustmop, line 55: Could we move all of the shell code strings into top-level consts, to keep the golang code straight-forward to read?

cmd/completion.go
```diff
@@ -0,0 +1,253 @@
...
+		cmd.Parent().GenBashCompletion(zshBody)
+
+		zshTail := `
+BASH_COMPLETION_EOF
+}
+
+__qri_bash_source <(__qri_convert_bash_to_zsh)
+`
+
+		o.Out.Write([]byte(zshHead))
```
dustmop, line 217: Instead:

```
o.Out.WriteString(zshHead)
```

cmd/completion.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+	return nil
+}
+
+const (
+        bash_completion_func = `
+__qri_parse_list()
+{
+    local qri_output out
+    if qri_output=$(qri list --simple 2>/dev/null); then
```
dustmop, line 229: Don't forget to update `--simple` to `--format simple`, assuming that change happens.

Arqu:
Another consideration would be to ship pre-generated autocomplete files so you can just `source ~/path/to/qri/source_file`

cmd/completion.go
```diff
@@ -0,0 +1,280 @@
...
+	"github.com/qri-io/ioes"
+	"github.com/spf13/cobra"
+)
+
+// NewAutocompleteCommand creates a new `qri complete` cobra command that prints autocomplete scripts
+func NewAutocompleteCommand(_ Factory, ioStreams ioes.IOStreams) *cobra.Command {
+	o := &AutocompleteOptions{IOStreams: ioStreams}
+	cmd := &cobra.Command{
+		Use:   "completion [bash|zsh]",
+		Short: "Generates shell completion scripts",
```
b5, line 17: keeping with our [CLI Help Style](https://github.com/qri-io/qri/blob/master/DEVELOPERS.md#cli-help-style), should be `generate shell auto-completion scripts`

cmd/completion.go
```diff
@@ -0,0 +1,280 @@
...
+# ~/.bashrc or ~/.zshrc
+source <(qri completion [bash|zsh])
+
+Alternatively you can pipe the output to a local script and
+reference that as the source for faster loading
+`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return o.Run(cmd, args)
+		},
+		ValidArgs: []string{"bash", "zsh"},
```
b5, line 33: Let's add the `Annotations` field to make this show up in `qri --help`:

```go
Annotations: map[string]string{
  "group": "other",
}
```

cmd/completion.go
```diff
@@ -0,0 +1,280 @@
...
+
+// NewAutocompleteCommand creates a new `qri complete` cobra command that prints autocomplete scripts
+func NewAutocompleteCommand(_ Factory, ioStreams ioes.IOStreams) *cobra.Command {
+	o := &AutocompleteOptions{IOStreams: ioStreams}
+	cmd := &cobra.Command{
+		Use:   "completion [bash|zsh]",
+		Short: "Generates shell completion scripts",
+		Long: `To load completion run
+
+source <(qri completion [bash|zsh])
```
b5, line 20: add a `$` for anything the user should type, and consider moving these into the `Examples` field.

Arqu:
> Ok just ran this, _so_ nice.
> 
> I think we'll need a few passes of fitting & cleanup–as you've noted there are additional commands not yet covered by this PR–but those can come in time. As an example `qri get ... <tab>` lists dataset names (yay!), but kinda obfuscates `qri get body <tab>`. Don't know if or how we want to handle that scenario.
> 
> All that said, wow do completions make life on the CLI easier. The benefits of autocompletion far outweigh any quirks.

Yeah, there's still quite a bit of work to do to have it complete, but did my best to document the things that work and that don't.

For the specific example with `qri get ... <tab>`, I need to get a full list of structure args and then figure out how to serve that in the completion.

Anyways, my intention was to keep the original issue alive and have it document the missing pieces so we can pick it up again later.

PR #1242, opened 2020-04-01
fix(cmd): general fixes to cmd behavior

Arqu:
This is the first of several things I want to do under this PR so will keep it generic for now and clean up later.
Starting off with the `--no-prompt` and `--no-color` flags.

### DIFF
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -68,6 +68,10 @@ func printItems(w io.Writer, items []fmt.Stringer, offset int) (err error) {
 }
 
 func printToPager(w io.Writer, buf *bytes.Buffer) (err error) {
+	if noPrompt {
+		fmt.Fprintln(w, buf.String())
+		return
+	}
 	// TODO (ramfox): This is POSIX specific, need to expand!
 	envPager := os.Getenv("PAGER")
 	if ok := doesCommandExist(envPager); !ok {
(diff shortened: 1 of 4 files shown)


cmd/print.go
```diff
@@ -68,6 +68,10 @@ func printItems(w io.Writer, items []fmt.Stringer, offset int) (err error) {
 }
 
 func printToPager(w io.Writer, buf *bytes.Buffer) (err error) {
+	if noPrompt {
```
dustmop, line 71: So, this is somewhat changing the meaning of `noPrompt`. Before, it meant "don't ask for my input", and was only used by `save` when accepting secrets. This is basically overloading it to mean "non-interactive", in order to avoid writing to the pager. That being said, our pager code could probably be improved anyway, for example by inspecting the file descriptor to determine if it's "smart" or not (I'm not sure what the correct technical term is). But, since this isn't that bad of an overloaded meaning, and is blocking another PR, I'm okay with it for now.

Arqu: Hmm good point. I was guiding myself by the `qri --help` where it's description for `--no-prompt` is `--no-prompt          disable all interactive prompts`. I'm fine with it for now though as it's a simple and not too intrusive overload, and we can split it off in the future if the need arises.

PR #1247, opened 2020-04-02
refactor(cmd): updating search and list to support --format simple

Arqu:
Simplified output formats for easy piping into other scripts. Is a prerequisite for dataset name resolution on most `qri` commands for the autocomplete feature https://github.com/qri-io/qri/pull/1240 https://github.com/qri-io/qri/issues/1236


Arqu:
> Any thoughts on instead doing: `qri list --format simple`? I think that's better than adding another flag, and there shouldn't be a need to use both `--simple` and `--format` at the same time.

Doh, I felt it was clunky but couldn't come up with anything better. Love it, will push an update soon.

Arqu:
GitHub had issues, is already merged

PR #1248, opened 2020-04-02
Fix #1245: Render error details when schema file can not be read

### LINKED ISSUE #1245: Provide reason (error details) when reading schema/structure file fails
## Expected behavior
If, when executing qri validate --structure filename`,  the supplied structure file can't be read, the underlying reason should be printed to the console

## Actual behavior
Only `error opening schema file:` is printed (neither detailed reason nor filename is reported)

## How to reproduce
For `me/<my_dataset>`, introduce an error in structure.json so it can't be parsed, then
run `qri validate --structure structure.json me/<my_dataset>`

## Further information
https://github.com/qri-io/qri/blob/master/lib/datasets.go#L1070 only returns the top level error without wrapping further error details. Note that besides that, the filename argument should read schemaFilename instead of p.SchemaFilename as other a structure filename would not be reported.

hbruch:
Fixes missing file name if structure file is provided instead of schema file and adds message of root cause error.

### DIFF
--- a/lib/datasets.go
+++ b/lib/datasets.go
@@ -1067,7 +1067,7 @@ func (r *DatasetRequests) Validate(p *ValidateDatasetParams, valerrs *[]jsonsche
 	} else {
 		data, err := ioutil.ReadFile(schemaFilename)
 		if err != nil {
-			return fmt.Errorf("error opening %s file: %s, error was: %s", schemaFlagType, schemaFilename, err)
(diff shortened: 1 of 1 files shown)


hbruch:
Removed error details again, as in most cases it should be file not found, I suppose. However, error message now displays filename for structure file, too.

dustmop:
Great, thanks for the fix! I will merge

PR #1253, opened 2020-04-03
fix(save): If body is too large to diff, compare checksums

dustmop:


dustmop:
Follow up on https://github.com/qri-io/qri/pull/1222. I realized that we already have a checksum in the structure of the previous version. We can just compare against that if the body is too large to diff.

PR #1256, opened 2020-04-06
fix(cmd): better error hints on 'qri connect'

Arqu:


cmd/connect.go
```diff
@@ -70,7 +70,7 @@ func (o *ConnectOptions) Complete(f Factory, args []string) (err error) {
 			return err
 		}
 	} else if !QRIRepoInitialized(qriPath) {
-		return fmt.Errorf("no qri repo exists")
+		return fmt.Errorf("no qri repo exists\nhave you run 'qri setup'")
```
b5, line 73: Ok, this one is going to sound like overkill, but I think it's a great chance to establish better patterns for both us developers and end users.

The `repo` package defines a bunch of error values:

https://github.com/qri-io/qri/blob/855ff8f1de6d492b98049e17740ecf4702a37ce8/repo/repo.go#L25-L46

I'd like to add `ErrNoRepo = fmt.Errorf("repo: no repo")` to that list, import our `errors` package here, and return 

```go
errors.New(repo.ErrNoRepo, "no qri repo exists\nhave you run 'qri setup'?")
```

This is going to add `repo` as an import to this package, which I think is an anti-pattern (we want `cmd` to import as few things outside of `lib` as possible). The long term fix for _that_ problem is to refactor `QRIRepoInitialized` to a new function defined in `/lib/setup.go`:

```go
// QriRepoExists checks if a qri repo is defined at the given path, without opening
// the repo itself
func QriRepoExists(path string) error {
  // for now this just checks for an existing config file
  _, err := os.Stat(filepath.Join(path, "config.yaml"))
  if !os.IsNotExist(err) {
    return nil
  }
  return errors.New(repo.ErrNoRepo, "no qri repo exists\nhave you run 'qri setup'?")
}
```

Arqu: Don't think it's overkill, I'd rather we keep moving in the right direction for the stable release even if that produces antipatterns untill we get there.

PR #1258, opened 2020-04-06
fix(transform): Re-open transform after friendly, so it writes to cafs

dustmop:


dustmop:
Besides un-breaking a pretty heinous bug, this also finally fixes https://github.com/qri-io/qri/issues/1217.

PR #1259, opened 2020-04-06
fix(watchfs): Methods need a mutex to avoid concurrent writes

dustmop:
See https://tip.golang.org/doc/go1.8#mapiter


dustmop:
I think this will fix some of the CI failures seen here: https://github.com/qri-io/qri/issues/1047

PR #1260, opened 2020-04-07
fix(dsfs): fix dag structure, defend against bad DAGs

b5:
we've been seeing a number of "bad dags" that are missing references to hashes that are intended to be internal to the dag, but are not linked. Use a context timeout to defend against this.

While we're at it, update the dag structure to write the proper filenames, which were erroneous. This may not address the central problem

Also while we're at it, remove `maybeAddDefaultViz` which is a source of problems.


base/dataset.go
```diff
@@ -35,27 +37,40 @@ func OpenDataset(ctx context.Context, fsys qfs.Filesystem, ds *dataset.Dataset)
 			return
 		}
 	}
+
 	if ds.Readme != nil && ds.Readme.ScriptFile() == nil {
-		if err = ds.Readme.OpenScriptFile(ctx, fsys); err != nil {
-			log.Debug(err)
-			return
+		readmeTimeoutCtx, cancel := context.WithTimeout(ctx, time.Millisecond*250)
```
Arqu, line 42: This seems error prone? I'd rather have this behavior within the `OpenScriptFile` or specifically in the `qfs.PathResolver` that actually implements this. On the `http` resolver it might make sense to have a timeout + retry mechanics etc. while on the `local` version this should definitelly get back in < 250ms.

base/dsfs/dataset.go
```diff
@@ -38,6 +38,16 @@ var BodySizeSmallEnoughToDiff = 20000000 // 20M or less is small
...
 // returning a fully-hydrated dataset
 func LoadDataset(ctx context.Context, store cafs.Filestore, path string) (*dataset.Dataset, error) {
+	// set a timeout to handle long-lived requests when connected to IPFS.
+	// if we don't have the dataset locally, IPFS will reach out onto the d.web to
+	// attempt to resolve previous hashes. capping the duration  to 700 MS yeilds
+	// quick results.
+	// TODO (b5) - The proper way to solve this is to feed a local-only IPFS store
+	// to this entire function, or have a mechanism for specifying that a fetch
+	// must be local
+	ctx, cancel := context.WithTimeout(ctx, time.Millisecond*700)
```
Arqu, line 48: Thinking about these timeouts, I'd actually prefer having them in the `config` this way if anybody has issues (specific to their setup/network etc) they can actually tune this (or -1 to turn off entirely?)

base/dsfs/package.go
```diff
@@ -53,8 +53,10 @@ const (
 	PackageFileViz
 	// PackageFileRenderedViz is the rendered visualization of the dataset
 	PackageFileRenderedViz
-	// PackageFileReadme is the raw readme of the dataset
+	// PackageFileReadme connects readme data to the dataset package
```
Arqu, line 56: Should this be a hidden file like `.qri-ref` if it's only a link?

base/dataset.go
b5: Definitely agree with this long term. mind opening an issue on qri-io/qfs?

base/dsfs/dataset.go
b5: Love that you've identified what we're doing here: request timeouts. I'll unify this value to a variable in this PR & we can make it a config value in the future. adding & removing from config is a little precarious this close to a release

base/dsfs/package.go
b5: This comment is somewhat misleading, it's more than a link, it also specifies the syntax ("markdown") readme is encoded with.

PR #1262, opened 2020-04-07
refactor(cmd): rename fsi to workdir, add working directory help section

### LINKED ISSUE #982: The fsi command is not listed in the command-line help
Running `qri` shows help about available commands, but does not list qri fsi. Relatedly, it might be worth breaking FSI commands out into their own section. This would move status, restore, checkout and init.

b5:
closes #982

biggest change shows up in the main help output, which cleans up the initial description, and adds a "Working Directory Commands" section:

```
$ qri
qri ("query") is a set of tools for building & sharing datasets: https://qri.io
Feedback, questions, bug reports, and contributions are welcome!
https://github.com/qri-io/qri/issues

Usage:
  qri [command]

Dataset Commands:
  add         add datasets from other peers
  diff        compare differences between two data sources
  export      copy datasets to your local filesystem
  get         get components of qri datasets
  list        show a list of datasets
  log         show log of dataset commits
  logbook     show a detailed list of changes on a dataset name
  remove      remove a dataset from your local repository
  rename      change the name of a dataset
  render      render a dataset readme or a dataset template
  save        save changes to a dataset
  sql         run a SQL query on local dataset(s)
  stats       get aggregated stats for a dataset
  update      schedule dataset updates
  use         select datasets for use with the qri get command
  validate    show schema validation errors

Working Directory Commands:
  checkout    create a linked directory and write dataset files to that directory
  init        initialize a dataset directory
  restore     restore a checked out dataset's files to a previous state
  status      show what components of a dataset have been changed
  workdir     file system integration tools

Network Commands:
  connect     connect to the distributed web by spinning up a Qri node
  fetch       show log of remote dataset commits
  peers       commands for working with peers
  publish     set dataset publicity
  registry    commands for working with a qri registry (qri.cloud)
  search      search the registry for datasets

Other Commands:
  completion  generate shell auto-completion scripts
  config      get and set local configuration information
  help        Help about any command
  setup       initialize qri and IPFS repositories, provision a new qri ID
  version     print the version number

Flags:
  -h, --help               help for qri
      --ipfs-path string   override IPFS path location (default "/Users/b5/.ipfs")
      --log-all            log all activity
      --no-color           disable colorized output
      --no-prompt          disable all interactive prompts
      --repo string        provide a path to load qri from (default "/Users/b5/.qri")

Use "qri [command] --help" for more information about a command.
```


cmd/qri.go
```diff
@@ -20,14 +20,14 @@ func NewQriCommand(ctx context.Context, pf PathFactory, generator gen.CryptoGene
...
-on the distributed web.
-
-https://qri.io
-
-Feedback, questions, bug reports, and contributions are welcome!
+		Long: `qri ("query") is a set of tools for building & sharing datasets: https://qri.io
+Feedback, questions, bug reports, and contributions are welcome! 
 https://github.com/qri-io/qri/issues`,
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			if ioStreams.IsCygwinTerminal() {
```
Arqu, line 27: This now superseeds my PR https://github.com/qri-io/qri/pull/1263 will close that out.

cmd/sql.go
```diff
@@ -13,10 +13,12 @@ func NewSQLCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	o := &SQLOptions{IOStreams: ioStreams}
 	cmd := &cobra.Command{
 		Use:   "sql QUERY",
-		Short: "run a SQL query on local dataset(s)",
+		Short: "run a SQL query on local dataset(s) (experimental)",
```
Arqu, line 16: Think we need to agree on a consistent "experimental" description. Specifically I saw @dustmop already has a commit in with the following wording: https://github.com/qri-io/qri/commit/c8c98d183f0d46e3f1005ce468dc68f58af021c6#diff-73ef6a20a16eceacdbdf948c7ad07c3eR61

go.mod
```diff
@@ -50,7 +50,7 @@ require (
 	github.com/qri-io/dataset v0.1.5-0.20200324184139-108a69072ede
 	github.com/qri-io/deepdiff v0.1.1-0.20200305020550-8173efebcaa1
 	github.com/qri-io/doggos v0.1.0
-	github.com/qri-io/ioes v0.1.0
+	github.com/qri-io/ioes v0.1.1-0.20200407033451-28794e5a4c83
```
Arqu, line 53: I see that we mostly cut a release on ioes updates since they are so infrequent. Considering the last addition actually introduced functionality maybe cut it again and pin to that. Regardless, fine with both options.

cmd/qri.go
Arqu: Though think we are both wrong:
https://github.com/fatih/color/pull/61#pullrequestreview-22449122
+ I think we should look into why `fatih/color` doesn't do this correctly, `v1.7.0` should handle it well, without us actually setting this.

b5: innnfteresting. looks like the latest release of `color` is `1.9.0`. We could try upgrading. I can revert this change until we figure it out?

cmd/sql.go
b5: very good point. I'll change to match his description.

go.mod
b5: totally. I'll cut 0.1.1 & add

cmd/qri.go
Arqu: Sounds good, we can keep the original issue continue tracking this.

b5:
look ok to merge @Arqu?

Issue #955, opened 2019-10-09
use mattn's is-a-tty package to detect if terminal should output color


b5:
We've talked about detecting TTYs in the past, looks like this package package fits the bill:
https://github.com/mattn/go-isatty

this would make windows output look less gross on the CLI, which currently contains some gross characters

Arqu:
Based on the linked PR, this is not a viable solution. The PR is a bandaid.

PR #1263, opened 2020-04-07
fix(cmd): tty should auto disable color

Arqu:
Sprint board didn't update so I got started on https://github.com/qri-io/qri/issues/955
Wound up on the https://github.com/qri-io/ioes/commit/28794e5a4c834251ed7e0fde143fd3c99bb58db3 commit.
Once we cut the release on `ioes` I'll update the deps here and we can merge.

### DIFF
--- a/cmd/qri.go
+++ b/cmd/qri.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/rpc"
 	"os"
+	"runtime"
 	"sync"
 
 	"github.com/qri-io/ioes"
@@ -17,6 +18,10 @@ import (
 
 // NewQriCommand represents the base command when called without any subcommands
 func NewQriCommand(ctx context.Context, pf PathFactory, generator gen.CryptoGenerator, ioStreams ioes.IOStreams) *cobra.Command {
+
+	qriPath, ipfsPath := pf()
+	opt := NewQriOptions(ctx, qriPath, ipfsPath, generator, ioStreams)
+
 	cmd := &cobra.Command{
 		Use:   "qri",
 		Short: "qri GDVCS CLI",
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -50,7 +50,7 @@ require (
 	github.com/qri-io/dataset v0.1.5-0.20200324184139-108a69072ede
 	github.com/qri-io/deepdiff v0.1.1-0.20200305020550-8173efebcaa1
 	github.com/qri-io/doggos v0.1.0
-	github.com/qri-io/ioes v0.1.0
+	github.com/qri-io/ioes v0.1.1-0.20200407033451-28794e5a4c83
```
b5, line 53: as requested, I've now published v0.1.1

cmd/qri.go
```diff
@@ -32,6 +32,10 @@ https://github.com/qri-io/qri/issues`,
 			if err == nil && noColor {
 				setNoColor(noColor)
 			}
+			// if it's a tty terminal disable color
+			if ioStreams.IsTerminal() {
```
b5, line 36: I think this is having the opposite effect of what's intended? Do we want `IsCygwinTerminal` instead? I'm not an expert here

cmd/qri.go
```diff
@@ -32,6 +32,10 @@ https://github.com/qri-io/qri/issues`,
...
+				setNoColor(noColor)
```
b5, line 37: What happens when the user supplies `qri --no-color=false`?

cmd/qri.go
Arqu: Still not sure on either option based on this: https://github.com/qri-io/qri/pull/1262#pullrequestreview-389020225

cmd/qri.go
Arqu: Good point, though I don't see how both of it should work at the same time. I guess leaving the option to the user makes sense. Though I will check if the flag was changed and not just default-ed.

Arqu:
Still not completely sure, want to address this during the day. I'll set up a Win environment and just play with it to get some field experience on it.

Arqu:
There was a lot of testing on this, in the end, the most sensible approach turned out to be:
- disable color on windows by default
- let users override it with `--no-color=false`

The reasoning being it's just too much effort to track and maintain different combinations of os/color support as they all revolve around different hacks to test for it. 

From some manual testing I found this (on a Win10 machine):

cmd
- IsTerminal = true
- IsCygwinTerminal = false
- Color does not work

PowerShell
- IsTerminal = true
- IsCygwinTerminal = false
- Color does not work

MSYS2
- IsTerminal = false
- IsCygwinTerminal = true
- Color does work

Bash
- IsTerminal = true
- IsCygwinTerminal = false
- Color does work

Cygwin
- IsTerminal = true
- IsCygwinTerminal = false
- Color does work

Here's a sample of what it generally takes to track color support: https://github.com/efekarakus/termcolor and even this is error prone.

PR #1274, opened 2020-04-13
refactor: changes to make testground simulation viable

b5:
I'm working on setting up [testground](http://testground.ipfs.team) for qri. This PR adds a few bits to make orchestrating tests possible. 

To make test node setup easier, I'd like to use `lib` directly, moving logic down from cmd into lib means we get profile creation without using cmd.

Next, we need a way to react to nodes connecting to program different behaviours for actors in the test. I've added events to `p2p` for exactly this, starting with just a connection event. I'm assuming more will come in the future.


ramfox:
@b5 
Having a bit of trouble with two tests, both seem to stem from the same issue: sometimes the libp2p identification of a peer failes. Which means we don't ever trigger `upgradeQriConnection`, and so the tests fail. 

In the case of `TestConnectedQriProfiles`, it fails because we are trying to ensure the profiles of each connected qri node exists in the peer list, and if the `upgradeQriConnection` fails we don't request the peer's profiles or their neighboring peer's profiles

In the case of `TestNodeEvents`: if we don't wait for the `ETP2PQriPeerConnected` event to trigger before shutting down the neighboring peer (so we can test the `ETP2PPeerDisconnected` event), then upgrading will fail because the connection was explicitly closed. But unfortunately, sometimes that event is never triggered, so the test hangs.

I added a `libp2pevent.EvtPeerIdentificationFailed` to our libp2pevent handler, and it looks like in these cases we are getting "stream reset" errors. All I can find about "stream reset" errors in libp2p2 is this: https://discuss.libp2p.io/t/what-causes-a-stream-reset/253 and https://discuss.libp2p.io/t/how-to-check-if-a-stream-is-already-open-with-peer/249/5, as well as the code where the error is possibly defined:  https://godoc.org/github.com/libp2p/go-libp2p-core/mux

So I'm not 100% sure about what's causing the reset. Hopefully it's an issue with everything being local and so responses are too fast and causing unexpected behaviors. But I do know that it's causing our tests to fail because we can't `upgradeToQriConnection` until the peer identification has been completed.

ramfox:
There is a lot of shenanigans happening down in the p2p tests after the libp2p and event upgrades. Mostly surrounding upgrades to qri connections, specifically because the timing of the upgrade has changed. I don't think this week is the time to fix it or straighten it out. 

Unfortunately, most of our tests that focus on p2p functionality are too flakey to rely on, since `libp2pevent.EvtPeerIdentificationCompleted` is not often successful.

The scope of this pr is supposed to get testgrounds viable, it is now viable! We can use this branch to successfully run a testground plan `remote push` @ github.com/qri-io/test-plan

lib/setup.go
```diff
@@ -9,24 +9,69 @@ import (
 
 	"github.com/qri-io/qfs/qipfs"
 	"github.com/qri-io/qri/config"
+	"github.com/qri-io/qri/repo"
 	"github.com/qri-io/qri/repo/gen"
 )
 
+// QriRepoExists retursn nile if qri repo is defined at the given path, without
+// opening the repo itself
```
b5, line 17:
```suggestion
// QriRepoExists returns nil if a qri repo is defined at the given path
// does not attempt to open the repo
```

p2p/node.go
```diff
@@ -361,12 +370,54 @@ func (n *QriNode) SendMessage(ctx context.Context, msg Message, replies chan Mes
...
 	n.handleStream(WrapStream(s), nil)
 }
 
+func (n *QriNode) libp2pSubscribe() error {
+	host := n.host
+	sub, err := host.EventBus().Subscribe([]interface{}{
+		new(libp2pevent.EvtPeerIdentificationCompleted),
+		new(libp2pevent.EvtPeerIdentificationFailed),
+	},
+	// libp2peventbus.BufSize(1024),
```
b5, line 398: did you research the purpose of this value? Any reason for keeping this commented out?

p2p/profile.go
```diff
@@ -5,12 +5,9 @@ import (
 	"encoding/json"
 	"fmt"
 
-	// "time"
```
b5, line 8: drop stray import

repo/buildrepo/build.go
```diff
@@ -107,9 +106,9 @@ func newLogbook(fs qfs.Filesystem, bus event.Bus, pro *profile.Profile, repoPath
 
 func newDscache(ctx context.Context, fs qfs.Filesystem, bus event.Bus, book *logbook.Book, username, repoPath string) (*dscache.Dscache, error) {
 	// This seems to be a bug, the repoPath does not end in "qri" in some tests.
-	if !strings.HasSuffix(repoPath, "qri") {
-		return nil, fmt.Errorf("invalid repo path: %q", repoPath)
-	}
+	// if !strings.HasSuffix(repoPath, "qri") {
+	// 	return nil, fmt.Errorf("invalid repo path: %q", repoPath)
+	// }
```
b5, line 111: this is my bad. mind reverting this change by commenting back in?

Issue #1277, opened 2020-04-13
naming users and datasets needs to be unified and enforced


dustmop:
The recently added `dsref` package should be the central location for parsing usernames and dataset names. In both cases we should only allow names that start with a lower case letter, then follow with one or more of (lower case letter, number, underscore, dash).

We also need to enforce this naming scheme everywhere. Here are some places I found that are behaving incorrectly:
- [ ] running `qri setup` doesn't validate that the provided username is valid
- [ ] logbook doesn't validate that the username for author init is valid
- [ ] modifying the profile.peername field in config.yaml can create invalid usernames
    * afterwards, running `qri save` will successfully create a dataset with an invalid username, but logbook will return the error "log: not found"
- [ ] `qri save --body my-body.csv` incorrectly turns "-" into "_" because package varName doesn't match dsrerf
- [ ] `qri checkout` also creates a directory name from the dataset name using varName
- [ ] dataset/validate.go uses ValidName which doesn't allow "-", but this appears to be unused
- [ ] peer messages received over p2p should also be validating received names

Desktop also needs to have parity with these checks

dustmop:
Reversing the fix for this issue: https://github.com/qri-io/qri/issues/1116.

Related: https://github.com/qri-io/qri/issues/1153 and https://github.com/qri-io/qri/issues/1132

dustmop:
Related issue that can be fixed at the same time. It's not possible to infer a name using a file that contains unicode:

```
> qri save --body=здравствуйте.csv 
invalid name: dataset name must start with a letter, and only contain letters, numbers, and underscore. Maximum length is 144 characters
```

PR #1284, opened 2020-04-16
feat(name): Generate valid dataset names, use in multiple places

dustmop:
Generate names from arbitrary strings using a state machine. Use this name
generator in multiple places to fix:

* qri setup's prompt for the username
* calling `qri save --body` which infers a dataset name
* qri checkout which used to not allow dashes in folder names

The improvements over varName are:

* Only one-pass over the string
* Dashes are allowed in names
* Generator and validator both live in dsref, so they stay synchronized
* Unicode awareness. Certain characters are normalized, others ignored
* Better handling of all capital abbreviations in CamelCase strings
* Numbers are not dropped if they appear at the start of a name
* Comma followed by space is turned into only one dash


dustmop:
Fixes most of https://github.com/qri-io/qri/issues/1277.

PR #1286, opened 2020-04-17
feat(save): Refactor save switches, use for file hint to inform commit messages

dustmop:


base/save.go
```diff
@@ -17,20 +17,11 @@ import (
...
-	Replace             bool
-	DryRun              bool
-	Pin                 bool
-	ConvertFormatToPrev bool
-	Force               bool
-	ShouldRender        bool
-	NewName             bool
-}
+// SaveSwitches is an alias for the switches that control how saves happen
+type SaveSwitches = dsfs.SaveSwitches
```
b5, line 21: I was going to argue against this alias but I'm delighted to see none of this is exposed at the lib level, so I think it's fine considering it's "internal"

base/dsfs/dataset.go
```diff
@@ -220,48 +221,69 @@ func DerefDatasetCommit(ctx context.Context, store cafs.Filestore, ds *dataset.D
...
+	// ConvertFormatToPrev is whether the body should be converted to match the previous format
+	ConvertFormatToPrev bool
+	// ForceIfNoChanges is whether the save should be forced even if no changes are detected
+	ForceIfNoChanges bool
+	// ShouldRender is deprecated, controls whether viz should be rendered
+	ShouldRender bool
+	// NewName is whether a new dataset should be created, guaranteeing there's no previous version
+	NewName bool
+	// FileHint is a hint for what file is used for creating this dataset
+	FileHint string
```
b5, line 241: string isn't _exactly_ a switch, but by using the name `Switch` we avoid colliding with the func-based `Options` convention we use elsewhere.

Issue #1209, opened 2020-03-19
TestRunner's TestRepo and CommandRunner disagree about repoPath


dustmop:
From cmd/test_runner_test.go:

There's a bug with TestRepo that I don't understand completely. The commands are run using a different refstore than the refstore returned by accessing the fields of the TestRepo directly. The command runner constructs a repo and then refstore which has a path similar to "/var/folders/tmpDir/T/qri_save_bad_case1234" with "qri" and "ipfs" directories within. However, trying to directly access the Repo object from TestRepo will return a refstore with the path "/var/folders/tmpDir/T/qri_save_bad_case1234" as the *qri repository*.

So doing:
`  run.RepoRoot.Repo()`
gives a refstore that saves to:
`  "/var/folders/tmpDir/T/qri_save_bad_case1234/refs.fbs"`
While the commandRunner is using:
`  "/var/folders/tmpDir/T/qri_save_bad_case1234/qri/refs.fbs"`
We work around this by constructing a lib.Instance, which uses the PathFactory to get the qri subfolder and correctly use the refstore at:
`  "/var/folders/tmpDir/T/qri_save_bad_case1234/qri/refs.fbs"`
This is probably the same bug that is handled in repo/buildrepo/build.go with a hack that
appends "/qri" to the repoPath.

b5:
closed by #1294

PR #1294, opened 2020-04-22
fix(repo): Construct temp repo using correct path

dustmop:
The TestRunner uses a TempRepo, which should pass the correct qri path, instead of the root path, when building the repo. This prevents two separate repos from being constructed in tests, cleaning up a bunch of subtle bugs that we were working around. Some test timestamps change because now some things that were constructed twice are now only constructed once.


dustmop:
Fixes https://github.com/qri-io/qri/issues/1209.

PR #1295, opened 2020-04-22
feat(save): add --drop flag to remove dataset components when saving

b5:
I'm not sure this should be merged as-is (which is why I've skipped tests for the moment).

We're in a situation where a few versions of qri have been creating bad DAG structures, referencing blocks not stored within the created DAG. Thanks to save being a patch operation, and FSI-linked save not recognizing the viz component, qri presents no way to remove a bad viz component. This --drop flag at least lets us drop entire component for now upon save.

Example uses:
```
# drop viz
$ qri save --drop viz

# add stuff, and drop readme
$ qri save --file body.csv --drop rm
```

### DIFF
--- a/base/save.go
+++ b/base/save.go
@@ -10,7 +10,6 @@ import (
 	"github.com/qri-io/qfs"
 	"github.com/qri-io/qfs/cafs"
 	"github.com/qri-io/qri/base/dsfs"
-	"github.com/qri-io/qri/dsref"
 	"github.com/qri-io/qri/logbook"
 	"github.com/qri-io/qri/repo"
 	"github.com/qri-io/qri/repo/profile"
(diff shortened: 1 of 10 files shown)


base/save.go
```diff
@@ -240,3 +246,43 @@ func GenerateAvailableName(r repo.Repo, peername, prefix string) string {
...
+		case "vz":
+			ds.Viz = nil
+		case "tf":
+			ds.Transform = nil
+		case "st":
+			ds.Structure = nil
+		case "bd":
+			ds.Body = nil
+		case "rd":
+			if ds.Viz != nil {
```
dustmop, line 277: Could you add a comment on why Viz is handled differently here.

lib/datasets.go
```diff
@@ -430,6 +430,8 @@ type SaveParams struct {
 	ConvertFormatToPrev bool
 	// string of references to recall before saving
 	Recall string
+	// string of references to delete before saving
```
dustmop, line 433: Not "references". Specify that it's a comma separated list of component names.

base/save.go
b5: "rd" is a stand in for "rendered", which isn't a component. I'm going to remove support for dropping `rd` for now, as it's not a component

base/save.go
```diff
@@ -240,3 +246,39 @@ func GenerateAvailableName(r repo.Repo, peername, prefix string) string {
...
+	revs, err := dsref.ParseRevs(revStr)
+	if err != nil {
+		return err
+	}
+
+	for _, rev := range revs {
+		if rev.Gen != 1 {
+			return fmt.Errorf("cannot drop specific generations")
+		}
+		switch rev.Field {
```
Arqu, line 265: Could we have this live in `base/component` like https://github.com/qri-io/qri/blob/ef62c71198c2110b5d4de5c5cf7245c78d8a9aed/base/component/field_id.go
And tag these for this specific usecase.

b5: don't think it should move into the component package. We haven't fully documented this, but the component package handles dynamic behavior on components using a new internal API. If anything those values `field_id.go` should move into the `dsref` package.

That said, I'm very much with you, `Drop` shouldn't be in `save.go`, so I've moved it to `base/revisions.go` and added tests.

PR #1304, opened 2020-04-23
feat(dscache): Dscache handles remove operations

dustmop:


logbook/action.go
```diff
@@ -10,8 +10,12 @@ type ActionType byte
...
 	// ActionDatasetNameInit is an action that inits a dataset name
 	ActionDatasetNameInit ActionType = iota
-	// ActionDatasetChange is an action for when a dataset changes
-	ActionDatasetChange
+	// ActionDatasetCommitChange is an action for when a dataset changes its newest commit
+	ActionDatasetCommitChange
+	// ActionDatasetDeleteAll is an action for when a dataset is entirely deleted
+	ActionDatasetDeleteAll
+	// ActionDatasetRename is when a dataset is renamed
+	ActionDatasetRename
```
b5, line 18: Still drives me nuts that we're re-implementing an event system here, but I can't think of a better solution 🤷‍♂️

Issue #1305, opened 2020-04-23
logbook needs refactoring


dustmop:
Some changes that need to happen for logbook:

- Logbook methods should not be working with dataset ref's. Using them requires name resolution, which should happen above the level of logbook. Instead, methods should use initID.
- Logbook has nothing in place that stores initIDs, they have to be calculated every time. There should be an (in-memory only, not serialized) table that maps initIDs to positions within the logbook structure. Maybe dscache could also store this to speed things up, need to see how it looks in code.
- InitName should not happen automatically inside other methods. If a caller has an initID, they use that, if not, they need to call InitName first to get an initID.
- As a stop-gap until dscache is fully in use, logbook could have a public method that converts a ref into an initID, but this should be the only method allowed to accept a ref parameter.
- Logbook has 3 levels of logs, and is essentially using "dynamic types" for each of them. The top-level is the user log, the second is the dataset log, the third is the branch log (I think this was done for easy storage as fbs). Implementation within logbook has to be careful to not mix these up, which is very error prone. We should introduce type aliases for each of these, to at least get some static type checking in places.

dustmop:
Another logbook problem, which related to the "dynamic typing" issue mentioned above: logbook isn't great about validating the data that it stores. Observe the following snippet from my logbook:

```
      {
        "type": "init",
        "model": "user",
        "name": "b5",
        "authorID": "QmSyDX5LYTiwQi861F5NAwdHrrnd1iRGsoEvCyzQMUyZ4W",
        "timestamp": "2020-01-13T19:58:33.918958-05:00"
      },
      {
        "type": "amend",
        "model": "user",
        "name": "fred",
        "authorID": "yzf43six37hhw26jai3brapivvfhndc4mivfhbwskzvyrxv6o7la",
        "timestamp": "2020-02-04T14:09:29.934139-05:00"
      },
      {
        "type": "amend",
        "model": "user",
        "name": "b5",
        "authorID": "yzf43six37hhw26jai3brapivvfhndc4mivfhbwskzvyrxv6o7la",
        "timestamp": "2020-02-04T16:12:00.46922-05:00"
      },
```

The first entry has a valid profileID for the authorID field. It's 46 bytes long, because it's a correctly base58 encoded peer.ID. The next two entries are incorrectly double-encoding this peer.ID, ending up with 52 byte long strings that can't be used without potentially causing bugs. Nothing in the code right now is validating that these fields contain intended data.

dustmop:
[...]

PR #1308, opened 2020-04-27
feat: Registry-Backed Reference Resolution

b5:
This PR proposes resolving a reference get it's own interface:

```go
type RefResolver interface {
  ResolveRef(ctx context.Context, ref *dsref.Ref) (error)
}
```

* Reference resolution returns a "canonicalized" dsref with the InitID and HEAD Path assigned. 
* Multiple subsystems can implement the RefResolver interface. 
* Reference resolution is it's own thing, independant of dataset loading or history fetching. 
* We've talked about just passing around ID's everywhere, we should be passing around dsref.Refs. @dustmop's resolver implementation fetches the InitID _and_ HEAD path. That feels exactly right to me.
* RefResolver implementations **must** be nil-callable, more on this below

### `lib.Instance.ResolveRef`

The primary "production" implementatinon of the `RefResolver` interface is a method defined on `*lib.Instance`, which is a composition of all `RefResolvers` available in a specific order:

|  # | resolver               | name is resolved if...          | note                    |
| ---| ---------------------- | ------------------------------- | ----------------------- |
|  1 | `dsref.Parse`          | the reference is complete       | should yeild perf bump  |
|  2 | `dscache.ResolveRef`   | the name & ID is in the cache   | cache updated with pull |
|  3 | `registry.ResolveRef`  | the registry has the identifier | requires network        |
|  4 | `p2p.ResolveRef`       | enough peers confirm the ID     | requires p2p network    |

Because implemetnations are nil-callable, an instance can just iterate them all calling `ResolveRef` until a match is found. once we hit the `registry` and eventually `p2p` implementations, we're doing reference resolution over the network.

### fsi: naming's problem child

@dustmop's `LoadDsRef` implementation gives really good food for thought here:

https://github.com/qri-io/qri/blob/26289349ddd3e2bb4438d3427ff64ff01d116b86/resolver/loader/loader.go#L49

I think this method does too much in the name of "resolving". In my mind resolving a reference should do only one thing: turn a provided reference string into a known identifier. If I'm just trying to run `qri log` on a dataset, qri should resolve the reference & fetch the log. That's it.

The way FSI works today complicates things, because it doesn't jive with reference resolution. When we detect something is FSI-linked, we _ignore_ the `Path` value, and use `FSIPath` instead, which has a separate set of functions for loading datasets. I suspect this is why our first cracks at reference resolution also load datasets.

The root problem here is FSI behaving differently than other `qfs.Filesystem` implementations. on the write side, this will need to be true for a long time because we do all kinds of file watching / web socket / event-driven stuff. On the _read_ side however, there's very little not to move towards the `fsi` package being another source that a `qfs.Filesystem` can delegate reads to. This PR introduces `/fsi` as a path prefix for data persisted using qri's file system integration (`/ipfs` is our most common path prefix), setting the stage for pushing fsi integration farther down the qri import hierarchy.

This helps bring FSI in-line with other systems, and most importantly drops the need for an `FSIPath` field when it comes to reference resolution.


dustmop:
> This PR proposes resolving a reference get it's own interface:
>
> ```
> type RefResolver interface {
>   ResolveRef(ctx context.Context, ref *dsref.Ref) (error)
> }
> ```

This should just be the Resolver interface that already exists. The work I've been focused on with name resolution is very much in flux, and it's okay to change the existing interface. The current methods `GetInfo...` aren't being used anywhere. They were originally going to be used by base/save.go, as I was exploring how to remove its calls to `CanonicalizeDatasetRef`, but through that work discovered that it's better to refactor base so that it uses initIDs instead, avoiding the need to canonicalize, but there's some work that needs to be done first before that can happen. For example, see https://github.com/qri-io/qri/issues/1305.

> The primary "production" implementation of the `RefResolver` interface is a method defined on `*lib.Instance`, which is a composition of all RefResolvers available in a specific order:

I don't think that we need to have an interface that is used by both the top-level intelligent resolver, *and* all of its constituent parts. The only benefit there would be if we need the top-level resolver to be completely pluggable and replaceable, but I don't think that's what we want. Yes, it has to be configurable (with regards to online vs offline, registry vs remote, p2p enabled or disabled), but there's never going to be a change in order of the subsystems it cares about. It's probably the right thing to have tight coupling within the implementation of the top-level resolver, because it stabilizes the implementation and makes it easier to reason about.

The interface at the top and the provided surface of each subsystem are actually different things. Keeping it this way allows for more flexibility in how each part returns its results, because it doesn't have to shoe-horn itself into this other interface that exists. For example, what if in the future we add reputation to p2p, and now have to handle that entirely within the p2p package, because we aren't able to return it up to the intelligent resolver, losing the opportunity to let that value inform its decision.

The benefit of having this top-level interface is that it simplifies testing, and we can get away from the "bag-of-stuff" approach that the `repo` currently has.

> ```
> func (dr *DatasetResolver) LoadDsref(ctx context.Context, refstr string (*dataset.Dataset, string, dsref.Ref, *dsref.VersionInfo, error) { 
> ```
>
> I think this method does too much in the name of "resolving". In my mind resolving a reference should do only one thing: turn a provided reference string into a known identifier. If I'm just trying to run qri log on a dataset, qri should resolve the reference & fetch the log. That's it.

Yes, this does too much, but it isn't meant to be a generic resolver for references. This is wrapping the pattern that appears in `Get` and `Validate` and `Stats` and `Diff`, which all resolve the ref then loads the dataset, either from the repo or FSI. Moving FSI down into the cafs.Filesystem simplifies things, but there's still a need in all of these places to load dataset heads, and this is for avoiding duplicating that work.

> This PR introduces /fsi as a path prefix for data persisted using qri's file system integration

I like this idea, but I think we should use a `fsi://` protocol instead. We're stuck with "/ipfs" for now, but should avoid compounding that existing complication.

dscache/dscache.go
```diff
@@ -173,6 +174,37 @@ func (d *Dscache) ListRefs() ([]reporef.DatasetRef, error) {
...
+// ResolveRef finds the identifier for a dataset reference
+// implements resolve.RefResolver interface
+func (d *Dscache) ResolveRef(ctx context.Context, ref *dsref.Ref) error {
+	// NOTE: isEmpty is nil-callable
+	if d.IsEmpty() {
+		return resolve.ErrCannotResolveName
+	}
+
+	// Handle the "me" convenience shortcut
+	if ref.Username == "me" && d.DefaultUsername != "" {
```
dustmop, line 185: I don't think this belongs in dscache. Handling "me" is a surface-level UI convenience, and it shouldn't be known about by lower-level components. While investigating the save path, I found tons of subtle bugs around how base handles blank and "me" usernames, including unit tests that rely on strange and broken behavior, as a consequence of "me" handling showing up throughout the stack. Better to keep the lower-level of the code simple, straight-forward, non-magical, and easy to reason about.

dsref/ref.go
```diff
@@ -2,6 +2,9 @@ package dsref
 
 // Ref is a reference to a dataset
 type Ref struct {
+	// ID is the canonical identifer for a dataset history, in Qri canonical IDs
+	// are logbook InitIDs
+	ID string `json:"id,omitempty"`
```
dustmop, line 7: I think it's probably a good thing to add this to ref, but calling it just `ID` is ambiguous. If it's an initID, it should be called that.

lib/resolve.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/qri-io/qri/dsref"
+	"github.com/qri-io/qri/fsi"
+	"github.com/qri-io/qri/resolve"
+)
+
+// assert at compile time that instance is a RefResolver
+var _ resolve.RefResolver = (*Instance)(nil)
+
+// ParseAndResolveRef combines reference parsing and resolution
+func (inst *Instance) ParseAndResolveRef(ctx context.Context, refStr string) (dsref.Ref, error) {
```
dustmop: Is this meant to be exposed to applications that embed qri as a library? Or by `cmd` or `api`? One of the reasons I put the Loader in its own package was so that `lib` could use it, but it wouldn't be accessible by consuming `lib`.

lib/resolve.go
```diff
@@ -0,0 +1,92 @@
...
+	ref, err := dsref.Parse(refStr)
+
+	// bad case references are allowed-but-warned for backwards compatibility
+	if errors.Is(err, dsref.ErrBadCaseName) {
```
dustmop, line 23: In the near future this should be a fatal error for every command *except* rename. Whether it's ignorable or not depends upon the caller, so it doesn't make sense to go this low in the stack.

lib/resolve.go
```diff
@@ -0,0 +1,92 @@
...
+			continue
+		}
+
+		return err
+	}
+
+	return resolve.ErrCannotResolveName
+}
+
+// TODO (b5) - this needs to move down into base, replacing base.LoadDataset with
```
dustmop, line 65: This sounds good, I'm a fan.

logbook/logbook.go
```diff
@@ -211,6 +212,57 @@ func (book *Book) DeleteAuthor() error {
 	return fmt.Errorf("not finished")
 }
 
+// ResolveRef finds the identifier & head path for a dataset reference
+// implements resolve.NameResolver interface
+func (book *Book) ResolveRef(ctx context.Context, ref *dsref.Ref) error {
```
dustmop: I don't think logbook should be relied on at all for resolving references. It conceivably can, by iterating over each branch and collapsing them, but it's not designed for that, and I can anticipate many problems arising from using this method. If this code is never meant to be called, it's effectively dead code, and its existence can lead to mistakes, confusion, and increased maintenance cost.

As mentioned in https://github.com/qri-io/qri/issues/1305, logbook should be accepting initIDs for its methods, and therefore needs some method in the interim that converts refs to initIDs, but that should be treated as temporary, and not an intrinsic part of how it is to be used.

Let's say, in the near future, we implement branching and rewinding to old versions. How well does logbook handle that?

repo/fs/fs.go
```diff
@@ -79,6 +83,35 @@ func NewRepo(store cafs.Filestore, fsys qfs.Filesystem, book *logbook.Book, cach
...
+		ref.Username = r.profile.Peername
+	}
+
+	match, err := r.GetRef(reporef.RefFromDsref(*ref))
+	if err != nil {
+		return resolve.ErrCannotResolveName
+	}
+
+	// TODO (b5) - repo doens't store IDs yet, breaking the assertion that ResolveRef
+	// will set the ID of a dataset. Need to fix that before we can ship this
```
dustmop, line 101: This is strong evidence that putting the same interface onto each subcomponent doesn't make sense. Repo fundamentally doesn't have the same surface area as the top-level resolver. It doesn't know about initID because it doesn't know about logbook.

dustmop:
Looking closer at the changes in this PR, no code anywhere (external to `resolve`) is calling `lib.Instance.ResolveRef`; there are only calls to `lib.Instance.ParseAndResolveRef` whose implementation then calls ResolveRef. This further backs up the idea that the top-level interface is not what callers actually want to be using, and is fundamentally different than the subcomponents, so there's no need to have a consistent interface between the top-level and the parts therein.

dsref/ref.go
b5: Sounds good! InitID it is.

dscache/dscache.go
b5: totally. felt icky while putting it in, I'll move it up

lib/resolve.go
b5: we could just expose `Instance.ResolveRef`, but yes I think reference resolution should be exported. It would be very helpful when instrumenting resolution times within a testground test, especially in the p2p context.

Other places where I'd want to see this exported are admittedly esoteric, but pinging the network for a dataset name you expect to exist at some point is another application. basically anything where you're looking to build an index from network calls without moving around any data beyond a reference the way `get` and `log` do.

repo/fs/fs.go
b5: No, this is evidence that `repo.RefStore` is an API we're intentionally letting rot on the vine. It doesn't have the same surface area as the same top level resolver because we're prioritizing work dscache, which _does_ have the the same surface area when it comes to this definition of reference resolution.

This is a deviation from the behavior that `ResolveRef` expects. We can solve this with an error to indicate different behaviour.

b5:
I totally take your point about not needing the same interface everywhere. In the long term I don't think we need strict adherence just this one Method, and if the behavior of different implementations needs to deviate, I think that's totally fine. But I'm trying to answer an "easier" problem first.

Reference resolution _must_ be the answer to a well-defined question: for a given dataset reference, what is the InitID and path to HEAD? Lots of subsystems in qri store and work with references. All of these subsystems need to "agree" on the current value of a reference. "Agree" is a very loose term here:

* FSI linking is supported in some subsystems, but not others. `dscache` will know about an FSI-linked directory, and can resolve to it. If you ask `dscache` and `logbook` for the HEAD Path of an FSI-linked dataset, you're going to get different answers, but if you get different InitIDs, that's an error.
* Dataset versions can be out on the network that we don't know about locally. Renames alter how to refer to an InitID, and pulling this can mess with local state (eg: if we put `InitID` in `.qri-ref` file is a the canonical source of InitID ultimately rests with the qri repo)
* Versions of histories can exist on the p2p network that don't exist on the registry
* Subsystems get added, removed, change, get corrupted, and rely on each other for recovery

So while I totally agree that long term each subsystem may have it's own, more nuanced input for reference resolution, we need a baseline to work from, for sorting heads from tails here. At a bare minimum something like `ResolveRef` should be used as an integrity check within integration tests. 

> Looking closer at the changes in this PR, no code anywhere (external to resolve) is calling lib.Instance.ResolveRef; there are only calls to lib.Instance.ParseAndResolveRef whose implementation then calls ResolveRef. This further backs up the idea that the top-level interface is not what callers actually want to be using, and is fundamentally different than the subcomponents, so there's no need to have a consistent interface between the top-level and the parts therein.

Let me add more examples to this PR before we make a call on this. I'm with you in not needing to enforce consistency between top-level and lower level reference resolution. I consider it a nice-to-have, and if it happens to survive, lovely. No tears will be shed if these deviate. One could argue that `ParseAndResolveRef` is already a deviation.

My concern is we're focusing on the wrong thing here, This PR is meant to be a stepping stone to doing reference resolution that can easily fall back to the network. One way to figure that out is to write a little more code & see what shakes out the other side. If all this does is get us some other, better interface, it'll be time well spent.

-- --

> I like this idea, but I think we should use a fsi:// protocol instead. We're stuck with "/ipfs" for now, but should avoid compounding that existing complication.

We're going to need an RFC to work this out. putting this in dataset names will break compatibility with URLS, but I totally agree we should be doing something along those lines. My concern: the reason `/ipfs` is antiquated is because IPFS's original pathing system didn't jive well with URLS. Let's not make the same mistake.

b5:
Oooook, after a bunch of thinking & work in other areas, back with an update to this "PRFC".

### `lib.ResolveRef` -> `lib.ResolveReference`
@dustmop, seems you were right to believe the top level resolver doesn't / shouldn't match the same `RefResolver` interface used by subsystems. Our work on recent RFCs have made many mentions of a `--remote` flag that allows the user to control where qri goes to perform requests. 

Because the top level resolver composes a number of potential sources, this `remote` flag should be fed directly to the resolver, telling it where it can and cannot look. to resolve a reference. Adding a "desired source" parameter to the top level resolver makes for a nice, expressive API. users can specify the empty string to get sensible default resolution, a specific remote address to resolve from, or one of a few keywords to get common resolver compositions like `local` or `network`.

To keep things fast, I've also added a `ParallelResolver` function that calls multiple resolvers at once.

logbook/logbook.go
```diff
@@ -207,6 +207,52 @@ func (book *Book) DeleteAuthor() error {
...
+	}
+
+	log, err := book.DatasetRef(ctx, *ref)
+	if err != nil {
+		return "", dsref.ErrNotFound
+	}
+	ref.InitID = log.ID()
+
+	if ref.Path == "" {
+		br, err := book.BranchRef(ctx, *ref)
```
dustmop, line 224: From https://github.com/qri-io/qri/blob/master/logbook/logbook.go#L708
"TODO(dustmop): Do not add new callers to this, transition away (preferring branchLog instead)"

Using branchLog will give you a strongly typed log, which can be passed to latestSavePath to make sure it's never called with a datasetLog instead.

Same with the call to `DatasetRef` above. That should use `datasetLog`, which can then use `dlog.InitID()` to get the initID.

dsref/ref.go
```diff
@@ -2,6 +2,9 @@ package dsref
 
 // Ref is a reference to a dataset
 type Ref struct {
+	// InitID is the canonical identifer for a dataset history, in Qri canonical
```
dustmop, line 5: A bit repetitive, just keep the first part before the comma

dsref/resolve.go
```diff
@@ -0,0 +1,137 @@
...
+type Resolver interface {
+	// ResolveRef uses ref as an outParam, setting ref.ID and ref.Path on success
+	// some implementations of name resolution may make network calls
+	// the returned "source" value should be either an empty string, indicating
+	// the source was resolved locally, or the multiaddress of the network
+	// address that performed the resolution
+	ResolveRef(ctx context.Context, ref *Ref) (source string, err error)
+}
+
+// MemResolver holds maps that can do a cheap version of dataset resolution,
```
dustmop, line 25: Feels like this should remain in its own source file "mem_resolve.go". A lot of our codebase follows that pattern, interface in one file and mem-version in another, and I think it's a good thing.

dsref/resolve.go
```diff
@@ -0,0 +1,137 @@
...
+
+	ref.InitID = id
+	if ref.Path == "" {
+		ref.Path = resolved.Path
+	}
+
+	return "", nil
+}
+
+type parallelResolver []Resolver
```
dustmop, line 82: This implementation could be put at the `lib` level, nothing else should be using it except for the top-level resolver.

keys/keys.go
```diff
@@ -0,0 +1,55 @@
+package keys
```
dustmop, line 1: This seems like it's from a separate PR. Isn't being used anywhere, and is unrelated to the rest of the change. Could you stash and save it for later?

lib/datasets.go
```diff
@@ -292,11 +248,11 @@ func (m *DatasetMethods) Get(p *GetParams, res *GetResult) error {
 			return err
 		}
 
-		if dr.Path == "" && res.FSIPath != "" {
+		if strings.HasPrefix(ref.Path, "/fsi") {
```
dustmop, line 251: Could this function, that checks if a ref.Path is an FSI path, and the one below, which trims the prefix, be put into a utility function? It's also being used in `loadDataset`, and I have a feeling it's going to show up a few other places as well.

keys/keys.go
b5: whoops, yes this shouldn't be here

lib/log.go
```diff
@@ -31,7 +32,9 @@ func NewLogMethods(inst *Instance) *LogMethods {
 type LogParams struct {
 	ListParams
 	// Reference to data to fetch history for
-	Ref string
+	Ref    string
+	Pull   bool
+	Remote string
```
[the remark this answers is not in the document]
b5: agreed, let's go with `Source`, which will often map to the `--remote` flag in `cmd` and `api`

lib/datasets.go
b5: yes, the right place for this should be `qfs.PathKind`: https://github.com/qri-io/qfs/blob/a6a7b98f0fa0969f6eae50fe14c2c6b5b32f4b4f/fs.go#L76-L87

Need to PR in a fix there first tho. I'll add a `TODO`

dsref/resolve.go
b5: I'm hoping to use this in `p2p` tests, possibly it's implementation as well

lib/resolve.go
```diff
@@ -0,0 +1,138 @@
...
+		return ref, "", fmt.Errorf("%q is not a valid dataset reference: %w", refStr, err)
+	}
+
+	resolvedSource, err := inst.ResolveReference(ctx, &ref, source)
+	return ref, resolvedSource, err
+}
+
+// ResolveReference finds the identifier & HEAD path for a dataset reference.
+// the source parameter determines which subsystems of Qri to use when
+func (inst *Instance) ResolveReference(ctx context.Context, ref *dsref.Ref, source string) (string, error) {
```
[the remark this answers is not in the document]
b5: > I guess this is part of my larger dislike of using strings for things that are actually structured data and/or enums.

We can't enumerate all cases at compile time because we want to support exact remote addresses, which in my mind means the set of all possible values to resolve from aren't properly captured by an enum or tuple. The inclination makes a lot of sense to me. Something akin to this:


```go
package lib

type ResolutionMode struct {
  Mode uint8  // specify how to do reference resolution
  Address string //
}

var (
  RMLocal = ResolutionMode{1, ""}
  RMNetwork = ResolutionMode{2, ""}
  RMRegistry = ResolutionMode{3, ""}
)
```

But we have strings coming from packages like `cmd`, which means we end up with something like this:

```
package lib

func ParseResolutionMode(str string) (Mode, error) {
  switch str {
    case "local":
      return RMLocal, nil
    //.. 
    default:
      // TODO - validate address
      return ResolutionMode{4,str}, nil
  }
}
```

So if we have to accept strings, and return Modes, why not just skip straight to strings? In my mind this lets `lib` users skip learning a data structure that can be explained via a comment.

The other issue, because users can specify remote names in addresses, we need configuration to be able to validate an address. If `ParseResolutionMode` from above could be a pure function I'd be more inclined to vote for the mode/enum variation

lib/resolve.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func (inst *Instance) resolveSources(source string) ([]dsref.Resolver, error) {
+	switch source {
+	case "":
+		return []dsref.Resolver{
+			inst.dscache,
+			inst.repo,
+			dsref.ParallelResolver(
+				inst.logbook,
```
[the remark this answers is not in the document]
b5: disagree that it causes problems. Yes it's slow, but logbook is only real source of truth here.

In recent refactoring I think I've found a nicer way to "deal with the slow": only falling back to `logbook.ResolveRef` within repo implementaionts. I want to avoid updating the `refstore` code to use InitIDs, having repo fall back to logbook means it'll only get called if dscache doesn't exist.

lib/resolve.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+
+		return resolvedSource, nil
+	}
+
+	return "", dsref.ErrNotFound
+}
+
+func (inst *Instance) resolveSources(source string) ([]dsref.Resolver, error) {
+	switch source {
```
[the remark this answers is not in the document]
b5: I think we're speaking to two different problems. To me the proposed code is great, and answers the problem of having fine grained control.

I wan to use this PR to answer some nagging issues of  inconsistent behaviour. Because users give us human reference strings, and we need to use both the network and local systems to get stable identifiers, we need a clear, enforceable behavior that's easy to test.

The problem is easier to see when at reference resolution over the network, where we want to measure both _internal_ consistency (all subsystems of a peer that retain references update with changes) and _network_ consistency (peers in the network agree on the same InitID). If we look at the networking parts of the codebase, lots of packages do some sort of reference resolution, they're just calling it by a different name:

* [`remote.ResolveHeadRef`](https://github.com/qri-io/qri/blob/2489d18cb091dbf381bf28acc259e9065316cbcc/remote/remote.go#L215-L224)
* [`remoteClient.ResolveHeadRefHTTP`](https://github.com/qri-io/qri/blob/2489d18cb091dbf381bf28acc259e9065316cbcc/remote/peer_sync_client.go#L247-L260)
* [`mockRemoteClient.ResolveHeadRef`](https://github.com/qri-io/qri/blob/2489d18cb091dbf381bf28acc259e9065316cbcc/remote/mock_client.go#L104-L107) 
* [`QriNode.RequestDataset`](https://github.com/qri-io/qri/blob/2489d18cb091dbf381bf28acc259e9065316cbcc/p2p/dataset.go#L16-L20)

There are more. If you broaden to include things like registry search (which implicitly resolves references), and logsync (which uses local resolution for comparison),
Many of these reference resolution functions are themselves compositions atop other subsystems, which if implemented incorrectly, will bubble up in strange ways. All of these are doing variations on the same behavior, inconsistently. We can't just _remove_ these because they're necessary. They're both ends of of a network call, or are a mock implementation intended to make testing easier.

By clearly defining an expected behaviour for any system that caches a reference, we can test that system for consistency against other subsystems of the same peer, and across the network. This is the real win that comes out of specified behavior: tests that make more sense.

I've updated this PR to include a spec test, an example of an internal consistency check (comparing dscache & logbook to confirm the hook is working), and a network consistency check (`lib.TwoActorRegistryIntegration` now checks all actors end in a consistent local state).

b5:
Alright, just finished another pass, some high-level changes:

#### Spec Tests
I've added `qri-io/qri/dsref/spec`, which defines a spec other tests can import to test that they properly implement `ResolveRef`

### Many things do some kind of reference resolution, all need to agree
I keep finding things that call `CanonicalizeDatasetRef` all over the codebase, and they need

### Resolving to an FSI path is a lib-level thing
I've changed the behavior of `ResolveRef` to only set the `Path` field to a content-addressed version, when it used to return `/fsi` if a linked directory exists. This removes lots of potential errors when resolving over a network. If a remote were to resolve a reference to a local FSI path... that would be bad.

Instead, I've added a new method to the `fsi` package: `ResolvedPath`, which is used by lib, and added `useFSI` as a flag on `lib.ParseAndResolveDatasetRef`. This PR includes examples where both make sense (_log_ _doesn't_ care about FSI, `get` _does_)

### Reference Resolution as a consistency check
This PR also uses new functions in the `dsrefspec` package to define a set of _consistency checks_, which confirm local subsystems and network nodes agree on head references. You can see examples in two places:
* `dsref.TestCacheRefConsistency` (`dscache/dscache_test.go`) which checks that logbook and dscache agree
* `lib.TwoActorRegistryIntegrationTest` (`lib/integration_test.go`) checks that all three nodes agree on the same reference after pushing & pulling

registry/regclient/client_test.go
```diff
@@ -0,0 +1,109 @@
...
...cBels3nm/fWScvjH4dMQXlavwcwPgKhy2NczDhr4X69oEw6Msd4hQiqJrlWd8juUg6vIsrl1wS/JAOCS65fuyJfV3Pw64rWbTPMwO3FOvxj+rFghZFQgjg/i45uHA2UUkM+h504M5Nzs6Arr/rgV7uPGR5e5OBw3lfiS9ZaA7QZiOq7sMy1L0qD49YO1ojqWu3b7UaMaBQx1Dty7b5IVOSYG+Y3U/dLjhTj4Hg1VtCHWRm3nMOE9cVpMJRhRzKhkq6gnZmni8obz2BBDF02X34oQLcHC/Wn8F3E8RiBjZDI66g+iZeCCUXvYz0vxWAQQKBgQDEJu6flyHPvyBPAC4EOxZAw0zh6SF/r8VgjbKO3n/8d+kZJeVmYnbsLodIEEyXQnr35o2CLqhCvR2kstsRSfRz79nMIt6aPWuwYkXNHQGE8rnCxxyJmxV4S63GczLk7SIn4KmqPlCI08AU0TXJS3zwh7O6e6kBljjPt1mnMgvr3QKBgQD6fAkdI0FRZSXwzygx4uSg47Co6X6ESZ9FDf6ph63lvSK5/eue/ugX6p/olMYq5CHXbLpgM4EJYdRfrH6pwqtBwUJhlh1xI6C48nonnw+oh8YPlFCDLxNG4tq6JVo071qH6CFXCIank3ThZeW5a3ZSe5pBZ8h4bUZ9H8pJL4C7yQKBgFb8SN/+/qCJSoOeOcnohhLMSSD56MAeK7KIxAF1jF5isr1TP+rqiYBtldKQX9bIRY3/8QslM7r88NNj+aAuIrjzSausXvkZedMrkXbHgS/7EAPflrkzTA8fyH10AsLgoj/68mKr5bz34nuY13hgAJUOKNbvFeC9RI5g6eIqYH0FAoGAVqFTXZp12rrK1nAvDKHWRLa6wJCQyxvTU8S1UNi2EgDJ492oAgNTLgJdb8kUiH0CH0lhZCgr9py5IKW94OSM6l72oF2UrS6PRafHC7D9b2IV5Al9lwFO/3MyBrMocapeeyaTcVBnkclz4Qim3OwHrhtFjF1ifhP9DwVRpuIg+dECgYANwlHxLe//tr6BM31PUUrOxP5Y/cj+ydxqM/z6papZFkK6Mvi/vMQQNQkh95GH9zqyC5Z/yLxur4ry1eNYty/9FnuZRAkEmlUSZ/DobhU0Pmj8Hep6JsTuMutref6vCk2n02jc9qYmJuD7iXkdXDSawbEG6f5C4MUkJ38z1t1OjA==`)
```
dustmop, line 81: You can replace this block of code with:

```
peerInfo := testPeers.GetTestPeerInfo(10)
pk1 := peerInfo.PrivKey

```

dscache/dscache_test.go
```diff
@@ -69,3 +74,81 @@ func TestDscacheAssignSaveAndLoad(t *testing.T) {
...
+		log.Fatal(err)
+	}
+	defer os.RemoveAll(tmpdir)
+
+	ctx := context.Background()
+	fs := localfs.NewFS()
+	path := filepath.Join(tmpdir, "dscache.qfb")
+	dsc := NewDscache(ctx, fs, nil, path)
+
+	dsrefspec.ResolverSpec(t, dsc, func(r dsref.Ref, author identity.Author, log *oplog.Log) error {
```
dustmop, line 90: For this situation where the callback doesn't use all of its parameters, could you rename them to `_`?

cmd/fetch_test.go
```diff
@@ -163,7 +163,7 @@ func TestFetchCommand(t *testing.T) {
...
+	expectBody := `{"data":[{"username":"peer_a","name":"test_movies","path":"/ipfs/QmXmnH1tFKyG493wsFiisZ14N4cjrymZZ6pqK3Vr9vWS2p","bodySize":720,"commitTime":"timeStampHere","commitTitle":"body changed by 70%","commitMessage":"body:\n\tchanged by 70%"},{"username":"peer_a","name":"test_movies","path":"/ipfs/QmNX9ZKXtdskpYSQ5spd1qvqB2CPoWfJbdAcWoFndintrF","bodySize":224,"commitTime":"timeStampHere","commitTitle":"created dataset from body_ten.csv","commitMessage":"created dataset from body_ten.csv"}],"meta":{"code":200},"pagination":{"nextUrl":"/history/peer_a/test_movies?page=2"}}`
```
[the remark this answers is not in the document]
b5: This is an important point we should discuss. This change drops the requirement that `ResolveRef` fill in the profileID. I'm going to look at all our code to see if this is even ok. If not, I can up the spec to require profileID be set. I'd like to keep requirements to only those which we're actually using.

@ramfox, does desktop rely on `profileID` values in datasets? I just took a quick look through the desktop codebase, couldn't find _any_ cases where it's used.

dsref/spec/resolve.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/multiformats/go-multiaddr"
+	"github.com/qri-io/dataset"
+	"github.com/qri-io/qfs"
+	"github.com/qri-io/qri/dsref"
+	"github.com/qri-io/qri/identity"
+	"github.com/qri-io/qri/logbook"
+	"github.com/qri-io/qri/logbook/oplog"
+)
+
+// PutFunc is required to run the ResolverSpec test, when called the Resolver
```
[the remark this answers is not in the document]
b5: v good point!

dsref/spec/resolve.go
```diff
@@ -0,0 +1,238 @@
+package spec
```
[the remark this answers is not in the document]
b5: sadly functions defined in `_test` files can't be imported into other packages 😢

fsi/fsi.go
```diff
@@ -64,6 +64,44 @@ func NewFSI(r repo.Repo, pub event.Publisher) *FSI {
 	return &FSI{repo: r, pub: pub}
 }
 
+// ResolvedPath sets the Path value of a reference to the filesystem
+// integration path if one exists, ignoring any prior Path value, returning true
+// if an FSI path exists.
+// setting an FSI path requires the reference already be resolved by a
+// dsref.Resolver
+func (fsi *FSI) ResolvedPath(ref *dsref.Ref) (bool, error) {
```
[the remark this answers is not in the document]
b5: fantastic! Even better given that `ErrNoLink` is already defined

lib/resolve.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/qri-io/dataset"
+	"github.com/qri-io/qri/base"
+	"github.com/qri-io/qri/base/dsfs"
+	"github.com/qri-io/qri/dsref"
+	"github.com/qri-io/qri/fsi"
+	"github.com/qri-io/qri/remote"
+)
+
+// ParseAndResolveRef combines reference parsing and resolution
+func (inst *Instance) ParseAndResolveRef(ctx context.Context, refStr, source string, useFSI bool) (dsref.Ref, string, error) {
```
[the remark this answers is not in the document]
b5: much prefer this change. Will make

api/log.go
```diff
@@ -83,6 +83,7 @@ func (h *LogHandlers) logHandler(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		if local && err == repo.ErrNotFound {
+			// TODO (b5): This should be 403 not 500
```
dustmop, line 86: Oops! I meant 404, my mistake! It's fine like this, just having a TODO is enough.

PR #1315, opened 2020-04-27
refactor(fetch into log): use `qri log` and `/history` endpoints for local and network log fetches

### LINKED ISSUE #1310: Merge `fetch` command into `log`
currently fetch serves one purpose: getting logbook data without saving it. We can make the log command do that, and drop the need for the command entirely.

what used to be:
```
$ qri fetch b5/world_bank_population
```

now becomes:
```
$ qri log b5/world_bank_population
```
If I don't have the reference, it'll be fetched.

ramfox:
TODO:
* [x] remove `fetch` command and `/fetch` endpoint
* [x] add pagination to fetch
* [x] fix confusing cloud/registry error when no log is found
* [x] need to open desktop pr to update actions to include `local` param when necessary

closes #1310 
related to qri-io/cloud#158
and qri-io/desktop#544 depends on this pr to merge before it can merge


cmd/log.go
```diff
@@ -80,29 +111,70 @@ func (o *LogOptions) Run() error {
 	// convert Page and PageSize to Limit and Offset
 	page := util.NewPage(o.Page, o.PageSize)
 
-	p := &lib.LogParams{
-		Ref: o.Refs.Ref(),
-		ListParams: lib.ListParams{
+	for _, ref := range o.Refs.RefList() {
```
b5, line 114: Just tried this to show multiple logs
```
qri log yashagarwal/all_week qri_eng/qri_cli_user_benchmarks
```
didn't work because max args is 1. I vote we stick with that and don't bother with the loop. listing multiple things at once is confusing.

lib/params.go
```diff
@@ -62,6 +62,12 @@ func ListParamsFromRequest(r *http.Request) ListParams {
 	return NewListParams(r.FormValue("orderBy"), page, pageSize)
 }
 
+// HasListParamsFromRequest returns true if either "page" or "pageSize" exist as form values
+// on an http.Request pointer
+func HasListParamsFromRequest(r *http.Request) bool {
+	return r.FormValue("page") != "" || r.FormValue("pageSize") != ""
+}
+
```
b5, line 70: Is this being used anywhere?

cmd/log_integration_test.go
```diff
@@ -16,7 +16,7 @@ func TestLogAndDeletes(t *testing.T) {
 	}
 
 	// Log should have exactly one version
-	err = run.ExecCommand("qri log me/log_test")
+	err = run.ExecCommand("qri log me/log_test --local")
```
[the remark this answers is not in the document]
ramfox: So you are right to be confused because it shouldn't affect anything aaaaannnnddd it doesn't actually affect anything. I just got a weird false positive when running tests yesterday, it kept hanging or just taking an extraordinarily long time to go, and for whatever reason, when I added the flags the hanging stopped. Removing the flags now.

lib/params.go
ramfox: ahh it was in a previous commit, removing

PR #1320, opened 2020-04-29
chore: remove update command, subsystem, api

b5:
Two problems that lead to the decision to remove `update`. First: it doesn't feel like part of the core value proposition of qri. There are better scheduler/task managers out there.

Second, it's too buggy, and generally unfinished. We haven't been able to make this work well enough to justify keeping it, and can't seem to find the bandwidth to justify improving it. We'll solve the update problem instead with docs on how to wire qri up to a cron job, and may one day revisit `update` as a plugin/wrapper.

Kill your darlings and whatnot.


b5:
apologies for beating you to this one @chriswhong, in the end I wanted to be the one to swing the axe

b5:
nice catch! I'm indeed looking to cover the config stuff in a follow up

PR #1321, opened 2020-05-01
refactor(logbook): Most logbook methods take an initID, not ref

dustmop:
Change most logbook methods, especially the ones that add operations to the logs, so that they accept initIDs instead. This will make it much easier to transition away from the old refstore.

Add a helper method that looks up the initID given a ref; this makes it possible to work with base before dscache is done.

Extract the initName helper out of mutations; callers should instead either have an initID or create one by using WriteDatasetInit.

Create strongly typed structs for each log level - AuthorLog, DatasetLog, BranchLog - to aid in understanding logbook code. They aren't used everywhere, but it's a start.

Remove CronJob Model now that update is gone.

### DIFF
--- a/fsi/init.go
+++ b/fsi/init.go
@@ -197,7 +197,7 @@ func (fsi *FSI) InitDataset(p InitParams) (name string, err error) {
 		}
 	}
 
-	if err = fsi.repo.Logbook().WriteDatasetInit(context.TODO(), ref.Name); err != nil {
+	if _, err = fsi.repo.Logbook().WriteDatasetInit(context.TODO(), ref.Name); err != nil {
 		if err == logbook.ErrNoLogbook {
 			rollback = func() {}
 			return name, nil
--- a/fsi/init.go
+++ b/fsi/init.go
@@ -38,7 +38,7 @@ var PrepareToWrite = func(comp component.Component) {
 }
 
 // InitDataset creates a new dataset
-func (fsi *FSI) InitDataset(p InitParams) (name string, err error) {
+func (fsi *FSI) InitDataset(p InitParams) (refstr string, err error) {
 	// Create a rollback handler
 	rollback := func() {
 		log.Debug("did rollback InitDataset due to error")
@@ -99,13 +99,13 @@ func (fsi *FSI) InitDataset(p InitParams) (name string, err error) {
 		return "", err
 	}
 
-	ref := &reporef.DatasetRef{Peername: "me", Name: p.Name}
+	datasetRef := reporef.DatasetRef{Peername: "me", Name: p.Name}
 
 	// Validate dataset name. The `init` command must only be used for creating new datasets.
 	// Make sure a dataset with this name does not exist in your repo.
-	if err = repo.CanonicalizeDatasetRef(fsi.repo, ref); err == nil {
+	if err = repo.CanonicalizeDatasetRef(fsi.repo, &datasetRef); err == nil {
 		// TODO(dlong): Tell user to use `checkout` if the dataset already exists in their repo?
-		return "", fmt.Errorf("a dataset with the name %s already exists in your repo", ref)
+		return "", fmt.Errorf("a dataset with the name %s already exists in your repo", datasetRef)
 	}
 
 	// Derive format from --source-body-path if provided.
@@ -123,8 +123,8 @@ func (fsi *FSI) InitDataset(p InitParams) (name string, err error) {
 
 	// Create the link file, containing the dataset reference.
 	var undo func()
-	if name, undo, err = fsi.CreateLink(targetPath, ref.AliasString()); err != nil {
-		return name, err
+	if refstr, undo, err = fsi.CreateLink(targetPath, datasetRef.AliasString()); err != nil {
+		return refstr, err
 	}
 	// If future steps fail, rollback the link creation.
 	rollback = concatFunc(undo, rollback)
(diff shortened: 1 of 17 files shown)


logbook/oplog/log.go
```diff
@@ -75,6 +75,12 @@ type Logstore interface {
 	// use Logstore.Children or Logstore.Descendants to populate missing children
 	HeadRef(ctx context.Context, names ...string) (*Log, error)
 
+	// RetrieveRef returns the log that matches the username and dataset name
+	RetrieveRef(username, dsName string) (*Log, error)
+
+	// RetrieveByInitID returns the log that matches the initID
+	RetrieveByInitID(initID string) (*Log, error)
```
b5, line 82: This is equivalent to `Logstore.Log(ctx, initID)`, please use that instead.

Given you couldn't find this method easily, it's clear to me the `Log` method on the `Logstore` interface is poorly named. Should we rename to `Logstore.Get`?

logbook/types.go
```diff
@@ -0,0 +1,85 @@
+package logbook
```
b5, line 1: this is _fantastic_. Feels like what the logbook package should have been doing all along

logbook/types.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/qri-io/qri/logbook/oplog"
+)
+
+// AuthorLog is the top-level log representing the book author
+type AuthorLog struct {
+	l *oplog.Log
+}
+
+// TODO(dustmop): Consider changing the "Append" methods to type-safe methods that are specific
+// to each log level, which accept individual parameters instead of type-unsafe Op values.
```
b5, line 13: in2 it

fsi/init.go
```diff
@@ -197,7 +197,7 @@ func (fsi *FSI) InitDataset(p InitParams) (name string, err error) {
 		}
 	}
 
-	if err = fsi.repo.Logbook().WriteDatasetInit(context.TODO(), ref.Name); err != nil {
+	if _, err = fsi.repo.Logbook().WriteDatasetInit(context.TODO(), ref.Name); err != nil {
```
b5, line 200: I'd like make sure we're returning this dropped InitID. Can we add a TODO saying `fsi.InitDataset` should return a `dsref.Ref` once it gets an `InitID` field, and set ref.InitID here?

logbook/types.go
```diff
@@ -0,0 +1,85 @@
+package logbook
+
+import (
+	"github.com/qri-io/qri/logbook/oplog"
+)
+
+// AuthorLog is the top-level log representing the book author
+type AuthorLog struct {
+	l *oplog.Log
+}
```
b5, line 10: I'd prefer this be `UserLog` and be generalized to _any_ top level user logbook entry, then have a method `authorLog` that returns the author, which is a preconfigured instance of a `UserLog`.

logbook/logbook.go
```diff
@@ -371,21 +360,72 @@ func (book *Book) WriteDatasetRename(ctx context.Context, ref dsref.Ref, newName
...
+// TODO(dustmop): Don't depend on this function permanently, use a higher level resolver and
+// convert all callers of this function to use that resolver's initID instead of converting a
+// dsref yet again.
+func (book *Book) RefToInitID(ref dsref.Ref) (string, error) {
+	if book == nil {
+		return "", ErrNoLogbook
+	}
+
+	// RetrieveRef is inefficient, iterates the top two levels of the logbook.
+	// Runs in O(M*N) where M = number of users, N = number of datasets per user.
```
b5, line 380: It's worse than this. Calculating the `Name()` of a log requires iterating all operations in the log.

Rather than bury a comment about time complexity in the function body, please declare it in the function description.

logbook/logbook.go
```diff
@@ -574,18 +591,19 @@ func (book *Book) WritePublish(ctx context.Context, ref dsref.Ref, revisions int
 
 // WriteUnpublish adds an operation to a log marking an unpublish request for a
 // count of sequential versions from HEAD
-func (book *Book) WriteUnpublish(ctx context.Context, ref dsref.Ref, revisions int, destinations ...string) error {
+// TODO(dustmop): Currently unused by codebase, only called in tests.
```
b5, line 594: nice catch!

logbook/oplog/log.go
```diff
@@ -75,6 +75,12 @@ type Logstore interface {
 	// use Logstore.Children or Logstore.Descendants to populate missing children
 	HeadRef(ctx context.Context, names ...string) (*Log, error)
 
+	// RetrieveRef returns the log that matches the username and dataset name
+	RetrieveRef(username, dsName string) (*Log, error)
```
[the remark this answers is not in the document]
dustmop: Not sure why I added this, will change to `HeadRef`.

logbook/oplog/log.go
dustmop: You're right, totally missed the `Log` method. Will rename it to `Get`.

fsi/init.go
dustmop: Will do. Avoided this at first because it still gets dropped, but having the TODO around is a good stop gap.

logbook/types.go
dustmop: :+1:

Arqu:
Love it, this will be a massive boost to next sprints work on cloud.
Also very much into the idea of being more deterministic and firm with logbook ops (for a lack of better words). It's our backbone and stability and clariti are 100% top priority.

PR #1333, opened 2020-05-08
refactor(base): base.SaveDataset takes an initID and path

dustmop:
Require an initID in base.SaveDataset. Now lib.Save has to create an initID first, or look one up for an existing dataset. Doing this requires deciding on a dataset name before entering base, meaning name inference moves up to lib.

CreateDataset does not take an initID, it is only responsible for putting the dataset into block storage and returning the new path. This means it no longer updates logbook. For now, it does update the refstore, so it needs a dataset name, but both of these facts will change in the near future.

Add todos about upcoming changes to transform and apply and dry-run.

Add a testRunner for base.


base/save.go
```diff
@@ -21,58 +21,45 @@ import (
...
 	if prevPath != "" {
-		log.Debugf("loading previous path: %s", prevPath)
-		if sw.NewName && inferredName != "" {
-			// Using --new flag, name was inferred, but it's already in use. Because the --new
-			// flag was given, user is requesting we invent a unique name. Increment a counter
-			// on the name until we find something that's available.
-			dsName = GenerateAvailableName(r, peername, dsName)
-			prev, mutable, prevPath, err = PrepareHeadDatasetVersion(ctx, r, peername, dsName)
+		// Load the dataset's most recent version, which will become the previous version after
+		// this save operation completes.
```
b5, line 39: lovely comment

base/save.go
```diff
@@ -269,3 +252,75 @@ func GenerateAvailableName(r repo.Repo, peername, prefix string) string {
...
+	if err == repo.ErrNotFound {
+		return false
+	}
+	return true
+}
+
+// DatasetNameAndStableIdentifers determines the final name for the dataset, by inferring one if
+// necessary, and returns stable identifiers for the full dataset history, and for the most
+// recent version.
+func DatasetNameAndStableIdentifers(ctx context.Context, r repo.Repo, peername string, dsName string, ds *dataset.Dataset, newName bool) (string, string, string, error) {
```
b5, line 276: I think this should return `(dsref.Ref, error)` and be destructured up in lib.

base/save.go
```diff
@@ -144,13 +132,24 @@ func SaveDataset(ctx context.Context, r repo.Repo, str ioes.IOStreams, changes *
 	// let's make history, if it exists
 	changes.PreviousPath = prevPath
 
-	// TODO(dustmop): Remove the need to assign this. See inside of CreateDataset for details
-	changes.Name = dsName
-	return CreateDataset(ctx, r, str, changes, prev, sw)
+	// Write the dataset to storage and get back the new path.
+	// TODO(dustmop): Only return the cafs path, since this function shouldn't know about references
```
b5, line 136: very into making this change

base/save.go
```diff
@@ -269,3 +252,75 @@ func GenerateAvailableName(r repo.Repo, peername, prefix string) string {
...
+		if newName && inferredName != "" {
+			// Using --new flag, name was inferred, but it's already in use. Because the --new
+			// flag was given, user is requesting we invent a unique name. Increment a counter
+			// on the name until we find something that's available.
+			dsName = GenerateAvailableName(r, dsName)
+		} else if newName {
+			// Name was explicitly given, with the --new flag, but the name is already in use.
+			// This is an error.
+			// TODO(dlong): Add a test for this case.
+			return "", "", "", fmt.Errorf("dataset name has a previous version, cannot make new dataset")
```
b5, line 291: use `github.com/qri-io/qri/errors` to distinguish user feedback from a machine-oriented error. I think we should have an `ErrNameTaken` defined in `base`:
```go
var ErrNameTaken = fmt.Errorf("name already in use")
```
And that should be used here:
```go
return dsref.Ref{}, qerr.New(ErrNameTaken, "dataset name has a previous version, cannot make new dataset")
```

base/save.go
```diff
@@ -269,3 +252,75 @@ func GenerateAvailableName(r repo.Repo, peername, prefix string) string {
...
+		} else if inferredName != "" {
+			// Name was inferred, and has previous version. Unclear if the user meant to create
+			// a brand new dataset or if they wanted to add a new version to the existing dataset.
+			// Raise an error recommending one of these course of actions.
+			return "", "", "", fmt.Errorf("inferred dataset name already exists. To add a new commit to this dataset, run save again with the dataset reference. To create a new dataset, use --new flag")
```
b5, line 296: Ditto for wrapping `ErrNameTaken` with this more readable error:

```go
return dsref.Ref{}, qerr.New(ErrNameTaken, "inferred dataset name already exists. To add a new commit to this dataset, run save again with the dataset reference. To create a new dataset, use --new flag")
```

base/save.go
dustmop: That's a very good idea. This requires added the `InitID` field to dsref, which https://github.com/qri-io/qri/pull/1308 also does, so I'm going to copy it over to this PR.

base/save.go
dustmop: Will do. Just to clarify, the parameter on the left is the machine-oriented error, and the parameter on the right is the user-destined message?

PR #1334, opened 2020-05-11
refactor(lib): rename RenderRequests -> RenderMethods, use an instance

### LINKED ISSUE #1235: cleanup: Rename lib methods containers away from "Requests"
The objects that contain the interface of lib used to be "[something]Requests" like DatasetRequests. We renamed some, such as ProfileMethods (used to be ProfileRequests), but did not finish the job.

Take a look at the function `Receivers` in `lib/lib.go`. All of the objects constructed in there should be using a name that ends in "Methods". We should also rename the `CoreRequestsName` interface function to `CoreMethodsName`.

b5:
our current plans have us keeping the `render` functionality around. This brings render up to pace.

Considering we'll remove `ExportRequests` with the PR that does the merging work, I think this PR closes #1235.

### DIFF
--- a/lib/render.go
+++ b/lib/render.go
@@ -3,7 +3,6 @@ package lib
 import (
 	"context"
 	"fmt"
-	"net/rpc"
 
 	"github.com/qri-io/dataset"
 	"github.com/qri-io/qri/base"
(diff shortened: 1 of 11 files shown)


lib/render.go
```diff
@@ -13,29 +13,23 @@ import (
 	reporef "github.com/qri-io/qri/repo/ref"
 )
 
-// RenderRequests encapsulates business logic for this node's
+// RenderMethods encapsulates business logic for this node's
 // user profile
-// TODO (b5): switch to using an Instance instead of separate fields
-type RenderRequests struct {
+type RenderMethods struct {
 	cli  *rpc.Client
```
dustmop, line 19: Can we also get rid of `cli` here?

b5: yes, we can

Issue #1176, opened 2020-03-12
Should we merge `export` command into get?


b5:
From discord chat:

> wondering  whether qri export has a future in general. The differences from qri checkout seem nuanced. (I think just that checkout also links the resulting output dir? Is there anything else?) It seems like you could accomplish the same with several calls to qri get. Not sure whether it’s an important enough use-case to have a shortcut that might be confusing to many users.
- @Mr0grog 

> It keeps being a problem that export is misunderstood which is a sign that it's not pulling its own weight
- @dustmop 

RFC0014 defined a need for Qri Export:
https://github.com/qri-io/rfcs/blob/master/text/0014-export.md

Then we shipped support for File System integration in RFC0025, which solved a lot of use cases export used to cover.

What If we merged all of export's functionality into `get`?

Qri Desktop relies on the `/export` endpoint to get versions, we'd need to migrate that to use get.

dustmop:
In my opinion, the RFC that is most relevant to this question is this one:
https://github.com/qri-io/rfcs/blob/master/text/0021-export_behavior.md

As this was written before we made FSI (file system integration), we can see that much of the "guide-level discussion" is no longer applicable. Most of these questions of allowing qri datasets to exist outside of our repository in a different representation is now easily doable by using `qri checkout`. The remaining capabilities could all be integrated into `qri get`, by allowing conversion into other file format. Even the high-fidelity export, which we never fully implemented, could be folded in as a new format that we define.

Export isn't doing enough work on its own to continue to exist, considering all of the confusion it causes.

> Qri Desktop relies on the `/export` endpoint to get versions, we'd need to migrate that to use get.

Is this true? I thought that the `/export` endpoint always writes to disk, and doesn't return any body. I thought it was only using `/export` for right-clicking and choosing "export".

Mr0grog:
Reading these RFCs and playing with the command, I now realize my mental model for what this command does is *way* off from reality. I think my understanding was fairly close to RFC 14, but actual functionality is closer to RFC 21, and is very different.

For instance, I thought these two commands were pretty close to the same:

```sh
# A basic checkout:
$ qri checkout me/awesome_data

# vs. export + link:
$ qri export -o awesome_data me/awesome_data
$ cd awesome_data
$ qri fsi link me/awesome_data
$ cd ..
```

[...]

PR #1335, opened 2020-05-11
feat(export): Can get a zip of a dataset using `get --format zip`

dustmop:
As part of https://github.com/qri-io/qri/issues/1176, a user can get a zip of their dataset like this:

```
qri get me/my_ds --format zip
```

By default, this will create a zip in the current directory named `my_ds.zip`. Also add an `-o` flag to get, so a user could instead do:

```
qri get me/my_ds --format zip -o export.zip
```

If the command is running through a pipe or redirect, and no `-o` flag is given, instead write the zip's bytes to stdout, so this command works:

```
qri get me/my_ds --format zip > export.zip
```

The same thing happens on the API, where a request like `/me/my_ds?format=zip` will receive an application/zip with a body response of the zip's bytes.

Some refactorings:
* The zip file contains a "qri-ref.txt" file, which is created the same way as FSI's ".qri-ref" file.
* Move hidden_file to it's own package to support this.
* Expand on API's TestRunner a bit so it's usable for the zip_test.


base/hiddenfile/hidden_file.go
```diff
@@ -1,6 +1,6 @@
 // +build !windows
 
-package base
+package hiddenfile
```
b5, line 3: how would you feel about merging both `linkfile` and this new `hiddenfile` into `fsi` instead? I feel like they're both part of our story around filesystem integration. Ended up needing to write a code spike to confirm no import cycles, here's a commit if you'd like to cherry-pick:

https://github.com/qri-io/qri/tree/hidden_and_link_file_to_fsi

base/archive/zip_test.go
```diff
@@ -101,34 +112,36 @@ func TestWriteZipFullDataset(t *testing.T) {
 	}
 }
 
-func TestUnzipDatasetBytes(t *testing.T) {
-	path := zipTestdataFile("exported.zip")
-	zipBytes, err := ioutil.ReadFile(path)
-	if err != nil {
-		t.Fatal(err)
-	}
+// TODO(dustmop): Rewrite zip importing
```
b5, line 115: Can we get an issue for this?

api/zip_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"io/ioutil"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/qri-io/dataset"
+)
+
+func TestGetZip(t *testing.T) {
```
b5, line 11: Would prefer to see all of this moved into `get_test.go`

base/archive/zip.go
```diff
@@ -5,124 +5,91 @@ import (
...
 
-		if ds.Viz.RenderedPath != "" {
-			if err = maybeWriteRenderedViz(ctx, store, zw, ds.Viz.RenderedPath); err != nil {
-				if errors.Is(err, context.DeadlineExceeded) {
-					err = nil
-				} else {
-					return err
-				}
-			}
+		// TODO(dustmop): The transform component outputs a json file, with a path string
```
b5, line 64: Would love to see an issue for this as well

api/zip_test.go
dustmop: Do you mean rename this to `get_test.go`? There isn't a file with that name in the `api/` package.

b5: Ooooops, yes rename then!

Issue #1249, opened 2020-04-03
Segmentation violation for qri diff structure.json


hbruch:
## Expected behavior
`qri diff structure.json` executed in the working dir of a dataset should return the diff compared to last saved version.

## Actual behavior
A  segmentation violation occurs:

```
$ qri diff structure.json
for linked dataset [mfdz/nvbw_haltestellen]

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x50c3d01]

goroutine 1 [running]:
github.com/qri-io/qri/lib.(*DatasetRequests).Diff(0xc000410000, 0xc0003e02a0, 0xc0002c9bc0, 0x0, 0xc000650120)
	/Users/hbruch/git/go-code/qri/lib/diff.go:229 +0xb11
github.com/qri-io/qri/cmd.(*DiffOptions).Run(0xc0002ce000, 0x5a2fb60, 0xc00031a370)
	/Users/hbruch/git/go-code/qri/cmd/diff.go:130 +0x12a
github.com/qri-io/qri/cmd.NewDiffCommand.func1(0xc0003bec80, 0xc000121170, 0x1, 0x1, 0x0, 0x0)
	/Users/hbruch/git/go-code/qri/cmd/diff.go:56 +0x97
github.com/spf13/cobra.(*Command).execute(0xc0003bec80, 0xc000121130, 0x1, 0x1, 0xc0003bec80, 0xc000121130)
	/Users/hbruch/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:826 +0x453
github.com/spf13/cobra.(*Command).ExecuteC(0xc00034c000, 0xc000138390, 0x57a3bb0, 0x5a024c0)
	/Users/hbruch/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914 +0x2fb
github.com/spf13/cobra.(*Command).Execute(...)
	/Users/hbruch/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864
github.com/qri-io/qri/cmd.Execute()
	/Users/hbruch/git/go-code/qri/cmd/cmd.go:53 +0x153
main.main()
	/Users/hbruch/git/go-code/qri/main.go:16 +0x20
```

## Additional info
qri: HEAD (c21419b3a0df46e)
OS: MacOS Mojave 10.14.6 (18G103)

dustmop:
[...]

Issue #1336, opened 2020-05-12
lib should take tests from cmd package


dustmop:
My bad on this one. I added a lot of tests to the `cmd` package that exercise tons of functionality in qri. This has prevented many regressions, and introduced the `testRunner` pattern, which has proven to be quite useful. However, by exclusively testing at the `cmd` level, we sometimes miss problems in the `api`, and have made the `cmd` package pretty slow to test, taking up to 50 seconds at times (assuming the tests complete at all). We also have avoided adding the same tests to `lib`, which both means our `lib` methods have weird or confusing interfaces, and technically leads to lower coverage of `lib` code.

We should move tests out of `cmd` and down to `lib`, utilizing the `testRunner` pattern whenever possible. `lib` should be the place that exercises "all possible edge cases" for the methods it surfaces. What remains at `cmd` should be:
* at least one test per command
* tests that specifically rely on how `cmd` outputs to stdout or stderr
* high level integration tests that are difficult to do at `lib` alone

VarsMolta:
I'd love to work on this issue. Seems like this involves a bunch of refactoring (moving test cases around) and ensuring coverage which I should be comfortable with.

PR #1341, opened 2020-05-13
refactor(diff): Refactor diff for clarity. Use lib.TestRunner

dustmop:
Diff needed a bit of work, as its implementation has been a bit tricky to understand, and there were some known issues reported recently. We also want to move some tests down from `cmd` to `lib` and this is a good opportunity to do so. In addition, make the lib.Diff interface easier to understand by making explicit the cases that it will handle.

Fixes https://github.com/qri-io/qri/issues/1249. Makes some in-roads on https://github.com/qri-io/qri/issues/1336. Adds more tests to diff that didn't exist anywhere.


dustmop:
Screen Shot 2020-05-13 at 6 39 44 PM

PR #1349, opened 2020-05-18
feat(dscache): Dscache used for fsi init and checkout

dustmop:
Dscache is updated by both logbook and fsi using a hook, which is defined in dsref/hook. This hook represents any sort of dataset change. In order to make this work properly, dscache should not live on the repo anymore, but instead should only be on lib.Instance. This is because the lib.Instance has an FSI struct, but the repo does not (it only has logbook). There are some hacks in place to workaround this until it is fixed.


dsref/hook/change.go
```diff
@@ -0,0 +1,39 @@
+package hook
```
b5, line 1: Ok, definitely don't want this as a subpackage of dsref, but aside from that, this is a welcome change IMHO.

This is evidence that our current event system isn't doing what we (primarily) need it to. Our current event system is only used in 2 places (`api.ServeWebsocket` and `watchfs`), and isn't useful to you here.

Seeing the implementation, I don't think these are hooks. They remind me of node.js's event system, which uses [synchronous, sequential callbacks](https://wanago.io/2019/02/18/typescript-node-js-eventemitter/) that inherit from a base `EventEmitter` definition.

I think we need _synchronous_ events that don't run through a bus. This `hook` package is much closer to that.

Would you mind moving this package to be a subpackage of event? That way we can unblock merging this PR, and take a crack at refactoring events later on

dsref/hook/change.go
```diff
@@ -0,0 +1,39 @@
...
+	Username   string
+	PrettyName string
+	HeadRef    string
+	Info       *dsref.VersionInfo
+	Dir        string
+}
+
+// ChangeNotifier is something that provides a hook which will be called when a dataset changes
+type ChangeNotifier interface {
+	SetChangeHook(func(change *DsChange))
```
b5, line 38: any reason to pass a pointer? Do notifiees need to be able to modify the `DsChange` argument?

dsref/hook/change.go
dustmop: Totally good to move this to a subpackage of `event`, I just had no clue where to put this and `dsref` was the first thing that came to mind, since it's about dataset changes. We'll definitely be using more (async) events in the future once we have progress bars, and handle command-line changes while Desktop is running. I don't think we know for sure what events will look like in the near future, but I think it's good to move halfway towards wherever that is.

dsref/hook/change.go
dustmop: Nope, was thinking they should be passed by pointer to avoid a copy, but that's not a good reason. Will change that.

PR #1351, opened 2020-05-18
test(benchmark): added benchmarks for save/validate

Arqu:
Benchmarks for the save code path with an added deep dive on the validation side.
Special note: this chokes on my machine for the 1G datasets without the optimizations that will follow on this.

Also will have to update the dependencies before merging, but want to get this up as a launch pad for the following PRs

### DIFF
--- a/base/dsfs/dataset_test.go
+++ b/base/dsfs/dataset_test.go
@@ -1,6 +1,7 @@
 package dsfs
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -12,7 +13,10 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/qri-io/dataset"
+	"github.com/qri-io/dataset/dsio"
 	"github.com/qri-io/dataset/dstest"
+	"github.com/qri-io/dataset/generate"
+	"github.com/qri-io/dataset/validate"
 	"github.com/qri-io/qfs"
 	"github.com/qri-io/qfs/cafs"
 	ipfs_filestore "github.com/qri-io/qfs/cafs/ipfs"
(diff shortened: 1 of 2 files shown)


base/dsfs/dataset_test.go
```diff
@@ -859,3 +864,203 @@ func TestGetDepth(t *testing.T) {
 		}
 	}
 }
+
+func GenerateDataset(b *testing.B, sampleSize int, format string) (int, *dataset.Dataset) {
```
b5, line 868: Refactored this to include some of the nicer aspects of the dataset/dsio API:

```go
func GenerateDataset(b *testing.B, sampleSize int, format string) (int, *dataset.Dataset) {
	ds := &dataset.Dataset{
		Commit: &dataset.Commit{
			Timestamp: time.Date(2017, 1, 1, 1, 0, 0, 0, time.UTC),
			Title:     "initial commit",
		},
		Meta: &dataset.Meta{
			Title: "performance benchmark data",
		},
		Structure: &dataset.Structure{
			Format: format,
			FormatConfig: map[string]interface{}{
				"headerRow":  true,
				"lazyQuotes": true,
			},
			Schema: map[string]interface{}{
				"type": "array",
				"items": map[string]interface{}{
					"type": "array",
					"items": []interface{}{
						map[string]interface{}{"title": "uuid", "type": "string"},
						map[string]interface{}{"title": "ingest", "type": "string"},
						map[string]interface{}{"title": "occurred", "type": "string"},
						map[string]interface{}{"title": "raw_data", "type": "string"},
					},
				},
			},
		},
	}

	gen, err := generate.NewTabularGenerator(ds.Structure)
	if err != nil {
		b.Fatalf("creating generator: %s", err.Error())
	}
	defer gen.Close()

	bodyBuffer := &bytes.Buffer{}
	w, err := dsio.NewEntryWriter(ds.Structure, bodyBuffer)
	if err != nil {
		b.Fatalf("creating entry writer: %s", err.Error())
	}

	for i := 0; i < sampleSize; i++ {
		ent, err := gen.ReadEntry()
		if err != nil {
			b.Fatalf("reading generator entry: %s", err.Error())
		}
		w.WriteEntry(ent)
	}
	if err := w.Close(); err != nil {
		b.Fatalf("closing writer: %s", err)
	}

	fileName := fmt.Sprintf("body.%s", ds.Structure.Format)
	ds.SetBodyFile(qfs.NewMemfileReader(fileName, bodyBuffer))

	return bodyBuffer.Len(), ds
}
```

This function is a great mini-program! High level feedback:
* skip the unmarshaling step, just make go types. Verbose? yes. Easier to plug variables tho, and prevents potential marshaling bugs from creeping into the test
* dsio package tries to let you abstract away from the data format
* use `Fatalf` in setup functions. Tests should fail on the spot if setup fails
* use `defer` statements near the opening call site to ensure stuff closes in all cases
* always handle the error from `entryWriter.Close`, that's where stuff usually breaks
* a `*bytes.Buffer` is a `io.ReadWriter`, we have methods for creating files from readers

Upside: ran this before & after with the same bench results, which  proves to me this PR isolates setup from benchmarks properly!

base/dsfs/dataset_test.go
```diff
@@ -859,3 +864,203 @@ func TestGetDepth(t *testing.T) {
...
+		b.Run(fmt.Sprintf("sample size %v", sampleSize), func(b *testing.B) {
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				b.StopTimer()
+
+				_, dataset := GenerateDataset(b, sampleSize, "csv")
+				if dataset == nil {
+					b.Skip("dataset == nil, skipping")
+					return
+				}
```
b5, line 993: by using `Fatal` in `GenerateDataset` we can drop this check

base/dsfs/dataset_test.go
```diff
@@ -859,3 +864,203 @@ func TestGetDepth(t *testing.T) {
...
+	for _, sampleSize := range []int{10000, 100000, 250000, 500000, 1000000, 10000000} {
+		b.Run(fmt.Sprintf("sample size %v", sampleSize), func(b *testing.B) {
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				b.StopTimer()
+				_, dataset := GenerateDataset(b, sampleSize, "csv")
+				if dataset == nil {
+					b.Skip("dataset == nil, skipping")
+					return
+				}
```
b5, line 1031: drop this check after switching to using `b.Fatal`

base/dsfs/dataset_test.go
```diff
@@ -859,3 +864,203 @@ func TestGetDepth(t *testing.T) {
...
+	for _, sampleSize := range []int{10000, 100000, 250000, 500000, 1000000, 10000000} {
+		b.Run(fmt.Sprintf("sample size %v", sampleSize), func(b *testing.B) {
+			b.ResetTimer()
+			for i := 0; i < b.N; i++ {
+				b.StopTimer()
+				_, dataset := GenerateDataset(b, sampleSize, "json")
+				if dataset == nil {
+					b.Skip("dataset == nil, skipping")
+					return
+				}
```
b5, line 1055: drop this check after switching to `b.Fatal`

base/dsfs/dataset_test.go
Arqu: Oh boy,  this is great stuff! Thx!

PR #1355, opened 2020-05-19
fix(dsref): Improve dsref parse error messages

dustmop:


dsref/parse.go
```diff
@@ -88,7 +94,7 @@ func Parse(text string) (Ref, error) {
 
 	if text != "" {
 		pos := origLength - len(text)
-		return r, fmt.Errorf("parsing ref, unexpected character at position %d: '%c'", pos, text[0])
+		return r, fmt.Errorf("unexpected char at position %d: '%c'", pos, text[0])
```
b5, line 97: any specific reason for the `char` phrasing? shorter line length? `char` feels a little more "programmer" than "character" to me

dustmop: Yeah, I guess shorter line length. I'm not married to the idea, will change it to "character".

PR #1360, opened 2020-05-22
refactor(*): adjust to fit with new `qfs` apis

### LINKED ISSUE #1375: bug(lib.NewInstance): repo lock problems blocking moving forward
Okay now that I've flipped back to upgrading `qri `to work with the latest `qfs` (which is blocking us from finishing the migration work)

We are running into repo lock issues in two specific circumstances. (We have new repo lock issues now because `qfs` now checks to see if the repo is locked before adding it as a filesystem).

#### run `qri connect` in one terminal and another command in another terminal:
To walk through the issue:
- run qri connect - this creates an instance of qri, which puts a lock on the underlying ipfs repo
- in a new terminal run any qri command `qri ls` for example, you will get a repo lock error
Why?
Let's use `qri list` as an example:
The qri `list` cmd, in the `RunE` section of the cobra command, runs `o.Complete()`. This fills out any options. For the `qri list` command, it fills out the `ListOptions` struct, which includes the `DatasetMethods` field. We use the `f.DatasetMethods()` function, which requires an instance in order to work. This creates a new instance, the instance tries to access the ipfs repo, which has a lock, and we error.
We don't look to see if we should actually be connecting via RPC when we are down in lib, after the instance is created. So we have a bit of a catch22 here, we can't check to see if we can communicate over RPC because we need an instance to do that, but we can't create an instance because the repo is locked!

#### cmd tests where we call two commands in a row using `run.ExecCommand` on the same TestRunner
Prime example of a test that calls multiple commands in a row:
https://github.com/qri-io/qri/blob/b0c585c9784c5af3bf332f913b84d496eac77eab/cmd/save_test.go#L1060-L1084
Looks like we don't release the Instance between running the commands.
Do we have a functional way of releasing the Instance, so we can do it when necessary?

ramfox:
This specifically addresses `lib.NewInstance`, but also sets the stage for the config Filesystem changes.

Still working through the different tests that rely on `qfs.Filesystem` in different ways and iterations.

relies on qri-io/qfs#31

closes #1375


cmd/connect.go
```diff
@@ -72,11 +74,28 @@ func (o *ConnectOptions) Complete(f Factory, args []string) (err error) {
...
+		if errors.Is(err, qipfs.ErrNeedMigration) {
+			err = nil
+			msg := `
+Your IPFS repo needs updating before qri can start. 
+Run migration now?`
+			if !confirm(o.Out, o.In, msg, false) {
+				return qerr.New(qipfs.ErrNeedMigration, `connect cancelled, requires migration.`)
+			}
+
+			if err = qipfs.Migrate(); err != nil {
```
Arqu, line 90: Not sure if we should trigger the migration here, especially as we will have other migration logic as well. Currently migration logic is trggered at https://github.com/qri-io/qri/blob/4159a259b679b70ffe374c3f2ae62f2b3852508a/lib/lib.go#L187
And runs https://github.com/qri-io/qri/blob/master/config/migrate/migrate.go which should "automagically" happen before we get to this point if I'm not mistaken.

cmd/factory_test.go
```diff
@@ -88,6 +89,7 @@ func NewTestFactoryInstanceOptions(opts ...lib.Option) (tf TestFactory, err erro
 		lib.OptQriNode(tnode.(*p2p.QriNode)),
 	}, opts...)
 
+	fmt.Println("new test factory instance options")
```
Arqu, line 92: This should be removed once finalized.

cmd/fetch_test.go
```diff
@@ -53,6 +53,7 @@ func TestFetchCommand(t *testing.T) {
 
 	ctx := context.Background()
 
+	fmt.Println("test fetch command")
```
Arqu, line 56: Remove when done.

cmd/fetch_test.go
```diff
@@ -138,6 +139,7 @@ func TestFetchCommand(t *testing.T) {
 	// TODO(dustmop): Try to add the below to a separate test in api/. Need to populate the peers
 	// in a fashion similar to api/fsi_test.go's `TestNoHistory`.
 
+	fmt.Println("another new isntance")
```
Arqu, line 142: Remove when done.

cmd/test_runner_test.go
```diff
@@ -182,56 +183,103 @@ func (run *TestRunner) MakeTmpDir(t *testing.T, pattern string) string {
...
+	}
+	cancel()
+	if doneCh != nil {
+		<-doneCh
+	}
+	return nil
 }
 
 // ExecCommandWithContext executes the given command with a context
 func (run *TestRunner) ExecCommandWithContext(ctx context.Context, cmdText string) error {
```
Arqu, line 235: Do we even need this as it all runs with the `run.Context`. Now it's pretty much the same as `ExecCommand` or should we work off the provided context.

lib/lib.go
```diff
@@ -337,22 +340,82 @@ func NewInstance(ctx context.Context, repoPath string, opts ...Option) (qri *Ins
 		}
 	}
 
+	if o.qfs != nil {
+		inst.qfs = o.qfs
+	} else if inst.qfs == nil {
+		// TODO (ramfox): adding a default mux config
```
Arqu, line 346: 👍

cmd/test_runner_test.go
```diff
@@ -45,6 +45,7 @@ type TestRunner struct {
 	XformVersion  string
 	CmdR          *cobra.Command
 	Teardown      func()
+	CmdDoneCh     chan struct{}
```
Arqu, line 48: This seems to be unused

cmd/connect.go
ramfox: Yea this is hold over/experimentation from before the `OneToTwo` function was planned.

b5:
superseded by #1387

PR #1361, opened 2020-05-22
feat(lib): subsystems can turn reference strings into datasets with a ParseLoadResolve function

b5:
builds on #1359

introduce a new interface: `dsfs.Loader` for loading datasets from a passed-in "source" parameter.

a high level `dsfs.ParseLoadResolve` function encomposes the common stack of parsing a reference, resolving, then loading (with a potential network pull for resolution and loading) makes a useful inversion of control flow for subsystems that need this highest-order functionality, while still providing configurability at the instance level.

### DIFF
--- a/base/dataset.go
+++ b/base/dataset.go
@@ -10,9 +10,35 @@ import (
 	"github.com/qri-io/qfs"
 	"github.com/qri-io/qfs/cafs"
 	"github.com/qri-io/qri/base/dsfs"
+	"github.com/qri-io/qri/dsref"
 	"github.com/qri-io/qri/repo"
 	reporef "github.com/qri-io/qri/repo/ref"
 )
+
+// NewLocalDatasetLoader creates a dsfs.Loader that operates on a filestore
+func NewLocalDatasetLoader(r repo.Repo) dsref.Loader {
+	return loader{r}
+}
+
+type loader struct {
+	r repo.Repo
+}
+
+func (l loader) LoadDataset(ctx context.Context, ref dsref.Ref, source string) (*dataset.Dataset, error) {
+	// LoadDataset fetches, derefences and opens a dataset from a reference
+	// implements the dsfs.Loader interface
+	if source != "" {
+		return nil, fmt.Errorf("only local datasets can be loaded")
+	}
+
+	ds, err := dsfs.LoadDataset(ctx, l.r.Store(), ref.Path)
+	if err != nil {
+		return nil, err
+	}
+
+	err = OpenDataset(ctx, l.r.Filesystem(), ds)
+	return ds, err
+}
 
 // OpenDataset prepares a dataset for use, checking each component
 // for populated Path or Byte suffixed fields, consuming those fields to
--- a/base/dataset.go
+++ b/base/dataset.go
@@ -24,9 +24,9 @@ type loader struct {
 	r repo.Repo
 }
 
+// LoadDataset fetches, derefernces and opens a dataset from a reference
+// implements the dsfs.Loader interface
 func (l loader) LoadDataset(ctx context.Context, ref dsref.Ref, source string) (*dataset.Dataset, error) {
-	// LoadDataset fetches, derefences and opens a dataset from a reference
-	// implements the dsfs.Loader interface
 	if source != "" {
 		return nil, fmt.Errorf("only local datasets can be loaded")
 	}
--- a/cmd/sql.go
+++ b/cmd/sql.go
@@ -56,6 +56,7 @@ PostgreSQL in a few ways:
 	}
 
 	cmd.Flags().StringVarP(&o.Format, "format", "f", "table", "set output format [table]")
+	cmd.Flags().BoolVar(&o.Offline, "offline", false, "prevent network access")
 
 	return cmd
 }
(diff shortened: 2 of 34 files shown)


dsref/resolve.go
```diff
@@ -27,7 +27,8 @@ type Resolver interface {
 }
 
 // ParallelResolver composes multiple resolvers into one resolver that runs
-// in parallel when called, returning the first valid response
+// in parallel when called, using the first resolver that doesn't return
+// ErrorNotFound
```
dustmop, line 31: ErrorNotFound -> ErrRefNotFound?

dsref/spec/load.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/qri-io/dataset"
+	"github.com/qri-io/qfs"
+	"github.com/qri-io/qri/dsref"
+)
+
+// PutDatasetFunc adds a dataset to a system that stores datasets
+// PutDatasetFunc is required to run the LoaderSpec test. When called the
+// Resolver should retain the dataset for later loading by the spec test
```
dustmop, line 16: Resolver -> Loader?

lib/resolve.go
```diff
@@ -51,107 +46,62 @@ func (inst *Instance) ParseAndResolveRefWithWorkingDir(ctx context.Context, refS
...
+		return "", dsref.ErrRefNotFound
 	}
 
 	// Handle the "me" convenience shortcut
 	if ref.Username == "me" {
 		ref.Username = inst.cfg.Profile.Peername
 	}
 
-	resolvers, err := inst.resolveSources(source)
+	resolver, err := inst.resolverMode(mode)
```
dustmop, line 60: This reads really strangely. `resolverMode` follows the "noun adjective" pattern, as if it's a getter that retrieves the mode of a resolver. But it actually does the opposite, it takes a mode and returns a resolver. Maybe `resolverByMode` would work?

base/dataset.go
```diff
@@ -10,10 +10,36 @@ import (
...
+func NewLocalDatasetLoader(r repo.Repo) dsref.Loader {
+	return loader{r}
+}
+
+type loader struct {
+	r repo.Repo
+}
+
+func (l loader) LoadDataset(ctx context.Context, ref dsref.Ref, source string) (*dataset.Dataset, error) {
+	// LoadDataset fetches, derefences and opens a dataset from a reference
```
dustmop, line 28: derefences -> dereferences

lib/load.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"github.com/qri-io/dataset"
+	"github.com/qri-io/qri/base"
+	"github.com/qri-io/qri/base/dsfs"
+	"github.com/qri-io/qri/dsref"
+	"github.com/qri-io/qri/fsi"
+	reporef "github.com/qri-io/qri/repo/ref"
+)
+
+// LoadDataset fetches, derefences and opens a dataset from a reference
```
dustmop, line 16: derefences -> dereferences

cmd/sql.go
```diff
@@ -82,9 +84,15 @@ func (o *SQLOptions) Complete(f Factory, args []string) (err error) {
 func (o *SQLOptions) Run() (err error) {
 	o.StartSpinner()
 
+	var rm string
```
dustmop, line 87: "rm" makes me think "remove". How about "mode".

dsref/load.go
```diff
@@ -0,0 +1,47 @@
...
+	LoadDataset(ctx context.Context, ref Ref, source string) (*dataset.Dataset, error)
+}
+
+// ParseResolveLoad is the function type returned by NewParseResolveLoader
+type ParseResolveLoad func(ctx context.Context, refStr string) (*dataset.Dataset, error)
+
+// NewParseResolveLoadFunc composes a username, resolver, and loader into a
+// higher-order function that converts strings to full datasets
+// pass the empty string as a username to disable the "me" keyword in references
+func NewParseResolveLoadFunc(username string, resolver Resolver, loader Loader) ParseResolveLoad {
```
dustmop, line 24: This implementation shouldn't be in this package. The purpose of `dsref` is that it's a low-dependency foundation that other things can build on, and so it shouldn't have any opinionated decisions or magic behavior in it. Even if this just looks like concatenating existing interfaces, the reliance of having a valid username, the "me" convention, and user-centric error messages are all business logic that should live higher up. As the comment above implies, lib has the useful implementation of the loader, so it should be constructing this function, and passing it down to the subsystems that need it.

startf/transform.go
```diff
@@ -327,51 +338,35 @@ func (t *transform) ModuleLoader(thread *starlark.Thread, module string) (dict s
 	return t.moduleLoader(thread, module)
 }
 
-// LoadDataset is a function
+// LoadDataset implements the starlark load_dataset function
```
dustmop, line 341: Need an integration test, similar to `TestSaveTransformUsingPrev` in `cmd`, otherwise this functionality will break again.

dsref/load.go
b5: v good points. moving to lib

PR #1363, opened 2020-05-22
fix(naming): Handle names with upper-case characters

dustmop:
Datasets that have names with upper-case characters should still work. But they should produce a warning when used, recommending a rename. The rename command doesn't warn about this.

Handle everything to do with this check inside GetRefSelect. Clean up how GetRefSelect is called so that it seems less magical, and works closer to how it was intended. Add constants to make it more readable.

Ignore bad case errors in lib.Resolve, since only cmd knows if such names are valid or not.

Follow-up to https://github.com/qri-io/qri/pull/1308. Makes major progress on https://github.com/qri-io/qri/issues/1132.


cmd/ref_link_ensurer.go
```diff
@@ -0,0 +1,42 @@
+package cmd
```
b5, line 1: can we move this whole file into `cmd/ref_select.go`?

cmd/ref_link_ensurer.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+// EnsureFSIAgrees should be passed to GetCurrentRefSelect in order to ensure that any references
+// used by a command have agreement between what their .qri-ref linkfile thinks and what the
+// qri repository thinks. If there's a disagreement, the linkfile wins and the repository will
+// be updated to match.
+// This is useful if a user has a working directory, and then manually deletes the .qri-ref (which
+// will unlink the dataset), or renames / moves the directory and then runs a command in that
+// directory (which will update the repository with the new working directory's path).
+func EnsureFSIAgrees(f *lib.FSIMethods) *FSIRefLinkEnsurer {
```
b5, line 15: Kinda a funny API, but it gets the job done! I'd say if we have to add another option fo `GetCurrentRefSelect`, we should consider re-structuring, I'd rather not force an API re-think right now

dustmop: This behavior has always existed, since EnsureRef was a thing, however it wasn't explicit. Callers would pass an fsiMethods pointer in, but it was unclear what that meant. I've seen at least one PR author confused by this fact. This type constructor clarifies the callsites of GetRefSelect a bit.

PR #1367, opened 2020-05-26
feat(migration): OneToTwo basic implementation

Arqu:
This is a skeleton implementation for the OneToTwo migration to align on the implementation details.


config/migrate/migrate.go
```diff
@@ -49,6 +61,92 @@ func ZeroToOne(cfg *config.Config) error {
...
+	// TODO(ramfox): qfs migration
+
+	if err := oneToTwoConfig(cfg); err != nil {
+		return err
+	}
+	if err := safeWriteConfig(cfg); err != nil {
+		rollbackConfigWrite(cfg)
+		return err
+	}
+
```
ramfox, line 85: can we add a comment here 
```
TODO(ramfox): remove original ipfs repo after all migrations were successful
```

config/migrate/migrate.go
```diff
@@ -49,6 +61,92 @@ func ZeroToOne(cfg *config.Config) error {
...
+		return err
+	}
+	
+	return nil
+}
+
+func oneToTwoConfig(cfg *config.Config) error {
+	if cfg.API != nil {
+		apiCfg := cfg.API
+		defaultApiCfg := config.DefaultAPI()
```
dustmop: As per the lint rules, this should be defaultAPICfg

config/migrate/migrate.go
```diff
@@ -49,6 +61,92 @@ func ZeroToOne(cfg *config.Config) error {
...
+		}
+		if apiCfg.WebsocketAddress == "" {
+			apiCfg.WebsocketAddress = defaultApiCfg.WebsocketAddress
+		}
+	} else {
+		return qerr.New(fmt.Errorf("invalid config"), "config does not contain API configuration")
+	}
+
+	if cfg.RPC != nil {
+		defaultRpcCfg := config.DefaultRPC()
```
dustmop: This should be defaultRPCCfg

config/migrate/migrate.go
```diff
@@ -49,6 +61,92 @@ func ZeroToOne(cfg *config.Config) error {
...
+		if cfg.RPC.Address == "" {
+			cfg.RPC.Address = defaultRpcCfg.Address
+		}
+	} else {
+		return qerr.New(fmt.Errorf("invalid config"), "config does not contain RPC configuration")
+	}
+
+	cfg.Revision = 2
```
dustmop: The ZeroToOne migration has this nice pattern that looks like:

```
  // conversion stuff happens
  cfg.Revision = 1
  if err := safeWriteConfig(...)
  return nil
```

Which makes it read really nicely; the Revision being incremented is very obvious. Could we make OneToTwo resemble the same shape? So the top-level would be:

```
  if err := oneToTwoConfig(...)
  cfg.Revision = 2
  if err := cfg.Validate(...)
  if err := safeWriteConfig(...)
  return nil
```

All it would take is moving the two blocks up to the calling function

config/migrate/migrate.go
```diff
@@ -49,6 +61,92 @@ func ZeroToOne(cfg *config.Config) error {
...
+	return nil
+}
+
+
+func rollbackConfigWrite(cfg *config.Config) {
+	cfgPath := cfg.Path()
+	if len(cfgPath) == 0 {
+		return
+	}
+	cfgPathComponents := strings.Split(cfgPath, "/")
```
dustmop, line 102: This won't work on Windows. Use `filepath.Dir` to get the path without the last component, and use `filepath.Join` to create the fullpath with the new basename. `filepath` is aware of Windows slashes, while `path` (and string splitting) is not.

lib/lib.go
```diff
@@ -184,14 +184,11 @@ func OptCheckConfigMigrations(cfgPath string) Option {
 			return fmt.Errorf("no config file to check for migrations")
 		}
 
-		migrated, err := migrate.RunMigrations(o.Streams, o.Cfg)
+		_, err := migrate.RunMigrations(o.Streams, o.Cfg)
```
dustmop, line 187: Remove the first return value if it's not needed. We can tell what happened from the `err` alone.

config/migrate/migrate.go
```diff
@@ -49,6 +61,67 @@ func ZeroToOne(cfg *config.Config) error {
...
+	}
+}
+
+func safeWriteConfig(cfg *config.Config) error {
+	cfgPath := cfg.Path()
+	if len(cfgPath) == 0 {
+		return qerr.New(fmt.Errorf("invalid path"), "could not determine config path")
+	}
+	cfgPathComponents := strings.Split(cfgPath, "/")
+	cfgPathComponents[len(cfgPathComponents)-1] = "config_updated.yaml"
```
dustmop, line 116: Same thing here. Actually, if this is being done in two places, please extract into a helper function, such as `configUpdateFilename(cfgPath)`

dustmop:
Hmmm, branch got modified just as I was submitted my comments, seems like some are out of date. Ah well.

config/migrate/migrate.go
Arqu: Really like this

Arqu:
> Hmmm, branch got modified just as I was submitting my comments, seems like some are out of date. Ah well.

Sorry :/ Good stuff anyways, I'll just carry over the comments to the other PR.

config/migrate/migrate.go
Arqu: Good point.

PR #1368, opened 2020-05-27
refactor(apiutil): upgrade to new apiutil api

b5:
builds on qri-io/apiutil#1, giving us a fighting chance to return proper pagination information


api/datasets.go
```diff
@@ -590,21 +590,14 @@ func getParamsFromRequest(r *http.Request, readOnly bool, path string) (*lib.Get
 	}
 
 	if !readOnly {
-		offset, offsetErr := util.ReqParamInt("offset", r)
-		limit, limitErr := util.ReqParamInt("limit", r)
+		p.Offset = util.ReqParamInt(r, "offset", 0)
+		p.Limit = util.ReqParamInt(r, "limit", lib.DefaultPageSize)
```
Arqu, line 594: Oh yes, kills so much boilerplate!

api/fsi_test.go
```diff
@@ -141,7 +141,7 @@ func TestNoHistory(t *testing.T) {
 	}
 
 	// Expected response for body of the dataset
-	expectBody = `{"data":{"path":"fsi_init_dir/body.csv","data":[["one","two",3],["four","five",6]]},"meta":{"code":200},"pagination":{"nextUrl":"/body/peer/test_ds?page=2"}}`
+	expectBody = `{"data":{"path":"fsi_init_dir/body.csv","data":[["one","two",3],["four","five",6]]},"meta":{"code":200},"pagination":{"page":1,"pageSize":50,"nextUrl":"/body/peer/test_ds?page=2","prevUrl":""}}`
```
Arqu, line 144: Shouldn't this also include result_count? Think -1 is good if it's "hard to get" or "N/A"

b5: Good point. I've filed #1374 as a follow up to set these. I'd like to do `pagination.resultsCount` in one big pass

PR #1369, opened 2020-05-27
refactor(transform): Move transform up out of base.Save

dustmop:
First, add more tests for transform. In the meantime, restore the ability to save with a transform by supplying a transform component, like:

```
qri save --file transform.json me/my_dataset
```

This was broken at some point, but there was no test for it, so it wasn't caught.

Next, move transform up, out of base.Save, and into base.TransformApply. Call that earlier, before the initID is created, closer to where filePaths are applied to the incoming dataset. This moves the functionality closer to how the upcoming `apply` command should work.


cmd/save_test.go
```diff
@@ -926,6 +925,122 @@ func TestSaveTwiceWithTransform(t *testing.T) {
 	}
 }
 
+func TestSaveTransformUsingPrev(t *testing.T) {
```
b5, line 928: grumble grumble grumble tests in lib. Then again these are integration tests. I take back the grumbling.

base/transform_apply.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		if head.BodyPath != "" {
+			var body qfs.File
+			body, err = dsfs.LoadBody(ctx, r.Store(), head)
+			if err != nil {
+				return err
+			}
+			head.SetBodyFile(body)
+		}
+	}
```
b5, line 54: I'm assuming the plan is to replace this when we rebase #1361?

cmd/save_test.go
```diff
@@ -131,12 +131,12 @@ func TestSaveRun(t *testing.T) {
...
-		{"add rows, dry run", "me/movies", "testdata/movies/dataset.json", "testdata/movies/body_twenty.csv", "Added 10 more rows", "Adding to the number of rows in dataset", false, true, true, "dataset saved: peer/movies@/map/QmPg2oEhBJSrJAVHgH7ptsmHQEVEio6JdcgHrDY45eYVtg\nthis dataset has 1 validation errors\n", "", ""},
-		{"add rows, save", "me/movies", "testdata/movies/dataset.json", "testdata/movies/body_twenty.csv", "Added 10 more rows", "Adding to the number of rows in dataset", true, false, true, "dataset saved: peer/movies@/map/QmTFfwHuWcXLbpL4b9Pb7BMM7zxhjqrgQRMiCSNub4ZX3E\nthis dataset has 1 validation errors\n", "", ""},
+		{"good inputs, dryrun", "me/movies", "testdata/movies/dataset.json", "testdata/movies/body_ten.csv", "", "", false, true, true, "dataset saved: peer/movies\n", "", ""},
+		{"good inputs", "me/movies", "testdata/movies/dataset.json", "testdata/movies/body_ten.csv", "", "", true, false, true, "dataset saved: peer/movies@/map/QmVDSrf4BpupXYmDJc6eXanFVW9DS6g3Pfkkx6YqkcUjEQ\nthis dataset has 1 validation errors\n", "", ""},
```
b5, line 135: what's breaking hashes on these tests? I'm assuming some sort of broken/bad/missing thing, just curious

base/transform_apply.go
dustmop: Yes, I think this is definitely a good place to use the loadDataset interface. No need for apply to know concretely how a dataset is being loaded.

cmd/save_test.go
dustmop: More subtle than that. The `dsfs.Timestamp` stub used by testRunner increments the minute by 1 every time it's called. The early return being added for dryRun (https://github.com/qri-io/qri/pull/1369/files#diff-8f260ff9153f41a66f62970bbc74ce9cR644) prevents `dsfs.Timestamp` from being called, which changes the timestamp and therefore hashes for future steps. If I were to add a call to `dsfs.Timestamp` just before that early return, the hash would stay the same. So the hashes changing is proof that no dsfs actions are occuring for dry-run.

Issue #1346, opened 2020-05-18
Migrating to configuration version 2


b5:
This PR should land the _capacity_ to run migration from configuration version 1 to 2, but leave running the migration itself off the default configuration so we can keep working without requiring `master` to run a migration.

The config migration should also handle migrating & updating IPFS deps & repo, but that should be handled elsewhere. This issue is to track migrating our configuration. The IPFS stuff should build be added to a migration function, and defined in `qri-io/qfs/qipfs`

### Steps to complete
* close #1345, adding support for multiaddresses
* Remove Stale configuration fields specified in RFC0029:  https://github.com/qri-io/rfcs/blob/master/text/0029-config_revision.md#remove-stale-configuration-fields
* Add in the `Filesystems` field, which can be passed down to a `muxfs.New`, unblocks the leet issue: #1337
* Test the migration, including a rollback

b5:
from our chat:

```go
package config

import (
	qipfs "github.com/qri-io/qfs/cafs/ipfs"
)


func OneToTwo(qripath string) error {

	// REQUIRE migrating into repo_path/ipfs
	// move IPFS folder inside Qri
	// upgrade IPFS version
	// change IPFS ports
	if err := qipfs.RunMigration(); err != nil {
		return qerr.New(err, "qri encountered an error while updating IPFS, there may be an issue with your repo")
	}
		
		// * copy folder
		// * run migration
		// * change ports

	// migrate qri configuration struct
	//    * use multiaddrs
	//    * remove stale fields
	//    * add filesystem field
	//      - now IPFS_PATH comes from qfs.Filesystem["ipfs"].options.path
	if err := oneToTwo(qripath); err != nil {
		return qerr.New(err, "qri encountered an error while updating your configuration file, please re-try")
	}

}

func oneToTwo(qripath string) error {
	ioutil.WriteFile("config_updated.yaml")
}

// package qipfs
func RunMigration() {
	rollback := func() {
		os.RemoveAll()
	}
		defer func() {
			if rollback != nil {
				rollback()
			}
		}

		rollback = nil
		return nil
}

var yalmConfig = `
filesystems:
- type: ipfs
	options:
		path: ./ipfs
- type: local
- type: http
`

`
QRI_PATH
	config.yaml
	ipfs/
`
```

ramfox:
Clarifying the responsibility of `qfs.RunMigrations`
`func RunMigrations(ipfsRepoPath, newRepoPath) error`
[...]

PR #1373, opened 2020-05-28
refactor(config): expose filesystem config

Arqu:
The basis is:
- Non-existing config means default to all (same as it was)
- If specified it creates all the named individual stores
- Empty array means disable all

Changes based on https://github.com/qri-io/rfcs/blob/master/text/0029-config_revision.md#remove-stale-configuration-fields
Addressing item No3 from #1346

Migration also needs to be implemented once the rest is done.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -40,20 +40,15 @@ type Config struct {
 	Logging *Logging
 }
 
-// Ignore implements the ignorer interface from base/fill/struct in order to safely
-// consume config files that have definitions beyond those specified in the struct
-// and are either deprecaated or no longer supported.
-// This simply ignores all defined ignore fields at read time.
-func (cfg *Config) Ignore(key string) error {
-	ignoredPaths := map[string]bool{
-		"update": true,
-		"webapp": true,
-		"render": true,
-	}
-	if ignoredPaths[key] {
-		return nil
-	}
-	return fmt.Errorf("key '%s' not found", key)
+// IgnoreFillField implements the FieldIgnorer interface from base/fill/struct
+// in order to safely consume config files that have definitions beyond those
+// specified in the struct and are either deorecated or no longer supported
+func (cfg *Config) IgnoreFillField(key string) bool {
+	return map[string]bool{
+		"Update": true,
+		"Webapp": true,
+		"Render": true,
+	}[key]
 }
 
 // NOTE: The configuration returned by DefaultConfig is insufficient, as is, to run a functional
--- a/config/config.go
+++ b/config/config.go
@@ -22,12 +22,13 @@ const CurrentConfigRevision = 2
 type Config struct {
 	path string
 
-	Revision int
-	Profile  *ProfilePod
-	Repo     *Repo
-	Store    *Store
-	P2P      *P2P
-	Stats    *Stats
+	Revision    int
+	Profile     *ProfilePod
+	Repo        *Repo
+	Store       *Store
+	P2P         *P2P
+	Stats       *Stats
+	Filesystems *Filesystems
 
 	Registry *Registry
 	Remotes  *Remotes
@@ -57,20 +58,18 @@ func (cfg *Config) SetArbitrary(key string, val interface{}) error {
 // DefaultConfig gives a new configuration with simple, default settings
 func DefaultConfig() *Config {
 	return &Config{
-		Revision: CurrentConfigRevision,
-		Profile:  DefaultProfile(),
-		Repo:     DefaultRepo(),
-		Store:    DefaultStore(),
-		P2P:      DefaultP2P(),
-		Stats:    DefaultStats(),
-
-		Registry: DefaultRegistry(),
-		// default to no configured remotes
-
-		CLI:     DefaultCLI(),
-		API:     DefaultAPI(),
-		RPC:     DefaultRPC(),
-		Logging: DefaultLogging(),
+		Revision:    CurrentConfigRevision,
+		Profile:     DefaultProfile(),
+		Repo:        DefaultRepo(),
+		Store:       DefaultStore(),
+		P2P:         DefaultP2P(),
+		Stats:       DefaultStats(),
+		Registry:    DefaultRegistry(),
+		CLI:         DefaultCLI(),
+		API:         DefaultAPI(),
+		RPC:         DefaultRPC(),
+		Logging:     DefaultLogging(),
+		Filesystems: DefaultFilesystems(),
 	}
 }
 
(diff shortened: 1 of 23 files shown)


config/store.go
```diff
@@ -4,9 +4,10 @@ import "github.com/qri-io/jsonschema"
 
 // Store configures a qri content addessed file store (cafs)
 type Store struct {
-	Type    string                 `json:"type"`
-	Options map[string]interface{} `json:"options,omitempty"`
-	Path    string                 `json:"path,omitempty"`
+	Type        string                 `json:"type"`
+	Options     map[string]interface{} `json:"options,omitempty"`
+	Path        string                 `json:"path,omitempty"`
+	Filesystems []string               `json:"Filesystems,omitempty"`
```
b5, line 10: Is this intended as a stepping stone for configuration? The RFC specs out `Filesystems` as a top-level config field structured like this:

```go
type Filesystem struct {
  Type string
  // Source string // TODO - enable once pluggable sources are a thing
  Config map[string]interface{}
}
```

```go
type Config struct {
  // ...
  Filesystems []Filesystem
  // ...
}
```

Arqu: Let's pretend I didn't zone out completely while doing this...

ramfox: should we pull in the `MuxConfig` struct for this?

Note also @Arqu we renamed `Options` to `Config` from the rfc

```
// MuxConfig contains the information needed to create a new filesystem
type MuxConfig struct {
	Type   string                 `json:"type"`
	Config map[string]interface{} `json:"config,omitempty"`
	Source string                 `json:"source,omitempty"`
}
```

https://github.com/qri-io/qfs/blob/8af27780e44bbeefa98491c7146450a256b246b9/muxfs/mux.go#L47-L52

config/config.go
```diff
@@ -63,14 +64,16 @@ func DefaultConfig() *Config {
...
-		CLI:     DefaultCLI(),
-		API:     DefaultAPI(),
-		RPC:     DefaultRPC(),
-		Logging: DefaultLogging(),
+		CLI:      DefaultCLI(),
+		API:      DefaultAPI(),
+		RPC:      DefaultRPC(),
+		Logging:  DefaultLogging(),
+		Filesystems: []*Filesystem{
+			DefaultFilesystemIPFS(),
```
Arqu, line 73: I'd like us to define what are the most basic defaults here.

config/config.go
```diff
@@ -234,6 +238,14 @@ func (cfg Config) Validate() error {
 		}
 	}
 
+	for _, filesystem := range cfg.Filesystems {
```
Arqu, line 241: I'm not a fan of separating this validation out. Some Go-fu help here would be nice to pull this up into the `validators`.

config/filesystem.go
```diff
@@ -0,0 +1,80 @@
+package config
+
+import (
+	"github.com/qri-io/jsonschema"
+)
+
+// Filesystem configures the filesystem for qri content
+type Filesystem struct {
```
Arqu, line 8: I've tried to compose this struct from `qfs/muxfs.MuxConfig` as @ramfox suggested, however the import was giving me a weird error :/

```
go: finding module for package github.com/qri-io/qfs/muxfs
go: github.com/qri-io/qri imports
        github.com/qri-io/qri/cmd imports
        github.com/qri-io/qri/config imports
        github.com/qri-io/qfs/muxfs: package provided by github.com/qri-io/qfs at latest version v0.1.0 but not at required version v0.1.1-0.20200207165332-a6a7b98f0fa0
```

config/store.go
```diff
@@ -54,6 +54,7 @@ func (cfg *Store) Copy() *Store {
 	res := &Store{
 		Type:    cfg.Type,
 		Options: cfg.Options,
+		Path:    cfg.Path,
```
Arqu, line 57: Will remove this (got carried over fom the pre-remove stale fields work)

p2p/test/ipfs.go
```diff
@@ -41,6 +41,7 @@ func MakeRepoFromIPFSNode(node *core.IpfsNode, username string) (qrirepo.Repo, e
 	}
 
 	memFS := qfs.NewMemFS()
+	// TODO(arqu) - verify if we need cafs/local here
```
Arqu, line 44: Was unsure on how/wether to handle this. I know we want to move away from cafs anyways.

repo/buildrepo/build.go
```diff
@@ -86,14 +86,29 @@ func New(ctx context.Context, path string, cfg *config.Config) (repo.Repo, error
...
-	mux := map[string]qfs.Filesystem{
-		"local": localfs.NewFS(),
-		"http":  httpfs.NewFS(),
-		"cafs":  store,
+	mux := map[string]qfs.Filesystem{}
+
+	if cfg != nil && cfg.Filesystems != nil && len(cfg.Filesystems) > 0 {
+		for _, fsCfg := range cfg.Filesystems {
+			if fs, ok := fsMap[fsCfg.Type]; ok {
+				mux[fsCfg.Type] = fs
```
Arqu, line 101: Need to work out how to pass the actual config into the constructors. Feels like `MuxConfig` for `Filesystem` is the way to go and then I can simply replicate this behavior https://github.com/qri-io/qfs/blob/8af27780e44bbeefa98491c7146450a256b246b9/muxfs/mux.go#L79

config/filesystem.go
```diff
@@ -0,0 +1,80 @@
...
+    "type": "object",
+    "required": ["type"],
+    "properties": {
+    	"type": {
+        "description": "Type of filesystem",
+        "type": "string",
+        "enum": [
+			"ipfs",
+			"local",
+			"http"
```
ramfox, line 63: also need "mem" and "map" as valid types for tests

p2p/test/ipfs.go
ramfox: Branch I'm working on to upgrade qri to the latest qfs makes these changes. I would leave this unless tests are failing.

repo/buildrepo/build.go
ramfox: This file gets decimated when we upgrade to the latest qfs, we no longer rely on `buildrepo` and almost solely on the `muxfs` package to add a filesystem to a qri instance. Again, as long as tests are passing, I think this should be fine!

config/filesystem.go
ramfox: Pretty sure this is because we are currently relying on an older version of qfs.

Maybe we refactor once we can rely on new `qfs`, but I think this is great!

config/config.go
ramfox: Ahh, I think we need to define `Validate` and `Copy` on `[]*Filesystem` structs, which then would call the `Validate` and `Copy` funcs on `Filesystem` structs when necessary

Then you can move the `cfg.Filesystems` into the above validators

config/store.go
Arqu: NVM this is correct.

Arqu:
Migration and changes implemented. Should be good to go now.

Arqu:
@dustmop Think now is the right time to actually do it, especially since the migration is waiting on other things to merge as well. Anyways the latest commit moves us to an ignorer interface.

Issue #1384, opened 2020-06-10
adding a dataset then trying to get it fails


dustmop:
```
> qri add b5/world_bank_population
🗼 fetched from registry

b5/world_bank_population
/ipfs/QmUHFmjhsCPNqWGvGDSydQeUGV7cHbGdttykuonRDUWk4N

Successfully added dataset b5/world_bank_population
```

```
qri get body b5/world_bank_population 
initID is required
```

```
qri get b5/world_bank_population 
initID is required
```

Seems like this is either caused by the recent RefResolve change, or the refactoring of base. I noticed a slight pause when I run `get` so maybe it's doing a network request, despite the dataset being in my repository?

dustmop:
We want removing a foreign dataset (one from outside my namespace, which I `add`ed) to erase the logbook log for that dataset. However, I can't find any evidence that this was ever working as described. I tried going through history and couldn't find anything.

I believe this should be performed by `RemoveLog` in logbook/logbook.go. But there's an explicit check in that function that prevents it from removing a log outside my own namespace, which I think means it won't work for this purpose. Currently, the only callers of `RemoveLog` are in "fsi/init.go" and "logbook/logsync.go", nothing in `lib` or `base` is using it.

I pushed a test to the branch `remove-foreign` which demonstrates what we want to happen. It is currently failing, and when it is passing this issue will be fixed.

b5:
closed by #1404, reinforced by #1408

PR #1386, opened 2020-06-10
fix(resolve): Workaround for resolve with a buggy logbook

dustmop:
Work-around for bug https://github.com/qri-io/qri/issues/1384. Logbook was broken for lookups by that bug.

Repo's Refstore doesn't fill in InitID, but `get` doesn't actually need that, so remove the check in fsi.ResolvedPath. Add a test for fsi.ResolvedPath that verifies it works without InitID.

Repo.fs should use its own GetRef to resolve references, instead of delegating to logbook.


b5:
superseded by #1404

PR #1389, opened 2020-06-18
refactor: require migration, remove path factory

### LINKED ISSUE #1296: Qri does not play nice with IPFS desktop
Qri wants the repo lock. Which means anyone running Qri has to turn off any IPFS daemons (IPFS desktop is a UI wrapper around go-ipfs), and because we don't perfectly re-implement the IPFS API, things get weird for outside applications.

The short-term fix is to make qri behave as closely as possible to go-ipfs from an API perspective.

### LINKED ISSUE #1319: Moving Qri's IPFS repo into the .qri folder
Background for why we're doing this: https://github.com/qri-io/qri/issues/1225

### Migrating Qri's configuration
We should time these changes with our overhaul of qri's configuration, updating the config RFC to talk about removing support for `IPFS_PATH`. Instead the ipfs repo path should be part of the new filesystem configuration, specific to the local IPFS `qfs.Filestore`. Doing it this way both removes configuration ambiguity, and allows users to get the old behavior back by specifying `~/.ipfs` as the repo path.

To make this transition as clean as possible, I think we should use a completely new IPFS config, with a new keypair. This'll mean users running on the libp2p network won't appear as the same peer. Yet another reason I'm thankful we connected user data to a separate keypair held by Qri.

### Moving Qri data from one IPFS repo to another
We're going to need a tool for detangling the hashes in an IPFS repo that Qri relies on, both to _move_ those hashes to another repo, and to unpin those hashes from the original repo.

### Using non-default ports for IPFS configuration
The other part of privatizing our IPFS instance is to move off of the default port configurations for our IPFS node.

We can leverage this move to a private configuration to set ports to non-standard ones. One option is to use `0` for port numbers, which defaults to an open port.

### Steps to complete:
_note: need to finish this list_

* [x] ratify the latest config RFC:  https://github.com/qri-io/rfcs/pull/51.
* [ ] write an IPFS repo migration tool for moving data between repositories
* [ ] write an IPFS-repo-move function:
  * create a new IPFS repo at `.qri/ipfs`, new keypair and all
  * edit port configuration to use nonstandard ports
  * run the IPFS repo migration, moving over any Qri hashes to the new repo
* [ ] write a qri config migration that also triggers the move 
* [ ] remove support for `IPFS_PATH` env var
* [ ] add documentation that for users that want to customize IPFS behavior: they need to edit the config within `.qri/ipfs`.

### LINKED ISSUE #1346: Migrating to configuration version 2

b5:
Put migration code to use, marking the migration as required before qri will start. This migration does a few things:
  * runs ipfs repo migration to support go-ipfs v0.5.0
  * converts ipfs configuration to use non-standard IPFS ports
  * moves the IPFS repo into the qri folder
  * adds a "Filesystems" configuration field

Qri now _only_ honors `QRI_PATH` by default. Drop cmd.EnvPathFactory in favour of qri path functions & string values.

closes #1296, closes #1319, closes #1346


config/migrate/migrate.go
```diff
@@ -3,18 +3,33 @@ package migrate
...
 // RunMigrations checks to see if any migrations runs them
-func RunMigrations(streams ioes.IOStreams, cfg *config.Config) (err error) {
+func RunMigrations(streams ioes.IOStreams, cfg *config.Config, interactive bool) (err error) {
 	if cfg.Revision != config.CurrentConfigRevision {
-		streams.PrintErr("migrating configuration...")
+		if interactive {
+			msg := `Your repo needs updating before qri can start. 
+Run migration now?`
+			if !confirm(streams.Out, streams.In, msg) {
+				return qerr.New(ErrNeedMigration, `your repo requires migration before it can run`)
```
feep, line 28: Something clearer.

'Your repo needs to be migrated before any qri commands can be run.'

Or, from line 25:

'Your repo needs updating before qri can start.'

They should probably say the same thing.

feep:
Notes [here](https://github.com/feep/qriscratch/blob/master/notes/0.9.9-dev.md#feat-enable-migration-f05780fb).

The [error for network failure](https://github.com/feep/qriscratch/blob/master/notes/0.9.9-dev.md#ugly-error-on-network-fail) is a little ugly.

Probably not worth doing anything about it right now. @ me if you want me to file an issue.

Arqu:
I've got this:

```
migrating configuration...done!
constructing "ipfs" filesystem: ipfs: need datastore migration
```

b5:
here's what I suspect happened @Arqu:

my guess: you had a config file with a `revision` value of `2`. The migration checked that file, assumed it was done, and moved on without actually doing anything. Mind setting config.yaml's `revision` to 1 & running any command like `qri list`?

> Also got: getting store from filesystem: resolver of kind 'ipfs' does not exist on this filesystem
> On a fresh pre-migration qri setup (ie just did qri setup and then updated)

This is a different (more concerning) issue, I'm going to try to reproduce now.

b5:
Ok, first effort to reproduce looks like it's working ok. Here's (roughly) what I ran:
```
$ export IPFS_PATH=/Users/b5/migration/test/ipfs
$ export QRI_PATH=/Users/b5/migration/test/qri
$ export PWD="$(pwd)"

# build from master@HEAD:
$ cd $QRI_REPO && git checkout 474151 && go install && cd $PWD

$ qri setup
choose a peername: [ruby_red_grand_basset_griffon_vendeen]:
set up qri repo at: /Users/b5/datasets/migration_test/qri

# build post-migration:
$ cd $QRI_REPO && git checkout 393473 && go install && cd $PWD

$ qri list
Your repo needs updating before qri can start.
Run migration now? [Y/n]:
migrating configuration...
  => Looking for suitable fs-repo-migrations binary.
  => None found, downloading.
  => Running: /var/folders/6p/sjn8bct92397mbgwmwwvcb4h0000gn/T/go-ipfs-migrate766176553/fs-repo-migrations -to 9 -y
Found fs-repo version 7 at /var/folders/6p/sjn8bct92397mbgwmwwvcb4h0000gn/T/ipfs_temp036298354
===> Running migration 7 to 8...
applying 7-to-8 repo migration
locking repo at "/var/folders/6p/sjn8bct92397mbgwmwwvcb4h0000gn/T/ipfs_temp036298354"
  - verifying version is '7'
> Upgrading config to new format
updated version file
===> Migration 7 to 8 succeeded!
===> Running migration 8 to 9...
applying 8-to-9 repo migration
updated version file
===> Migration 8 to 9 succeeded!
  => Success: fs-repo has been migrated to version 9.
done!
you have no datasets
```

Arqu:
Yup, changing the config revision to 1 and re-doing the migration did the trick.

cmd/qri.go
```diff
@@ -36,8 +34,7 @@ https://github.com/qri-io/qri/issues`,
 	cmd.SetUsageTemplate(rootUsageTemplate)
 	cmd.PersistentFlags().BoolVarP(&opt.NoPrompt, "no-prompt", "", false, "disable all interactive prompts")
 	cmd.PersistentFlags().BoolVarP(&opt.NoColor, "no-color", "", false, "disable colorized output")
-	cmd.PersistentFlags().StringVar(&opt.RepoPath, "repo", qriPath, "provide a path to load qri from")
-	cmd.PersistentFlags().StringVar(&opt.IpfsPath, "ipfs-path", ipfsPath, "override IPFS path location")
+	cmd.PersistentFlags().StringVar(&opt.qriPath, "path", qriPath, "provide a path to load qri from")
```
dustmop, line 37: This description and flag name seems too generic and/or confusing. This sounds like it could be used to load an executable or framework or something. And `path` doesn't narrow down what is happening. What if we use `qri-path` and describe it as "path where qri data and config lives"

cmd/qri.go
```diff
@@ -162,6 +155,11 @@ func (o *QriOptions) Instance() *lib.Instance {
 	return o.inst
 }
 
+// QriPath returns the path to the qri directory
```
dustmop, line 158: Would prefer some more specificity. This could mean the "qri directory" where the binary is installed, or where Desktop exists, or where I store my datasets. How about "QriPath returns the path where qri data and configuration exist"

lib/lib.go
```diff
@@ -291,7 +287,7 @@ func NewInstance(ctx context.Context, repoPath string, opts ...Option) (qri *Ins
 		opts = []Option{
 			OptStdIOStreams(),
 			OptSetIPFSPath(""),
-			OptCheckConfigMigrations(),
+			OptCheckConfigMigrations(false),
```
dustmop, line 290: What are your thoughts on changing the migration UI such that after migrations run successfully the qri binary exits with a message like "migration successful, please run command again". As it is now, if I run `qri list` then I get the migration prompt, confirm, see some messages about progress, then soon after the pager pops up and obscures what happened. I think it would be fine to bail out inside and ask the user to start over, it's less cognitive load than having to remember what was going on before the migration started.

repo/test/temp_repo.go
```diff
@@ -51,8 +51,14 @@ func newTempRepo(peername, prefix string, g gen.CryptoGenerator) (r TempRepo, er
 		return r, err
 	}
 
+	// Create directory for new Qri repo.
+	QriPath := filepath.Join(RootPath, "qri")
```
dustmop, line 55: lower-case this var

Hmmm, why is `RootPath` upper-case?

cmd/factory.go
```diff
@@ -42,33 +41,22 @@ type Factory interface {
...
 }
 
-// NewDirPathFactory creates a path factory that sets qri & ipfs paths to
-// dir/qri & qri/ipfs
-func NewDirPathFactory(dir string) PathFactory {
-	return func() (string, string) {
-		return filepath.Join(dir, "qri"), filepath.Join(dir, "ipfs")
-	}
+// RootedQriPath gives a "/qri" directory from a given root path
+func RootedQriPath(root string) string {
```
dustmop, line 60: This is only used in one place, might as well remove it and inline at the call site.

lib/lib.go
Arqu: +1

b5:
> There's a big move away from the term "repoPath" here. Is that motivated by deemphasizing the repo package? Because I think referring to the place that stores block data, logbook, refs, config, etc as the "repository" is still useful as a piece of terminology. In fact, it's even more useful once "ipfs/" lives inside "qri/", since then there's no confusion over whether "repo" refers to "qri/" or "ipfs/" or the parent of one or both of these.

Definitely motivated by de-emphasizing the term "repo". I was hoping to make the change in terminology incrementally. After reading your note I agree this PR doesn't even address that change in the notes. which isn't good. If we're going to drop the "repo" term, we should do it explicitly. I'll revert changes away from the `repoPath` term

lib/lib.go
b5: > "migration successful, please run command again"
fully agree

repo/test/temp_repo.go
b5: uhhh... dunno. will fix

config/config.go
```diff
@@ -305,3 +313,17 @@ func (cfg *Config) WithPrivateValues(p *Config) *Config {
...
 	return res
 }
+
+// DefaultFilesystems is the default filesystem stack
+func DefaultFilesystems() []qfs.Config {
+	return []qfs.Config{
+		{
+			Type: "ipfs",
+			Config: map[string]interface{}{
+				"path": "./ipfs",
```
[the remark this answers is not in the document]
b5: really good point about windows. At a minimum we should audit to make sure this value is always generated with `file path.Join(cfg.Path(), "ipfs")`. 

Regarding "correct", this is what we said in [RFC0029](https://github.com/qri-io/rfcs/blob/master/text/0029-config_revision.md#relative-paths-in-configuration-files-are-relative-to-the-file):

> Relative Paths in configuration files are relative to the file
> If a configuration value specifies a filepath, and the given path is relative, that path should be relative to the location of the config file. We should add a note of this to $ qri config --help.

config/migrate/migrate.go
```diff
@@ -87,7 +108,6 @@ func OneToTwo(cfg *config.Config) error {
 	}
 
 	// TODO(ramfox): remove original ipfs repo after all migrations were successful
```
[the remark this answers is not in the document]
b5: It's not yet, and I agree we need to figure out an answer here. whatever we come up with, I'd like to handle it in a separate PR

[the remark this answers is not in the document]
Arqu: Works for me, it's safe for the end user even in this state, just that we waste some disk space currently.

lib/lib.go
b5: nm, managed to fit this in

dustmop: Yes totally

cmd/cmd.go
```diff
@@ -60,6 +61,9 @@ func Execute() {
 // ErrExit writes an error to the given io.Writer & exits
 func ErrExit(w io.Writer, err error) {
 	log.Debug(err.Error())
+	if errors.Is(err, migrate.ErrMigrationSucceeded) {
+		printSuccess(w, "migration succeeded, re-run your command to continue")
```
dustmop, line 65: Add `os.Exit(0)` here? Otherwise looks like the err gets printed again. Also, return code should be 0 on success. Though I'm not sure if "the original command didn't run" should count as success?

b5: damn nice catch. I vote we exit with status 0 (success), a thing did succeed, and only because the user chose `Yes`. I've intentionally carved around a _non_ interactive version of this so we can write `qri connect --migrate` which will auto-migrate and continue

PR #1398, opened 2020-06-23
refactor: qri config version 2 migration

### LINKED ISSUE #1296: Qri does not play nice with IPFS desktop

### LINKED ISSUE #1319: Moving Qri's IPFS repo into the .qri folder

### LINKED ISSUE #1346: Migrating to configuration version 2

b5:
This branch has become the new de-facto master, I vote we merge it & keep working on cleanup from master

closes #1296, closes #1319, closes #1346


b5:
@Arqu any reason we shouldn't merge this?

Arqu:
Don't have any real objections. Master is mostly for dev and bleeding edge anyways. 

Only thing I want to do is test extensively before cutting the release.

PR #1403, opened 2020-06-24
feat(migrate): remove qri-only IPFS repo on config migration

### LINKED ISSUE #1399: Qri should delete ipfs repos that only contain qri data when migrating to version 2
this cleanup is a last step in decoupling qri from the "standard" IPFS installation. Putting this formally: migration should remove IPFS repo is `ipfs pin ls` is a subset of root hashes qri cares about. 

### Steps to complete:
* [x] need a function that logbook, building a map of all referenced root hashes
* [x] build a map of "common" root hashes (IPFS adds two root hashes by default)
* [x] add a method to `qipfs.Filesystem` that returns a channel of recursive (root) pins
* [x] write a function `RepoContainsOnlyQriData` that builds the logbook root hash map, range over the channel of root hashes, return `ErrContainsNonQriData` or something on the first result that isn't within the root hash map or common root hash map
* [x] call `RepoContainsOnlyQriData` in the 1-to-2 migration, if it doesn't error, remove the folder & warn the user

b5:
closes #1399


cmd/cmd.go
```diff
@@ -51,8 +51,9 @@ func Execute() {
 	root.SilenceErrors = true
 	// Execute the subcommand
 	if err := root.Execute(); err != nil {
-		printErr(os.Stderr, err)
-		os.Exit(1)
+		// printErr(os.Stderr, err)
```
dustmop, line 54: remove dead code

config/migrate/migrate.go
```diff
@@ -117,7 +123,14 @@ func OneToTwo(cfg *config.Config) error {
 		return err
 	}
 
-	// TODO(ramfox): remove original ipfs repo after all migrations were successful
+	// we're intention
```
dustmop, line 126: This comment seems incomplete

config/migrate/migrate.go
```diff
@@ -117,7 +123,14 @@ func OneToTwo(cfg *config.Config) error {
...
+	if err := maybeRemoveIPFSRepo(cfg, oldIPFSPath); err != nil {
+		fmt.Printf("error removing IPFS repo at %q:\n\t%s", oldIPFSPath, err)
+		fmt.Printf(`qri has successfully internalized this IPFS repo, and no longer 
+		needs the folder a %q. you may want to remove it
```
dustmop, line 130: "folder a" -> "folder at"

PR #1404, opened 2020-06-24
fix(remove): When removing foreign dataset, remove the log

dustmop:


dustmop:
Fixes https://github.com/qri-io/qri/issues/1384.

base/remove.go
```diff
@@ -42,15 +41,22 @@ func RemoveEntireDataset(ctx context.Context, r repo.Repo, ref dsref.Ref, histor
...
 	}
-	if err := book.WriteDatasetDelete(ctx, initID); err == nil {
-		didRemove = appendString(didRemove, "logbook")
-	} else {
-		// If the logbook is missing, it's not an error worth stopping for, since we're
-		// deleting the dataset anyway. This can happen from adding a foreign dataset.
-		if err != oplog.ErrNotFound {
+	if ref.Username == book.AuthorName() {
+		// TOOD(dustmop): Logbook should validate the fact that author's should only be able to
+		// write to their own logs. Trying to write to another user's log should throw an error.
```
b5, line 46: so much agree. I'm filing an issue

logbook/logbook.go
```diff
@@ -861,6 +861,15 @@ func (book *Book) RemoveLog(ctx context.Context, sender identity.Author, ref dsr
...
 }
 
+// RemoveAnyoneLog removes a log owned by anyone.
+func (book *Book) RemoveAnyoneLog(ctx context.Context, ref dsref.Ref) error {
+	if book == nil {
+		return ErrNoLogbook
+	}
+	book.store.RemoveLog(ctx, dsRefToLogPath(ref)...)
+	return book.save(ctx)
+}
```
b5, line 871: Ah I see, a bunch of this confusion is my fault. I wrote the current `logbook.RemoveLog` at the same time as `logsync`, and was attempting to prevent situations where foreign users shouldn't be allowed remove logs they don't have write access to. The _only_ place this is needed is when operating as a remote that supports delete requests

Given that you've just had to write a new `RemoveLog` that doesn't have this requirement, clearly we need a method that allows arbitrary log removal. To me this means `RemoveAnyoneLog` is the "real" `RemoveLog`, and we should move the author-only check to either a method on logbook that's called on `logbook/logsync/logsync.go#L273`, or externalized to that function.

Either way, in the name of a better API, we should rename `RemoveAnyoneLog` to `RemoveLog`. That's what it's doing.

remote/mock_client.go
```diff
@@ -44,7 +51,39 @@ func (c *MockClient) FetchLogs(ctx context.Context, ref dsref.Ref, remoteAddr st
 
 // CloneLogs is not implemented
 func (c *MockClient) CloneLogs(ctx context.Context, ref dsref.Ref, remoteAddr string) error {
```
b5, line 53: Stale comment.

Happy to see MockClient get a new trick

dustmop:
Updated. Earlier, I saw the logsync tests failing, but the fix you described, checking at the callsite, worked fine.

PR #1409, opened 2020-06-25
feat(cmd): add migrate flag to connect command

b5:
this is important for zero-input environments like docker and qri desktop, which need a surefire way to get to a running qri node without interruption. callers that want no prompts at all should now use:

```
$ qri connect --setup --migrate
```

To auto-run setup when the qri repo path is empty, and auto run migrations before startup.

### DIFF
--- a/cmd/connect.go
+++ b/cmd/connect.go
@@ -42,6 +42,7 @@ peers & swapping data.`,
 	}
 
 	cmd.Flags().BoolVarP(&o.Setup, "setup", "", false, "run setup if necessary, reading options from environment variables")
+	cmd.Flags().BoolVarP(&o.Migrate, "migrate", "", false, "automatically run migrations if necessary")
 	cmd.Flags().StringVarP(&o.Registry, "registry", "", "", "specify registry to setup with. only works when --setup is true")
 
 	return cmd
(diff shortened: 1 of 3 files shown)


b5:
This is a great example of where being able to open & close a repo twice in the same process comes in handy ;)

cmd/connect.go
```diff
@@ -76,6 +78,15 @@ func (o *ConnectOptions) Complete(f Factory, args []string) (err error) {
 		return errors.New(repo.ErrNoRepo, "no qri repo exists\nhave you run 'qri setup'?")
 	}
 
+	if o.Migrate {
+		// any required migration by default interrupts all commands with a prompt
+		// users setting the migrate flag are indicating we should skip the migration
+		// prompt, which is this only possible prompt in the `connect` command.
+		// setting noPrompt to true prior to calling init will have the effect
+		// of auto-running migrations
+		noPrompt = true
```
Arqu, line 87: I'm fine with it for now, however this might interfere with future commands which might want an independent prompt/no-prompt switch, but perfectly fine with it for now.

b5: agree. I left a big note for this exact reason

PR #1413, opened 2020-06-25
fix(migrate): wait for repo to shut down after migration

ramfox:
We successfully wait for the ipfs repo to shut down after the migration,
only if we have no unknown pins to report. Small adjustment so that we
wait for the repo to shut down in both cases.


config/migrate/migrate.go
```diff
@@ -300,15 +300,14 @@ longer requires the repo at %q
...
-		return err
+	} else {
+		fmt.Printf("moved IPFS repo from %q into qri repo\n", oldPath)
+		if err := os.RemoveAll(oldPath); err != nil {
+			return err
+		}
 	}
 
-	log.Error("successfully migrated repo, shutting down")
+	log.Info("successfully migrated repo, shutting down")
```
b5, line 310: hrm, this is a good start, but I think we can go deeper, making sure we _always_ close gracefully after opening the repo. Here's what I'm thinking as a refactor:

```go
func maybeRemoveIPFSRepo(cfg *config.Config, oldPath string) error {
	fmt.Println("\nChecking if existing IPFS directory contains non-qri data...")
	repoPath := filepath.Dir(cfg.Path())
	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)

	r, err := buildrepo.New(ctx, repoPath, cfg)
	if err != nil {
		cancel()
		return err
	}

	defer gracefulShutdown(cancel, r)

	// Note: this is intentionally using the new post-migration IPFS repo to judge
	// pin presence, because we can't operate on the old one
	fs := r.Filesystem().Filesystem(qipfs.FilestoreType)
	if fs == nil {
		return nil
	}

	logbookPaths, err := r.Logbook().AllReferencedDatasetPaths(ctx)
	if err != nil {
		return err
	}

	paths := map[string]struct{}{
		// add common paths auto-added on IPFS init we can safely ignore
		"/ipld/QmQPeNsJPyVWPFDVHb77w8G42Fvo15z4bG2X8D2GhfbSXc": {},
		"/ipld/QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn": {},
		"/ipld/QmS4ustL54uo8FzR9455qaxZwuMiUhyvMcX9Ba8nUH4uVv": {},
	}

	for p := range logbookPaths {
		// switch "/ipfs/" prefixes for "/ipld/"
		p = strings.Replace(p, "/ipfs", "/ipld", 1)
		paths[p] = struct{}{}
	}

	unknownPinCh, err := fs.(*qipfs.Filestore).PinsetDifference(ctx, paths)
	if err != nil {
		return err
	}

	log.Debugf("checking pins...%#v\n", paths)

	unknown := []string{}
	for path := range unknownPinCh {
		log.Debugf("checking if unknown pin is a dataset: %s\n", path)
		path = strings.Replace(path, "/ipld", "/ipfs", 1)
		// check if the pinned path is a valid qri dataset, looking for "dataset.json"
		// this check allows us to ignore qri data logbook doesn't know about
		if f, err := fs.Get(ctx, fmt.Sprintf("%s/dataset.json", path)); err == nil {
			f.Close()
		} else {
			unknown = append(unknown, path)
		}
	}

	if len(unknown) > 0 {
		fmt.Printf(`
Qri left your original IPFS repo in place because it contains pinned data that 
Qri isn't managing. Qri has created an internal copy of your IPFS repo, and no
longer requires the repo at %q
`, oldPath)
		if len(unknown) < 10 {
			fmt.Printf("unknown pins:\n\t%s\n\n", strings.Join(unknown, "\n\t"))
		} else {
			fmt.Printf("\nfound %d unknown pins\n\n", len(unknown))
		}
	} else {
		if err := os.RemoveAll(oldPath); err != nil {
			return err
		}
		fmt.Printf("moved IPFS repo from %q into qri repo\n", oldPath)
	}

	log.Info("successfully migrated repo, shutting down")

	return nil
}

func gracefulShutdown(cancel context.CancelFunc, r repo.Repo) {
	var wg sync.WaitGroup
	go func() {
		<-r.Done()
		wg.Done()
	}()

	wg.Add(1)
	cancel()
	wg.Wait()
}

```

ramfox: beautiful!

PR #1422, opened 2020-06-30
fix(rpc): shut down gracefully after completed RPC call

b5:
CLI commands running over RPC were hanging after completion thanks to the RPC client side
instance never firing on the shutdown channel.

The fix for this was one line: add `go inst.waitForAllDone()` to lib.NewInstance when returning
an RPC-backed instance. All other changes are in service of testing that line.

* Moved the `ServeRPC` method down from the API onto the instance itself, making testing easier
* colocated all the rpc stuff into one file: rpc.go, which should make it easier to refactor
  in the future
* re-worked some of the NetworkIntegrationTestRunner in lib, then realized it was overkill
  for the RPC test.

### DIFF
--- a/lib/rpc.go
+++ b/lib/rpc.go
@@ -48,7 +48,7 @@ func Receivers(inst *Instance) []Methods {
 	}
 }
 
-// ServeRPC checks for a configured RPC port, and registers a listner if so
+// ServeRPC checks for a configured RPC port, and registers a listener if so
 func (inst *Instance) ServeRPC(ctx context.Context) {
 	cfg := inst.cfg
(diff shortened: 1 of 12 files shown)


lib/rpc.go
```diff
@@ -0,0 +1,85 @@
...
+		NewSQLMethods(inst),
+		NewRenderMethods(inst),
+		NewFSIMethods(inst),
+
+		// TODO (b5) - deprecate ExportRequests
+		NewExportRequests(inst.Node(), nil),
+	}
+}
+
+// ServeRPC checks for a configured RPC port, and registers a listner if so
```
dustmop, line 51: listner -> listener

PR #1423, opened 2020-06-30
fix(windows): Don't detect terminal type

dustmop:


dustmop:
This fixes compilation on Windows. On Posix, syscall.Stdout is an int, and terminal.IsTerminal takes an int. On Windows, terminal.IsTerminal still takes an int, but syscall.Stdout returns a Handle. Maybe some day we'll try and detect pipes on Windows, but for now, just always return true.

PR #1426, opened 2020-07-01
feat(logbook): write push & pull operations, add logbook diagnostic output format

b5:
Prior to this commit we were _not_ recording push & delete operations in logbook. As per [RFC0030](https://github.com/qri-io/rfcs/blob/master/text/0030-replace_publish_clone_with_push_pull.md) we're now recording pushes of specific versions to a given remote within the oplog.

To do this, we make an optimistic write _before_ pushing or deleteing logs that adds an operation assuming it's successful. This way the receiver of the log gets the most up-to-date log that is signed with the operation included. If the push or delete operation fails, we roll back the log, removing the operation.

This approach tolerates faults better. the result of having an extra push / delete operation is less problematic than having a log that is short by one operation on a remote.

With this in place, logbook can now accurately report which versions of a dataset have been published, and has a running record of _where_ those versions have been published.

This PR also adapts a few more logbook functions to work with initIDs instead of references.

I've also added a "Diagnostic" method that's turned out to be very useful for quickly understanding the contents of a logbook by printing `MODEL ID OPCOUNT NAME` in an indented hierarchy. This was valuable enough to merit adding to the logbook command as a flag. Sample output:

```
user icm3gkoaoqn4vpcgzx7rmqsupcdcuec7fgqdq5z4kinvjprshqqq 2 nyc-transit-data
  dataset s3cql3s7fbay7abohm3qqdas7cm4ymu47jplqog5hxos5ytgaywa 2 turnstile_daily_counts_2020
    branch 2bz6mglcqlxa72nb3b7hnz4akrn6swi6iaredvzbgiwf5lp7fg5q 12 main
  dataset w5nl6h7fgwvu6gtfuxjo7fx4lwnerl7uuqycmu2ydrvdx4n76c7a 3 turnstile_daily_counts_2019
    branch 3ave4lm42mym2iuzyazv5b3ackdpvmtredcy4gdgbdsf3ulqxxoq 4 main
  dataset zngx523lbvukv4vebomvpzwsccvab5nuxf5c6h3svjdakpblotea 5 turnstiles_station_list
    branch vyznxsyekkjcfwfpor2svc7v5if2ec7bfq2p5u65xr3xkvnuhqoq 3 main
```

### DIFF
--- a/logbook/logbook.go
+++ b/logbook/logbook.go
@@ -54,8 +54,8 @@ const (
 	BranchModel
 	// CommitModel is the enum for a commit model
 	CommitModel
-	// PublicationModel is the enum for a publication model
-	PublicationModel
+	// PushModel is the enum for a push model
+	PushModel
 	// ACLModel is the enum for a acl model
 	ACLModel
 )
@@ -76,8 +76,8 @@ func ModelString(m uint32) string {
 		return "branch"
 	case CommitModel:
 		return "commit"
-	case PublicationModel:
-		return "publication"
+	case PushModel:
+		return "push"
 	case ACLModel:
 		return "acl"
 	default:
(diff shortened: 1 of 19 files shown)


logbook/logbook.go
```diff
@@ -604,59 +604,115 @@ func (book *Book) WriteVersionDelete(ctx context.Context, initID string, revisio
...
+	if err = book.save(ctx); err != nil {
+		return nil, nil, err
+	}
+
+	// after successful save calling rollback drops the written push operation
+	// rollback can only be called once. it will overwrite itself with a noop func
+	// when called
+	rollback = func(ctx context.Context) error {
+		defer func() {
+			rollback = func(ctx context.Context) error { return nil }
```
Arqu, line 640: This is sweet!

logbook/logbook.go
```diff
@@ -604,59 +604,115 @@ func (book *Book) WriteVersionDelete(ctx context.Context, initID string, revisio
...
```
dustmop, line 698: Same deal here.

logbook/logbook.go
b5: Damn. Super glad you caught this. I wrote a test for this exact purpose, and didn't compose the test properly.

logbook/logbook.go
```diff
@@ -757,43 +813,25 @@ func (book *Book) latestSavePath(branchLog *oplog.Log) string {
...
 //       branch
 //       branch
 //       ...
-func (book Book) UserDatasetRef(ctx context.Context, ref dsref.Ref) (*oplog.Log, error) {
-	if ref.Username == "" {
-		return nil, fmt.Errorf("logbook: reference Username is required")
-	}
-	if ref.Name == "" {
-		return nil, fmt.Errorf("logbook: reference Name is required")
+func (book Book) UserDatasetBranchesLog(ctx context.Context, datasetInitID string) (*oplog.Log, error) {
```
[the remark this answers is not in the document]
b5: Thinking the same thing. This API is getting stretched a little thin for my taste, and while I'd like to move over to the strong-typed variants, I'm not sure we have the implantation right yet. Writing methods like write access a little while back was a real pain.

b5:
@dustmop, flipped the method name to `Summary` instead of `Diagnostic`. Look ok?

PR #1430, opened 2020-07-06
refactor(event): re-work events to be synchronus

b5:
Change our event bus to operate synchronusly, taking inspiration from node.js `EventHandler`'s, but keeping a central bus singleton.

Using synchronus events allows us to write code that terminates sequentially, making events useful for one-off function calls that need to wait for all event subscribers to terminate before closing. Subscribers that want the old async format can create goroutines within handlers, shown in the updated `watchfs` eventHandler.

I've removed the unsubsribe interface in favour of a simplified closing mechanism on the bus. Closing the event bus context will send ErrBusClosed to any publish after closure.


event/event_test.go
```diff
@@ -3,7 +3,6 @@ package event
 import (
 	"context"
 	"fmt"
-	"testing"
 )
 
 func Example() {
```
b5, line 8: @Arqu & @ramfox the example function shows how event subscription is re-worked to use a function argument instead of a return read-only channel. Consumers are welcome to subscribe to multiple event Topics in the same handler to keep the old multiplexing pattern: 

```go
var b event.Bus

func handler(ctx context.Context, topic event.Topic, payload interface{}) error {
  switch type {
    // place all your topic-oriented handlers here ...
    case event.ETDatasetPulled:
    case event.ETPeerConnected:
      // ...
  }
  return nil
}

b.Subscribe(handler, event.ETDatasetPulled, event.ETPeerConnected)
```

b5:
@dustmop I think we can merge the `hooks` package into the `events` package with this change to synchronous events. Would love your thoughts.

event/event_test.go
Arqu: Love it! Especially turning to sequential executiion first.

One thing I'd like to squeeze in here is the ability to subscribe to different "topic levels". Ie we have some topics like `main:fsi:linked`, `main:fsi:unlinked`. I'd like to be able to subscribe to either the specific one (`main:fsi:linked`) or subscribe to `main:fsi` and get the handler invoked on both events.

Happy to take this on, we can do a separate PR for it.

b5: I'd like to avoid implementing hierarchical events until we see a clear need for them, mainly out of performance concerns. 

Switching to synchronous events means the performance of event handlers affects the performance of qri generally. For that reason I'd like to err on the side of being _very_ careful about what events we incorporate, and how we expose those events.

I think we can get lots of the effects of "levels" by composing multiple event publishers where needed. If we wanted to test your example now, we could wire up an `fsi` event topic, and have the FSI subsystem fire them on all method calls. Put another way: I think we can have our cake & eat it too if we're careful about what event topics we create & fire

Arqu: Fair enough, this was mostly an ergonomic feature at this point with no clear need. Considering the single handler multi topic subscribe, I'm fairly good with it for the time being.

api/test_runner_test.go
```diff
@@ -27,11 +28,11 @@ type APITestRunner struct {
 }
 
 func NewAPITestRunner(t *testing.T) *APITestRunner {
-	run := APITestRunner{}
-	run.Node, run.NodeTeardown = newTestNode(t)
-
 	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
```
b5, line 34: this line had `cancel` being called after setup completion, which was giving `bus closed` errors. re-worked to close the context in the `Delete` method

api/websocket.go
```diff
@@ -86,15 +79,6 @@ func (s Server) ServeWebsocket(ctx context.Context) {
...
 				select {
-				case e := <-busEvents:
-					log.Debugf("bus event: %s\n", e)
-					if fce, ok := e.Payload.(event.FSICreateLinkEvent); ok {
-						s.Instance.Watcher.Add(watchfs.EventPath{
-							Path:     fce.FSIPath,
-							Username: fce.Username,
-							Dsname:   fce.Dsname,
-						})
-					}
```
b5, line 97: This subscription is already set up in the `watchfs` constructor

event/dataset.go
```diff
@@ -0,0 +1,35 @@
...
+type DsChange struct {
+	InitID     string
+	TopIndex   int
+	ProfileID  string
+	Username   string
+	PrettyName string
+	HeadRef    string
+	Info       *dsref.VersionInfo
+	Dir        string
+}
```
b5, line 35: I think this can be refactored in a follow-up. This struct grew organically, and now many of these fields are present in `dsref.VersionInfo`

event/event.go
```diff
@@ -5,35 +5,39 @@ package event
...
 	golog "github.com/ipfs/go-log"
 )
 
-var log = golog.Logger("event")
+var (
+	log = golog.Logger("event")
+
+	// ErrBusClosed indicates the event bus is no longer coordinating events
+	// because it's parent context has closed
+	ErrBusClosed = fmt.Errorf("event bus is closed")
```
Arqu, line 19: Is this something we should handle on every subscription?

b5: If we see this error, it's a bug. The event bus should be wired to the same long-running context that all subsystems are wired to, and all should close at the same time. In that scenario this error will appear if a subsystem doesn't obey context closure & continues to publish events.

So no, I don't think it should need to be explicitly handled.

PR #1436, opened 2020-07-08
feat(lib): `OptNoBootstrap` option func for `NewInstance`

### LINKED ISSUE #1434: feat(`lib`): add `OptNoBootstrap` option for NewInstance
## What feature or capability would you like?

When we test Qri using `testground` we want the option to only test connections between nodes we explicitly want to be connected. This prevents chatter in the logs and can help us isolate issues. To get this behavior, we need a way to disable bootstrapping on nodes. We can do this by removing the bootstrap addrs when we configure the new instance.

`OptNoBootstrap` will return a `lib.Option` func that will configure the node to not have any bootstrap addrs

### Do you have a proposed solution?

```go
func OptNoBootstrap(cfg *config.Config) Option {
    return func(o *InstanceOptions) err {
        // if ipfs filesystem option exists, add NoBootstrap param to config
        // zero out qri bootstrap addrs
    }
}
```

### Please link any related issues.

relies on qri-io/qfs#41

ramfox:
Using this option will disable bootstrapping to any qri or ipfs nodes

This has been confirmed to work via testground

depends on qri-io/qfs#42
closes #1434

### DIFF
--- a/lib/lib.go
+++ b/lib/lib.go
@@ -21,6 +21,7 @@ import (
 	"github.com/qri-io/qfs"
 	"github.com/qri-io/qfs/cafs"
 	"github.com/qri-io/qfs/muxfs"
+	"github.com/qri-io/qfs/qipfs"
 	"github.com/qri-io/qri/config"
 	"github.com/qri-io/qri/config/migrate"
 	"github.com/qri-io/qri/dscache"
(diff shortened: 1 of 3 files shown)


lib/lib.go
```diff
@@ -178,6 +179,22 @@ func OptCheckConfigMigrations(interactive bool) Option {
 	}
 }
 
+// OptNoBootstrap ensures the node will not attempt to bootstrap to any other nodes
+// in the network. Should only be used when in a testing environment, like testgrounds
```
b5, line 183: I think this works in a number of contexts, doesn't need the "tests only" note. For example: when you want to do local discovery only

ramfox: good point!

Issue #1437, opened 2020-07-09
finish transition away from `repo.CanonicalizeDatasetRef`


b5:
I'm concerned about the possibility of datasets resolving to different `InitIDs` if different codepaths use `CanonicalizeDatasetRef ` and/or `ResolveDatasetRef`. The easiest way to avoid any bugs that may arise here is to move past the call

### Steps to complete
* [ ] remove `CanonicalizeDatasetRef` from the Save path
* [ ] remove `CanonicalizeDatasetRef` from base completely, using either resolved dataset refs or a passed-in resolver func
* [ ] switch all lib `CanonicalizeDatasetRef` calls to `ResolveDatasetRef`
* [ ] remove `CanonicalizeDatasetRef` from `p2p` package
* [ ] remove `CanonicalizeDatasetRef` from `remote` package
* [ ] finally, delete `repo.CanonicalizeDatasetRef`

PR #1439, opened 2020-07-09
refactor(save): overhaul save path InitID preparation

b5:
related to #1437 by switching save to use a `dsref.Resolver`.

overhaul `lib.DatasetMethods.Save` with the goal of establishing a stable dataset identifier as early as possible. This PR introduces a new `base` package function `PrepareSavingReference`, which returns a fully resolved reference & stable dataset name in all valid cases. For new dataset creation `PrepareDatasetReference` creates an oplog to get an InitID. use a rollback function `lib`  to clean up logbook when saving is unsuccessful.

This change removes _many_ TODOs in the save codepath, but does imply subtle changes that may alter edge-case behaviors of save. I'm confident those changes are for the better given this


base/dataset_prepare.go
```diff
@@ -12,66 +13,117 @@ import (
...
+		badCaseErr error
+		isNew      = false
+	)
+
+	ref, err := dsref.ParseHumanFriendly(refStr)
+	if errors.Is(err, dsref.ErrBadCaseName) {
+		// save bad case error for later, will fail if dataset is new
+		badCaseErr = err
+	} else if errors.Is(err, dsref.ErrEmptyRef) {
+		if bodyPathStr == "" {
```
dustmop, line 48: Add a comment to this block describing what it's doing. "User is calling save but didn't provide a dataset reference. Try to infer a usable name.".

base/dataset_prepare.go
```diff
@@ -12,66 +13,117 @@ import (
...
+		basename := filepath.Base(bodyPathStr)
+		basename = strings.TrimSuffix(basename, filepath.Ext(bodyPathStr))
+		basename = strings.TrimSuffix(basename, ".")
+		ref.Name = dsref.GenerateName(basename, "dataset_")
+
+		// need to use profile username b/c resolver.ResolveRef can't handle "me"
+		// shorthand
+		check := &dsref.Ref{Username: pro.Peername, Name: ref.Name}
+		if _, resolveErr := resolver.ResolveRef(ctx, check); resolveErr == nil {
+			if !newName {
```
dustmop, line 60:
```suggestion
			if !newName {
			// Name was inferred, and has previous version. Unclear if the user meant to create
			// a brand new dataset or if they wanted to add a new version to the existing dataset.
			// Raise an error recommending one of these course of actions.
```

base/dataset_prepare.go
```diff
@@ -12,66 +13,117 @@ import (
...
+// successful calls to PrepareSaving reference always have an InitID, and may
+// have a Path value
+func PrepareSaveRef(
+	ctx context.Context,
+	pro *profile.Profile,
+	book *logbook.Book,
+	resolver dsref.Resolver,
+	refStr string,
+	bodyPathStr string,
+	newName bool,
```
dustmop, line 35: I don't think `newName` differentiates this value well (especially since the introduction of `isNew`), it's unclear exactly what it means. That is, it represents if the user is explicitly requesting a new dataset be created, whose name is not in use yet. I would rename it to `wantNewName` or `requestingNewName`.

base/dataset_prepare.go
```diff
@@ -12,66 +13,117 @@ import (
...
+// new logbook history & InitID to write to
+// successful calls to PrepareSaving reference always have an InitID, and may
+// have a Path value
+func PrepareSaveRef(
+	ctx context.Context,
+	pro *profile.Profile,
+	book *logbook.Book,
+	resolver dsref.Resolver,
+	refStr string,
+	bodyPathStr string,
```
dustmop, line 34: To enforce the idea that this isn't being used as a path (no file is being loaded), and it's only used to generate a reasonable dataset name, how about renaming this to `bodyPathHint`.

base/dataset_prepare.go
```diff
@@ -12,66 +13,117 @@ import (
...
-// MaybeInferName infer a name for the dataset if none is set
-func MaybeInferName(ds *dataset.Dataset) string {
-	if ds.Name == "" {
-		filename := ds.BodyFile().FileName()
-		basename := strings.TrimSuffix(filename, filepath.Ext(filename))
-		return dsref.GenerateName(basename, "dataset_")
+// GenerateAvailableName creates a name for the dataset that is not currently in use
+func GenerateAvailableName(ctx context.Context, pro *profile.Profile, resolver dsref.Resolver, prefix string) string {
+	lookup := &dsref.Ref{Username: pro.Peername, Name: prefix}
+	counter := 1
```
dustmop, line 119: Before, this used `counter := 0` so that the first attempted name is "dataset_1" instead of "dataset_2". Wondering if this is an intended change. So if I drag a csv file named "stats.csv" into Desktop (for example), I get "me/stats" then doing it again gives me "me/stats_2"? I was thinking it was better to start generating names with "..._1" but maybe that's programmer-think getting in the way.

Regardless, please add a comment stating the assumption (whichever way we go) to make it explicit that it's expected behavior.

lib/datasets.go
```diff
@@ -489,88 +490,85 @@ func (m *DatasetMethods) Save(p *SaveParams, res *reporef.DatasetRef) error {
...
+		if isNew && (!success || p.DryRun) {
+			log.Debugf("removing unused log for new dataset %s", ref)
+			ctx, cancel := context.WithTimeout(context.Background(), time.Second*2)
+			if err := m.inst.logbook.RemoveLog(ctx, ref); err != nil {
+				log.Errorf("couldn't cleanup unused reference: %q", err)
+			}
+			cancel()
+		}
+	}
+	defer rollback()
```
dustmop, line 554: This is a potential bug. I realize you're not assigning to `rollback` in this function, but it'll be easy for someone in the future to make the mistake of thinking that modifying `rollback` will change how it acts. However, that wouldn't work, see https://gist.github.com/dustmop/f91e66692f4e874331bdab0d5b339602

I would suggest either getting rid of the `rollback` variable and just calling `defer func() ...` directly, relying on the `success` check inside the function, or just changing this call to:

```
defer func() {
  rollback()
}()
```

lib/datasets.go
```diff
@@ -489,88 +490,85 @@ func (m *DatasetMethods) Save(p *SaveParams, res *reporef.DatasetRef) error {
...
-		// When saving in an FSI directory, the ref should exist (due to `qri init`), and we
-		// need to load the previous version from the working directory.
-		if datasetRef.FSIPath != "" {
-			ds, err = fsi.ReadDir(datasetRef.FSIPath)
+	var fsiPath string
+	if !isNew {
+		// check for FSI linked data
+		fsiRef := ref.Copy()
+		if err := m.inst.fsi.ResolvedPath(&fsiRef); err == nil {
+			fsiPath = strings.TrimPrefix(fsiRef.Path, "/fsi")
```
dustmop, line 564: This `TrimPrefix` call appears in 3 places in this file, could we get a short helper function `refPathToFilesysPath` to collect the logic in one place.

lib/datasets_test.go
```diff
@@ -117,14 +117,7 @@ func TestDatasetRequestsSave(t *testing.T) {
...
-		// {&dataset.Dataset{Peername: "bad", Name: "path", Commit: &dataset.Commit{Qri: "qri:st"}}, nil, "decoding dataset: invalid commit 'qri' value: qri:st"},
-		// {&dataset.Dataset{Peername: "bad", Name: "path", BodyPath: "/bad/path"}, nil, "error with previous reference: error fetching peer from store: profile: not found"},
-		// {&dataset.Dataset{BodyPath: "testdata/q_bang.svg"}, nil, "invalid data format: unsupported file type: '.svg'"},
-		// {&dataset.Dataset{Peername: "me", Name: "cities", BodyPath: "http://localhost:999999/bad/url"}, nil, "fetching body url: Get http://localhost:999999/bad/url: dial tcp: address 999999: invalid port"},
-		// {&dataset.Dataset{Name: "bad name", BodyPath: jobsBodyPath}, nil, "invalid name: error: illegal name 'bad name', names must start with a letter and consist of only a-z,0-9, and _. max length 144 characters"},
-		// {&dataset.Dataset{BodyPath: jobsBodyPath, Commit: &dataset.Commit{Qri: "qri:st"}}, nil, "decoding dataset: invalid commit 'qri' value: qri:st"},
+		{"empty params", SaveParams{}, "no changes to save"},
```
dustmop, line 120: Definitely think this is a better error. However, it reminds me of this problem:
https://github.com/qri-io/qri/blob/master/cmd/save_test.go#L297

We used to allow saving with just a structure and no body. We don't need to necessarily fix that in this PR, but is there a plan for whether the newly refactored save path will allow for this behavior to be restored in the future?

base/dataset_prepare.go
```diff
@@ -12,66 +13,117 @@ import (
...
+	book *logbook.Book,
+	resolver dsref.Resolver,
+	refStr string,
+	bodyPathStr string,
+	newName bool,
+) (dsref.Ref, bool, error) {
+
+	var (
+		badCaseErr error
+		isNew      = false
```
dustmop, line 40: I don't think this variable helps readable, since it appears to be stateful but only changes according to the function's natural control flow. I'm guessing this is to avoid returning plan `false` and `true` values throughout the function. If that's so, how about instead defining two local constants `notNew = false` and `isNew = true` and using those.

dustmop:
Getting rid of all of these TODOs is awesome. It feels like the save path is nearly arriving at its final form.

base/dataset_prepare.go
```diff
@@ -12,66 +13,117 @@ import (
...
-	// head (if it exists) by using the initID, and pass it into base.Save.
-
-	// Determine if the save is creating a new dataset or updating an existing dataset by
-	// seeing if the name can canonicalize to a repo that we know about
-	lookup := &reporef.DatasetRef{Name: name, Peername: peername}
-	if err = repo.CanonicalizeDatasetRef(r, lookup); err == repo.ErrNotFound || lookup.Path == "" {
-		return &dataset.Dataset{}, &dataset.Dataset{}, "", nil
+// PrepareSaveRef works out a dataset reference for saving a dataset
+// version. When a dataset exists, resolve the initID & path
+// When no history exists, ensure a locally-unique dataset name and create a
```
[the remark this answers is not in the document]
b5: > By "no history exists" do you mean "no dataset with that name exists"?

Yep, that's exactly what I mean. will change.

base/dataset_prepare.go
```diff
@@ -12,66 +13,117 @@ import (
...
+// new logbook history & InitID to write to
+// successful calls to PrepareSaving reference always have an InitID, and may
+// have a Path value
+func PrepareSaveRef(
+	ctx context.Context,
```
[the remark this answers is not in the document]
b5: I started doing this after reading about line breaks in the [thanos](https://thanos.io/contributing/coding-style-guide.md/#only-two-ways-of-formatting-functions-methods) style guide.

I keep a ruler at 80 characters in my editor, and if the arguments of a function exceed that length, switch to one arg per line. What do you think about making that a rule? If you're into it we can get this going in our new `.github` repo.

cc @feep

base/dataset_prepare.go
b5: Yeah, here I'm proposing matching the OS X convention of starting with `_2`. From a UX perspective, the `_1` is implied as the first "stats" dataset in your example. I'll add a comment

lib/datasets.go
b5: very good point. dropping the variable assignment in favour of a direct `defer func() { ...` call

lib/datasets_test.go
b5: yes. This PR shouldn't inhibit no-body saves. Most of that testing needs to occur in `dsfs.CreateDataset`, then get propagated up.

base/dataset_prepare.go
b5: I've dropped the variable declaration, replaced it with a commend & direct value returns

lib/datasets.go
b5: sounds good. I've added `fsi.FilesystemPathToLocal` for this purpose

PR #1445, opened 2020-07-14
refactor(base): remove all calls to repo.CanonicalizeDatasetRef from base package

b5:
makes progress on https://github.com/qri-io/qri/issues/1437 building on save path refactoring to move the `base` package away from `repo.CanonicalizeDatasetRef`


cmd/save.go
```diff
@@ -184,19 +185,21 @@ continue?`, true) {
...
 	if err = o.DatasetMethods.Save(p, res); err != nil {
 		return err
 	}
 
 	o.StopSpinner()
-	printSuccess(o.ErrOut, "dataset saved: %s", res)
-	if res.Dataset.Structure != nil && res.Dataset.Structure.ErrCount > 0 {
-		printWarning(o.ErrOut, fmt.Sprintf("this dataset has %d validation errors", res.Dataset.Structure.ErrCount))
+	ref := dsref.ConvertDatasetToVersionInfo(res).SimpleRef()
+	ref.ProfileID = ""
```
Arqu, line 195: Why do we remove the profileID here?

remote/client.go
```diff
@@ -453,73 +452,15 @@ func (c *client) PullDataset(ctx context.Context, ref *dsref.Ref, remoteAddr str
...
-	success := false
-	for i := 0; i < tasks; i++ {
-		res := <-responses
-		err = res.Error
-		if err == nil {
-			cancelFetch()
-			success = true
-			*ref = *res.Ref
-			break
-		}
```
Arqu, line 517: 🤩

base/dataset.go
```diff
@@ -260,45 +260,6 @@ func RawDatasetRefs(ctx context.Context, r repo.Repo) (string, error) {
 	return builder.String(), nil
 }
 
-// FetchDataset grabs a dataset from a remote source
-func FetchDataset(ctx context.Context, r repo.Repo, ref *reporef.DatasetRef, pin, load bool) (err error) {
```
Arqu, line 264: What replaces FetchDataset, didn't seem to find the the alternative?

b5: Yeah there _isn't_ an alternative right now. FetchDataset doesn't currently fetch logbook data, and this logic really needs to be moved into the `p2p` package. I'd like to follow up in the very near future with a PR that replaces `FetchDataset`. Will open an issue

cmd/save.go
b5: to make the call to `ref.String()` below not print such a verbose reference.

without profileID:
```
dataset saved rm_me_3/saxaphones@/ipfs/QmRL5Wq4K1JXYZsgZKWJpdFJvQEsNCc6n9twRkzSGsCLbF
```

with:
```
dataset saved  rm_me_3/saxaphones@QmcPmrQYqFamdHYFL8qbfpbT3dRNRVanLmFEY9srWsddt9/ipfs/QmRL5Wq4K1JXYZsgZKWJpdFJvQEsNCc6n9twRkzSGsCLbF
```

Long term it'd be better to add a method to `dsref` that does this

PR #1447, opened 2020-07-16
fix(logbook,logsync,cmd): round of fixes prior to release

b5:
we were missing these, which is bad.

### DIFF
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -168,7 +168,7 @@ func confirm(w io.Writer, r io.Reader, message string, def bool) bool {
 		return def
 	}
 	input = strings.TrimSpace(strings.ToLower(input))
-	return (input == "y" || input == "yes") == def
+	return (input == "y" || input == "yes")
 }
 
 func usingRPCError(cmdName string) error {
--- a/cmd/qri.go
+++ b/cmd/qri.go
@@ -32,6 +32,7 @@ https://github.com/qri-io/qri/issues`,
 	}
 
 	cmd.SetUsageTemplate(rootUsageTemplate)
+	cmd.PersistentFlags().BoolVarP(&opt.Migrate, "migrate", "", false, "automatically run migrations if necessary")
 	cmd.PersistentFlags().BoolVarP(&opt.NoPrompt, "no-prompt", "", false, "disable all interactive prompts")
 	cmd.PersistentFlags().BoolVarP(&opt.NoColor, "no-color", "", false, "disable colorized output")
 	cmd.PersistentFlags().StringVar(&opt.repoPath, "repo", repoPath, "filepath to load qri data from")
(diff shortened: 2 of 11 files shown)


b5:
@dustmop, would love another look since I've pushed two additional fix-commits to this branch. One is a quick fix that un-brakes `$ qri publish --unpublish`, the other is a re-work that adds granularity & clarity to migration execution.

@ramfox, I'd love to get your review on the migration execution.

I've tacked these two commits onto this PR so we can keep moving on a patch release

cmd/qri.go
```diff
@@ -92,6 +93,8 @@ type QriOptions struct {
 	repoPath string
 	// generator is source of generating cryptographic info
 	generator gen.CryptoGenerator
+	// automatically run migrations if necessary
```
ramfox, line 96: So after testing a few different variations, I think this is the subtly that needs to be defined:

`qri list` -> if there is a migration needed, we will prompt the user for permission to run a migration
`qri list --no-prompt` -> if there is a migration needed, we will inform the user there is a migration needed, but not prompt them to execute one
`qri list --migrate` -> essentially the same as `qri list`, if there is migration we will prompt the user to run one
`qri list --migrate --no-prompt` -> if there is a migration needed, we will begin the migration without prompting the user

So the `migrate` flag really only does anything if the `--no-prompt` flag is also present.

Wondering if we should explain this subtly in the comments and potentially in the flag description => "when used with the `no-prompt` flag, will automatically run migrations if necessary"

cmd/print.go
```diff
@@ -168,7 +168,7 @@ func confirm(w io.Writer, r io.Reader, message string, def bool) bool {
 		return def
 	}
 	input = strings.TrimSpace(strings.ToLower(input))
-	return (input == "y" || input == "yes") == def
+	return (input == "y" || input == "yes")
```
ramfox, line 171: curious why the comparison to the default was removed. Is there a logic that I'm not getting?

cmd/qri.go
ramfox: More worried about future devs understanding the subtleties rather than users. Can see future confusion about "But I have the `--migrate` flag on!" & the description saying that we "automatically run migrations"

b5: > qri list --migrate -> essentially the same as qri list, if there is migration we will prompt the user to run one

hrm, that's def not what I intended, and thinking back I didn't properly vet that combination. I think running with the `--migrate` flag should:

* skip the prompt
* run the migration
* continue with the command, not asking the user to re-run.

The last step is up for debate, but I think it honors the "automatic" term in the `migrate` flag description.

ramfox: Yea, what you've outlined as your intention for the command feels correct for the "automatic" term.

b5:
ok changes made. would love a confirm @ramfox

ramfox:
confirmed! yay!

Issue #1441, opened 2020-07-13
desktop should switch to /get?format=zip instead of /export 


b5:
filing an issue here for tracking before we cut a release. Let's make sure `/get` is both usable and in use for desktop before cutting v0.9.10

ramfox:
assuming from convo w/ dustmop that should be `format=zip`

dustmop:
Fixed by https://github.com/qri-io/desktop/pull/573.

PR #1448, opened 2020-07-16
feat(get): Fix root handler for `get` so zip works, remove export handler

dustmop:


dustmop:
See https://github.com/qri-io/qri/issues/1441 and https://github.com/qri-io/desktop/issues/568

PR #1453, opened 2020-07-20
refactor(export): Remove export command

dustmop:


dustmop:
Finishes off https://github.com/qri-io/qri/issues/1176.

PR #1455, opened 2020-07-21
refactor(p2p): `n.ResolveRefLocally` allows a `QriNode` to resolve a ref locally

ramfox:
This replaces `CanonicalizeDatasetRef`. If a peer gets a request to `Get` or `Resolve` a dataset from another peer, we need a way to attempt to resolve that dataset locally. `n.ResolveRefLocally` only attempts to resolved using the `n.Repo` and `n.Repo.Dscache()` to ensure the resolution is happening locally.

note: This is currently difficult to test the validity of as our p2p tests are extremely flakey.

ticks of a box in #1437 

Edit: 

Since these functions are not referenced anywhere, and the p2p resolve ref functionality needs to be better documented and planned, this pr now removes the unused `DatasetRequest` and `ResolveDatasetRef` functions, handlers, and tests.


p2p/dataset.go
```diff
@@ -14,7 +14,7 @@ import (
 const MtDatasetInfo = MsgType("dataset_info")
 
 // RequestDataset fetches info about a dataset from qri peers
-// It's expected the local peer has attempted to canonicalize the reference
+// It's expected the local peer has attempted to resolve the reference
 // before sending to the network
 // ref is used as an outparam, populating with data on success
 func (n *QriNode) RequestDataset(ctx context.Context, ref *reporef.DatasetRef) (err error) {
```
b5, line 20: Oh boy is this method out of date!

First, we have the comment  `It's expected the local peer has attempted to resolve the reference`, and then we're calling `ResolveRefLocally` within the method... yeah.

But before we go fixing that, this method has zero callers within the codebase (other than a test). mainly because this logic has been split & superseded by two new concepts:

* reference resolution
* dataset previews

Let's remove this method entirely 😄

p2p/node.go
```diff
@@ -231,6 +232,13 @@ func (n *QriNode) GoOffline() error {
 	return err
 }
 
+// ResolveRefLocally attempts to resolve a dataset using the local repo
+func (n *QriNode) ResolveRefLocally(ctx context.Context, ref *dsref.Ref) error {
+	localResolver := dsref.SequentialResolver(n.Repo.Dscache(), n.Repo)
+	_, err := localResolver.ResolveRef(ctx, ref)
+	return err
+}
+
```
b5, line 241: Ok I'm going to go into great detail here because clearly you've used this PR to build up a bunch of knowledge around reference resolution & I wanna run with it.

It's really tempting to write methods like this that provide wrappers around a common stack of resolvers, but this leads to an anti pattern we've been working hard to clean out of the `base` package: _subsystems that can resolve references lead to duplicated resolution work_.

Instead we've been trying to enforce a new rule: **subsystems should work with initIDs whenever possible.** Instead of writing this function, prefer refactoring functions in the `p2p` package to either accept `dsref.Ref` arguments that fail if the `InitID` field is blank, or just take a direct `initID string` param. This forces reference resolution to happen both earlier and _only once_.

In a nearby-but-separate issue, we want `p2p.QriNode` to implement the `dsref.RefResolver` interface so it can help with reference resolution. If we had this function, the p2p subsystem would work like this in practice: Let's assume we want to do a dataset preview purely over p2p:

```
$ qri preview ramfox/lol_stats --remote p2p
```

1. we'd call `lib.DatasetMethods.Preview()`, which would call [`lib.Instance.ParseAndResolveRef`](https://github.com/qri-io/qri/blob/master/lib/resolve.go#L13), passing in `p2p` as the source.
2. `ParseAndResolveRef` would call `resolverForMode`, which would return (not yet written) `p2p.QriNode.ResolveRef` as the only resolver.
3. That resolver would get called, cool now we have a resolved ref.
4. lib will feed that resolved reference to a (not yet written) `p2p.QriNode.Preview` method, with  a resolved initID
5. preview gets pulled from the p2p network.
6. party.

So the whole thing happens over p2p, but is _orchestrated_ at the lib level.


### But what about subsystems that need to resolve stuff?

This is also great chance to talk about _control flow inversion_ when it comes to reference resolution. There are times where a subsystem needs to turn string dataset refs like `ramfox/lol_stats` into datasets, and for those we invert the control flow by building up a tidy little function `dsref.ParseResolveLoad` for the subsystem. SQL is the first command to implement this. Here's the _entire_ `lib.SQLMethods.Exec` method:

https://github.com/qri-io/qri/blob/65c5878ec52444f694d85a5236c7159443d87719/lib/sql.go#L33-L59

If we look at the SQL service, it has a `dsref.ParseResolveLoad` field, and requires a PRL function in it's constructor:

https://github.com/qri-io/qri/blob/65c5878ec52444f694d85a5236c7159443d87719/sql/sql.go#L29-L33

It calls that whenever it wants to turn a string into a dataset. This way the `lib` package gets to decide how datasets should be resolved and loaded (which keeps it user configurable), but the subsystem itself gets to decide _what_ and _when_ to do dataset loading.

I don't think `p2p` package should need to do this, but if it did, this is how we would do it.

ramfox: Okay, this is great to read through & makes sense! I understand the control flow inversion pattern.

I have a few thoughts/questions:

we've already talked about removing `DatasetRequest`, because it isn't being used anywhere. Should we also remove `ResolveDatasetRef` (our p2p precursor to `ResolveRef`), as it is not being referenced anywhere either? This would essentially remove all need for `Canonicalizing` or `Resolving` in p2p (as it stands currently). OR should I do the reactors to get `DatasetRequest` & `ResolveDatasetRef` off of `repo.DatasetRef` and onto `dsref.Ref`... which in the latter case is... I guess just `ResolveRef`.

The issue with control flow inversion in p2p, is that the requests to resolve a ref don't come from lib, they come from a peer, so we don't have lib to coordinate. Should our QriNodes have a `resolver` (guaranteed local) that they receive from lib when they are created & we can call on it when we get a handle a request from a peer? Should it not live in the QriNode struct, and instead be something that is passed to our handlerFuncs when we create a node? (`type HandlerResolverFunc (resolver) handlerFunc` & then the `MakeHandlers` func passes down resolver to the handlerFuncs that need it appropriate)

Should we just convert these p2p functions now, or should we just document the plan forward, remove them, and move on?

b5: > Should we also remove ResolveDatasetRef (our p2p precursor to ResolveRef), as it is not being referenced anywhere either? 

Yes! pumped.


> is that the requests to resolve a ref don't come from lib, they come from a peer, so we don't have lib to coordinate

First, I think we should save this as work for a separate PR, where we really sit down and think about reference resolution in a p2p setting, but to answer your question: yes this is the exact spot for this "control flow inversion" (I've  misused the term a bit), and you're exactly right here:

> Should our QriNodes have a resolver (guaranteed local) that they receive from lib when they are created & we can call on it when we get a handle a request from a peer? 

You've got the pattern exactly right: a bunch of connected peers should all use a resolver func provided when `QriNode` is constructed to answer questions coming in over p2p. The reason we should stake this out as separate work surrounds message coordination. I'd like to hash out a plan, write some code, and test it with a testground plan before merging it.

b5: to be specific about "the plan", I think we can do better than just blasting messages to all connected peers asking for a reference resolution. Maybe we create some set number of channels and feed them to each connected peer in rounds.

We should also keep in mind: this is the first real place where the returned "source" value will be _very_ important information. A peer that can resolve a reference is more likely to have logbook and block data.

b5, line 241: I don't think we should keep this method

ramfox: ahhh thank you for catching this

PR #1458, opened 2020-07-23
fix(download): adding component selector to root handler

Arqu:
This still needs to be finished by extending the get handler to write/return the result bytes instead of the current `maybeWriteOutfile`

### DIFF
--- a/api/root.go
+++ b/api/root.go
@@ -54,9 +54,11 @@ func (mh *RootHandler) Handler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	format := r.FormValue("format")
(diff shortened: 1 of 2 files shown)


b5:
@dustmop & @Arqu I'm not sure on the status of this PR. Still needed?

Arqu:
This is just some preparation. Still valid AFAIK just that it doesn't change much as the output is still being written to file instead of returned.

Happy to either abandon this and get back at a better time or keep it hanging as a reminder. Just move it into the todo column.

Arqu:
Closing as stale.

PR #1467, opened 2020-07-27
refactor(lib): remove CanonicalizeDatasetRef calls from render & diff

b5:
makes progress on #1437


lib/log.go
```diff
@@ -137,16 +136,13 @@ func (m *LogMethods) Logbook(p *RefListParams, res *[]LogEntry) error {
 	}
 	ctx := context.TODO()
 
-	ref, err := repo.ParseDatasetRef(p.Ref)
+	ref, _, err := m.inst.ParseAndResolveRef(ctx, p.Ref, "local")
```
ramfox, line 139: Should the dsref package formalize the non-multiaddr sources into constants ("local", "registry")?

b5: I think it should be either the empty string for local resolution or a multiaddr. It's a temporary issue that the registry comes back as a URL that we should get fixed ASAP, but we'll need some URL-to-multiaddr translation work

PR #1473, opened 2020-07-29
feat(oplog): extension interface for sparse-parent logs

b5:
logbook was relying on an under-specified property of the `opstore.Logstore` interface: some  implementations _always_ fetch the Parent log as part of a Get request.

`logbook.UserDatasetBranchLog` was using this, which blew up when used with a different implementation. The solution to this is to push the logic needed by `logbook.UserDatasetBranchLog` into the oplog package, and implement it using explicit calls to a `oplog.Logstore`.

In the future I'd like to both work on the `Logstore` API and write a set of spec tests that implementations can validate against, much like we've done with `dsref.RefResolver`.

At the same time I've defined an exention interface that holds these explicit calls and an optional method logstores can implement to speed up this process. There are situations where this can be done with a single trip to a database. The extension interface provides this as an upgrade path without any changes to the logbook or opstore packages

### DIFF
--- a/logbook/logbook.go
+++ b/logbook/logbook.go
@@ -15,7 +15,7 @@ import (
 	"sync"
 	"time"
 
-	logger "github.com/ipfs/go-log"
+	golog "github.com/ipfs/go-log"
 	crypto "github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/qri-io/dataset"
 	"github.com/qri-io/qfs"
@@ -42,8 +42,7 @@ var (
 	// This is mainly here for tests to override
 	NewTimestamp = func() int64 { return time.Now().UnixNano() }
 
-	// package logger
-	log = logger.Logger("logbook")
+	log = golog.Logger("logbook")
 )
 
 const (
@@ -848,11 +847,13 @@ func (book *Book) latestSavePath(branchLog *oplog.Log) string {
 //       branch
 //       ...
 func (book Book) UserDatasetBranchesLog(ctx context.Context, datasetInitID string) (*oplog.Log, error) {
+	log.Debugf("UserDatasetBranchesLog datasetInitID=%q", datasetInitID)
 	if datasetInitID == "" {
 		return nil, fmt.Errorf("%w: cannot use the empty string as an init id", ErrNotFound)
 	}
 	dsLog, err := book.store.Get(ctx, datasetInitID)
 	if err != nil {
+		log.Debugf("store error=%q datasetInitID=%q", err, datasetInitID)
 		return nil, err
 	}
 
--- a/logbook/logbook.go
+++ b/logbook/logbook.go
@@ -851,15 +851,14 @@ func (book Book) UserDatasetBranchesLog(ctx context.Context, datasetInitID strin
 	if datasetInitID == "" {
 		return nil, fmt.Errorf("%w: cannot use the empty string as an init id", ErrNotFound)
 	}
-	dsLog, err := book.store.Get(ctx, datasetInitID)
+
+	dsLog, err := oplog.GetWithSparseAncestorsAllDescendants(ctx, book.store, datasetInitID)
 	if err != nil {
 		log.Debugf("store error=%q datasetInitID=%q", err, datasetInitID)
 		return nil, err
 	}
 
-	// construct a sparse oplog of just user, dataset, and branches
-	sparseLog := &oplog.Log{Ops: dsLog.Parent().Ops, Logs: []*oplog.Log{dsLog}, Signature: dsLog.Signature}
-	return sparseLog, nil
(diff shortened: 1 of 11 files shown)


logbook/oplog/log.go
```diff
@@ -92,6 +93,68 @@ type Logstore interface {
 	Descendants(ctx context.Context, l *Log) error
 }
 
+// SparseAncestorsAllDescendantsLogstore is a an extension interface to
+// Logstore with an optimized method for getting a log with sparse parents and
+// all descendants
+type SparseAncestorsAllDescendantsLogstore interface {
```
dustmop, line 99: Not a huge fan of these names, because they're talking about implementation details of what is being returned, rather than the intention of why a caller would want to use them. That said, I think the reason they should be used is that we don't really have the vocabulary for what they should be called instead, and hopefully that sort of question will be answered in future design work. This is fine for now, but we should keep in mind that we need to come up with some name for what's really going on here when deciding on future logbook interfaces.

logbook/logbook.go
```diff
@@ -848,17 +847,18 @@ func (book *Book) latestSavePath(branchLog *oplog.Log) string {
 //       branch
 //       ...
 func (book Book) UserDatasetBranchesLog(ctx context.Context, datasetInitID string) (*oplog.Log, error) {
```
dustmop, line 849: My inclination is that more of these methods should move away from oplog.Log and towards something strongly typed, like what is in logbook/types.go, but this is fine for now.

logbook/oplog/log.go
```diff
@@ -92,6 +93,68 @@ type Logstore interface {
...
+	if sparseLS, ok := store.(SparseAncestorsAllDescendantsLogstore); ok {
+		return sparseLS.GetSparseAncestorsAllDescendants(ctx, id)
+	}
+
+	l, err := store.Get(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO (b5) - this is error prone
```
dustmop, line 123: Could you expand on this a bit? Do you mean its bad if this condition is true, and if so what might cause it? Or do you mean it's bad how this condition is being handled?

logbook/oplog/log.go
```diff
@@ -368,6 +431,7 @@ type Log struct {
 	authorID string // authorID value cache. not persisted
 	parent   *Log   // parent link
 
+	ParentID  string // init id of parent, if one exists
```
dustmop, line 434: Does "one" refer to initID or parent? If the latter, could you change this to "init id of parent, if a parent exists"

logbook/oplog/log.go
b5: totally agreed. I've intentionally made it something really idiosyncratic so it won't be a name we'll want to use. That said, It's also part of the oplog package, so I'm ok with obtuse names. The logbook interface is more important for qri, but this one definitely needs more thought

logbook/oplog/log.go
b5: I'm realizing this isn't always the case it's more of a field that _may_ hold the initID of the parent log if one exists, but that really isn't being strictly enforced at this point. I'm going to change this to just `init id of parent Log`

logbook/oplog/log.go
b5: I'll update that comment with this:

```
        // TODO (b5) - this is error prone. logs may exist but not be fetched. This
	// check is here now because not all implementations support the Descendants
	// method properly.
	// in the future remove this check once all implementations we know of have
	// a working `Descendants` implementation
```

b5:
@Arqu to 

> Instead of fetching ALL the logs we now fetch the specific log by ID and then all descendants and the parent 1 at a time

Yes, but only in special circumstances. This method is only used once, but it's a crucial use: when we're pulling a subset of a logbook that pertains to a dataset. logsync needs this to effectively "export" details about a given log to send over the wire.

> This is now possible as we have sort of a weak doubly linked list implementation?

Yes. The interface implies that any implementation of an `oplog.Store` must maintain a doubly-linked list. I'm hoping to formalize this in a later refactor by writing spec tests that get specific about what _exactly_ methods like `oplog.Logstore.Get` should return, and update the base implementation to use a doubly-linked list under the hood
> GetSparseAncestorsAllDescendants is just an interface definition at this point which will become meaningful later once propperly implemented and the current work is a fallback essentially?

Exactly this is an example of the "extension pattern", a term [Russ Cox has recently coined](https://youtu.be/yx7lmuwUNv8?t=928). Anyone that uses the  `oplog.GetSparseAncestorsAllDescendants` function will get the upgrade for free if they pass in the store that implements the extension interface. My hope is we can come up with a way to implement this interface with a single database call when the opstore is backed by a database to avoid many round trips.

PR #1486, opened 2020-08-05
refactor(p2p): add `QriIdentityService` to to replace `upgradeToQriConnection`

ramfox:
- mark old `QriProtocolID` (`/qri`) as deprecated, rename to `depQriProtocolID`
- add new `QriProtocolID` (`/qri/0.1.0`)
- add new `QriIdentityService` that manages upgrading qri connection, requesting profiles, and handling profile requests
- `QriIdentityRequest` checks if the connection supports the list of expected protocols, protects the connection, and requests the peer's qri profile

- added a test that:
  - spins up a 5 node network & connects each node
  - adds one more node with a bus that we pass it
  - the bus is subscribed to the `ETP2PQriPeerConnected` and `ETP2PQriPeerDisconnected` events
  - we test that we go from no connections to 5 expected peers whose profiles we have, back to no connections again after we disconnect

- tests are passing locally, but failing on ci and I'm at a bit of a loss


event/p2p.go
```diff
@@ -27,4 +27,22 @@ var (
...
+	ETP2PProfileSendSuccess = Type("p2p:Profile sent successfully")
+	// ETP2PProfileSendFailed fires when the profile exchange service
+	// failes tp completes a profile exchange
+	ETP2PProfileSendFailed = Type("p2p:Profile send failed")
+	// ETP2PProfileReceiveSuccess fires when the profile exchange service
+	// successfully completes a profile exchange
+	ETP2PProfileReceiveSuccess = Type("p2p:Profile received successfully")
+	// ETP2PProfileReceiveFailed fires when the profile exchange service
+	// failes tp completes a profile exchange
+	ETP2PProfileReceiveFailed = Type("p2p:Profile receive failed")
```
b5, line 47: These events are too fine-grained in my opinion. My rule when defining an event: come up with a concrete use case that can't make use of an existing event, and isn't a test. To me these events are exposing too much of the internals of the profile exchange service.

We should aim to make events as general as possible, and err on the side of _not_ defining them because they're easy to add, hard to remove, and a potential source of slowness.

How about this instead:

```go
const (
  // ...

  // ETP2PQriPeerConnected fires whenever a peer-2-peer connection that
  // supports the qri protocol is established
  // payload is a profile.Profile
  // subscribers cannot block the publisher
  ETP2PQriPeerConnected = Type("p2p:QriPeerConnected")
  // ETP2PQriPeerDisconnected fires whenever a qri peer-2-peer connection
  // is closed
  // payload is a profile.Profile
  // subscribers cannot block the publisher
  ETP2PQriPeerDisconnected = Type("p2p:QriPeerConnected")
)
```

We can emit `ETP2PQriPeerConnected` in what is presently `ETP2PProfileReceiveSuccess`, and wire up a new `libp2p.Event` handler for disconnects that calls a method on `*ProfileService` that checks if the disconnected peer is a qri peer and emits the event

p2p/profile_service.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/qri-io/qri/repo"
+	"github.com/qri-io/qri/repo/profile"
+)
+
+// ProfileProtocolID is the protocol id for the profile exchange service
+const ProfileProtocolID = "/qri/profile/0.1.0"
+
+// ProfileTimeout is the length of time we will wait for a response in a profile
+// exchange
+const ProfileTimeout = time.Minute * 2
```
b5, line 24: group yo constants

p2p/profile_service.go
```diff
@@ -0,0 +1,218 @@
...
+
+// ProfileService manages the profile exchange. This exchange should happen
+// whenever a node connects to a new peer
+type ProfileService struct {
+	Host    host.Host
+	Repo    repo.Repo
```
b5, line 30: it'd be be better if you could keep these unexported

p2p/profile_service.go
```diff
@@ -0,0 +1,218 @@
...
+	defer func() {
+		helpers.FullClose(s)
+		if err != nil {
+			ps.pub.Publish(ctx, event.ETP2PProfileSendFailed, p)
+			return
+		}
+		ps.pub.Publish(ctx, event.ETP2PProfileSendSuccess, p)
+	}()
+
+	log.Debug("%s received a profile request from %s %s", ProfileProtocolID, p, s.Conn().RemoteMultiaddr())
```
b5, line 80: `log.Debug` -> `log.Debugf`

p2p/profile_service.go
```diff
@@ -0,0 +1,218 @@
...
+
+	if err := ps.pub.Publish(ctx, event.ETP2PProfileRequestReceived, p); err != nil {
+		log.Errorf("%s publishing event %s", ProfileProtocolID, err)
+	}
+
+	pro, err = ps.Repo.Profile()
+	if err != nil {
+		log.Errorf("%s error getting this node's profile: %s", ProfileProtocolID, err)
```
b5, line 88: I've been using recent PRs to transition logging within error handling to debug level. error-level messages create additional noise that confuses users.

p2p/profile_service.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// receiveAndStoreProfile takes a stream, receives a profile off the stream,
+// and stores it in the Repo's ProfileStore
+func (ps *ProfileService) receiveAndStoreProfile(ctx context.Context, s network.Stream) {
+	defer func() {
+		go helpers.FullClose(s)
+	}()
+
+	log.Debug("%s received profile message from %s %s", s.Protocol(), s.Conn().RemotePeer(), s.Conn().RemoteMultiaddr())
```
b5, line 176: debugf

p2p/profile_service.go
```diff
@@ -0,0 +1,218 @@
...
+// exchange
+const ProfileTimeout = time.Minute * 2
+
+// ProfileService manages the profile exchange. This exchange should happen
+// whenever a node connects to a new peer
+type ProfileService struct {
+	Host    host.Host
+	Repo    repo.Repo
+	pub     event.Publisher
+	peersMu *sync.Mutex
```
b5, line 32: whenever you declare a mutex lock, add a comment about the fields it's governing: `// lock for peers map`

p2p/profile_service.go
```diff
@@ -0,0 +1,218 @@
...
+		err error
+		pro *profile.Profile
+	)
+
+	ctx := context.Background()
+
+	p := s.Conn().RemotePeer()
+
+	defer func() {
+		helpers.FullClose(s)
```
b5, line 72: if I understand this correctly, this closes the stream, but not the connection, right?

p2p/profile_service.go
```diff
@@ -0,0 +1,218 @@
...
+func (ps *ProfileService) ProfileRequest(ctx context.Context, p peer.ID) {
+	<-ps.ProfileWait(ctx, p)
+	return
+}
+
+// ProfileWait checks to see if a request to this peer is already in progress
+// if not, it sends a profile request to the peer
+// either way it returns a channel to wait on until the profile request has
+// been completed
+func (ps *ProfileService) ProfileWait(ctx context.Context, p peer.ID) <-chan struct{} {
```
b5, line 116: Does this need to be exported? Looks like  `ProfileService` users should be calling `ProfileRequest`

event/p2p.go
ramfox: Yes, I hear you. Definitely using them at crutches for the test-plan test, wanting as much incite into what is going on with the profile exchange as possible to debug. Will alter.

p2p/profile_service.go
ramfox: Exactly. It will close the stream, and wait for the other end of the stream to close as well. But it won't close the underlying connection.

go.sum
```diff
@@ -1071,6 +1071,10 @@ github.com/qri-io/jsonschema v0.2.0/go.mod h1:g7DPkiOsK1xv6T/Ao5scXRkd+yTFygcANP
 github.com/qri-io/qfs v0.1.1-0.20200623175200-23690284798a/go.mod h1:NP4Va1b0/V4nbyPdX3FmZ4cJTPatn0d5FZF5GBxcIJA=
 github.com/qri-io/qfs v0.5.1-0.20200722134009-7b5ac5539ce7 h1:sQS0Xx53fjvZGKF7+/S47zVoZJmFY8k+UyGCA8ek6sY=
 github.com/qri-io/qfs v0.5.1-0.20200722134009-7b5ac5539ce7/go.mod h1:NP4Va1b0/V4nbyPdX3FmZ4cJTPatn0d5FZF5GBxcIJA=
+github.com/qri-io/qfs v0.5.1-0.20200810200435-d477dc1dde1e h1:1j5YM2lb2OxBWSmKadeR/t0IVX+XE0QSUxxjiAvR3/8=
+github.com/qri-io/qfs v0.5.1-0.20200810200435-d477dc1dde1e/go.mod h1:NP4Va1b0/V4nbyPdX3FmZ4cJTPatn0d5FZF5GBxcIJA=
+github.com/qri-io/qfs v0.5.1-0.20200810213433-eb06cdd4b298 h1:+OkHcMupnS/IRUScpS3/4SkCpYQOZFFepRObXNGXjzI=
+github.com/qri-io/qfs v0.5.1-0.20200810213433-eb06cdd4b298/go.mod h1:NP4Va1b0/V4nbyPdX3FmZ4cJTPatn0d5FZF5GBxcIJA=
```
b5, line 1077: run `go mod tidy` to drop the old dep.

lib/lib.go
```diff
@@ -683,8 +683,8 @@ type Instance struct {
 
 // Connect takes an instance online
 func (inst *Instance) Connect(ctx context.Context) (err error) {
-	oldRemoteClientExisted := inst.remoteClient != nil
-	log.Debugf("inst.Connect oldRemoteClientExisted=%t", oldRemoteClientExisted)
+	// oldRemoteClientExisted := inst.remoteClient != nil
+	// log.Debugf("inst.Connect oldRemoteClientExisted=%t", oldRemoteClientExisted)
```
b5, line 687: any reason why this is removed? Mind clarifying why we no longer need to track & decrement the releaser on `remoteClient`?

This might be why `TestConnect` is failing in CI

p2p/profile_service.go
```diff
@@ -0,0 +1,227 @@
...
+
+	var (
+		err error
+		pro *profile.Profile
+	)
+
+	p := s.Conn().RemotePeer()
+
+	defer func() {
+		helpers.FullClose(s)
```
b5, line 80: Mind adding your clarification from review as a comment here? I find the `FullClose` method name confusing. (`FullCloseStream` would have been nicer)

```go
// close the stream, and wait for the other end of the stream to close as well
// this won't close the underlying connection
helpers.FullClose(s)
````

Adding explanatory comments will help others understand our p2p code better. six-months-from-now us will be thankful

File: p2p/node.go [hunk 13]
```diff
@@ -166,10 +171,15 @@ func (n *QriNode) GoOnline(ctx context.Context) (err error) {
...
 
+	n.profileService.StartProfileService(n.host)
+
 	// add multistream handler for qri protocol to the host
 	// setting a stream handler for the QriPrtocolID indicates to peers on
 	// the distributed web that this node supports Qri. for more info on
 	// multistreams  check github.com/multformats/go-multistream
+	// note: even though we are phasing out the old qri protocol, we
+	// should still handle it for now, since this is what old nodes will
+	// be relying on
```
b5, line 182: let's plan to drop support for the old protocol in v0.10.0

p2p/node.go
```diff
@@ -53,6 +53,10 @@ type QriNode struct {
 	// ipfs node provided by repo
 	Repo repo.Repo
 
+	// ProfileService listens for requests for profile exchanges and allows
+	// your node to request profile exchanges
+	profileService *ProfileService
```
b5, line 58: I'd personally prefer this field be named `profiles`. Unexported fields get to assume the user knows it's a `ProfileService` and will be referring to that API when working. I find it makes method definitions a little easier to read. For example:

```go
n.profileService.IsQriPeer(pi.ID)
```

becomes:
```go
n.profiles.IsQriPeer(pi.ID)
```

I find the latter easier to think about, and it makes me think more about the `ProfileService` API by creating a distinct use case.

p2p/node.go
```diff
@@ -405,10 +423,11 @@ func (n *QriNode) libp2pSubscribe() error {
 		for e := range sub.Out() {
 			switch e := e.(type) {
 			case libp2pevent.EvtPeerIdentificationCompleted:
-				log.Debugf("libp2p identified peer: %s\n", e.Peer)
-				n.upgradeToQriConnection(e.Peer)
+				log.Debugf("libp2p identified peer: %s", e.Peer)
+				// err := n.upgradeToQriConnection(e.Peer)
```
b5, line 427: drop ded code

p2p/node.go
```diff
@@ -381,6 +392,13 @@ func (n *QriNode) disconnected(_ net.Network, conn net.Conn) {
 	log.Debugf("disconnected from peer: %s", conn.RemotePeer())
 	pi := n.Host().Peerstore().PeerInfo(conn.RemotePeer())
 	n.pub.Publish(context.Background(), event.ETP2PPeerDisconnected, pi)
+
+	if n.profileService.IsQriPeer(pi.ID) {
+		pro, err := n.Repo.Profiles().PeerProfile(pi.ID)
+		if err == nil {
+			n.pub.Publish(context.Background(), event.ETP2PPeerDisconnected, pro)
+		}
+	}
```
b5, line 401: We're trying to reduce / remove direct use of the `Repo` interface so we can one day remove repo entirely. To that end, the new profiles service should own access to peer profiles. It'd passed a `Repo` instance on construction, and can manage these `PeerProfile` calls. 

Let's refactor `IsQriPeer` to be a `PeerProfile` accessor method that returns nil when no profile exists for a given peer.ID:

```suggestion
	if pro := n.profiles.PeerProfile(pi.ID); pro != nil {
		n.pub.Publish(context.Background(), event.ETP2PPeerDisconnected, pro)
	}
```

p2p/profile_service.go
```diff
@@ -0,0 +1,227 @@
...
+	ps.peersMu.Unlock()
+	return ok
+}
+
+// StartProfileService adds a host and a profile handler to the host
+// to the profile service
+func (ps *ProfileService) StartProfileService(h host.Host) {
+	ps.host = h
+	h.SetStreamHandler(ProfileProtocolID, ps.ProfileHandler)
+}
```
b5, line 65: I think this could just be `Start`

p2p/profile_service.go
```diff
@@ -0,0 +1,227 @@
...
+// NewQriProfileService creates an profile exchange service and adds
+// the `ProfileHandler` to the host
+func NewQriProfileService(r repo.Repo, p event.Publisher) *ProfileService {
+	ps := &ProfileService{
+		repo:    r,
+		pub:     p,
+		peersMu: &sync.Mutex{},
+		peers:   map[peer.ID]chan struct{}{},
+	}
+	// h.SetStreamHandler(ProfileProtocolID, ps.ProfileHandler)
```
b5, line 47: drop ded code

p2p/profile_service.go
```diff
@@ -0,0 +1,227 @@
+package p2p
```
b5, line 1: Generally, we should aim to get a local test for this code up as well as the testground work. We can do that in a follow-up

lib/lib.go
ramfox: ahh boo, I thought I was so cool and added a comment here ala uhLeeshUh, but it was a "pending" review, not a comment, here it is:

I am getting "negative wait group" errors in test-plans from the `oldRemoteClientExisted` 
```
  if oldRemoteClientExisted {
      inst.releasers.Done()
  }
```
I can't find a difference in the way we connect our normal qri instances than our test-plans qri instances, so I'm wondering if maybe this doesn't work exactly the way we expect? What was it added for?

p2p/node.go
ramfox: I'm planning on doing a larger rename in the next iteration, since it is going to grow past just a profile service and into more of a qri identification/qri connection manager type service. qriIDService? qriConns?

Happy to do this rename now tho.

p2p/node.go
ramfox: Hmmm, what about ConnectedPeerProfile? Because we are checking if this is a connected peer, not a peer that we have potentially seen before. Or is it enough to say in the comment of `PeerProfiles` "only returns profiles of peers that we have connected to this session"

ramfox: Actually, refactored to have `profiles.HandleQriPeerDisconnect` and doing all the work of emitting the events and looking for the profile into the profile service. Now the `ETP2PPeerDisconnected ` and `ETP2PPeerConnected` events both get emitted from the same service.

lib/lib.go
b5: Interesting. My hunch is an interaction within the `cmd` package, probably along the `qri connect` codepath. I'm guessing there may be 2 invocations of `QriNode.GoOnline` when calling `qri connect`, and only 1 when using qri directly as a library.

lib/lib.go
```diff
@@ -694,8 +691,11 @@ func (inst *Instance) Connect(ctx context.Context) (err error) {
 	// for now if we have an IPFS node instance, node.GoOnline has to make a new
 	// instance to connect properly. If remoteClient or remote retains the reference to the
 	// old instance, we run into issues where the online instance can't "see"
-	// the additions. We fix that by re-initializing the client and remote with the new
-	// instance
+	// the additions. We fix that by shutting down the previous client and
+	// re-initializing the client and remote with the new instance
+	if inst.remoteClient != nil {
+		<-inst.remoteClient.Shutdown()
+	}
```
ramfox, line 698: Added `Shutdown` method to `remote.Client`. This can also be refactored into lib, and have lib control the shutdown, but this felt cleaner. Would love opinions tho.

ramfox:
smoke tested using `qri connect` & other commands, all looks good!

ramfox:
`TestConnect` is failing on ci, but passing locally. Taking another look.

ramfox:
`TestConnect` & `TestAnnounceConnected` both failing on CI but passing locally, even with `-count 10`
At a bit of a loss here. Voting to `t.skip` and move on.

ramfox:
Added `TestQriIdentityService`!

  - spins up a 5 node network & connects each node
  - adds one more node with a bus that we pass it
  - the bus is subscribed to the `ETP2PQriPeerConnected` and `ETP2PQriPeerDisconnected` events
  - we test that we go from no connections, to 5 expected peers whose profiles we have, back to no connections again after we disconnect

ramfox:
hmmm, now getting this error in ci tests:
```
Failed
panic: qtls.ClientSessionState not compatible with tls.ClientSessionState

goroutine 1 [running]:
github.com/lucas-clemente/quic-go/internal/handshake.init.2()
github.com/lucas-clemente/quic-go/internal/handshake.init.2()
```

When trying to figure out why this exists:
```
$ go mod why github.com/marten-seemann/qtls
github.com/marten-seemann/qtls
github.com/qri-io/qri/p2p
github.com/ipfs/go-ipfs/core
github.com/ipfs/go-ipfs/core/node/libp2p
github.com/libp2p/go-libp2p-quic-transport
github.com/lucas-clemente/quic-go
github.com/lucas-clemente/quic-go/internal/handshake
github.com/marten-seemann/qtls
```

Not totally sure what this is coming from. Will dig in tomorrow, but @b5 if you have any off-the-cuff thoughts I'd appreciate it.

dustmop:
quic is basically http 3.0, probably libp2p uses it to communicate. I would guess by this error that either our tlc or quic dependency need to be updated, are the go.mod entries for either of those far behind the latest?

ramfox:
so all our deps referencing `quic` or `tls` or `qtls` are indirect deps, so nothing we have explicitly pulled that would make a parent dependency out of date.

Also, this is only happening on circle ci, not locally. And the error showed up after re-running the tests on the same commit. I'm wondering if there is something messed up in a cache?

ramfox:
Also `TestConnect` error, that was fixed earlier in this pr is now back. Will look into it again in the morning!

b5:
after googling `panic: qtls.ClientSessionState not compatible with tls.ClientSessionState`, a number of other projects seem to come across this issue when running go v1.11.0, and the solution was to upgrade to go v0.12+

There's a new patch of golang out. A few users on twitter have pointed to regressions in the networking package. We run CI against golang:latest, and that image was upgraded 14 hours ago. @ramfox you ran tests 9 hours ago. My guess: latest version of go may be the issue. wanna try explicitly changing our CI target version to go 1.14.6?

p2p/connected_test.go
```diff
@@ -26,47 +29,71 @@ func TestAnnounceConnected(t *testing.T) {
 		nodes[i] = node.(*QriNode)
 	}
 
+	// set up bus & listener for `ETP2PQriPeerConnected` events
```
ramfox, line 32: I'm looking back on this test, and wondering what exactly it is supposed to be testing.

First, we set up a network of nodes & explicitly connect them.
Then we create a separate node that hasn't been connected explicitly, and only connect it to a single node.
Then we "AnnounceConnected" and wait until we get a message from each node.

Are we testing that AnnounceConnected works? Or are we trying to tell if we can get connections to each node just by connecting to one node in the network? 

This also ignores the part that sometimes nodes are connecting without us explicitly connecting them, but they do it through dhs, not because of a 2nd degree qri connection.

I'm wondering if this is doing too much, and we should be explicitly connecting the main test node to each node in the test network. The test would be much more reliable.

b5: > Are we testing that AnnounceConnected works? Or are we trying to tell if we can get connections to each node just by connecting to one node in the network?

Both. The definition of working for `AnnounceConnected` is to propagate that message to other qri peers.

> This also ignores the part that sometimes nodes are connecting without us explicitly connecting them, but they do it through dhs, not because of a 2nd degree qri connection.

I'm assuming you mean through the DHT here, and yes there are lots of ways qri peers can connect. This test is checking one of those ways.

Is this test causing problems? I'm happy to disable it for now if so. To me the bigger issues I'm seeing in CI is the resurfaced `TestConnect` failure in the cmd package, and making sure`TestQriIdentityService` passes consistently. Those two enough are a huge slog, so I'd be happy to disable this to focus on those

ramfox: re: dht, yea typo

I'm a bit confused about exactly how AnnounceConnected is supposed to fit in with everything.

From following the code right now, we have 
- `AnnounceConnected` which sends a message to each connected qri peer w/ "my" addr info. It is being used in one place: after we've had one successful connection to a bootstrap peer, we `AnnounceConnected`
- `handleConnected` checks to see if we've received this message before, if not, it decodes the message and requests the profile from the original node.

But this is just what `upgradeQriConnection` and now `QriIdentityService` does except from the other direction, we wait until the connection has been fully identified by libp2p and then we request the peer's profile. In this case, we are saying instead "Hey I'm here, request my profile", but to peers that we have already established on (our side) that are qri connections. Those peers don't turn around and further propagate our info to their peers.

If the goal is that we should be propagating the existence of newly connected peers, I think we should be doing it another way. 

Once we have "upgraded" to a qri connection, we should send a message to our other connections "hey this peer exists". There should be a loop in the `QriIdentityService` that listens for this messages. If you receive one of these messages, check if you are already connected the peer. If so, drop ignore it. If not, establish a connection, which will trigger the QriIdentityService, which will propagate that peer to your already connected peers.

I think this is what we are trying to achieve, and our code may have done that at one point, but we've changed so much that it's not currently doing that.

b5: Completely agree. Your most recent comment clarifies: this isn't so much about what the test is doing as: we have overlapping functionality between `AnnounceConnected` and the new identity service. Let's remove `AnnounceConnected`

ramfox:
Okay so back in the world of not understanding why sometime the peers won't connect when you dial them and sometimes they disconnect without reason.

Locally, when I run the `TestQriIdentityService` test individually, it always passes. When I run it in a count of, for example, 10, it fails multiple times with peers sometimes not connecting at all, and sometimes disconnecting for no reason.

ramfox:
UGH DISABLING DISCOVERY JUST FIXED IT UGH

ramfox:
hahaaaaomg @b5#1924 I am at a loss now. So got everything working no race conditions, no failing connections, no random disconnects...

running this on circle ci:
```
expected list of connected peers different then the given list of connected peers: 
      expected: [
QmPeFTNHcZDr3ZFEfFfepxS5PqHAmfBRGQNPJ389Cwh1as
QmVCuW9TryE7EMi6KDpu6fipmVp2wATp2pPunTsH86LL3k
QmWYgD49r9HnuXEppQEq1a7SUUryja4QNs9E6XCH2PayCD
QmZn1v9w5mkGMWdThF7spirifjWEa58k1zP7Ksxn4hyziG
QmbHm5S9hRsvZWE56XWA7dksfXDNABRwK6LaH6KS1z7uEw
]
got: [
QmPeFTNHcZDr3ZFEfFfepxS5PqHAmfBRGQNPJ389Cwh1as
QmWYgD49r9HnuXEppQEq1a7SUUryja4QNs9E6XCH2PayCD
QmZn1v9w5mkGMWdThF7spirifjWEa58k1zP7Ksxn4hyziG
QmbHm5S9hRsvZWE56XWA7dksfXDNABRwK6LaH6KS1z7uEw
QmcRG1rLTs2cE7v9JLzBT8vHJ381D73LxzdnXWj275hyjg
]
```
... before node `QmVCuW9TryE7EMi6KDpu6fipmVp2wATp2pPunTsH86LL3k` could connect
SOME RANDOM NODE `QmcRG1rLTs2cE7v9JLzBT8vHJ381D73LxzdnXWj275hyjg` connected
where the fuck did this one come from?!?!
This means even if I control for making sure we don't continue on until the expected nodes are the ones that connected... we will still have this random `QmcRG1rLTs2cE7v9JLzBT8vHJ381D73LxzdnXWj275hyjg` node in our list of connections!!!

b5:
So if I understand properly what's going on here, you needed rename every test peer in the suite to figure out which test was "leaking" & accepting local connections? If so, ouch, but worth it. 

If that's the case, did you find the stray test repo?

ramfox:
I DID!

It was our other problem child `TestConnect`!!!

Any resource in the instance that DOES NOT have the same context as the instance (aka wasn't created/launched at the same time as we created the instance itself) does not get cleaned up/closed at the same time as the instance. This doesn't affect us when we use qri in the wild, because the contexts are tied to the life of the process, so when the process closes down the resources are released. But in our tests, that's not necessarily true. 

We were not explicitly closing the QriNode when we do `instance.Shutdown`, because when we `node.GoOnline` we create a new context, one that's not tied to the context of the instance.

To fix, I've explicitly added `node.GoOffline` to our `inst.Shutdown`, and refactored `node.GoOffline` to not panic if we have never brought the peer online in the first place.

This has been such a learning experience. It's making we wondering if we should do another refactor (at some point) to the way we clean up instance. I've found that it's been really simple if each subsystem has its own way of "shutting down",  a `Shutdown` func that returns its `Done()` channel.
```
func (inst *Instance) Shutdown() {
  // non-parallel version
  <-inst.qfs.Shutdown()
  <-inst.node.Shutdown()
  <-inst.remoteClient.Shutdown()
  ... etc
  return errCh
}
```

OR just refactor so we are waiting on each subsystem's `Done()` channel in the `Shutdown` func, rather than spreading it throughout the build process, and require that any time we use an instance, we expect it to be `Shutdown`.

Anyway, just some thoughts as I was working through this. Very very fascinating to learn about and my experience with contexts & channels is now through the roof :)

ramfox:
One other nice side effect of me renaming the peer on every test, is a second set of eyes on the test names able to go in and adjust when what the test was testing wasn't actually reflected in the name of the test runner. I bet it's super easy to copy and  paste a bunch of this testing boiler plate and forget to adjust the name.

p2p/identity_service.go
```diff
@@ -0,0 +1,286 @@
...
+	// ProfileTimeout is the length of time we will wait for a response in a
+	// profile exchange
+	ProfileTimeout = time.Minute * 2
+)
+
+var (
+	// expectedProtocols is the list of protocols we expect a node that has a
+	// profile service to speak
+	expectedProtocols = []string{
+		string(QriProtocolID),
```
b5, line 32: do we need the `QriProtocolID` service? I'd love to be able to deprecate it entirely.

p2p/identity_service.go
```diff
@@ -0,0 +1,286 @@
...
+	pro, err := q.profiles.PeerProfile(pid)
+	if err != nil {
+		log.Debugf("error getting peer profile: %s", err)
+		return nil
+	}
+	return pro
+}
+
+// Start adds a host and a profile handler to the host
+// to the profile service
```
b5, line 120:
```suggestion
// Start adds a profile handler to the host, retains a local reference to the host
```

p2p/ping_test.go
```diff
@@ -8,6 +8,7 @@ import (
 )
 
 func TestPing(t *testing.T) {
+	t.Skip("for TestQriIdentity test")
```
b5: still need to skip?

p2p/node_test.go
```diff
@@ -59,6 +60,7 @@ func TestNewNode(t *testing.T) {
 }
 
 func TestNodeEvents(t *testing.T) {
+	t.Skip("for TestQriIdentity test")
```
b5: still need to skip?

p2p/test/p2ptest_test.go
```diff
@@ -12,6 +12,7 @@ import (
 // - we have a tag on each peer
 // - we have the protocols each peer supports
 func TestConnectNodes(t *testing.T) {
+	t.Skip("for TestQriIdentity test")
```
b5: still need to skip?

p2p/identity_service.go
```diff
@@ -0,0 +1,286 @@
...
+		string(ProfileProtocolID),
+	}
+	// ErrPeerNotFound is returned when the profile service cannot find the
+	// peer in question
+	ErrPeerNotFound = fmt.Errorf("peer not found")
+)
+
+// QriIdentityService manages the profile exchange. This exchange should happen
+// whenever a node connects to a new peer
+type QriIdentityService struct {
```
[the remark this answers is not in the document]
ramfox: would you prefer `QriProfileService` as a name, because it would be way easier to change the protocol to `/qri/identity/0.1.0`

The reason I prefer `IdentityService` is it is doing more than just exchanging profiles, it's verifying the identity of the remote peer & ensuring a lasting connection.

But this is not a hill I care about dying on, so if you prefer `QriProfileService` that's fine!

p2p/identity_service.go
ramfox: `QriProtocolID` is now the new `/qri/0.1.0` protocol id, not the old `/qri` protocol id.
My logic is it can be a protocol that we basically reserve to use, but probably won't ever use to send messages on, but rather as a way to potentially to control protocol versioning. If we see that the `QriProtocolID` is the wrong version, we can immediately back off or respond in a specific way

ramfox: again, not married to this idea, if you'd rather we remove it entirely, just a potential logic for keeping it

p2p/ping_test.go
ramfox: yea crossed wires, only keeping 2 skips of the bunch I had to figure out this bug, as they were more flakey than the others and need some attention, but aren't critical

p2p/identity_service.go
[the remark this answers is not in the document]
b5: Yeah totally agree it's doing more than profile exchange. My concern is I don't see "Identity" as being a broader concept than "Profile". We have other places within the codebase that do this profile / identity mixup, and I think it's a source of confusion.

To me the method names on this service do a really good job of articulating what it's doing. I know it's more work, but I think it's worth it.

p2p/identity_service.go
b5: love it. Let's keep it!

p2p/identity_service.go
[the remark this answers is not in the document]
ramfox: no prob :)

PR #1487, opened 2020-08-06
fix(setup): Fix setup, add many unit tests

dustmop:
Due to a recent refactoring, the `qri setup` command was no longer prompting for a username. This caused the username validation to fail. Passing a username via the `--username` flag was also ignored, as the code in lib/setup.go would overwrite it. Fix these problems, add tests to make sure they won't happen again.

Rename the `--peername` flag to `--username` for `qri setup`

### DIFF
--- a/cmd/setup.go
+++ b/cmd/setup.go
@@ -145,7 +145,7 @@ func (o *SetupOptions) DoSetup(f Factory) (err error) {
 	if o.Username != "" {
 		cfg.Profile.Peername = o.Username
 	} else if !o.Anonymous {
-		cfg.Profile.Peername = inputText(o.Out, o.In, "choose username (leave empty to generate a default name):", "")
+		cfg.Profile.Peername = prompt(o.Out, o.In, "choose username (leave empty to generate a default name):")
 	}
 
 	// If a username was passed with the --username flag or entered by prompt, make sure its valid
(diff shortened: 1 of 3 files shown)


cmd/setup.go
```diff
@@ -136,21 +134,25 @@ func (o *SetupOptions) DoSetup(f Factory) (err error) {
...
 	}
 
-	if o.Peername != "" {
-		cfg.Profile.Peername = o.Peername
-	} else if cfg.Profile.Peername == doggos.DoggoNick(cfg.Profile.ID) && !o.Anonymous {
-		cfg.Profile.Peername = inputText(o.Out, o.In, "choose a peername:", doggos.DoggoNick(cfg.Profile.ID))
+	// Handle the --username flag, or prompt the user for a username
+	if o.Username != "" {
+		cfg.Profile.Peername = o.Username
+	} else if !o.Anonymous {
```
b5, line 147: we should also obey `noPrompt` here.

dustmop: `noPrompt` is not supported by `setup`, because the function that assigns `noPrompt` is Qri.Init which runs when an instance is created. An instance isn't created for `setup` because it can't be created before the repo exists. Should we add `noPrompt` handling here that copies what Qri.Init is doing? If so, is there any point in having both `--no-prompt` and `--anonymous`?

b5: ah that makes sense. We also have the `--setup` flag itself, which is a no-prompt setup. Ok my bad ignore me, Let's merge!

PR #1490, opened 2020-08-10
chore(deepdiff): Bump deepdiff to get recent fixes

dustmop:


b5:
mind running `go mod tidy`? it should remove the prior versions of deep diff from `go.sum`.

PR #1493, opened 2020-08-19
refactor(p2p): add p2p `ResolveRef` & `NewP2PRefResolver`

ramfox:
- add a p2p `ResolveRef` that asks all connected peers to resolve the given reference
- add `handleResolveRef` that handles `ResolveRefProtocolID` streams & returns a hydrated ref, if it has one, or a nil ref if it doesn't
- add `NewP2PRefResolver` that returns a `p2pRefResolver`
- adjust `inst.resolverForMode` to use `inst.p2pResolver` & add `inst.p2pResolver` as a `dsref.ParallelResolver` for mode "network"
- add `localResolver` as a field for `QriNode`, we need a localResolver to help when we handle ref resolution requests - we need to try and resolve that ref locally & return what we find (if we find anything) to the request node
- adjust `NewQriNode` signature to take a `localResolver dsref.Resolver`


p2p/node.go
```diff
@@ -113,10 +117,17 @@ func NewQriNode(r repo.Repo, p2pconf *config.P2P, pub event.Publisher) (node *Qr
...
-		// later
+		// Make sure we always have proper IOStreams, this can be set later
 		LocalStreams: ioes.NewDiscardIOStreams(),
 	}
+
+	// TODO(ramfox): when we remove the repo, we need to be sure that the node
+	// has a reference to the dscache in order for it to be possible to resolve
+	// any references locally
+	// use repo and dscache reference inside repo to create a local resolver
+	node.localResolver = dsref.SequentialResolver(r.Dscache(), r)
```
ramfox, line 128: Okay, want an opinion here. When I first coded this up, I had the `localResolver` come from a param passed into `NewQriNode`. I remembered we didn't really want other systems, other than the instance, to have control over resolution.

However, we need the qri node to have a local resolver when we get a request from a peer to see if we have the ref in question on our node.

I ended up removing the passed in `localResolver` param, and creating a local resolver when we create the node. Honestly, I think mostly for aesthetic reasons. 

I'm wondering if we should be switching over to a system closer to the way we do NewInstance: adding repo, bus, and localResolver fields to the p2p config, and setting them dynamically through options.

p2p/ping_test.go
```diff
@@ -20,6 +20,9 @@ func TestPing(t *testing.T) {
 
 	// Convert from test nodes to non-test nodes.
 	peers := asQriNodes(testPeers)
+	for _, peer := range peers {
+		peer.Discovery.Close()
```
ramfox, line 24: One major discovery during this process: we need to ensure that peer's Discovery capabilities are turned off for certain tests, and we need to make sure we `node.GoOffline` when we finish the tests.

p2p/node.go
b5: > Honestly, I think mostly for aesthetic reasons.

I know the feel. It feels like adding complexity to the `NewQriNode` constructor, but it's _necessary_ complexity. We need to be able to configure the notion of "local" at the instance layer. qri cloud uses a different set of resolvers as it's local stack, making this code a no-go as it's currently written.

I think `NewQriNode` should just take the extra param for now. This is a low level API, where it's supposed to be OK to have very complex constructors.

p2p/node.go
```diff
@@ -113,10 +117,17 @@ func NewQriNode(r repo.Repo, p2pconf *config.P2P, pub event.Publisher) (node *Qr
...
+
+	// TODO (ramfox): remote `MakeHandlers` when we phase out older p2p functions
```
b5, line 130: remote -> remove

p2p/resolve_ref.go
```diff
@@ -2,88 +2,186 @@ package p2p
...
+			numReqs--
+			if !res.ref.Complete() && numReqs == 0 {
+				return "", dsref.ErrRefNotFound
+			}
+			if res.ref.Complete() {
+				*ref = *res.ref
+				return res.source, nil
 			}
+		case <-streamCtx.Done():
+			log.Debug("p2p.ResolveRef request timed out before resolving ref")
```
b5, line 77: this might not be a timeout. the parent context might be cancelled. I say this because that's really what we should be doing in the long run: creating a sub context that's passed to all parallel resolvers and then cancelling when any of them returns with a resolved reference.

This way, if the registry is in a parallel resolver with the p2p resolver, and it comes back first, all this p2p work will be cleaned up faster

p2p/resolve_ref.go
```diff
@@ -2,88 +2,186 @@ package p2p
...
+		refMu.Lock()
+		reqRef := refCp.Copy()
+		refMu.Unlock()
+		go func(pid peer.ID, reqRef *dsref.Ref) {
+			source := rr.resolveRefRequest(streamCtx, pid, reqRef)
+			resCh <- resolveRefRes{
+				ref:    reqRef,
+				source: source,
+			}
+		}(pid, &reqRef)
```
b5, line 62: not sure we need the mutex lock here. passing copies should suffice:

```suggestion
		go func(pid peer.ID, reqRef dsref.Ref) {
			source := rr.resolveRefRequest(streamCtx, pid, &reqRef)
			resCh <- resolveRefRes{
				ref:    reqRef,
				source: source,
			}
		}(pid, reqRef.Copy())
```

This would mean dropping the pointer from the  `resolveRefRequest.ref` struct field.

p2p/resolve_ref.go
```diff
@@ -2,88 +2,186 @@ package p2p
...
+	"github.com/qri-io/qri/dsref"
 )
 
-// MtResolveDatasetRef resolves a dataset reference
-const MtResolveDatasetRef = MsgType("resolve_dataset_ref")
+const (
+	// p2pRefResolverTimeout is the length of time we will wait for a
+	// RefResolverRequest response before cancelling the context
+	// this can potentially be a config option in the future
+	p2pRefResolverTimeout = time.Second * 60
```
b5, line 20: let's drop this to 20 seconds. We're only firing requests to all connected peers, not yet doing any fancy re-firing as new peers connect or anything, so there's very little reason to have this be longer than 20 seconds.

Also, ref resolution is supposed to (hopefully) be hella fast.

p2p/resolve_ref.go
ramfox: 🤣  got a little overzealous fixing a race condition

p2p/resolve_ref_test.go
```diff
@@ -2,66 +2,150 @@ package p2p
...
+				}
+			}
+			connectedPeers = append(connectedPeers, pid)
+			if len(connectedPeers) == numNodes {
+				close(qriPeerConnWaitCh)
+			}
+		}
+		return nil
+	}
+	bus.Subscribe(watchP2PQriEvents, event.ETP2PQriPeerConnected)
```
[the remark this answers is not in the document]
ramfox: Yes I agree! Been thinking about that. We have a bunch of variables about how a specific test might want their network set up:
- # of nodes
- are they connected or not
- do they have discovery active or not
- do we need event handlers to coordinate specific situations
- do we have one single node that is the main node we are looking at? Does that one have special behavior regarding the above: connected to all nodes, connected to one node, discovery active, specific event handler

Seems like we should have two main functions, both of which should return a `shutdown` function that allows us to spin down the network at the end of the test:
NewQriTestNetwork - takes a factory, number of nodes, and some sort of params structs that describes the specifics of the network (discovery, connection, bus & event handlers)
NewQriTestNode - takes a factory, an optional list of TestPeers, and a params struct that describes the setup of the node (discovery, # of connected nodes, bus & event handlers)

These may replace functions that already exist.

We can generate a bus & event handler that ensures the nodes all get connected, and then replace them with the specified bus and event handler. And if you need a reallllly specific setup, then you have to spin up the network on your own.

p2p/resolve_ref_test.go
```diff
@@ -2,66 +2,150 @@ package p2p
...
+			Name:      r.Name,
+		})
+		cache := builder.Build()
+		nodeWithRef.Repo.Dscache().Assign(cache)
+		testResolveRef := &dsref.Ref{
+			Username: r.Username,
+			Name:     r.Name,
+		}
+		nodeWithRef.Repo.Dscache().ResolveRef(ctx, testResolveRef)
+		t.Logf("test resolve ref %v\n", testResolveRef)
```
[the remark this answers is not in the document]
ramfox: I actually left it on purpose so you have more details about what happened, since the logging only show unless you specifiy -v or the test fails. But, happy to remove it!

PR #1494, opened 2020-08-19
fix(resolve): Fix ref resolution when given divergent logbook data

dustmop:
This is the first step in fixing a really tricky problem. Here's what is going wrong right now:

1) `user_a` creates a qri repository, at time `t0`. This generates their private key, profile ID, ipfs repo, refstore, and logbook. The profile ID is a hash of the private key, so it's going to be unique for this user and their repository. The initial logbook entry includes the profile ID, and the hash of this entry creates what I'll call the "UserCreateID".

2) That user then saves a dataset called `my_ds` and pushed it to remote `R`. The remote now has the ipfs blocks and logbook log for `user_a/my_ds`, as well as the logbook op for the creation of `user_a`, which is a parent of the dataset's log. So the logbook looks like this:

```
  user_a (the hash of this is UserCreateID)
    user_a/my_ds (the hash of this is my_ds's initID)
      user_a/my_ds/main (the main branch)
```

3) `user_b` points at `R` and then runs `qri pull user_a/my_ds`. They get the ipfs and logbook data.

4) At time `t1`, `user_a` now creates a new qri repository using the same ProfileID, either by using the QRI_CONFIG_SETUP_DATA environment variable, or by copying their repo and running `qri registry prove`. Thought they have the same ProfileID, they are creating a new logbook, so the "UserCreateID" has now changed.

5) `user_a` adds a new version to the existing dataset `my_ds`. They push this to `R`.

6) `user_b` pulls the dataset again using `qri pull user_a/my_ds`. Because of how logbook.MergeLog works, it merges elements based on the "UserCreateID", or the hash of the first op in the logbook. This essentially orphans data in the logbook, which usually orphans the dataset, making it unable to resolve locally.

There's a couple of problems that need to be fixed here, but this PR addresses only (6). When merging logs, they now fallback to comparing ProfileIDs, instead of just the UserCreateID, and local resolution of references is once again using the refstore, since logbook now has invalid data and will fail to calculate the ids necessary to create a fully-populated dsref.

There were also some disabled tests and bugs in tests that needed to be fixed in order to correctly test this. The MockRemoteClient in particular now creates actual IPFS blocks and logbook data that matches it, in order to test how we merge logs when pulls happen.

The longer term fixes will involved a redesign of logbook, and changes to how remotes handle disagreements in data, but I'll file a different issue for that.


Arqu (approved):
Very nice breakdown, thanks for that. I left a nit but can live without it.

Since on some points I'm out of my depth, out of curiosity, is it possible long term we live without `UserCreateID` entirely and simply work off `ProfileID` especially as we move towards pkey rotations in the future? This presumably should reduce the fragmentation of logs in general.

PR #1496, opened 2020-08-24
feat: remote access control

b5:
Once this PR is merged users can _optionally_ define a `remote_access_control.yaml` file within their qri repo. That YAML file  is a _policy_ file that will control actions on the remote. 

I owe a little more documentation about how the access control grammar works, but have a look at the tests & especially the example test for a demo of how it works. Some key points:
* a _policy_ is a set of rules that define access control
* a _rule_ has an effect (allow/deny) for a person (subject) trying to do something (action) to one or more resources (datasets) 
* `subject` values are users, identified by peerID 
* `resource` and `action` types can be arranged into hierarchies with the `*` globbing pattern. `dataset:*` will match `dataset:b5:world_bank_population` and `dataset:ramfox:lol_stats:body`, but not, say `user:ramfox`
* action names in policy documents will closely match their name within lib. `remote:push` corresponds to `lib.RemoteMethods.Push`
* resources a little trickier. For now `dataset` is a catch-all to refer to all supporting data for a dataset, versions, logbook data, previews, etc.

### Steps to complete
* [x] access control package defining grammar
* [x] add an unexported `*access.Policy` field to `remote.Remote`, and a `Policy` field to remote options that sets the field in `NewRemote`
* [x] within each remote hook, call `rem.Policy.Enforce`, passing in the requesting profile and the resource/action combo. As an example the `dsPushPreCheck` should check if r.Policy is non-nil. If it is, read the profile.Profile of the user that's pushing from their peerID and call `r.Policy.Enforce(pro, "dataset:REF_USERNAME:REF_NAME", "remote:push')`, returning any non-nil error
* [x] write a big old test that proves access can be denied
* [x] define a `OptLoadFileIfExists(filename string)` remote option function that checks for a policy file at the given path, populates `remote.Options.Policy` if so.
* [x] call that `OptLoadFileIfExists` in `lib.NewInstance` constructor, passing in the calculated repo path + the `remote_access_control.yaml` constant


ramfox:
So I'm guessing we need the "matchSubject" and the "subjectUsername" comparison because we want to be able to add a policy that says "For any subject, allow them to do xyz to their own datasets".

Also, we can potentially have policy combinations that allow a subject to push/pull/delete resources that aren't in their namespace. Because of this, we need, inside the hook functions, a way to get both the profile information of the subject that is doing the action, as well as the information of the resource itself. Currently, the `meta` field for remote hooks only has the `pid` of the subject requesting the action. The Username/Peername that exists in the `meta` param is the Username/Peername of the reference, not the subject.

In order to satisfy our needs for access control, we need each implementation of dsync push pull (and maybe logsync push pull) to send over the peername/username of the pushee/pullee.

b5:
> So I'm guessing we need the "matchSubject" and the "subjectUsername" comparison because we want to be able to add a policy that says "For any subject, allow them to do xyz to their own datasets".

Yep. 

> In order to satisfy our needs for access control, we need each implementation of dsync push pull (and maybe logsync push pull) to send over the peername/username of the pushee/pullee.

We should chat about this, but I don't think we should modify dsync / logsync for this. I'm hoping we can use something like `repo.Profiles().PeerProfile(pid)` within the hook functions to fetch profile details for the request. For this to work properly, we need to make sure that store of profile information is as up-to-date as possible, and IIRC our new profile service isn't updating this (admittedly antiquated) service.

The fun bit here: your new p2p profile service emits events that profile.Store implementations could subscribe to & record.

ramfox:
You are right dsync and logsync are not the right place for this anyway.

`Client` is where the `params` that we use in the remote/remote hooks are coming from.

Potential way forward: Adding `username` field to `Client`. When `NewClient` is created, take the `username` from the `node.Repo.Profile()`. We are already relying on `node.Repo.PrivateKey()` to store the private key in the Client. Both will have to be altered if we remove a reference to Repo in the node at some point.

Alter the function that creates the params for the remote request:
`func sigParams(pk priv.Key, ref *dsref.Ref) (map[string]string, error)` => `func sigParams(pk priv.Key, username string, ref *dsref.Ref) (map[string]string, error)`
map would now have an additional `subject` field containing the client's username

function that parses the params:
`func (r *Remote) pidAndRefFromMeta(meta map[string]string) (profile.ID, dsref.Ref, error)` =>
`func (r *Remote) subjAndRefFromMeta(meta map[string]string) (*profile.Profile, dsref.Ref, error)`
Now returns a `profile.Profile`, rather than just a `profile.ID`

ramfox:
Let me know what you think of the above, otherwise, can alter so the remote itself needs to grab the profile in the hook.

ramfox:
Also just confirming: do we need to `policy.Enforce` during every hook, or just the preChecks? Was assuming it was the preChecks, but wondering if I'm missing a nuance.

ramfox:
Can't forget about those dirty liars!

ramfox:
hmmm looks like we don't have `dsync.PullPreCheck` or a `dsync.RemovePreCheck`.

We have a `GetDagInfoCheck`, and `GetDagInfo` looks like the first step in which the remote is involved in a `pull`. However, the `GetDataInfoCheck` happens after we've already taken some actions. Proposing we need to add a `remote.dsPullPreCheck` that gets passed into the `dsyncConfig.PullPreCheck`, which actually gets set as`dsync.GetDagInfoPreCheck` & called in `GetDagInfo`

As for `RemovePreCheck`, dsync doesn't really handle remove, it handles `RemoveCID`, which is called to remove each block individually. Each time `dsync.RemoveCheck` is called, but this isn't really what we are looking for in this iteration. We want to stop the remove request before the remote even gets to the `RemoveCID` action. We have a `remote.RemoveDataset` func. We should call the `policy.Enforce` inside `remote.RemoveDataset` at the start of the function.

ramfox:
@b5 thoughts: should a remote be expected to have a policy, and so we use a `DefaultRemotePolicy` when we create the remote, if no policy was added, or should it be okay to have a `nil` policy that permits everything?

ramfox:
Revisiting the profile/username problem, the method I came up with doesn't work. Logsync doesn't have the same `params` that are sent over, so this only works for dsync

However, I'm not sure that the method of having remote look for the profile of the associated profile ID will work either. The new profile service DOES add the profiles to the profile store in the repo! However, this is only happening over p2p, if the nodes are connecting via HTTP then I don't think there is a guarantee that the profile will exist in the profile store.

Also, via p2p, there is also no guarantee that at this point in the connection process, we have a copy of their profile. We can always wait on this exchange to finish (which is easy to do with the new profile service), but again, we would have to be dividing behavior between p2p and HTTP & in the hook itself there is no way of knowing which way we are connecting.

Since logsync is the first thing happening in all of these actions, the enforcement inside log is the critical part for the policies to get enforced.

I can think of two ways forward:
- adjust logsync so that "author" includes a Username... even if we move policy enforcement to reside in the logsync pre hooks, we need to know the Username of the subject in order to enforce any policies
- adjust the remote/client push/pull/remove functions to first have a "Hey can I do this" interaction, before proceeding with pushing/pulling/removing. I think this is simpler for now, and will have the least adjustment period especially since we are still debating what identity means.

This means as the first action inside of `PushDataset`, `PullDataset`, `RemoveDataset`, the client would call something like `remote.Permission()`, and pass in the profile, resource, and action. If we error, we stop early.

The remote would handle "Permission" calls, and call `policy.Enforce()` using the send profile, resource, and action, returning `ErrAccessDenied` if the policy doesn't allow that action, and nil if otherwise.

Obviously this isn't the most secure way, and we should still think about how to handle enforcing policy inside logsync and clean up the paths that need `preCheck`s in dsync, but this can get us on track, at least.

b5:
> adjust logsync so that "author" includes a Username... even if we move policy enforcement to reside in the logsync pre hooks, we need to know the Username of the subject in order to enforce any policies

I think we should go with this.

I very much feel the temptation to write an "orchestration" protocol/endpoint that describes the higher-order push/pull process, and attach access control to that. I'm concerned about this for two reasons:
* separating permissions requests diminishes the value of having access control in the first place. access control only works if  it's sitting directly in front of the privileged function
* I'm concerned this is an anti-pattern that will creep into our codebase & never go away, and will cause problems if we don't model it with care

As an example: it's kinda irritating that we don't have a "dataset pushed" event, because logsync and dsync are separate, so you get to subscribe to logs being pushed and version data being pushed, but not both. It's tempting to attach this this event to the end of this theoretical "/permission" endpoint and publish _dataset pushed_ events when that succeeds, but this will break the capacity for clients to make more sophisticated pushes. Say: pushes that send multiple versions, or logs only, or fill historical versions.

Instead I think we should save this for later, and model this work around a higher-order notion of a _session_ that gets created at the lib level. and passed with all lower-level network requests. Sessions might include, say, 1 logsync push, 5 version pushes, and a version pull.

But I digress. I think we should go with the first and only attach permissions to logsync. That'll be enough that when logsync returns an error the push will abort client-side

ramfox:
Dope, got it!

ramfox:
Ended up having to alter the interface of `logsync.remote`, changing the `put` func so that it also passed over a `ref`, like the other `get`, `del` funcs. Otherwise, we would have to read and parse the `io.Reader` sent over, before we can determine if we should reject this request. By adding a `ref` param, we have everything we need to determine if the request has permission to be made before we do anything with the content in the reader.

ramfox:
Okay, last part that I am leaving until I get back on Thursday is adding the policy file option to the `lib.NewInstance`. 

Each other bit is done including tests that prove we can deny requests!

@b5 Would love a mid-pr review, as I am out tomorrow & want some guidance so I can wrap this pr up on Thursday!

go.mod
```diff
@@ -26,6 +26,7 @@ require (
 	github.com/libp2p/go-libp2p-connmgr v0.2.4
 	github.com/libp2p/go-libp2p-core v0.5.7
 	github.com/libp2p/go-libp2p-peerstore v0.2.6
+	github.com/libp2p/go-libp2p-quic-transport v0.8.0 // indirect
```
b5, line 29: side note, this fixes the compilation errors we were seeing when using go 1.15

identity/identity.go
```diff
@@ -13,19 +13,22 @@ import (
 type Author interface {
 	AuthorID() string
 	AuthorPubKey() crypto.PubKey
+	AuthorName() string
```
b5, line 16: should be `Username() string`.  This entire `author` package should get refactored away in the near future, ideally removing the `Author` terminology entirely.

identity/identity.go
```diff
@@ -13,19 +13,22 @@ import (
...
 }
 
 type author struct {
 	id     string
 	pubKey crypto.PubKey
+	name   string
```
b5, line 22: `name` -> `username`

remote/client.go
```diff
@@ -122,8 +124,14 @@ func NewClient(ctx context.Context, node *p2p.QriNode, pub event.Publisher) (c C
 		})
 	}
 
+	pro, err := node.Repo.Profile()
+	if err != nil {
+		log.Debug("cannot get profile from repo, need username for access control on the remote to function")
+	}
```
b5, line 130: this is delightful. I was initially going to ask for a `profile` argument to `NewClient` to preserve a scenario where a user may be offline, but want to use a remote in HTTP-only mode.  But looking at this constructor, almost all of it relies on the node _not_ being nil. So this whole thing would need to be rewritten.

I'm also noticing that I'm asking for you to add arguments to constructors a lot in review lately, and want to make a note of a possible future pattern instead: Having multiple constructors.

If we want this "HTTP only" mode, we could instead define a new constructor:

```go
func NewHTTPClient(r repo.Repo, pub event.Publisher) (*Client, error)
```

I want to point this pattern out so you're aware of it. We have a tendency to emphasize one constructor doing many things. This means sometimes arguments can be nil, and other times not so much, and it's hard to tell when it is and isn't ok to pass nil as args. Having additional constructor functions might help make this less painful

remote/remote.go
```diff
@@ -32,6 +35,9 @@ var log = golog.Logger("remote")
 // hook contexts may be populated with request parameters
 type Hook func(ctx context.Context, pid profile.ID, ref dsref.Ref) error
 
+// OptionsFunc is a type of function that takes an Options
+type OptionsFunc func(o *Options)
```
b5, line 39:
```suggestion
// RemoteOptionsFunc adjusts the behavior of Remote when passed to NewRemote
type RemoteOptionsFunc func(o *Options)
```

No more fight left in me on the spelling of `behavior`. 😄

remote/remote.go
```diff
@@ -104,10 +112,43 @@ type Remote struct {
 	datasetPulled         Hook
 	FeedPreCheck          Hook
 	PreviewPreCheck       Hook
+
+	// policy defines the access control for the remote
+	policy *access.Policy
+}
+
+// OptPolicy adds a policy to the remote options
+// primarily used for testing
```
b5, line 121: meh, I think this is generally useful, I'd drop the line about testing

remote/remote.go
```diff
@@ -235,12 +277,19 @@ func (r *Remote) GoOnline(ctx context.Context) error {
...
-	pid, ref, err := r.pidAndRefFromMeta(params)
+	subj, ref, err := r.subjAndRefFromMeta(params)
 	if err != nil {
 		return err
 	}
 	log.Debugf("remove dataset %s", ref)
 
+	pid := subj.ID
+	if r.policy != nil {
+		if err := r.policy.Enforce(subj, strings.Join([]string{"dataset", ref.Username, ref.Name}, ":"), "remote:remove"); err != nil {
```
b5, line 288: I'd pull the construction of `dataset:username:name` into a variable one line up & pass the var here. There's a very high chance we'll define helper functions in the `access` package for this.

Apply this advice to all hooks

remote/remote.go
```diff
@@ -432,6 +505,41 @@ func (r *Remote) logHook(name string, h Hook) logsync.Hook {
 	}
 }
 
+func (r *Remote) logPreCheckHook(name string, action string, h Hook) logsync.Hook {
```
b5, line 508: gorge

remote/remote_test.go
```diff
@@ -276,6 +278,133 @@ func TestFeeds(t *testing.T) {
...
+			"actions": [
+				"remote:push"
+			]
+		}
+	]
+`
+	allowPushPolicy := &access.Policy{}
+	if err := json.Unmarshal([]byte(allowPushJSON), allowPushPolicy); err != nil {
+		t.Fatalf("error unmarshalling 'allowPushPolicy': %q", err)
+	}
```
b5, line 317: I'm a fan of defining a `mustJSON` helper function for these in tests:

```go
func mustJSON(data string, v interface) {
  if err := json.Unmarshal([]byte(data), v); err != nil {
    panic(err)
  }
}
```

```suggestion
	allowPushPolicy := &access.Policy{}
	mustJSON( `
	[
		{
			"title": "allow subject to push its own datasets",
			"effect": "allow",
			"subject": "*",
			"resources": [
				"dataset:_subject:*"
			],
			"actions": [
				"remote:push"
			]
		}
	]
`, allowPushPolicy)
```

remote/signature.go
```diff
@@ -35,7 +35,10 @@ func sigParams(pk crypto.PrivKey, ref dsref.Ref) (map[string]string, error) {
 		"profileID": ref.ProfileID,
 		"path":      ref.Path,
 
-		"pid":       pid,
+		"pid": pid,
+		// subject is the client node's username, will be used
+		// on the remote side to determine access control
+		"subject":   subject,
```
b5, line 41: I think this should be `username`. The entire map is talking about the `subject`

remote/remote_test.go
ramfox: LOVE IT

remote/signature.go
ramfox: Ahh, yes, however, we are also transmitting the `ref` over, which has the field `username`. Can change to `subject_username` so it is more clear?

remote/remote.go
ramfox: I created a helper function in `access` for this purpose. Used it in `logPreCheckHook` and nowhere else 🤦‍♀️

remote/signature.go
b5: ugh, sorry, PR preview cut the rest of this map off. let's use `requester` whenever we're talking about who is making the request, and pass _requester_ values as the "subject" for access control enforcement

b5:
I can't approve b/c I started this PR. If you agree it's good to go @ramfox, Let's get this merged!

b5:
Please be sure to smoke test first, just as a sanity check

PR #1497, opened 2020-08-25
refactor(api): Remove root handler, introduce /get, disallow /me

dustmop:


dustmop:
Added another commit that allows the "Accept: text/csv" header in order to download just the body bytes. Also put a TODO for some things that need to be fixed involving http headers.

@ramfox this is the change that desktop needs to make to its api usage. "/peer/my_dataset" needs to become "/get/peer/my_dataset". "/body" is unchanged.

PR #1502, opened 2020-09-01
feat(access): API Authentication with JSON Web Tokens

b5:
When finished this should:
* [ ] define interfaces for both creating and storing JWTs
* [ ] make any node running with an API a source of JSON web tokens
* [ ] support authenticated API requests with an `Auth: Bearer [token]` header
* [ ] support a local store of JWTS, keyed by remote location

### DIFF
--- a/access/context.go
+++ b/access/context.go
@@ -1,0 +1,27 @@
+package access
+
+import (
+	"context"
+)
+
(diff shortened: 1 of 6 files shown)


Arqu:
Closed by https://github.com/qri-io/qri/pull/1539

PR #1506, opened 2020-09-03
docs(readme): Update version of go-ipfs used in qri in README badge

machawk1:


b5 (approved):
oh wow nice catch! Thanks @machawk1!

Issue #1504, opened 2020-09-02
API /get should support component=body&format=csv


b5:
right now getting a CSV from the API requires `/body/{username}/{dataset}`, we should make the same thing possible with `/get/{username}/{dataset}?component=body&format=csv`.

we should aim to deprecate & remove the `/body/` endpoint in favor of this approach. It's similar to the  "get vs. export" debate we had a while back. There should be ONE way to "get" stuff: the `/get` endpoint

cc @chriswhong

dustmop:
Big fan of this idea.

b5:
@hbruch brought up an important point on [discord](https://discordapp.com/channels/497052800778502152/497052802275737611/752631582241587270): there are external systems that rely on file extensions in URLS to determine it's a CSV file. His use case is with [kepler.gl](https://kepler.gl), where they state directly in their docs: [URL has to end in .csv](https://github.com/keplergl/kepler.gl/blob/master/docs/user-guides/b-kepler-gl-workflow/a-add-data-to-the-map.md#load-map-using-url)

If our API supported `/get/{user}/{dataset}/body.csv`, and that served CSV data, kepler.gl would be able to use Qri data. From an implementation standpoint, it's just a little extra parsing logic, so I think we should support it in the general form, allowing either `json` and `csv` file extensions: `/get/{user}/{dataset}/body.{format}`

b5:
closed by #1510

Issue #1382, opened 2020-06-05
We should merge qri-io/apiutil into this repo


dustmop:
We should copy the code from `qri-io/apiutil` to `qri-io/qri/api/apiutil`, then deprecate that old repo. The code in apiutil isn't generic enough to justify keeping it as a stand-alone repo, and we keep needing to make changes to it, motivated by feature requests in qri and registry code.

PR #1510, opened 2020-09-09
refactor(api): Combine /get and /body endpoints

dustmop:
First, move the `apiutil` repo into `api/util`, fixing https://github.com/qri-io/qri/issues/1382. There's some changes in this repo that especially want to have some generic error handling, and `api/util` is the right place to put it.

Move the functionality of the /body endpoint into /get. The old /body endpoint still exists, but it's doing most of the same work. Use a single function `parseGetReqArgs` to parse the relevant values from the http.Request. Introduce
a generic error handler for the API, which will send better http status codes if it knows the cause of the specific error it is given.

Users can download a dataset body using one of three methods:

```
localhost:2503/get/peer/my_ds?download=true&format=csv
localhost:2503/get/peer/my_ds/body.csv
-H "Accept: text/csv" localhost:2503/get/peer/my_ds?component=body
```


dustmop:
Fixes https://github.com/qri-io/qri/issues/1504.

Also, read-only mode has been unused and undocumented for a while, this PR removes some of that code from `/get`, such as disallowing `?all=true` from read-only mode.

Aside from that, this *should not* break any existing functionality, but it's also a relatively large change. We should do some manual testing of the API, especially with FSI and desktop, to make sure nothing major was missed.

b5 (approved):
played with it locally, was delighted. This is a major rubber-stamp review, but this week is just so nuts. LGTM

PR #1511, opened 2020-09-09
feat(sql): auto-alias table names when only one dataset is referenced

b5:
this makes simple statements much easier to write, as users don't have to juggle dataset aliasing.

* the first un-aliased dataset will be automatically aliased to `ds`
* subsequent datasets will be `ds2`, `ds3` ... and so on
* BUT! select statements that reference columns without aliases will error

this error is because preprocessing happens before we have any schema information, making it a non-trivial operation to automatically disambiguate column references in select statements.


dustmop (approved):
LGTM

PR #1516, opened 2020-09-10
fix(get): get command auto-fetches

### LINKED ISSUE #1501: qri get for a dataset ref not in repo returns "logbook: not found" or "reference not found"
```
> qri get other_user/data_i_dont_have
logbook: not found
```

for a new user (I think the root cause here is not being cloud connected, unsure):

```
> qri get other_user/data_i_dont_have
reference not found
```

My understanding is that this should instead be resolved by the registry. Don't believe there's any tests for this functionality, so it's easy to end up broken.

b5:
From a [lengthy chat on discord](https://discordapp.com/channels/497052800778502152/571427701475377152/753380113445421148), this PR makes get do the same thing as `sql`, with the caveat that we'll iterate on this interaction in the future

closes #1501

### DIFF
--- a/cmd/get_test.go
+++ b/cmd/get_test.go
@@ -369,7 +369,7 @@ func TestGetDatasetCheckedOutUsingDscache(t *testing.T) {
 }
 
 func TestGetRemoteDataset(t *testing.T) {
-	run := NewTestRunnerWithMockRemoteClient(t, "test_peer_pull_and_list", "pull_and_list")
+	run := NewTestRunnerWithMockRemoteClient(t, "test_get_remote_dataset", "get_remote_dataset")
 	defer run.Delete()
 
 	expect := "cannot use '--offline' and '--remote' flags together"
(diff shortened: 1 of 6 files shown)


cmd/get_test.go
```diff
@@ -367,3 +367,29 @@ func TestGetDatasetCheckedOutUsingDscache(t *testing.T) {
 		t.Errorf("unexpected (-want +got):\n%s", diff)
 	}
 }
+
+func TestGetRemoteDataset(t *testing.T) {
+	run := NewTestRunnerWithMockRemoteClient(t, "test_peer_pull_and_list", "pull_and_list")
```
dustmop, line 372: There's already a test using this testname and peername. Change to something like "test_get_remote_dataset" and "get_remote_dataset".

cmd/get_test.go
```diff
@@ -367,3 +367,29 @@ func TestGetDatasetCheckedOutUsingDscache(t *testing.T) {
...
+		t.Errorf("response mismatch\nwant: %q\n got: %q", expect, err)
+	}
+
+	expect = "reference not found"
+	err = run.ExecCommand("qri get --offline other_peer/their_dataset")
+	if expect != err.Error() {
+		t.Errorf("response mismatch\nwant: %q\n got: %q", expect, err)
+	}
+
+	// TODO (b5) - this fails at the moment with:
```
dustmop, line 387: Make this change to remote/mock_client.go:

```
@@ -196,6 +196,7 @@ func (c *MockClient) createTheirDataset(ctx context.Context, ref *dsref.Ref) err
        other.resolver[ref.Human()] = ref.InitID
        other.dscache[ref.InitID] = path
        ref.ProfileID = other.info.EncodedPeerID
+       ref.Path = path
 
        // Add a save operation to logbook
        err = other.book.WriteVersionSave(ctx, ref.InitID, &ds)
```

dustmop (approved):
LGTM, fix for the unit test within comments

PR #1517, opened 2020-09-10
feat(pull): set pulling source with remote flag on CLI & API

b5:
I very much owe tests for this.

### DIFF
--- a/lib/datasets.go
+++ b/lib/datasets.go
@@ -917,10 +917,10 @@ func (m *DatasetMethods) Remove(p *RemoveParams, res *RemoveResponse) error {
 
 // PullParams encapsulates parameters to the add command
 type PullParams struct {
-	Ref        string
-	LinkDir    string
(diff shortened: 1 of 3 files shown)


lib/datasets.go
```diff
@@ -935,20 +935,24 @@ func (m *DatasetMethods) Pull(p *PullParams, res *dataset.Dataset) error {
 
 	ctx := context.TODO()
 
-	ref, source, err := m.inst.ParseAndResolveRef(ctx, p.Ref, "network")
+	if p.Remote == "" {
+		p.Remote = "network"
```
dustmop, line 939: Is it a problem to modify the input parameters? We don't usually do that. Might be worth creating a local var instead.

dustmop (approved):
LGTM, delta one possible problem

lib/datasets.go
b5: shouldn't be, but you're right it's a bad practice. Will change

PR #1518, opened 2020-09-10
fix(api): Preview endpoint can handle IPFS paths

dustmop:


dustmop:
Apologies, I'm going to self-merge this without review; it's needed to fix the release and folks are not around to review.

PR #1520, opened 2020-09-11
fix(remote): be more lenient with `ResolveRef` error when trying to remove a dataset

ramfox:
When we go to remove a dataset from a remote, we very well may have already removed identifying information that exists in other systems that the `localResolver.ResolveRef` relies on.

However, even if `ResolveRef` comes back with an error, it still may have enough identifying information for the remote to continue removing the dataset.

In this case, we were coming up against the way that the `repo` resolves the dataset reference. It first checks to see if the reference can be found in the repo, and if so, it asks logbook for the `InitID`. However, at this point in the process, we have already removed the log data for this reference, so we error that the log is not found. However, the reference has been populated. Not fully populated, since we need the InitID to be considered "fully populated", but populated enough that we have what we need to continue removing the dataset.


remote/remote.go
```diff
@@ -306,7 +306,17 @@ func (r *Remote) RemoveDataset(ctx context.Context, params map[string]string) er
...
+		// already removed the associated log or some other identifying
+		// information that another system relies on to resolve
+		// the reference. However, the ResolveRef process might have
+		// worked sufficiently to continue trying to fully remove
+		// the dataset.
+		// we include these err exceptions to say that even if the
+		// reference wasn't FULLY resolved, it may have been resolved
+		// enough that we can continue to remove the dataset in other
+		// qri systems
+		if err == dsref.ErrRefNotFound || err == logbook.ErrNotFound {
```
b5, line 319:
```suggestion
		// At this point in the dataset removal process, we may have
		// already removed the associated logbook data or other identifying
		// information that another system relies on to resolve
		// the reference. However, the ResolveRef process might have
		// partially resolved the reference, enough that other subsystems
		// can still perform the delete
		// we nullify these errors to give other subsystems a chance to delete
		if  errors.is(err, dsref.ErrRefNotFound) || errors.is(err, logbook.ErrNotFound) {
		  log.Warnf("couldn't resolve %q before removing the dataset. attempting to remove anyway.", ref)
```

ramfox: THATS SO MUCH BETTER

b5 (approved):
Siiiiiiiiick

uhLeeshUh:
Ahh so we've allowed the delete to continue by catching the `ResolveRef` error and allowing the process to continue. Nice work here & thanks for cluing me in @ramfox!

PR #1529, opened 2020-09-16
docs(api): update OpenAPI docs for dataset handlers

### LINKED ISSUE #1528: docs(api): update OpenAPI for dataset handler functions
Ensure OpenAPI documentation for dataset handlers functions is up to date

uhLeeshUh:
Closes #1528 

Update OpenAPI yaml for our dataset handler functions


api/open_api_3.yaml
```diff
@@ -158,6 +158,25 @@ paths:
         '500':
           $ref: '#/components/responses/StatusInternalServerError'  
   /list:
+    parameters:
```
uhLeeshUh, line 161: Note: I'm planning to refactor the `page` & `pageSize` params into some sort of shared `Paginate` component that we can reference here instead since I imagine we share these in a lot of places -- was just playing around here to learn the OpenAPI spec

api/open_api_3.yaml
```diff
@@ -179,7 +198,7 @@ paths:
         schema:
           type: string
     get:
-      summary: Get the list a specific peer’s datasets
+      summary: Get the list of a specific peer’s datasets
```
uhLeeshUh, line 201: I didn't end up making any changes to the `/list/:peername` endpoint but am getting the `listing datasets on a peer is not implemented` error when I try to hit the endpoint locally. I tracked down to where the error is triggered in code [here](https://github.com/qri-io/qri/blob/master/lib/datasets.go#L128) but am having trouble following the code path to get there -- am I missing where `UseDscache` would be set to true somewhere ?

b5: this is a perfect example of where we've dropped support for a concept along the way, but haven't bothered to update the documentation of it. currently `/list/:peername` doesn't work.

Another note while we're here: we're trying to harmonize around `username` instead of `peername`. It'd be great to switch any reference to `peername` in this doc to `username`

b5: I think the right thing to do here would be to drop the documentation of this endpoint entirely. We can re-add it once it's a thing we can do again

uhLeeshUh: oh great to know, ok will do. Thanks!

api/open_api_3.yaml
```diff
@@ -170,28 +179,27 @@ paths:
...
         '403':
           $ref: '#/components/responses/StatusForbidden'
         '404':
           $ref: '#/components/responses/StatusNotFound'
         '500':
-          $ref: '#/components/responses/StatusInternalServerError'  
-  /save:
-    post:
+          $ref: '#/components/responses/StatusInternalServerError'
+    put:
```
uhLeeshUh, line 202: @b5 another quick question here -- I noticed that we often support multiple HTTP verbs for an endpoint (like POST and PUT here, or DELETE and POST for /remove) but it seems like we tend to only document the former (I'd assume more common?) case. I added in this documentation for PUT /save but should I remove? Trying to figure out how thorough we'd like to be here

chriswhong:
Heads up, this is not a valid openAPI yaml file according to https://apitools.dev/swagger-parser/online/, and `redoc` (docs display component we are going to use on qri.io) doesn't like it either.

chriswhong (approved):
LGTM

PR #1537, opened 2020-09-21
refactor(init): rename flag `source-body-path` to `body`

### LINKED ISSUE #1524: feat(CLI): init CLI command should be able to work with existing files
### What feature or capability would you like?
Should be able to init a new Qri dataset with an existing file in the target directory

Feedback from user Paul Crickard in Sprint R: “I put my csv in a folder I made for the project before seeing that the next command added it. I was wondering if it was possible to run init with files already in and not using the flag like the example. I think I already have files in folders a lot of times when I start a project.”

ramfox:
Also, added examples to the `qri init --help` message

PR adding `body` flag to quickstart cli tutorial forthcoming.

closes #1524

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -18,9 +18,22 @@ func NewInitCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "init [PATH]",
 		Short: "initialize a dataset directory",
-		Long: `'initialize' creates a new dataset, links it to the current or specified
-directory, and creates starter files for the dataset's components.`,
-		Example: ``,
+		Long: `'initialize' creates a new dataset, links it to the current directory,
+and creates starter files for the dataset's components. You can also specify an
+already existing body file using the 'body' flag.`,
+		Example: `  # initialize a new dataset, linking it to the current directory:
+
+  $ mkdir earthquakes && cd earthquakes
+  $ qri init
+  Name of new dataset [earthquakes]: 
+  Format of dataset, csv or json [csv]: csv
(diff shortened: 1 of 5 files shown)


cmd/init.go
```diff
@@ -18,9 +18,22 @@ func NewInitCommand(f Factory, ioStreams ioes.IOStreams) *cobra.Command {
...
+		Long: `'initialize' creates a new dataset, links it to the current directory,
+and creates starter files for the dataset's components. You can also specify an
+already existing body file using the 'body' flag.`,
+		Example: `  # initialize a new dataset, linking it to the current directory:
+
+  $ mkdir earthquakes && cd earthquakes
+  $ qri init
+  Name of new dataset [earthquakes]: 
+  Format of dataset, csv or json [csv]: csv
+  initialized working directory for new dataset ramfox/earthquakes
```
dustmop, line 30: Would suggest using a placeholder username here instead. A few lines down as well.

dustmop (approved):
LGTM, with one suggestion to fix

cmd/init.go
ramfox: do you mean like

`[USERNAME]/earthquakes`

or `my_username/earthquakes`

dustmop: Either, though I guess I have a preference for the second option. I don't think the docs have an explicit "example username" but maybe we should decide on one some point soon.

ramfox: after looking a bit, we use `b5`, `ramfox`, `chriswhong`, in the `qri log` examples. We also use `user` in `qri preview` and `qri remove` examples, so I'm going to go with `user`!

PR #1539, opened 2020-09-22
feat(access): TokenSource & TokenStore interfaces, implementations

Arqu:
Token infrastructure for qri/cloud


access/token_test.go
```diff
@@ -0,0 +1,66 @@
...
+	defer func() { access.Timestamp = prevTs }()
+
+	peerInfo := cfgtest.GetTestPeerInfo(0)
+	tokens, err := access.NewPrivKeyTokenSource(peerInfo.PrivKey)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	pro := &profile.Profile{
+		ID:       profile.IDB58MustDecode(peerInfo.EncodedPeerID),
```
dustmop, line 27: Also here, use `peerInfo.PeerID` instead

access/token.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+type pkTokenSource struct {
+	pk            crypto.PrivKey
+	signingMethod jwt.SigningMethod
+	verifyKey     *rsa.PublicKey
+	signKey       *rsa.PrivateKey
+}
+
+// assert pkTokenSource implements tokens at compile time
```
dustmop, line 66: "implements TokenSource"

dustmop (approved):
This is rad, great tech! Have just a few small fixes to request, but aside from that this seems good to me.

Arqu:
> This is rad, great tech! Have just a few small fixes to request, but aside from that this seems good to me.

Thanks but 99.9% is b5s work, so all credits to him. Updates and merge incoming.

access/token.go
```diff
@@ -0,0 +1,299 @@
...
+
+	// set our claims
+	t.Claims = &TokenClaims{
+		StandardClaims: &jwt.StandardClaims{
+			Subject: pro.ID.String(),
+			// set the expire time
+			// see http://tools.ietf.org/html/draft-ietf-oauth-json-web-token-20#section-4.1.4
+			ExpiresAt: exp,
+		},
+		Username: pro.Peername,
```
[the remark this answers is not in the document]
Arqu: This specific PR is less concerned with the implementation specifics and more on raw mechanics. Also this will evolve a bit as we deepen the integration with token based out between systems. 

With that said, this particular spot does push the notion of a username, however all the integration work done so far only has it included for convenience reasons and is not used as a real reference to anything. A lot more detail on the more real use cases can be seen in [this integration](https://github.com/qri-io/cloud/pull/433). 

As such, I expect this to go away or be modified to suit a more general usecase.

access/token_test.go
Arqu: PeerID an ProfileID are not interchangeable here.

PR #1540, opened 2020-09-22
bug(profile): add check for qri config before running code against it

### LINKED ISSUE #1492: bad error message when running a command without a profile
Before, if a user had no profile on their machine, downloaded `qri` core and tried to run a command, they would get a message about needing to run `qri setup` first. The migration check now overrides this, so the error message is `no config file to check for migrations`, which is confusing for new users.

We should make sure to check first if a profile exists, and if not, recommend running `qri setup`.

Also, setup has real tests now! Thanks to https://github.com/qri-io/qri/commit/6ea4ddb16162b053a1b8ee63b34fe90111a2a8ec. So we can add a test for this to make sure it doesn't regress. The lack of test before is why the breakage went unnoticed.

uhLeeshUh:
closes #1492 

- [x] Add code to confirm a config exists before executing code against it & display useful error to user

- [x] Add test

### DIFF
--- a/cmd/cmd_test.go
+++ b/cmd/cmd_test.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -15,6 +16,7 @@ import (
 	"github.com/qri-io/dataset"
 	"github.com/qri-io/qfs"
 	"github.com/qri-io/qri/config"
+	repotest "github.com/qri-io/qri/repo/test"
 )
 
 func init() {
@@ -660,5 +662,26 @@ func TestBadCaseIsJustWarning(t *testing.T) {
 	err := run.ExecCommand("qri get test_peer_qri_get_bad_case/a_New_Dataset")
 	if err != nil {
 		t.Errorf("expect no error (just a warning), got %q", err)
+	}
+}
+
+func TestSetupHappensBeforeOtherCommands(t *testing.T) {
+	ctx := context.Background()
+
+	qriHome := createTmpQriHome(t)
+	stdinText := "qri_test_name"
+	cmd, shutdown := newCommandWithStdin(ctx, qriHome, stdinText, repotest.NewTestCrypto())
+
+	cmdTextList := [8]string{"qri add", "qri connect", "qri diff", "qri get", "qri init", "qri list", "qri pull", "qri search"}
+	expect := "no qri repo exists\nhave you run 'qri setup'?"
+
+	for _, cmdText := range cmdTextList {
+		noRepoErr := executeCommand(cmd, cmdText)
+		if noRepoErr.Error() != expect {
+			timedShutdown(fmt.Sprintf("ExecCommand: %q\n", cmdText), shutdown)
+			err := fmt.Sprintf("expected error for command %v:\n %v\n but received error:\n %v", cmdText, expect, noRepoErr)
+			t.Fatal(err)
+		}
+
 	}
 }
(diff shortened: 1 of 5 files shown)


cmd/qri.go
```diff
@@ -128,6 +129,11 @@ func (o *QriOptions) Init() (err error) {
 	}
 	setNoPrompt(o.NoPrompt)
 
+	repoErr := lib.QriRepoExists(o.repoPath)
```
uhLeeshUh, line 132: I added this check in the `Init` function here because it seems like all our Qri CLI commands either call this directly in their respective `Complete` function (like [here](https://github.com/qri-io/qri/pull/1540/files#diff-8ff128e26e876899f9196f8b77b3e358R73)) or many call it indirectly via other methods like `DatasetMethods` [here](https://github.com/qri-io/qri/blob/master/cmd/init.go#L72).

Is it a fair assumption that `Init` will be called before every command is run, though? I assume yes since initializing instance state seems like a pre-req to running any command but want to be sure. Is there also any function that we always run before each cmd call? I'd like to put something like this there instead of assuming `Init` will be called in one way or another (and skip the logic if the command is `setup`) but I couldn't track one down other than something like `NewQriCommand` which feels too soon to check for a config since the user wouldn't have a chance to run `qri setup` yet

dustmop: This is exactly the right spot. I don't think it's explicitly stated, but it's certainly the intention that `Init` will setup state before any command is run, except for `setup`. Could you modify the comment describing `Init` to state as much, and then switch this from Draft to Ready?

cmd/cmd_test.go
```diff
@@ -662,3 +664,24 @@ func TestBadCaseIsJustWarning(t *testing.T) {
 		t.Errorf("expect no error (just a warning), got %q", err)
 	}
 }
+
+func TestSetupHappensBeforeOtherCommands(t *testing.T) {
```
uhLeeshUh, line 668: Hey @dustmop do you agree with this location as the right spot for this test? I read the issue as alluding to us including this as a part of `setup_test.go` but that felt a little off to me because we're testing this for any command that's not `qri setup`. I also didn't love the idea of popping a similar test in each of the `_test.go` files for the respective qri commands since that was super duplicative, so I settled on this file which seemed to encompass tests which span different commands? Let me know if I'm wrong about its intent, though

cmd/cmd_test.go
```diff
@@ -662,3 +664,24 @@ func TestBadCaseIsJustWarning(t *testing.T) {
...
+	ctx := context.Background()
+
+	qriHome := createTmpQriHome(t)
+	stdinText := "qri_test_name"
+	cmd, shutdown := newCommandWithStdin(ctx, qriHome, stdinText, repotest.NewTestCrypto())
+
+	cmdTextList := [8]string{"qri add", "qri connect", "qri diff", "qri get", "qri init", "qri list", "qri pull", "qri search"}
```
uhLeeshUh, line 675: Def welcome some feedback here if there's a better way of handling this -- I ended up listing the commands that users would likely run before realizing they need to set up their profile

cmd/cmd_test.go
dustmop: I think this is a good spot for it. My initial instance was also perhaps `setup_test.go`, but you're right that this is testing non-setup commands, so this generic file is a better location.

cmd/cmd_test.go
dustmop: This works well enough! I don't think we need to exhaustively list everything, especially since the command list will change, and the important thing is to test that the error happens at all, to catch any future change that might break that behavior (more migrations works?).

Two things though:
1) No need to pass anything to "standard in", none of these commands read it. Use `newCommand` instead of `newCommandWithStdin` and get rid of `stdinText`.
2) `qri add` is an alias of `qri pull`, but we're favoring `pull` and want to eventually get rid of `add`. Drop `add` and keep `pull`.

uhLeeshUh: Ah gotcha, ok great will change now

uhLeeshUh:
@dustmop thanks for the feedback! Should be ready for a re-review when you have a moment

Issue #1532, opened 2020-09-17
feat: get inferred structure of a body without creating a dataset


b5:
@aborruso brought up a great idea in [chat on discord](https://discordapp.com/channels/497052800778502152/497052802275737611/756148792209703033):

> A lot of a time, working with data in csv format, it's useful to write some doc or same email in which it's necessary to add a schema.
Because CSV does not have it by default.
> For this kind of jobs, I use one of the tools of my box. For this kind of job I use goodtables or xsv. To have it in qri, it would be great, because it would become more and more a data swiss knife, to do great and proper things as create properly described dataset, but also a tool to do in right way small things with a CSV

I ❤️ whenever someone sees qri ad a "data swiss [army] knife", and think this is a great example. We should definitely have a way of getting an inferred schema without creating a version. In chat I'd proposed adding this functionality to `get` by adding the save flags to get to "hot build" a dataset:

```
$ qri get schema --body file.csv
```

dustmop:
What's interesting is that `get` already has a need for this "hot build" functionality. If a dataset is checked out in a working directory, and has a `body.csv` file but no `structure.json`, we infer a schema in order to read the body file. This is currently handled by `fsi.GetBody` in `fsi/body.go`, and there's a long-standing TODO in `lib.Get` about needing to handle this special case, and how the code path could be simplified. If we generalize this functionality so that `get` can infer a schema in a modular fashion, both problems will get solved.

The biggest question for me is how exactly to handle the UI, and whether there's use cases that would be covered aside from getting the structure / schema. Right now `get` does not have a `--body` flag, and this would be the first time adopting it from `save`. What would this change imply? Can I override other components or just the body? Does this handle the use case of wanting to test a different schema against the same body to catch validation errors? Would like to sort that out before making a final decision.

b5:
Agreed. I think the next step here is to use our new spec site to define the exact role of "get", and use that as a place to examine how this fits with existing tech.

Issue #1371, opened 2020-05-28
Consider moving qri-io/dataset/detect to mainline repo


dustmop:
The code in [https://github.com/qri-io/dataset/tree/master/detect](https://github.com/qri-io/dataset/detect) is used to detect schemas for datasets that don't have them, including types and column names, amongst some other things. We've recently starting doing these sorts of "intelligent inference" tasks more over in the mainline qri repo (for example: [dsref/generate](https://github.com/qri-io/qri/blob/master/dsref/generate.go)), and personally I feel like that's a good thing: perhaps the dataset repo should be less fancy, more pure data representation, and leave opinionated stances to other packages.

Moving detect into qri would allow us to use dsref/generated instead of varName, which would enable us to fix https://github.com/qri-io/desktop/issues/508. It would also make it easier to fix https://github.com/qri-io/qri/issues/1322 and other such problems.

dustmop:
Making this change would also make it easier to implement this suggestion: https://github.com/qri-io/qri/issues/1536.

PR #1541, opened 2020-09-23
fix(`Validate`): allow validation of FSI dataset w/ no history

### LINKED ISSUE #1535: bug: running validate in FSI-linked directory requires history
### What is your OS and version?
Mac OS X

### What version of qri are you using (`qri version`)?
v0.9.13-dev

### What did you do?
tried to validate before saving:

```
$ mkdir validate_test
$ cd validate_test
$ qri init
$ qri validate
for linked dataset [b5/validate_example]

repo: no history
```

### What happened?

got `repo:no history`

### What did you expect to happen?

I expected qri to validate based on the data in the working directory

ramfox:
- Pull out logic from `InferValues` and create `InferStructure` function that infers a structure from a body. Adjust tests to test `InferStructure` directly.
- Adjust `lib.Validate` to use `ParseAndResolveRefWithWorkingDir` & `fsi.FilesystemPathToLocal` (to catch any fsiPath), rather then `CanonicalizeDatasetRef`
- Add `TestDatasetRequestValidateFSI` to catch errors when trying to validate a fsi dataset w/ no history

closes #1535

### DIFF
--- a/base/dataset_prepare.go
+++ b/base/dataset_prepare.go
@@ -145,46 +145,66 @@ func InferValues(pro *profile.Profile, ds *dataset.Dataset) error {
 	// if we don't have a structure or schema then attempt to determine one
 	body := ds.BodyFile()
 	if body != nil && (ds.Structure == nil || ds.Structure.Schema == nil) {
-		// use a TeeReader that writes to a buffer to preserve data
-		buf := &bytes.Buffer{}
-		tr := io.TeeReader(body, buf)
-		var df dataset.DataFormat
-
-		df, err := detect.ExtensionDataFormat(body.FileName())
-		if err != nil {
-			log.Debug(err.Error())
-			err = fmt.Errorf("invalid data format: %s", err.Error())
+		if err := InferStructure(ds); err != nil {
 			return err
 		}
-
-		guessedStructure, _, err := detect.FromReader(df, tr)
-		if err != nil {
-			log.Debug(err.Error())
-			err = fmt.Errorf("determining dataset structure: %s", err.Error())
-			return err
-		}
-
-		// attach the structure, schema, and formatConfig, as appropriate
-		if ds.Structure == nil {
-			ds.Structure = guessedStructure
-		}
-		if ds.Structure.Schema == nil {
-			ds.Structure.Schema = guessedStructure.Schema
-		}
-		if ds.Structure.FormatConfig == nil {
-			ds.Structure.FormatConfig = guessedStructure.FormatConfig
-		}
-
-		// glue whatever we just read back onto the reader
-		// TODO (b5)- this may ruin readers that transparently depend on a read-closer
-		// we should consider a method on qfs.File that allows this non-destructive read pattern
-		ds.SetBodyFile(qfs.NewMemfileReader(body.FileName(), io.MultiReader(buf, body)))
 	}
 
 	if ds.Transform != nil && ds.Transform.ScriptFile() == nil && ds.Transform.IsEmpty() {
 		ds.Transform = nil
 	}
 
+	return nil
+}
+
+// InferStructure infers the Structure field of the dataset, guarenteeing
+// that the structure will contain a Format, FormatConfig, and Schema, based
+// on the given dataset body. It will not write over any Structure, Format,
+// FormatConfig, or Schema that already exists.
+func InferStructure(ds *dataset.Dataset) error {
+	if ds == nil {
+		return fmt.Errorf("empty dataset")
+	}
+
+	body := ds.BodyFile()
+	if body == nil {
+		return fmt.Errorf("empty body")
+	}
+	// use a TeeReader that writes to a buffer to preserve data
+	buf := &bytes.Buffer{}
+	tr := io.TeeReader(body, buf)
+	var df dataset.DataFormat
+
+	df, err := detect.ExtensionDataFormat(body.FileName())
+	if err != nil {
+		log.Debug(err.Error())
+		return fmt.Errorf("invalid data format: %s", err.Error())
+	}
+
+	guessedStructure, _, err := detect.FromReader(df, tr)
+	if err != nil {
+		log.Debug(err.Error())
+		return fmt.Errorf("determining dataset structure: %s", err.Error())
+	}
+
+	// attach the structure, schema, and formatConfig, as appropriate
+	if ds.Structure == nil {
+		ds.Structure = guessedStructure
+	}
+	if ds.Structure.Schema == nil {
+		ds.Structure.Schema = guessedStructure.Schema
+	}
+	if ds.Structure.FormatConfig == nil {
+		ds.Structure.FormatConfig = guessedStructure.FormatConfig
+	}
+	if ds.Structure.Format == "" {
+		ds.Structure.Format = guessedStructure.Format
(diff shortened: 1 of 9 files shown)


lib/datasets.go
```diff
@@ -1009,16 +1009,24 @@ func (m *DatasetMethods) Validate(p *ValidateParams, res *ValidateResponse) erro
...
-	if err != nil && err != repo.ErrEmptyRef {
-		if err == repo.ErrNotFound {
-			return fmt.Errorf("cannot find dataset: %s", ref)
+	fsiPath := ""
+	var err error
+	ref := dsref.Ref{}
+
+	// if there is both a bodyfilename and a schema/structure
+	// we don't need to resolve any references
+	if !(p.BodyFilename != "" && schemaFlagType != "") {
```
ramfox, line 1018: The `Validate` code path, with its variety of options, is pretty confusing. Tests were very helpful here, and I'm not sure this is the right time, but a streamlined refactor that does a better job of illustrating the flow and minimizing the 'gotchya's' would be helpful here.

lib/datasets.go
```diff
@@ -1009,16 +1009,24 @@ func (m *DatasetMethods) Validate(p *ValidateParams, res *ValidateResponse) erro
...
+		// TODO (ramfox): we need consts in `dsref` for "local", "network", "p2p"
```
ramfox, line 1019: Again, not sure this is the time, but keep seeing "local" "network" and "p2p" in a bunch of places. Might be nice to have some codified consts that we can rely on.

lib/datasets.go
```diff
@@ -1009,16 +1009,24 @@ func (m *DatasetMethods) Validate(p *ValidateParams, res *ValidateResponse) erro
...
+		ref, _, err = m.inst.ParseAndResolveRefWithWorkingDir(ctx, p.Ref, "local")
+		if err != nil {
+			return err
+		}
+
+		// TODO (b5) - replace this prefix check with a call to qfs.PathKind when it
+		// supports the fsi prefix
```
ramfox, line 1026: Would `qfs` need anything more than just adding `fsi` to the `PathKind` func? Would we have to guarantee any other `fsi` functionality?

base/dataset_prepare.go
```diff
@@ -145,46 +145,66 @@ func InferValues(pro *profile.Profile, ds *dataset.Dataset) error {
...
 
-		// glue whatever we just read back onto the reader
-		// TODO (b5)- this may ruin readers that transparently depend on a read-closer
-		// we should consider a method on qfs.File that allows this non-destructive read pattern
-		ds.SetBodyFile(qfs.NewMemfileReader(body.FileName(), io.MultiReader(buf, body)))
+// InferStructure infers the Structure field of the dataset, guarenteeing
+// that the structure will contain a Format, FormatConfig, and Schema, based
+// on the given dataset body. It will not write over any Structure, Format,
+// FormatConfig, or Schema that already exists.
+func InferStructure(ds *dataset.Dataset) error {
```
dustmop, line 164: This is awesome! A super helpful function to have, I bet this can also be used to solve https://github.com/qri-io/qri/issues/1532 pretty easily.

Looking at https://github.com/qri-io/qri/blob/master/fsi/body.go, it seems fsi does some extra tricks, by using `Seek`, which is can do as it knows it's working with local files. If we use `InterStructure` to also handle "get body" for fsi, we can either drop that extra efficiency gain, or add some dynamic interface checking to this function.

base/dataset_prepare.go
```diff
@@ -145,46 +145,66 @@ func InferValues(pro *profile.Profile, ds *dataset.Dataset) error {
...
+	tr := io.TeeReader(body, buf)
+	var df dataset.DataFormat
+
+	df, err := detect.ExtensionDataFormat(body.FileName())
+	if err != nil {
+		log.Debug(err.Error())
+		return fmt.Errorf("invalid data format: %s", err.Error())
+	}
+
+	guessedStructure, _, err := detect.FromReader(df, tr)
```
dustmop, line 184: I realize this is pretty much a pure extraction refactor, but just wanted to leave a note for myself. Our usage of `detect` is getting more and more coupled to qri core, and there's evidence (such as ignoring the middle return value) that more changes should be made to it to bring it more in line with how qri specifically uses it. So, I'd like to bring up https://github.com/qri-io/qri/issues/1371 again, which proposes that `detect` move out of the `dataset` repo and into `base/detect` so that it's easier to work with.

base/dataset_prepare.go
```diff
@@ -145,46 +145,66 @@ func InferValues(pro *profile.Profile, ds *dataset.Dataset) error {
...
-			ds.Structure.FormatConfig = guessedStructure.FormatConfig
-		}
+	return nil
+}
 
-		// glue whatever we just read back onto the reader
-		// TODO (b5)- this may ruin readers that transparently depend on a read-closer
-		// we should consider a method on qfs.File that allows this non-destructive read pattern
-		ds.SetBodyFile(qfs.NewMemfileReader(body.FileName(), io.MultiReader(buf, body)))
+// InferStructure infers the Structure field of the dataset, guarenteeing
```
dustmop, line 160: guarenteeing -> guaranteeing

lib/datasets.go
dustmop: I think this would be great to have, I would even go further and say that the resolution source should be it's own type, which is either some known instance like `LocalSource` or `P2PSource`, or something wrapped custom address like `RemoveSource("http://my-remote-url/").

lib/datasets.go
dustmop: So many negatives! For this condition, you can apply demorgan's law to make it read a little easier, and update the comment at the same time:

```
// If either bodyfile or schema is not provided, need to resolve the reference
if p.BodyFilename == "" || schemaFlagType == "" {
```

lib/datasets.go
dustmop: Would love to have this as a common utility function, like `fsi.IsFSIPath` so that we don't expose the hard-coded string constant everywhere. I don't know the specific plan for `qfs`, but having a utility function for now would make an eventual switch easier.

lib/datasets_test.go
```diff
@@ -1116,6 +1116,50 @@ Pirates of the Caribbean: At World's End ,foo
...
+	inst := NewInstanceFromConfigAndNode(ctx, config.DefaultConfigForTesting(), node)
+
+	// we need some fsi stuff to fully test remove
+	methods := NewFSIMethods(inst)
+	// create datasets working directory
+	datasetsDir, err := ioutil.TempDir("", "QriTestValidateFSIDataset")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(datasetsDir)
```
dustmop, line 1141: Check out TestRenameNoHistory in this file; you can replace all of this code above by creating a testRunner:

```
	tr := newTestRunner(t)
	defer tr.Delete()
	datasetsDir := tr.CreateAndChdirToWorkDir("QriTestValidateFSIDataset")
        methods := NewFSIMethods(tr.Instance)
```

dustmop (approved):
LGTM! This is great! Have just a small comment or two on fixes, and a bunch of TODOs mostly for future reference

PR #1553, opened 2020-10-05
fix(connect): only run `DoSetup` if there is existing repo

ramfox:
If a user runs `qri connect --setup`, but already has a repo, proceed
forward with spinning up the qri daemon. Only run `DoSetup` if a
repository cannot be found.

### DIFF
--- a/cmd/connect.go
+++ b/cmd/connect.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/qri-io/ioes"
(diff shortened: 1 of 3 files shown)


cmd/connect.go
```diff
@@ -55,7 +55,17 @@ type ConnectOptions struct {
...
+	repoPath := f.RepoPath()
+
+	repoErr := lib.QriRepoExists(repoPath)
+
+	// If we have a repo, we should not acknowledge the `setup` flag. This is
+	// specifically important for any automated running of the qri backend that
+	// cannot know before hand if the user has a repository setup or not.
+	// The `--setup` flag, in this case, is intended as a catch-all: if the user
+	// has not set up a repository yet, run the setup before running `qri connect`,
+	// otherwise, continue to spin up the daemon.
```
ramfox, line 67: Went back and forth with the content of this comment. Would love feedback.

ramfox, line 67: Went back and forth with the content of this comment. Would love feedback.

b5: First, don't think we need the comment. Instead let's focus on making the logic clear.
I think this should be inverted, where we check for repo existence only if `o.Setup` is true, and the solution should be ever-so-slightly more thorough:

```go
if o.Setup {
  repoErr := lib.QriRepoExists(f.RepoPath())
  if errors.Is(repoErr, lib.ErrNoRepo) {
     // make a repo...
  } else if repoErr != nil {
    return err
  }
```

this'll require modifying `lib/setup.go`:
```go
var ErrNoRepo = fmt.Errorf("no repo exists")

// QriRepoExists returns nil if a qri repo is defined at the given path
// does not attempt to open the repo
func QriRepoExists(path string) error {
	// for now this just checks for an existing config file
	_, err := os.Stat(filepath.Join(path, "config.yaml"))
	if !os.IsNotExist(err) {
		return nil
	}
	return ErrNoRepo
}
```

note the return value is no longer `repo.ErrNoRepo`. This might break a test or two, but really shouldn't. We're declaring the `ErrNoRepo` in lib to avoid having to import repo within `cmd`.

b5 (approved):
LGTM

PR #1554, opened 2020-10-06
fix(watchfs): `WatchAllFSIPaths` needs to call `watchPaths`

ramfox:
also adds test confirming the fix

### DIFF
--- a/watchfs/watchfs_test.go
+++ b/watchfs/watchfs_test.go
@@ -10,6 +10,8 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/qri-io/qri/event"
+	reporef "github.com/qri-io/qri/repo/ref"
+	repotest "github.com/qri-io/qri/repo/test"
 )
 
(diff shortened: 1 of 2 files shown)


watchfs/watchfs_test.go
```diff
@@ -10,6 +10,8 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/qri-io/qri/event"
+	reporef "github.com/qri-io/qri/repo/ref"
+	repotest "github.com/qri-io/qri/repo/test"
```
ramfox, line 14: a bit wary of how much this relies on repo, but until dscache can reliable give us a list of references, this is what we got!

Arqu (approved):
Looks good, but TBH don't really understand what's under the hood for FSI so would rather yield to someone else. Giving the rubber stamp anyways as everything looks good/makes sense and tests cover it.

b5 (approved):
looks good for now. This could potentially be watching a _lot_ of files, but that's a separate issue

PR #1556, opened 2020-10-08
test(windows): Fix lib/ unit tests for Windows platform

dustmop:
Some FSI tests check explicit paths. For now, always use POSIX style forward slashes, instead of mixing them with Windows slashes. Increase the timeout for event bus publishing.

Treat "event bus is closed" event as debug, instead of error

### DIFF
--- a/lib/fsi.go
+++ b/lib/fsi.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


dustmop:
TODO: Setup CI to test Windows. We'll have to disable `cmd/` to do so, as they run too slowly to pass on circleCI.

b5 (approved):
looks great!

PR #1559, opened 2020-10-12
fix(sql): Truncate unexpected columns during join

dustmop:


dustmop:
This is a duplicate PR of https://github.com/qri-io/qri/pull/1543, I added a commit with a TODO saying we want unit tests for the sql command. Want to get this merged before we do the next release.

PR #1562, opened 2020-10-15
fix(fill): Map keys are case-insensitive, handle maps recursively

dustmop:
Ran into this problem while working on logbook. I found that parsing config.yaml using yaml.Unmarshal, then trying to call GetPathValue("profile.privkey") didn't work. The reason is that map used to require case-exact string matching. This fixes that, along with a TODO about maps not being traversed the same way other collections are.


b5 (approved):
Looks great!

PR #1564, opened 2020-10-19
fix(list): List datasets even if some refs have bad profileIDs

dustmop:
This works around a problem where users with invalid profileIDs in their repository were unable to list refs at all. The functionality put in place here is only temporary, and will go away once we get rid of CanonicalizeProfile and switch to dscache. For that reason, it doesn't change the interface to lib.List, and instead returns a well-known error for this situation, which cmd uses to output a warning.


b5 (approved):
LGTM!

PR #1570, opened 2020-10-29
refactor(api): better health check handling for API

b5:
This change makes qri play nicer with liveness probes. Having the default (empty) endpoint respond with 200 makes default liveness check settings happy. Also drop logging of the liveness checks. They pollute logs like crazy.

Fix "versionzz" key in response -> "version".


b5 (commented):
In a rush. Merging

PR #1572, opened 2020-11-03
chore(p2p): update bootstrap addresses

Arqu:
This is to be merged AFTER the cloud migration. All new clients will default to the new bootstrap addresses. Old ones will continue using the current ones until we have the next config migration where we aim to switch to DNS backed addresses.


b5 (approved):
Looks great! Thanks!

PR #1573, opened 2020-11-03
fix(api): health & root endpoints use middleware, which handles OPTIONS

b5:
reviewing a frontend PR caught a regresssion from a recent change that dropped
logging from the `/health` and `/` endpoints: it also dropped CORS headers set
by middlware on each of those routes. Add it back with refactoring, and do a
bunch of cleanup to make routes work a little more consistently.

Now OPTIONS requests are handled entirely by middleware, which means middleware
MUST be added to all routes.


dustmop (approved):
LGTM!

PR #1574, opened 2020-11-06
fix(init): TargetDir for init, can be absolute, is created if needed

dustmop:
Prior to this change, lib.InitDataset would take two parameters, Dir which would need to exist already, and Mkdir which would be created but could only be a single directory name instead of a relative path. Fix this problem, and also add multiple tests, and cleanup some TODOs.


dustmop:
Also changes the allowed parameters for the api's init method. It looks like Desktop is not using this currently, so the old form values are being dropped.

b5:
cc @ramfox for the Desktop API issue

b5 (approved):
Looks good!

PR #1576, opened 2020-11-06
chore(dockerfile): use two-image build

b5:
fix an issue in makefiles that's super prominent when make is used in conjunction with dockerfiles, drop size of production docker images from ~586MB to ~75MB

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,29 +1,70 @@
-FROM golang:1.14.6
+# build image to construct the binary
+FROM golang:1.14.6-alpine AS builder
 LABEL maintainer="sparkle_pony_2000@qri.io"
 
-ADD . /go/src/github.com/qri-io/qri
+RUN apk update && apk upgrade && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,32 +1,54 @@
...
+    && apk upgrade \
+    && apk add --no-cache \
+       bash git make openssh
+
+# build environment variables:
+#   * enable go modules
+#   * use goproxy
+#   * disable cgo for our builds
+#   * ensure target os is linux
+ENV GO111MODULE=on \
```
dustmop, line 15: Pretty sure we can stop setting GO111MODULE, I think that option is on by default starting with go 1.12. Maybe also GOPROXY, but not sure.

dustmop (approved):
LGTM

Dockerfile
b5: good point. I'd like to do a follow-up that works out our new min/max golang version support

PR #1579, opened 2020-11-13
feat(dsfs): overhaul save path to use bounded memory, calculate stats component at save time

b5:
This PR is, _ahem_, large.

### Switch to  Rabin Chunking


lib/lib.go
```diff
@@ -52,6 +51,14 @@ var (
 	log = golog.Logger("lib")
 )
 
+// func init() {
```
Arqu, line 54: dead code?

lib/integration_test.go
```diff
@@ -21,7 +21,7 @@ import (
 
 func TestTwoActorRegistryIntegration(t *testing.T) {
 	tr := NewNetworkIntegrationTestRunner(t, "integration_two_actor_registry")
-	defer tr.Cleanup()
+	// defer tr.Cleanup()
```
Arqu, line 24: dead

cmd/save_test.go
```diff
@@ -320,7 +331,12 @@ func TestSaveBasicCommands(t *testing.T) {
 				return
 			}
 			actual := parseDatasetRefFromOutput(run.GetCommandOutput())
-			if diff := cmp.Diff(c.expect, actual); diff != "" {
+			expect := dstest.Template(t, c.expect, tmplData)
+			// vi := run.LookupVersionInfo(t, "test_peer_save_basic/ten_movies")
+			// ds := run.MustLoadDataset(t, vi.Path)
```
Arqu, line 336: dead

base/remove.go
```diff
@@ -103,7 +104,11 @@ func RemoveNVersionsFromStore(ctx context.Context, r repo.Repo, curr dsref.Ref,
 		// blank PreviousPath is found.
 		i--
 		// unpin dataset, ignoring "not pinned" errors
-		if err = UnpinDataset(ctx, r, dest.Path); err != nil && !strings.Contains(err.Error(), "not pinned") {
+		// if err = UnpinDataset(ctx, r, dest.Path); err != nil && !strings.Contains(err.Error(), "not pinned") {
```
Arqu, line 107: dead code as this is now handeled deeper

base/preview.go
```diff
@@ -23,19 +23,24 @@ const (
 )
 
 // CreatePreview generates a preview for a dataset version
-func CreatePreview(ctx context.Context, r repo.Repo, ref dsref.Ref) (ds *dataset.Dataset, err error) {
+// TODO (b5) - previews currently don't include the new stats component, because
```
Arqu, line 26: 👍

base/dsfs/doc.go
```diff
@@ -6,3 +6,8 @@ import (
 )
 
 var log = logger.Logger("dsfs")
+
+func init() {
+	// logger.SetLogLevel("dsfs", "debug")
```
Arqu, line 11: dead

base/dsfs/dataset_test.go
```diff
@@ -999,3 +1047,41 @@ func BenchmarkValidateJSON(b *testing.B) {
 		})
 	}
 }
+
+// func BenchmarkPrepareDataset1000000Rows(b *testing.B) {
+// 	ctx := context.Background()
+// 	_, ds := GenerateDataset(b, 1000000, "csv")
+// 	fs := qfs.NewMemFS()
```
Arqu, line 1054: I guess this will be uncommented before the merge?

Arqu (approved):
Well you definitely won the yak shaving contest for Q4 with this one :)
I mostly reviewed from a pretty high level, but I'm pretty fine with it as it mostly relies on refactoring the codebase to leverage the guts that moved down to `dataset` and `qfs`.

I tagged a few nit picks along the way (by no means exhausitve) but at this PR size, feel free to ignore those.

base/dsfs/dataset_test.go
```diff
@@ -384,8 +431,8 @@ func TestWriteDataset(t *testing.T) {
 		}
 
 		if ref.Transform != nil {
-			if !ref.Transform.IsEmpty() {
-				t.Errorf("expected stored dataset.Transform to be a reference")
+			if ref.Transform.IsEmpty() {
+				t.Errorf("expected stored dataset.Transform to populated")
```
dustmop, line 435: to populated -> to be populated

dustmop (commented):
Got up to and including base/dsfs/dataset_test.go, have to review base/dsfs/doc.go forward still.

api/datasets_test.go
```diff
@@ -461,21 +465,18 @@ func TestDatasetGet(t *testing.T) {
...
 	assertStatusCode(t, "get dataset", actualStatusCode, 200)
-	if diff := cmp.Diff(expectBody, actualBody); diff != "" {
+	expect := dstest.LoadGoldenFile(t, "testdata/expect/TestGetDataset.test_ds.json")
```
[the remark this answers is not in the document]
b5: >  However, if somehow the api handler were to break and not return the meta.title or structure.format, nothing in the buggy PR would reveal that. Yes, the hash would change, but that's opaque just as abide's binary files were. It would be invisible in the diff.

But... the golden file would change, right? and because it's in plaintext we can actually _review_ the changes to the file, which includes not just hashes, but the entire dataset response.

I can see the upside of an API that lets us fuzz-up golden files. I think another way to do this would be to write functions that remove parts of the dataset we don't want to compare to and write the result of calling those functions to golden files.

> it appears that this test could also benefit by combining boilerplate. Is there ever going to be a case where we want to CompareDatasets but not UpdateGoldenFileIfEnvVarSet?

There are cases where we're using them  separately, but I love the idea of a convenience function. I'm going to put it in detest because I'm tired of copying shit all over the place for tests

base/dsfs/commit.go
```diff
@@ -0,0 +1,349 @@
...
+			if err != nil {
+				log.Error("prev.Transform.ScriptPath %q read err: %s", prev.Transform.ScriptPath, err)
+			}
+		}
+	}
+	if ds.Transform != nil && ds.Transform.ScriptPath != "" {
+		log.Debugf("inlining next transform ScriptPath=%q", ds.Transform.ScriptPath)
+		// TODO(dustmop): The ipfs filestore won't recognize local filepaths, we need to use
+		// local here. Is there some way to have a cafs store that works with both?
+		fs, err := localfs.NewFS(nil)
```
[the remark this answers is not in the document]
b5: Yes! I ended up removing the `localfs.NewFS` call and just using the passed in fs directly... and it cleared up a failing test in the cmd package

base/dsfs/compute_fields.go
```diff
@@ -0,0 +1,349 @@
...
+
+	crypto "github.com/libp2p/go-libp2p-core/crypto"
+	"github.com/qri-io/dataset"
+	"github.com/qri-io/dataset/dsio"
+	"github.com/qri-io/dataset/dsstats"
+	"github.com/qri-io/jsonschema"
+	"github.com/qri-io/qfs"
+)
+
+type computeFieldsFile struct {
```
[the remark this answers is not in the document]
b5: I'm rather proud of this thing, and will try to add a comment describing it. It takes the place of the three goroutines that used to split up the work of calculating fields, but this one is set up to behave as a normal file, but it's "internally teed", generating a dsio.entryReader _as the file is being read & persisted to the filesystem_.

qfs does consume this thing as if it were a regular file. the filesystem interface is calling read on the file. itself, which is writing to the tee'd file, where sidecar calculations like error and byte counts are happening. We used to do this as two passes over the body file in series, now those steps happen in parallel.

We need write hooks to actually take persisted these derived fields in other component files, but I think this is generally a big win for the Filesystem interface, that happens to compose nicely without needing to make explicit additions to the write hook interface

base/dsfs/dataset_test.go
b5: actually, needs to be deleted. `PrepareDataset` no longer exists

PR #1581, opened 2020-11-19
feat(dsync): update dsync to use multi-block streams over HTTP

b5:
updating this dependency brings multi-block streaming for dsync


Arqu (approved):
👍

PR #1583, opened 2020-11-20
fix(api): Fix unmarshal bug in api test

dustmop:
My comments on the previous PR (dsfs overhaul) were misinformed. I thought dstest was doing work to drop components from datasets, or just not comparing them. Turns out it was actually a bug in how the api tests were calling json.Unmarshal

The API response looks like this:
```
{
  "data": {
    "peername": ...
    "name": ...
    "dataset": {
      "peername": ...
      "name": ...
```

By unmarshalling response.Data, the top-level fields for peername, name, etc were honored, but the result would have no components. Instead, we should unmarshal response.Data.Dataset, in order to get everything.

Previously the golden file had only 4 top-level fields. With this fix, now it has everything.


b5 (approved):
nice catch!

PR #1586, opened 2020-11-23
feat(api): change report API

Arqu:


lib/changereport.go
```diff
@@ -0,0 +1,460 @@
...
+	err = inlineScriptsToBytes(leftDs)
+	if err != nil {
+		return err
+	}
+	err = inlineScriptsToBytes(rightDs)
+	if err != nil {
+		return err
+	}
+
+	if leftDs.Commit != nil || rightDs.Commit != nil {
```
dustmop, line 376: Given that all of these fields have the same type, perhaps, it's better to use a map from "component name" -> ChangeReportComponent? There doesn't seem to be a need from the type system to unfold all of these blocks for the sake of type-safety. If the problem is that the `leftDs` and `rightDs` have strongly typed structs, perhaps `base/component/component.go` can help, it provides a type-safe way to generically map over the dataset components. It's also supposed to abstract over datasets being from the repository vs FSI, but there's some bugs in how it does so.

Arqu: Nice one, I'll use the component map and refactor this to be a bit more streamlined! The component compare is handy.

lib/changes.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/qri-io/qri/changes"
+	"github.com/qri-io/qri/dsref"
+)
+
+// ChangeReportParams defines parameters for diffing two sources
+type ChangeReportParams struct {
+	LeftRefstr  string
+	RightRefstr string
+}
```
ramfox, line 14:
```suggestion
type ChangeReportParams struct {
	LeftRefstr  string `json:"left"`
	RightRefstr string `json:"right"`
}
```

adding this would be really helpful for desktop!

Arqu: Totally, this was just a rough mashup of params to get things going

ramfox:
This is actually falling together... way better than expected. Getting a response from the backend and actually loading the `Changes` page!!

The only thing that doesn't flow well is the `Commit`. I'm realizing as I look at the (typescript) interface for that object, that it's actually a `VersionInfo`, not a `Commit`. I think we should adjust the API response model to have `VersionInfo` rather than `Commit`! This should include the `username`, `name`, `path`. If we still want to call it `Commit`... I guess we can. We can also still include a `Commit` in the response & just ignore it on the frontend.

Then the model will basically line up exactly with what we expect!

Exciting!

Arqu:
Ok so this has ways to go, but is now in much better shape. The code is super defensive, but we will unwrap more as we finish up some of the underlying stuff. Would love to start getting a more thorough review on this.
Tests are comming shortly.

changes/changes.go
```diff
@@ -0,0 +1,602 @@
...
+	ds.Stats, statsErr = svc.stats.Stats(ctx, ds)
+	if statsErr != nil {
+		return qerr.New(statsErr, "missing stats components")
+	}
+	return nil
+}
+
+// parseStats uses json serializing > deserializing to easily parse the stats
+// interface as we have little type safety in the dataset.stats component right now
+func (svc *Service) parseStats(ds *dataset.Dataset) ([]EmptyObject, error) {
```
b5, line 115: into it. We should get on this ASAP

b5 (approved):
Looks good to me. There's lots to do here, but the high priorities in my eyes are:

* TESTS
* drop all double-serialization to json by:
  * adding strong typing to the stats component (as you've noted)
  * moving a small api for dynamic component comparison down from the `component` package into the dataset package
* define functions that operate on dynamic components for "meta" generation 

Generally, this whole package feels like it really needs the dynamic expression of datasets that we started to work on in `github.com/qri-io/qri/base/component`. The sooner we get a cleaner API extracted from that package, the better

Arqu:
Excellent :) Highly agree on the above. Will cover tests in this pass, and have the remaining 2 points in upcoming PRs as we work further down the ladder.

changes/changes_test.go
```diff
@@ -0,0 +1,550 @@
...
+	const alteredBodyDataColumns1 = `city,avg_age,in_usa
+toronto,55.0,false
+new york,44.0,false
+chicago,440.4,false
+chatham,650.25,false
+raleigh,5000.65,false`
+
+	ds.Name = "cities"
+
+	updateDataset(t, mr, ds, alteredBodyDataColumns1)
```
Arqu, line 496: Need some help with this, apparently this method doesn't properly update the `ds.Structure` while all else I needed seems fine. Functionally speaking things seems to be correct when I force the right type of problem for it to pass. @dustmop @b5 any hints on how to improve my `updateDataset` method?

changes/changes_test.go
```diff
@@ -0,0 +1,550 @@
...
+		return res.Columns[i].Title < res.Columns[j].Title
+	})
+
+	expect := &StatsChangeComponent{
+		Summary: &ChangeReportDeltaComponent{
+			ChangeReportComponent: ChangeReportComponent{
+				Left:  StatsChangeSummaryFields{Entries: 5, Columns: 4, TotalSize: 157},
+				Right: StatsChangeSummaryFields{Entries: 5, Columns: 4, TotalSize: 157}, // is this right?
+			},
+			Delta: StatsChangeSummaryFields{},
```
Arqu, line 111: This seems to be empty, think we need to populate the delta on the summary too. Leaving this mostly as a note to myself.

changes/changes_test.go
b5: ah I think I understand. Are you expecting structure to change because you changed the body structure? currently we only infer a structure _if one doesn't exist_. So on the first commit a structure gets created, then needs to be explicitly overridden in future commits.

One easy solution: have your `updateDataset` function explicitly set ds.Structure to nil. `base.CreateDataset` will re-infer a new structure for you that'll match

changes/changes.go
```diff
@@ -0,0 +1,612 @@
...
+
+// StatsChangeComponent represents the stats change report
+type StatsChangeComponent struct {
+	Summary *ChangeReportDeltaComponent   `json:"summary"`
+	Columns []*ChangeReportDeltaComponent `json:"columns"`
+}
+
+// ChangeReportResponse is the result of a call to changereport
+type ChangeReportResponse struct {
+	VersionInfo *ChangeReportComponent `json:"version_info,omitempty"`
```
dustmop, line 47: I missed the earlier discussion about VersionInfo last week. Is this really the right interface here? VersionInfo's intention (apologies, it's not very well spelled out in the code) is to return a sparse version of a dataset at a specific commit, such that it contains exactly what the `list` operation needs to show minimal information about a dataset. Everything in a VersionInfo, as it is defined by `dsref/version_info`, should be available in the other components below. Or is it just here for the "username", "name", "path", "profileID"? If so, a `dsref.Ref` would be more appropriate.

Perhaps it really is better to use VersionInfo and I'm missing the larger context. But at first glance it seems redundant and over-specified.

changes/changes.go
```diff
@@ -0,0 +1,612 @@
...
+		} else if leftDs.Commit == nil && rightDs.Commit != nil {
+			res.Commit.Meta["status"] = fsi.STAdd
+		} else if leftDs.Commit != nil && rightDs.Commit != nil {
+			if leftDs.Commit.Path == rightDs.Commit.Path {
+				res.Commit.Meta["status"] = fsi.STUnmodified
+			} else {
+				res.Commit.Meta["status"] = fsi.STChange
+			}
+		} else {
+			res.Commit.Meta["status"] = fsi.STParseError
```
dustmop, line 484: Are we sure this case means some file has failed to parse? Couldn't it also be both left and right have a nil component? Maybe it's better to have another status code in `fsi` such as `fsi.STMissing`.

changes/changes.go
```diff
@@ -0,0 +1,612 @@
...
+	log = golog.Logger("changes")
+	// ErrCompareStats indicates an error with stats comparison
+	ErrCompareStats = errors.New("stats compare error")
+)
+
+// ChangeReportComponent is a generic component used to populate the change report
+type ChangeReportComponent struct {
+	Left  interface{}            `json:"left"`
+	Right interface{}            `json:"right"`
+	Meta  map[string]interface{} `json:"meta,omitempty"`
```
dustmop, line 28: Correct me if I'm wrong, but is this only ever used to hold a `"status"` key to value? Since `"status"` appears so often (always?), maybe it's better to make that top level on this struct.

Also, I would prefer not to have this field named `Meta` since there's a dataset component with that same name, but it means something totally different. Not sure how this is used aside from the already mentioned `"status"` value, but maybe `About` would be better if it's talking about how the component has changed.

changes/changes.go
```diff
@@ -0,0 +1,612 @@
...
+			return nil, qerr.New(err, fmt.Sprintf("failed to calculate stats column delta for %q", columns[i].Title))
+		}
+		columns[i].Delta = deltaCol
+		i++
+	}
+
+	return columns, nil
+}
+
+func (svc *Service) columnStatsDelta(left, right interface{}, lCol, rCol *tabular.Column, hasLeft, hasRight bool) (map[string]interface{}, error) {
```
dustmop, line 272: This function is really dense. Wondering if there's some way to split it up into smaller pieces, or refactor or combine code.

changes/changes.go
dustmop: What is the shape of `Stats`? It looks to be declared as `interface{}`, but is it always actually a list of json like objects? In other words `[]map[string]interface{}`?

If so, it is probably worth adding a method to `base/fill` that operates on a list of structs. I'm thinking `fill.StructList` or something, that just loops over the input and calls `fill.Struct` for each element. Seems like a good fit, this is exactly what `base/fill` was designed to do.

dustmop (commented):
Apologies, didn't get to do an in depth review in regards to the `base/component` question, but gave this another passover with some suggestions and fixes.

changes/changes.go
Arqu: This is less functional and more utility, @ramfox asked for me to include VersionInfo as it's easier to manage internally on the FE side. No real usecase beyond that as the Commit right bellow it contains all the neccessary info.

changes/changes.go
Arqu: True, I just used a blanket error and tried to conform to what exists already due to existing integrations, happy to update.

changes/changes.go
Arqu: Agree on the rename. Also for the first pass, it's true that it only holds the `"status"` but the intention is to use it for more contextual data once we get there. In that vein, I'll keep the object structure but will rename. (Depending on timelines might separate that out in another PR for timing with FE integrations)

changes/changes.go
Arqu: Most of them are :( Will try another pass to make things a bit leaner.

changes/changes.go
Arqu: This is mainly a shortcut to work around type casting issues between `interface{}` and `[]something{}` hence the double serialization. In this specific usecase the top-level is indeed a `[]map[string]interface{}` however it is not *always* the case as far as I understood it from the dataset package especially in the realm of JSON datasets. 

Will look into it if it makes sense doing now or simply working with this untill we introduce the propper typing in `Stats`

changes/changes.go
Arqu: So we had another discussion on this topic, and for now we're going to keep both in favor of fast iterating and err on the side of sending more than we need till we formalize the structures and essentially re-do some of these structures.

ramfox:
@b5 @Arqu
So looks like I was wrong, `commitTitle` and `commitMessage` ARE NOT in the `VersionInfo` definition on backend. The frontend type definition has both... along with a TODO from b5 - "These are not yet supplied by the backend", referencing a number of fields, all of which exist in the backend definition of `VersionInfo` today, except for `commitTitle` and `commitMessage`.

So if they are supposed to exist in the backend definition of a `VersionInfo`, I'm happy to file an issue. If not, I've already made a work around pulling the commit title and message from the change report object anyway.

Arqu:
Don't have an answer on what's the right approach for `VersionInfo` in regards to backend. I'd guess that considering the weight of such a change compared to this usecase, it's not justified and we should work towards a more consolidated `changes` struct which includes the appropriate fields instead of shaping the core pieces to fit this specific feature.

b5 (approved):
Looks great. Instead of reviewing I added two commits to your branch. The first is just a test refactor that adds in the _test runner_ pattern we've been using in a number of other packages to cut down on boilerplate & give us clues for how other package APIs can be refactored.

The second change is to make the exported `Service` an interface with an unexported implementation. This clarifies the API a little and gives package consumers fewer things to tinker with.

This PR does make me realize that we need to change the `loader` interface to accept paths instead of refs. If we did that, the `Report` function could change to this:
```go
Report(ctx context.Context, leftRef, rightRef, loadSource string) (*ChangeReportResponse, error)
```

which means that the `changes` package can truly be lower-level, where it knows nothing about dsrefs at all (cc @dustmop for this last bit)

Arqu:
Yes, there's so much to improve, but glad it's starting to take shape and we sort of have a path forward from here.
Also, appreciate the refactor, much smoother now :)

PR #1587, opened 2020-11-24
perf(dsfs): don't calculate commit descriptions if title and message are set

b5:
diffing can be expensive, especially if we're going to throw away the results. Add a set of byte-level checks before dropping down to diffing to ensure we in fact have something to diff.


base/dsfs/commit.go
```diff
@@ -95,13 +100,89 @@ func commitFileAddFunc(privKey crypto.PrivKey) addWriteFileFunc {
...
+		return nil
+	}
+	if !ds.Readme.ShallowCompare(prev.Readme) {
+		log.Debugf("byte changes exist. readme components are inequal")
+		return nil
+	}
+	if !ds.Transform.ShallowCompare(prev.Transform) {
+		log.Debugf("byte changes exist. transform components are inequal")
+		return nil
+	}
```
Arqu, line 142: What about structure (and should add future stats). I know they are qri calculated and not user supplied. Or maybe just hold that off for the change reports and use that for a medium level check before diving into diffing.

Arqu (approved):
Neat :)

base/dsfs/commit.go
b5: structure and stats both have path values at this point, and are covered below in the `PathMap` comparisons

PR #1588, opened 2020-11-30
Fix fsi stats

b5:
reify the purpose of the `stats` package in qri repo as _service_ that wraps a cache & uses `github.com/qri-io/dataset/dsstats` for stats calculations, re-working the `stats` package interface in the process.

finish work we started a while ago on a local stats cache, expanding it to be aware of keys that match paths on the local filesystem.

The end result is a cache that will calculate stats components, cache the result, and invalidate the cache whenever the body file in an FSI-linked directory changes.


stats/cache.go
```diff
@@ -3,79 +3,262 @@ package stats
...
+func (c *localCache) addAndPurgeExpired(cacheKey string, statProps, targetProps didmod.Props) {
+	c.infoLk.Lock()
+	defer c.infoLk.Unlock()
+	log.Debugw("adding stat props", "cacheKey", cacheKey, "statProps", statProps, "targetProps", targetProps)
+	c.info.StatFileProps[cacheKey] = statProps
+	c.info.TargetFileProps[cacheKey] = targetProps
+
+	var (
+		lowestKey     string
+		lowestModTime int
```
dustmop, line 183: Time.Unix is guaranteed to return an int64. Could we change lowestModTime to be int64, and initialize it to math.MaxInt64 instead.

cmd/search_test.go
```diff
@@ -164,7 +164,7 @@ func TestSearchRun(t *testing.T) {
 	}))
 	rc := regclient.NewClient(&regclient.Config{Location: server.URL})
 
-	f, err := NewTestFactoryInstanceOptions(ctx, lib.OptRegistryClient(rc))
+	f, err := NewTestFactoryInstanceOptions(ctx, run.RootPath, lib.OptRegistryClient(rc))
```
dustmop, line 167: This looks a bit confusing, but works okay. The intention with TestRunner is that the RootPath is *not* the RepoPath, but rather the parent directory that everything temporary lives inside of (and this difference has bitten before, causing subtle bugs in tests). However, TestRunner in general needs some cleanup, documentation, and clarification of intention. Could you add a TODO here to add an explicit RepoPath to the SearchTestRunner?

dustmop (approved):
LGTM, good stuff all around, just a few requests for clarification / comments

stats/cache.go
b5: lovely idea. will incorporate

stats/cache.go
```diff
@@ -3,79 +3,262 @@ package stats
...
 
-var _ Cache = (*nilCache)(nil)
+func (c *localCache) cacheKey(key string) string {
+	return b32Enc.EncodeToString([]byte(key))
+}
 
-// PutJSON places stats in the cache, keyed by path
-func (nilCache) PutJSON(ctx context.Context, path string, r io.Reader) error {
-	return ErrNoCache
+const uintSize = 32 << (^uint(0) >> 32 & 1)
```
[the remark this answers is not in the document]
b5: ended up nixing the need for this with your  suggestion below

PR #1591, opened 2020-11-30
feat(version): add info from git and golang to qri version command

b5:
before:
```
$ qri version
0.9.14-dev
```

after:
```
$ qri version
version:	0.9.14-dev
build date:	2020-11-30T20:32:33Z
git summary:	v0.9.13-56-gb6834436-dirty
git branch:	feat_version_add_git_info
git commit:	b6834436
git state:	dirty
golang version:	go1.15
```

Also adds  `$ qri version --format=json` flag to get this info printed as JSON, which tools like qri desktop's `main` process may be able to make use of

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,6 @@ QRI_VERSION?="0.9.14-dev"
 BUILD_FLAGS?=CGO_ENABLED=0
 PKG=$(shell go list ./version)
 GOLANG_VERSION=$(shell go version | awk '{print $$3}')
-GOVVV_FLAGS=$(shell govvv -flags -pkg $(PKG) -version $(QRI_VERSION))
 
 require-goversion:
 	$(eval minver := go1.13)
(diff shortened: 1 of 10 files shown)


dustmop:
When I try to `make`, I get `/bin/sh: govvv: command not found`, am I missing a step?

Also, that `command not found` error doesn't fail the build, the go build still continues unabated, but it probably *should* fail.

Makefile
```diff
@@ -1,6 +1,13 @@
...
 default: build
 
+QRI_VERSION?="0.9.14-dev"
+
+BUILD_FLAGS?=CGO_ENABLED=0
+PKG=$(shell go list ./version)
+GOLANG_VERSION=$(shell go version | awk '{print $$3}')
+GOVVV_FLAGS=$(shell govvv -flags -pkg $(PKG) -version $(QRI_VERSION))
+
 require-goversion:
```
dustmop, line 11: Would recommend also adding:

```
require-govvv:
	go install github.com/ahmetb/govvv
```

and adding that dependency to `build` and `install`.

version/version.go
```diff
@@ -1,4 +1,47 @@
...
+		"gitBranch":     GitBranch,
+		"gitState":      GitState,
+		"gitSummary":    GitSummary,
+		"buildDate":     BuildDate,
+		"golangVersion": GolangVersion,
+	}
+}
+
+// Summary prints a summary of all build info.
+func Summary() string {
```
dustmop, line 36: Could we add a message here if the binary was built with `go install` by mistake? So, compare "Version" to "n/a", and if equal, return something like "please build using `make install` instead of `go install`".

dustmop (commented):
Figured it out...

Makefile
```diff
@@ -24,11 +23,17 @@ require-goversion:
 		fi; \
 	fi;
 
-build: require-goversion
-	$(BUILD_FLAGS) go build -ldflags="-X ${PKG}.GolangVersion=${GOLANG_VERSION} ${GOVVV_FLAGS}" .
+require-govvv:
+	$(eval govvv_loc := $(shell which govvv))
+	@if [ "$(govvv_loc)" == "" ]; then go get github.com/ahmetb/govvv; fi;
```
dustmop, line 28: `go get` will download the source code, but won't build govvv and put it onto the PATH, it needs `go install`.

b5: very good point, just force-pushed this change

dustmop (approved):
LGTM

PR #1592, opened 2020-12-02
refactor(profiles): move profile package up, merge identity into profiles

b5:
the beginning of a bunch of refactoring work. When this round of refactoring is done the `profile` package should be a service that wraps a new 'keys' package that stores keypair information separate from profile info.

Moving `profile` up out of repo gives us a good starting point for re-working the `profile.Store` interface


profile/author.go
```diff
@@ -0,0 +1,50 @@
...
+		pubKey:   pubKey,
+		username: username,
+	}
+}
+
+func (a author) AuthorID() string {
+	return a.id
+}
+
+func (a author) AuthorPubKeyID() crypto.PubKey {
```
dustmop, line 40: Is this needed? It's not part of the interface.

repo/fs/fs.go
```diff
@@ -89,9 +78,10 @@ func NewRepo(path string, fsys *muxfs.Mux, book *logbook.Book, cache *dscache.Ds
 		return nil, err
 	}
 
+	own := pro.Owner()
 	// add our own profile to the store if it doesn't already exist.
```
dustmop, line 82: I feel like a better approach would be to have the interface guarantee that the owner is always in the profile list. Would probably simplify understanding how to use it. This is fine for now, but it definitely feels like more a work-around than a semantic we actually want to support.

repo/mem_repo.go
```diff
@@ -35,13 +34,23 @@ type MemRepo struct {
 
 var _ Repo = (*MemRepo)(nil)
 
+// NewMemRepoProfile creates a new in-memory repository and an empty profile
+// store owned by the given profile
+func NewMemRepoProfile(ctx context.Context, owner *profile.Profile, fs *muxfs.Mux, bus event.Bus) (*MemRepo, error) {
```
dustmop, line 39: The way this names reads makes it seems like it's returning a "repo profile" (a profile where repo is an adjective describing it), when its really more of a "repo and profile", or "repo with profile". Would prefer something like `NewMemRepoWithProfile` or perhaps `NewOwnedMemRepo` or something like that.

dustmop (approved):
This is cool! Definitely puts things more in line with where we want them to be.

PR #1594, opened 2020-12-07
feat(save): support custom timestamps on commit

b5:
This will let [qri-git](https://github.com/qri-io/qri-git) match timestamps from git on import.

setting the input dataset commit.Timestamp value will now be preserved instead of being overwritten with the current time. Input timestamp will ALWAYS be converted to UTC time zone.

This change makes it much easier to generate dataset histories with commit timestamps that aren't in chronological order, which is perfectly fine according to spec, but will cause bugs for any code that relies on such an assumption.

### DIFF
--- a/base/dsfs/compute_fields.go
+++ b/base/dsfs/compute_fields.go
@@ -168,8 +168,12 @@ func (cff *computeFieldsFile) handleRows(ctx context.Context) {
 	}
 
 	cff.Lock()
-	// assign timestamp early. saving process on large files can take many minutes
-	cff.ds.Commit.Timestamp = Timestamp()
+	if cff.ds.Commit.Timestamp.IsZero() {
+		// assign timestamp early. saving process on large files can take many minutes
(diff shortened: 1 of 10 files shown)


lib/peers.json
```diff
@@ -0,0 +1 @@
+{"QmeL2mdVka1eahKENjehK6tBxkkpk5dNQ1qMcgWi7Hrb4B":{"id":"QmeL2mdVka1eahKENjehK6tBxkkpk5dNQ1qMcgWi7Hrb4B","peername":"default_profile_for_testing","created":"2020-12-07T09:41:29.273167-05:00","updated":"2020-12-07T09:41:29.273167-05:00","type":"peer","email":"","name":"","description":"","homeurl":"","color":"","thumb":"","photo":"","poster":"","twitter":""}}
```
dustmop, line 1: Remove this file, and peers.json.lock as well

dustmop (approved):
LGTM

PR #1597, opened 2020-12-09
fix(save): remove dataset field computing deadlock

b5:
@chriswhong was getting a spinning-wheel on save that was hanging forever on a very small dataset. Turns out it's a concurrency problem.

The dsfs.computeFieldsFile `Read` and `handleRows` methods were both contending for the same mutex lock, which would deadlock the program. Fix this by using a `dsio.TrackedReader` to count bytes instead of local state, which drops the need for taking the lock on calls to `Read`. Should yeild a microscopic performance bump as well.

Another problem that crops out of this state: users understandably use SIGKILL to get out of the hung process, which is leaving the dataset name "taken" and causing errors on re-running save.

If a user starts the save process but abandons it midway with SIGKILL qri's currently structured to terminate without any cleanup. This means the 'rollback' procedure in an errored save never fires, and the dataset name is left taken but not in the user's list of datasets. This (admittedly hacky) solution checks for stray histories in this exact state (name and branch logs declared, no version save) and considers them safe to remove before creating a new replacement log. Fix that too

### DIFF
--- a/base/dsfs/commit.go
+++ b/base/dsfs/commit.go
@@ -90,7 +90,6 @@ func commitFileAddFunc(privKey crypto.PrivKey, pub event.Publisher) addWriteFile
 					log.Debugf("ensureCommitTitleAndMessage: %s", err)
 					return nil, fmt.Errorf("error saving: %w", err)
 				}
-
 			}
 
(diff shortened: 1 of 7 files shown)


chriswhong (approved):
Tested it locally with my problematic dataset, worked like a charm.

PR #1599, opened 2020-12-11
fix(changes): column renames are properly handled now

Arqu:
Change reports now allocate column diffs based on max unique columns detected not by length of the left or right side.


dustmop (approved):
LGTM!

PR #1600, opened 2020-12-15
fix(logbook): Logsync validates that ref has correct profileID

dustmop:
It should always be the case that the data within a logbook matches the reference that logsync says it is pushing. No need to keep this behavior inside of a hook.

This change causes `TestP2PLogsync` to hang. I don't know what is causing this, but could dive into it in the near future.


b5 (approved):
LGTM.

PR #1602, opened 2021-01-14
feat(keystore): keystore implmenetation

Arqu:


key/keybook.go
```diff
@@ -0,0 +1,175 @@
...
+
+type memoryKeyBook struct {
+	sync.RWMutex // same lock. wont happen a ton.
+	pks          map[peer.ID]ic.PubKey
+	sks          map[peer.ID]ic.PrivKey
+}
+
+var _ pstore.KeyBook = (*memoryKeyBook)(nil)
+
+// noop new, but in the future we may want to do some init work.
```
b5, line 22: allocation of the maps is initialization work

key/key.go
```diff
@@ -0,0 +1,23 @@
...
+var log = logger.Logger("key")
+
+// ID identifies a key, usually it's determined by hash-of-publickey
+type ID = peer.ID
+
+// Public is the public half of a cryptographic key, plus an identifier
+type Public struct {
+	ID  ID
+	Key crypto.PubKey
+}
```
b5, line 18: While I'm glad to see a distinct `key.ID` type, I don't think we need this struct. Instead I think we should add a `KeyID` field to `profile.Profile` and require users access the keystore to get keys.

_but_, you can get lots of the value of this struct by adding / keeping `PrivKey` and `PubKey` fields on `profile.Profile`, which will save the overhead of repeated lookups. Given that `ProfileStore` is going to wrap `KeyStore`, we can have the profile store populate whatever key into we need.

We should later add a function to the key package that defines the canonical method for getting a n identifier from a given key, which should lower the need for this struct even more

key/store.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/qri-io/qri/config"
+	"github.com/theckman/go-flock"
+)
+
+// Store is an abstraction over a libp2p.KeyBook
+// In the future we may expand this interface to store symmetric encryption keys
+type Store interface {
+	peerstore.KeyBook
+
+	Owner() (peer.ID, crypto.PrivKey)
```
b5, line 24: I know I wrote this, but I think we should remove the `Owner` method, and the concept of storing "owner" state with the `key.Store` interface entirely. I'd rather see the current user cached in configuration / on an instance, and keep the key store as in-the-dark as possible.

If all this interface is doing is abstracting `keystone.KeyBook` and adding persistence, that's a great step IMHO.

profile/store.go
```diff
@@ -24,6 +25,8 @@ type Store interface {
 	Owner() *Profile
 	// SetOwner handles updates to the current user profile at runtime
 	SetOwner(own *Profile) error
+	// profile stores are built atop a store of keypairs
+	KeyStore() key.Store
```
b5, line 29: I'm not sure we want to surface the KeyStore in the interface. This starts to recreate the "repo problem" where we have different functions relying on different sources for interface accessors.

Obvi the store constructor / internal state will still need a keystore tho

key/keybook.go
Arqu: yeah, this was simply ripped out from the peerstoremem package. I have to work on this piece a bit to get it in line with our code.

key/key.go
Arqu: perfect

key/store.go
Arqu: glad to see this, the owner provides little to no value here and is just extra stuff to carry around. It's not even persisted in the current context but loaded from config every time.

profile/profile.go
```diff
@@ -143,6 +152,7 @@ func (p Profile) Encode() (*config.ProfilePod, error) {
 	for _, maddr := range p.NetworkAddrs {
 		addrs = append(addrs, maddr.String())
 	}
+	// TODO(aqru): see if we should also show pubkey/keyID here
```
Arqu, line 155: @b5 This is still TODO before I merge

key/keybook.go
```diff
@@ -0,0 +1,179 @@
...
+// PubKey returns the public key for a given ID if it exists
+func (mkb *memoryKeyBook) PubKey(p peer.ID) ic.PubKey {
+	mkb.RLock()
+	pk := mkb.pks[p]
+	mkb.RUnlock()
+	if pk != nil {
+		return pk
+	}
+	pk, err := p.ExtractPublicKey()
+	if err == nil {
```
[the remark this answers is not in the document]
Arqu: Good point. This code was essentially taken out the libp2p peerstore code for the memKeybook and they generally do a best effort to provide a public key if a private key is present. Unsure how that should work on our end. But think it's safer to comment out and put a TODO we can revisit once we settle on the mechanics as this introduces a bit of low level magic we might not be expecting above.

key/keybook.go
```diff
@@ -0,0 +1,179 @@
...
+		mkb.pks[p] = pk
+		mkb.Unlock()
+	}
+	return pk
+}
+
+// AddPubKey inserts a public key for a given ID
+func (mkb *memoryKeyBook) AddPubKey(p peer.ID, pk ic.PubKey) error {
+	// check it's correct first
+	if !p.MatchesPublicKey(pk) {
```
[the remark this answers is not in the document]
Arqu: Yes, this is a first pass based on the libp2p implementation to get it going. 
But, yes this should go away, multiple keys will be supported and will use key identifiers which might not be the same as peerIDs, but are for now.

key/store.go
```diff
@@ -0,0 +1,196 @@
...
+	peerstore.KeyBook
+}
+
+// NewMemStore constructs an in-memory key.Store
+func NewMemStore() (Store, error) {
+	return NewMemStoreKeybook(newKeyBook())
+}
+
+// NewMemStoreKeybook creates an in-memory key.Store with a custom keybook
+func NewMemStoreKeybook(kb peerstore.KeyBook) (Store, error) {
```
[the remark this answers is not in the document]
Arqu: This was left here mostly to have the ability to mix and match the store with the keybook (mainly for testing purposes).
I'll move it out and we can bring it back if needed.

profile/store.go
```diff
@@ -99,14 +107,25 @@ func (m *MemStore) SetOwner(own *Profile) error {
...
 		return fmt.Errorf("profile.ID is required")
 	}
 
 	m.Lock()
-	m.store[profile.ID] = profile
+	m.store[p.ID] = p
 	m.Unlock()
+
+	if p.PubKey != nil {
+		if err := m.keyStore.AddPubKey(peer.ID(p.GetKeyID()), p.PubKey); err != nil {
```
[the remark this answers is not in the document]
Arqu: Yeah, I've been fighting it myself, and only reason I carried it is cause I started out with it and then build around it. Since we're sporting our own keybook anyways, I'll clean this all up.

profile/profile.go
Arqu: @dustmop any opinion on this?

dustmop: I think this is a no? We have the profileID, and can use that to look up the key{public,id} from the keystore. No reason to store it twice, better to keep it normalized.

dustmop (approved):
LGTM!

PR #1603, opened 2021-01-15
feat(apply): Apply command, and `qri save --apply`

dustmop:


dustmop:
Implements the non-FSI parts of https://github.com/qri-io/rfcs/blob/master/text/0027-transform_application.md

lib/transform.go
```diff
@@ -0,0 +1,102 @@
...
+	"fmt"
+	"io"
+
+	"github.com/qri-io/dataset"
+	"github.com/qri-io/qri/base"
+	"github.com/qri-io/qri/dsref"
+)
+
+// TransformMethods encapsulates business logic for transforms
+// TODO (b5): switch to using an Instance instead of separate fields
```
b5, line 14: ded copy-paste

lib/transform.go
```diff
@@ -0,0 +1,102 @@
...
+
+// ApplyParams are parameters for the apply command
+type ApplyParams struct {
+	Refstr    string
+	Transform *dataset.Transform
+	Source    string
+	Secrets   map[string]string
+
+	// TODO(dustmop): add `Wait bool`, if false, run the save asynchronously
+	// and return events on the bus that provide the progress of the save operation
```
b5, line 37: great idea

cmd/save.go
```diff
@@ -65,13 +65,14 @@ commit message and title to the save.`,
 	cmd.Flags().StringVarP(&o.Message, "message", "m", "", "commit message for save")
 	cmd.Flags().StringVarP(&o.BodyPath, "body", "", "", "path to file or url of data to add as dataset contents")
 	cmd.MarkFlagFilename("body")
-	cmd.Flags().StringVarP(&o.Recall, "recall", "", "", "restore revisions from dataset history")
 	// cmd.Flags().BoolVarP(&o.ShowValidation, "show-validation", "s", false, "display a list of validation errors upon adding")
+	cmd.Flags().BoolVar(&o.Apply, "apply", false, "apply a transformation and save the result")
+	cmd.Flags().BoolVar(&o.NoApply, "no-apply", false, "don't apply any transforms that are added")
```
b5, line 70: this flag feels a little clunky, but playing with this branch in the wild makes sense. Forcing the user to explicitly say what they want to do with the transform is good, but the ergonomics of these two flags _and_ the needed `--file` flag feels like it could use improvement. How do you feel about making `--apply` and `--no-apply` accept path args that must be transforms?

```
$ qri save --apply tf.star me/dataset
$ qri save --no-apply tf.star me/dataset
```

On one hand it violates the multiple `--file` flag convention, but we're already doing that with `--body`. 🤷

b5 (commented):
Looks great. I have one question about flags, and a few nits. Once we come to a decision on the flags & get those changes in, we should get this merged!

cmd/save.go
dustmop: I've gone back and forth on the need for a `--file` flag. Looking at the [RFC](https://github.com/qri-io/rfcs/blob/master/text/0027-transform_application.md), which also explains that exact rational for `--no-apply`, there is also no `--file` flag. The reason I thought it should be put in place is the ergonomics of:

1) Crafting an apply command to develop a transform.
2) Running the transform iteratively and eventually finishing it.
3) Changing it to make a commit by simply insert the string `save --` between `qri` and `apply`.

Example:
```
qri apply --file dev_transform.star me/my_dataset
```
becomes:
```
qri save --apply --file dev_transform.star me/my_dataset
```

Removing the need for `--file` for `apply` would break that 3rd step, since it would turn a command with 2 positions arguments into a command with only 1. However, I agree that requiring `--file` makes step 1 a bit more awkward. Looking to see if you have any thoughts on this specifically because I'm rather split-minded.

b5: I'm with you on the fence. Ok, let's keep the PR as-written then

Issue #1609, opened 2021-01-18
transform.Apply emit events


b5:
_depends on #1607, #1608_

Once we've landed the transform package & event bus additions, we can use them to make `transform.Apply` emit events that represent state transitions, creating a log of transform execution.

###Steps
* If we haven't already added this: Adjust transform.Apply to return a new runID parameter: `(runID string, err error)`. Apply should generate a new runID on each call.
* transform.Apply accepts a “wait” argument that when passed true early returns with a run Identifier. In this PR `lib` will always pass `true` for wait, keeping current behaviour
* Augment transform.Apply with events that write on the event bus. See code spike for details
* Events should use a channel to enforce a total order when being sent to the event bus

b5:
closed by #1623

PR #1614, opened 2021-01-20
feat(event): Subscribe by event topic, id, or to all

dustmop:
Expand the abilities of the eventBus in preparation for transform's needs. Can subscribe to events by "sessionID", or to all events. Rename the old `Subscribe` method to `SubscribeTopics`. The old name `Type` was a bit ambiguous, and lead to using unfortunate variable names like `typ`; I've changed it to `Topic`, a term that already in use in some of the comments and documentation.

### DIFF
--- a/base/dsfs/dataset_test.go
+++ b/base/dsfs/dataset_test.go
@@ -334,9 +334,9 @@ func TestDatasetSaveEvents(t *testing.T) {
 	privKey := testPeers.GetTestPeerInfo(10).PrivKey
 	bus := event.NewBus(ctx)
 
-	fired := map[event.Topic]int{}
-	bus.SubscribeTopics(func(ctx context.Context, e event.Event) error {
-		fired[e.Topic]++
+	fired := map[event.Type]int{}
+	bus.SubscribeTypes(func(ctx context.Context, e event.Event) error {
+		fired[e.Type]++
 		return nil
 	},
 		event.ETDatasetSaveStarted,
@@ -356,7 +356,7 @@ func TestDatasetSaveEvents(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	expect := map[event.Topic]int{
+	expect := map[event.Type]int{
 		event.ETDatasetSaveStarted:   1,
 		event.ETDatasetSaveProgress:  2,
 		event.ETDatasetSaveCompleted: 1,
--- a/base/dsfs/dataset_test.go
+++ b/base/dsfs/dataset_test.go
@@ -334,9 +334,9 @@ func TestDatasetSaveEvents(t *testing.T) {
 	privKey := testPeers.GetTestPeerInfo(10).PrivKey
 	bus := event.NewBus(ctx)
 
-	fired := map[event.Type]int{}
-	bus.Subscribe(func(ctx context.Context, t event.Type, payload interface{}) error {
-		fired[t]++
+	fired := map[event.Topic]int{}
+	bus.SubscribeTopics(func(ctx context.Context, e event.Event) error {
+		fired[e.Topic]++
 		return nil
 	},
 		event.ETDatasetSaveStarted,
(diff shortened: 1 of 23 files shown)


dustmop:
Part of https://github.com/qri-io/qri/issues/1609.

event/event.go
```diff
@@ -105,17 +137,47 @@ func NewBus(ctx context.Context) Bus {
...
+		if err := handler(ctx, e); err != nil {
+			return err
+		}
+	}
+
+	if sessionID != "" {
+		for _, handler := range b.idSubs[sessionID] {
+			if err := handler(ctx, e); err != nil {
+				return err
+			}
```
Arqu, line 175: I think this bit can be a bit risky. We run the risk of double subscription. Ie a given code piece does `SubscribeAll` and in a separate step does an ID subscription to be able to isolate a given session stream. If not careful at the moment of writing we might end up with a handler being added at two points in which case weird stuff can happen.

Now as a solution I wouldn't automagically start de-duping request handlers here and would leave it to the user to keep track of what they're doing, however adding some sort of handler ID and logging it would be very useful for instrumentation and tracing now that we're going to rely so much more on the event bus. 

Thinking of it, this might not even be the right place to start solving this but maybe attach a trace to the context that's passed in?

event/event.go
```diff
@@ -124,16 +186,32 @@ func (b *bus) Publish(ctx context.Context, topic Type, data interface{}) error {
...
+func (b *bus) SubscribeAll(handler Handler) {
+	b.lk.Lock()
+	defer b.lk.Unlock()
+	log.Debugf("Subscribe All")
+	b.allSubs = append(b.allSubs, handler)
+}
+
 // NumSubscribers returns the number of subscribers to the bus's events
 func (b *bus) NumSubscribers() int {
 	b.lk.Lock()
```
Arqu, line 217: I know it's not your code and this is used pretty infrequently but this should be an `RLock/RUnlock`

Arqu (commented):
Pretty much happy as it is, only thing I think would be useful is some basic tracing instrumentation for the handler/publisher relationship so we can follow the event back to the source as things get a bit more evolved with the event bus.

base/dsfs/dataset_test.go
```diff
@@ -334,9 +334,9 @@ func TestDatasetSaveEvents(t *testing.T) {
 	privKey := testPeers.GetTestPeerInfo(10).PrivKey
 	bus := event.NewBus(ctx)
 
-	fired := map[event.Type]int{}
-	bus.Subscribe(func(ctx context.Context, t event.Type, payload interface{}) error {
-		fired[t]++
+	fired := map[event.Topic]int{}
```
b5, line 337: we switched away from `Topic` to `Type` to match react's (ok, redux) event system, which uses the `type` keyword to demux action types. Let's keep `Type`, and ideally remove any erroneous references to `topic` still hanging around

event/event.go
```diff
@@ -17,39 +18,56 @@ var (
...
+// matching event is published. Handler calls are blocking, called in order
 // of subscription. Any error returned by a handler is passed back to the
 // event publisher.
 // The handler context originates from the publisher, and in practice will often
 // be scoped to a "request context" like an HTTP request or CLI command
 // invocation.
 // Generally, even handlers should aim to return quickly, and only delegate to
 // goroutines when the publishing event is firing on a long-running process
-type Handler func(ctx context.Context, t Type, payload interface{}) error
+type Handler func(ctx context.Context, e Event) error
```
b5, line 47: Big fan of this new Handler signature

event/event.go
b5: I think this is an issue we can skip dealing with for the time being, but I think you're right on both counts asmir: using the context param is a great place to start the instrumentation process, and having an answer for the double subscription problem would be nice to have.

For now, let's just be careful when Subscribing 😄

PR #1615, opened 2021-01-20
refactor(api): adjust `Server` & middleware funcs to make the qri api more extensible by outside packages

### LINKED ISSUE #1611: Make API package extensible
We’d like qrimatic to add to the existing qri core API. Rework the `AddRoutes` method to operate on a passed-in `http.ServeMux` that qrimatic can own. Export functions like `Middleware` and `NoLogMiddleware` so outside packages can use them. When finished, it should be possible to write the following (pseudo) program:

```go
package main

import (
	"github.com/qri-io/qri/lib"
	"github.com/qri-io/qri/api"
)

func main() {
  inst, _ := lib.NewInstance(/* ... */)
  apiServer := api.New(s.inst)
  apiServer.Mux.Handle("/extra_endpoint", apiServer.Middleware(MyCustomHandler))

	// block 
	_ := apiServer.Serve(ctx)
}
```

ramfox:
Adds `Mux` field to `Server` struct (type `http.ServeMux`). A new `Mux`
is created in the `api.New` function when it returns a new `Server`.

The `Middleware` and `NoLogMiddleware` functions are now exported.

closes #1611

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -35,11 +35,14 @@ func init() {
 // Create one with New, start it up with Serve
 type Server struct {
 	*lib.Instance
+	Mux *http.ServeMux
 }
 
 // New creates a new qri server from a p2p node & configuration
 func New(inst *lib.Instance) (s Server) {
-	return Server{Instance: inst}
+	s = Server{Instance: inst}
+	s.Mux = NewServerRoutes(s)
+	return s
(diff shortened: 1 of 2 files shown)


b5 (approved):
LGTM

Issue #1601, opened 2021-01-11
test TestNewInstanceWithAccessControlPolicy creates files in source directory


dustmop:
Running the test TestNewInstanceWithAccessControlPolicy in lib/lib_test.go creates two files in the source directory:

```
lib/peers.json
lib/peers.json.lock
```

This is probably a consequence of not using a temporary directory when it should be used.

PR #1616, opened 2021-01-22
fix(repo): Don't use blank path for new repo in tests

dustmop:
Real runs of qri read the config from a path, which sets that path. Most tests use a mem repo. The test TestNewInstanceWithAccessControlPolicy uniquely uses a fs repo but that config doesn't know the repo path. By not setting that path, running this test would create json and json.lock files when creating a KeyStore and ProfileStore.

Fixes https://github.com/qri-io/qri/issues/1601


b5 (approved):
LGTM!

PR #1617, opened 2021-01-25
refactor(ws): refactoring websockets to use same port as http API

Arqu:
Updating websockets to work over the same port.

### DIFF
--- a/lib/websocket.go
+++ b/lib/websocket.go
@@ -2,12 +2,8 @@ package lib
 
 import (
 	"context"
-
 	"net/http"
-	"time"
 
-	ma "github.com/multiformats/go-multiaddr"
-	manet "github.com/multiformats/go-multiaddr-net"
 	"github.com/qri-io/qri/event"
 	"github.com/qri-io/qri/watchfs"
 	"nhooyr.io/websocket"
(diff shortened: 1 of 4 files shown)


lib/websocket.go
```diff
@@ -16,11 +12,25 @@ import (
 
 const qriWebsocketProtocol = "qri-websocket"
 
-// ServeWebsocket creates a websocket that clients can connect to in order to
-// get realtime events
-func (inst *Instance) ServeWebsocket(ctx context.Context) {
-	apiCfg := inst.cfg.API
+// Collect all websocket connections
+var wsConnections = []*websocket.Conn{}
```
b5, line 16: Not a fan of having this as a global, especially since we'll need to do dead connection mgmt for long-lived processes. Let's do an interface with a single `Serve` method & + private implementation:

```go
type WebsocketHandler interface {
  WSConnectionHandler(w http.ResponseWriter, r *http.Request)
}

type wsHandler struct {
  conns []*websocket.Conn
}

var _ WebsocketHandler = (*wsHandler)(nil)

func NewWebsocketHandler(ctx context.Context, bust event.Bus, r repo.Repo) WebsocketHandler {
 // ...
}

func (h *wsHandler)  WSConnectionHandler(w http.ResponseWriter, r *http.Request) {
  // ...
}
```

While we're here, looks like the constructor doesn't need a full instance, just `event.Bus` & `repo.Repo` args.

b5 (approved):
looks great! thx

Issue #1608, opened 2021-01-18
Extend `event` package to meet needs of transform logging


b5:
We need to teach the event package some new tricks to make it work as a foundation for transform execution logging. First, we need to be able to order events, we’ll do that by having the bus datestamp events as they come in, then clients can order by date. Second we need  to add identifiers to log events so we can disambiguate different logs from each other. We’ll do that by adding a `PublishID` method to the event.Publisher interface that attaches an ID to an event.

* Define new set of `Transform` events & structure (See code spike)
* Event bus generates timestamps using an overridable `NowFunc`. 
* New `event.Event` struct is passed to handlers, can be used outside of event package for event encapsulation (transform package will use `event.Event` in it’s event-ordering channel (see transform.Apply emits events project)
* Add new “PublishID” method to event.Publisher interface that allows publishing an event with an identifier
*  Write tests for `PublishID`, augment existing tests to check timestamps with a fake timestamp function that just adds 1 on each call.
* Add a method to the bus for subscribing to all events. Switch the websocket to subscribe to all events instead of a predefined list.

PR #1623, opened 2021-01-27
feature(transform): Transform step execution

dustmop:


dustmop:
Implements the rest of https://github.com/qri-io/qri/issues/1609 and most of https://github.com/qri-io/qri/issues/1608.

b5 (approved):
@dustmop I opted to pick this branch up & run with it. I've added a test that illustrates a few changes I've made to the events emitted. Most notable are using `TransformStop` instead of `TransformSuccess/Fail`, and defining a bunch of constants that enumerate run states & message levels. I'm going to merge this as-is, but the API changes are definitely still up for discussion IMHO

PR #1627, opened 2021-02-01
refactor(profile): add active profile method stub and thread context …

Arqu:
…through to profile store

Since it's a bit hard on the eyes to review, a few quick higlights:
- Added `.Active(ctx)` which will later be fleshed out to return the "selected" profile instead of the owner
- Exposed an explicit `.Owner()` on repo to be able to perform work which is related to the repo owner
- All p2p stuff is linked to the `.Owner`, the owner should carry all the p2p stuff not temporary profiles
- The context is threaded through to all the `repo.Profile/Key` -> `profile_store...` actions but most of the entry points are either `context.TODO()` or `ctx.Background()` which need to be resolved and double checked in a subsequent refactor

### DIFF
--- a/profile/store.go
+++ b/profile/store.go
@@ -1,6 +1,7 @@
 package profile
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -25,6 +26,8 @@ type Store interface {
 	Owner() *Profile
 	// SetOwner handles updates to the current user profile at runtime
 	SetOwner(own *Profile) error
+	// Active is the active profile that represents the current user
+	Active(ctx context.Context) *Profile
 
 	// put a profile in the store
 	PutProfile(profile *Profile) error
@@ -100,7 +103,7 @@ func NewMemStore(owner *Profile, ks key.Store) (Store, error) {
 	}, nil
 }
 
-// Owner accesses the current user profile
+// Owner accesses the current owner user profile
 func (m *MemStore) Owner() *Profile {
 	return m.owner
 }
@@ -109,6 +112,11 @@ func (m *MemStore) Owner() *Profile {
 func (m *MemStore) SetOwner(own *Profile) error {
 	m.owner = own
 	return nil
+}
+
+// Active is the curernt active profile
+func (m *MemStore) Active(ctx context.Context) *Profile {
+	return m.Owner()
 }
 
 // PutProfile adds a peer to this store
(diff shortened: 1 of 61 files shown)


profile/store.go
```diff
@@ -25,6 +26,8 @@ type Store interface {
 	Owner() *Profile
 	// SetOwner handles updates to the current user profile at runtime
 	SetOwner(own *Profile) error
+	// Active is the active profile that represents the current user
+	Active(ctx context.Context) *Profile
```
b5, line 30: I'm assuming a follow-up will define something a lot like [this](https://github.com/qri-io/qri/blob/63fadf8339de9a603aff382f95bac89b92b0361b/remote/context.go#L1-L26) in the `key` package, adding a `context.go` file & functions for adding & getting key identifier (and eventually public key) from a context, and that this `Active` function will effectively be:

```go
func ActiveProfile(ctx context.Context) (*Profile, error) {
  keyID, err := key.KeyIDFromContext(ctx)
  return store.GetByKeyID(keyID)
}
```

If so, let's make `ActiveProfile` a function that accepts a store, reducing the need for implementations to add boilerplate:

```go
package profile

func ActiveProfile(ctx context.Context, store Store) (*Profile, error) {
  keyID, err := key.KeyIDFromContext(ctx)
  return store.GetByKeyID(keyID)
}
```

When embedding values in contexts, it's best to keep the context accessors high up the call stack. I'd personally love to see pseudocode like this:

```go
package lib

func (m *DatasetMethods) Save(ctx context.Context, params *SaveParams) (*dataset.Dataset, error) {

  keyID, privKey, err := key.PrivateKeyFromContext(ctx, m.Keystore)
  // error earliest if we don't have the private key
  
  // this API isn't worked out yet, but something like...
  cred, err := access.CredentialTokenForResource(p.DatasetID, keyID)
  if err != nil {
    return ErrAccessDenied
  }

  return base.Save(ctx, params, privKey /* ... */)
}
```

For the next PR if you haven't already had a read through the golang blog post on context, def worth starting there: https://blog.golang.org/context

b5 (approved):
Lovely. Looks like you have a legitimate failing case on a Websocket test. Get that to pass & then feel free to merge!

profile/store.go
Arqu: Yes, you're pretty close to the target on what the goal is.
There is one key distinction I want to throw out there:
- There is managing multiple keys for a profile
- And then there is managing multiple profiles with their keys

We technically need to cover for both. I will first be covering the multiple profiles path as we don't have the multiple key stuff sorted yet.

PR #1628, opened 2021-02-02
spike(http): transitioning from rpc to http

Arqu:
DO NOT MERGE!

This is a quick spike to test out some concepts around moving from RPC to HTTP and flesh out some work ahead.

Among other things:
- Generalize the `lib.HTTPClient.Call` method
- Substitute all the RPC callers
- Ensure param/result object compatibility
- Ensure everything we need is JSON serializable
- Encoding/Decoding of `context.Context` into the request (headers)
- Map out all the HTTPClient actions
- Probably fill in some blanks for API
- Finish bubbling context to `qri/cmd` & `qri/api`

Things that definitely don't work with this setup yet:
- Console output is not passed along
- Events are swallowed

This will require for us to also solve the WebSocket streaming between client/node


lib/transform.go
```diff
@@ -35,7 +35,7 @@ type ApplyParams struct {
 	Wait      bool
 
 	Source       string
-	ScriptOutput io.Writer
+	ScriptOutput io.Writer `json:"-"`
```
dustmop, line 38: I wonder if it would be not too hard to wire up an io.Writer that forwards output to the websocket?

lib/http.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func NewHTTPClient(addr string) *HTTPClient {
+	return &HTTPClient{
+		Addr: addr,
+	}
+}
+
+func (c HTTPClient) Call(action string, ctx context.Context, params interface{}, result interface{}) error {
+	if action == "TransformMethods.Apply" {
```
dustmop, line 27: Totally workable for a code spike, just wanted to point to how go rpc works. It has a function "suitableMethods" (see https://golang.org/src/net/rpc/server.go) that uses reflection to find all methods on an interface that matches a certain pattern, and then automatically exposes them over rpc. I hope we can do something similar for the http case, though we would need some way to map the method names to their respective endpoints.

lib/http.go
```diff
@@ -0,0 +1,71 @@
...
+		log.Debugf("transform.apply body")
+		body, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			return err
+		}
+
+		log.Debugf("transform.apply.body: %s", body)
+		resData := struct {
+			Data interface{}            `json:"data"`
+			Meta map[string]interface{} `json:"meta"`
```
dustmop, line 58: Is it the case that Meta will always have known fields? I think the api is always creating the same "envelope" containing the status code, optional error message, etc.

lib/http.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+const jsonMimeType = "application/json"
```
dustmop, line 12: Probably nearly always true that we always use application/json, but I think zip output doesn't, there might be one or two other methods.

lib/lib.go
```diff
@@ -408,6 +408,11 @@ func NewInstance(ctx context.Context, repoPath string, opts ...Option) (qri *Ins
 			// we have a connection
 			log.Debugf("using RPC address %s", cfg.RPC.Address)
 			inst.rpc = rpc.NewClient(conn)
+
+			// TODO(arqu): I blindly assume this means we have a working API connection
+			// till the HTTPClient is fleshed out
+			inst.http = NewHTTPClient("localhost:2503")
```
dustmop, line 414: Port should come from the config.

dustmop (commented):
Cool as a proof-of concept! Excited to see this develop.

lib/http.go
b5: agreed. I think grouping MIME types supported by a package into a set of constants like this is a healthy practice.

lib/http.go
b5: I firmly agree with this point of automatic/programmatic mapping the API layer from lib. I thought it'd be good to start with the apply command for this reason: the `/apply` endpoint is ripe for formalizing. So far it's just a JSON serialization wrapper to a lib method over HTTP.

Need to put more thought into how such a generalization would work, and we'll most certainly have endpoints that won't fit neatly into some "make an HTTP endpoint for this lib method" box, but I think we should still move in that direction.

b5 (commented):
Looking really good. Let's break lib.TransformMethods.Apply & see how many things explode. I think this is a good first pass though!

Arqu:
This is now old based on https://github.com/qri-io/qri/pull/1629

PR #1629, opened 2021-02-04
feat(httpClient): introducing an httpClient

Arqu:
This is the first step towards moving from RPC > HTTP.

So far this is non-breaking and simply introduces the HTTPClient and semi-defines the API endpoints we have.
Next series of PRs will convert us from RPC to the HTTPClient and funnel down the context from above to lib.

Also used the opportunity to update our ma/net package as the original one was deprecated and moved in as a subpackage.


lib/lib.go
```diff
@@ -408,6 +408,19 @@ func NewInstance(ctx context.Context, repoPath string, opts ...Option) (qri *Ins
...
+			// TODO(arqu): I blindly assume this means we have a working API connection
+			// till the HTTPClient is fleshed out
+			maAddr, err := ma.NewMultiaddr(cfg.API.Address)
+			if err != nil {
+				return nil, err
+			}
+			httpAddr, err := manet.ToNetAddr(maAddr)
+			if err != nil {
+				return nil, err
+			}
```
b5, line 421: zomg I've wanted this function for a long time.

We should move these into the `NewHTTPClient` constructor though, and pass `cfg.API.Address` directly

lib/lib.go
```diff
@@ -863,6 +877,14 @@ func (inst *Instance) RPC() *rpc.Client {
 	return inst.rpc
 }
 
+// HTTP accesses the instance HTTP client if one exists
+func (inst *Instance) HTTP() *HTTPClient {
```
b5, line 881: Any reason not to name this method `HTTPClient`?

config/api.go
```diff
@@ -3,6 +3,7 @@ package config
 import (
```
b5, line 3: Not a fan of having all this defined in config. Can we move it to a new file `/lib/api.go`?

lib/http.go
```diff
@@ -0,0 +1,102 @@
...
+	// TODO(arqu): inject context values into headers
+
+	return c.do(ctx, addr, httpMethod, mimeType, params, result)
+}
+
+func (c HTTPClient) do(ctx context.Context, addr string, httpMethod string, mimeType string, params interface{}, result interface{}) error {
+	var req *http.Request
+	var err error
+
+	if httpMethod == "GET" || httpMethod == "DELETE" {
```
b5, line 48:
```go
	if httpMethod == http.MethodGet || httpMethod == http.MethodDelete {
```

lib/http.go
```diff
@@ -0,0 +1,102 @@
...
+// NewHTTPClient instantiates a new HTTPClient
+func NewHTTPClient(addr string) *HTTPClient {
+	return &HTTPClient{
+		Addr: addr,
+	}
+}
+
+// Call resolves the action by mapping it to the correct API endpoint and passes on parameters and context info
+// currently this is a fairly flat process but should evolve in the future
+func (c HTTPClient) Call(ctx context.Context, apiEndpoint config.APIEndpoint, methodOverride string, params interface{}, result interface{}) error {
```
b5, line 30: Change `Call`'s signature to:

```go
func (c HTTPClient) Call(ctx context.Context, apiEndpoint config.APIEndpoint, params interface{}, result interface{}) error {
```

and have it proxy a second exported method `CallMethod`

```
func (c HTTPClient) CallMethod(ctx context.Context, apiEndpoint config.APIEndpoint, params interface{}, result interface{}) error {
```

passing in a default `http.MethodPost` when using `Call`

lib/http.go
```diff
@@ -0,0 +1,102 @@
...
+
+// NewHTTPClient instantiates a new HTTPClient
+func NewHTTPClient(addr string) *HTTPClient {
+	return &HTTPClient{
+		Addr: addr,
+	}
+}
+
+// Call resolves the action by mapping it to the correct API endpoint and passes on parameters and context info
+// currently this is a fairly flat process but should evolve in the future
```
b5, line 29: Comments like `currently this is...` get stale pretty quickly. I'd prefer we not add 'em

lib/http.go
```diff
@@ -0,0 +1,102 @@
...
+func (c HTTPClient) Call(ctx context.Context, apiEndpoint config.APIEndpoint, methodOverride string, params interface{}, result interface{}) error {
+	// post should be the default method as we are almost always passing on a call from one instance to another
+	httpMethod := http.MethodPost
+	if methodOverride != "" {
+		httpMethod = methodOverride
+	}
+	// TODO(arqu): work out mimeType configuration/override per API endpoint
+	mimeType := jsonMimeType
+	addr := fmt.Sprintf("http://%s/%s", c.Addr, apiEndpoint)
```
b5, line 38: we don't always use `HTTP`. I think this means the `NewHTTPClient` constructor is going to need to be more robust.

dustmop:
No comments aside from b5's feedback.

b5 (approved):
Looks great! v excited about this one

PR #1630, opened 2021-02-05
feat(prove): Prove a new keypair for an account, set original profileID

dustmop:


registry/regclient/profile.go
```diff
@@ -38,9 +38,44 @@ func (c *Client) CreateProfile(p *registry.Profile, pk crypto.PrivKey) (*registr
...
+	if p.PublicKey == "" {
+		return nil, fmt.Errorf("ProveKeyForProfile: PublicKey required")
+	}
+	if p.Signature == "" {
+		return nil, fmt.Errorf("ProveKeyForProfile: Signature required")
+	}
+
+	// Send proof request to the registry
+	res := RegistryResponse{}
+	err := c.doJSONRegistryRequest("PUT", "/registry/provekey", p, &res)
```
Arqu, line 70: Can we have this path as a constant somewhere, it will align better with the upcoming API work.

registry/regclient/profile.go
```diff
@@ -135,3 +170,48 @@ func (c Client) doJSONProfileReq(method string, p *registry.Profile) (*registry.
 	env.Data.Peername = env.Data.Username
 	return env.Data, nil
 }
+
+// RegistryResponse is a generic container for registry requests
+// TODO(dustmop): Only used currently for ProveKey
+type RegistryResponse struct {
```
Arqu, line 176: I'm also doing this for the HTTPClient work, would be good to have a common place for it with the shifting emphasis to HTTP APIs. `apiutil` definitely feels like the best place for it.

If you feel like it, you can leave a `// TODO(arqu): update with common API response object` as I'll soon be doing a large swoop of my own TODOs and can cover this too.

registry/regclient/profile.go
```diff
@@ -135,3 +170,48 @@ func (c Client) doJSONProfileReq(method string, p *registry.Profile) (*registry.
...
+	Data map[string]string
+	Meta struct {
+		Error  string
+		Status string
+		Code   int
+	}
+}
+
+// doJSONProfileReq sends a json body to the registry
```
Arqu, line 185: This can potentially also be moved to the HTTPClient.
I'd be happy if this merged soon and I can refactor accordingly.

Arqu (approved):
LGTM. Just some nits/comments.

registry/regclient/profile.go
Arqu: I've added a PR to `apiutil` for this. Also one more note the "status" response is not strictly correct per our API specs.

dustmop:
Thanks! Pushed a commit with fixes, merging now.

PR #1631, opened 2021-02-05
refactor(apply): RPC > HTTP

Arqu:
First of several to come for transitioning from RPC to HTTP and moving `context` up to `api` & `cmd`

### DIFF
--- a/api/transform.go
+++ b/api/transform.go
@@ -27,8 +27,8 @@ func (h TransformHandlers) ApplyHandler(prefix string) http.HandlerFunc {
 			return
 		}
 
-		res := lib.ApplyResult{}
-		if err := h.TransformMethods.Apply(&p, &res); err != nil {
+		res, err := h.TransformMethods.Apply(r.Context(), &p)
+		if err != nil {
 			util.WriteErrResponse(w, http.StatusBadRequest, err)
 			return
 		}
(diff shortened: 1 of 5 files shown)


api/transform.go
```diff
@@ -27,8 +27,8 @@ func (h TransformHandlers) ApplyHandler(prefix string) http.HandlerFunc {
 			return
 		}
 
-		res := lib.ApplyResult{}
-		if err := h.TransformMethods.Apply(&p, &res); err != nil {
+		res, err := h.TransformMethods.Apply(r.Context(), &p)
+		if err != nil {
 			util.WriteErrResponse(w, http.StatusBadRequest, err)
 			return
```
b5, line 33: I'm delighted to see we don't need to do any shenanigans with type assertions here. I tend to forget the raw `interface{}` return type has merit if you're passing it to a thing that accepts `interface{}`

b5 (approved):
Looks good. It's nice to see that our first step in refactoring is a straight up replacement for the `net/rpc` package.

api/transform.go
Arqu: Yeah we kinda managed to isolate the shenanigans down into the httpClient so it doesn't feel weird and makes all the above code follow patterns from rest of the codebase.

And yeah, we're literally forced to replace the RPC here because we're braking the contract by passing in context.

PR #1633, opened 2021-02-08
spike(dataset): RPC > HTTP

Arqu:
So I realized the PR could use some added comments:
- Current API has a lot of inconsistent behavior based on the contents that are passed in to the request
- We have troubles encoding and shuffling around query and path params into the `lib.Params`
- Due to the duality of request processing eg `/get/` & `/save/` we still have to track whether the request was "proxied"

Things that are broken or untested with the new changes:
- scriptOutput doesn't work (and it didn't work with RPC either) - this should be moved to websockets
- `lib.GetRawRefs` doesn't have an API endpoint/handler at all
- `/get/` with download/zip responses should work but are not properly tested yet
- `/save/` is subtly broken in places which might become more apparent with other clients like Desktop. For one, all requests now require the dataset to be passed in as a value to the key `dataset` as part of the `lib.SaveParams` instead of as a raw dataset.
- save mimeType tests are broken and probably the API too as we don't have a nice way to nest the key/value pairs or alternatively a separate parser on the `api.saveHandler` for multipart forms
- `/rename/` has backwards compatibility for a `new` param which should be `next` and will be removed
- `/remove/` `remote.Remove` is untested 

Aggregation of input so far on how to move forward:
- Ensure ALL endpoints work strictly by consuming  `lib.Params` and adjust all clients to fit that model
- Ensure less/no branching in "consuming" logic
- Common interface to share or convert the path params like `/get/{peername}/{dsname}/body`
- Remove `Proxy` logic once we have a pattern of single endpoint <> single code path

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -25,6 +25,7 @@ const (
 	DefaultTemplateHash = "/ipfs/QmeqeRTf2Cvkqdx4xUdWi1nJB2TgCyxmemsL3H4f1eTBaw"
 	// TemplateUpdateAddress is the URI for the template update
 	TemplateUpdateAddress = "/ipns/defaulttmpl.qri.io"
+	jsonContentType       = "application/json"
 )
 
 func init() {
@@ -186,8 +187,6 @@ func NewServerRoutes(s Server) *http.ServeMux {
 	m.Handle(lib.AERename.String(), s.Middleware(dsh.RenameHandler))
 	m.Handle(lib.AEDiff.String(), s.Middleware(dsh.DiffHandler))
 	m.Handle(lib.AEChanges.String(), s.Middleware(dsh.ChangesHandler))
-	// Deprecated, use /get/username/name?component=body or /get/username/name/body.csv
-	m.Handle(lib.AEBody.String(), s.Middleware(dsh.BodyHandler))
 	m.Handle(lib.AEStats.String(), s.Middleware(dsh.StatsHandler))
 	m.Handle(lib.AEUnpack.String(), s.Middleware(dsh.UnpackHandler))
 
--- a/api/api.go
+++ b/api/api.go
@@ -28,7 +28,7 @@ const (
 )
 
 func init() {
-	golog.SetLogLevel("qriapi", "info")
+	golog.SetLogLevel("qriapi", "debug")
 }
 
 // Server wraps a qri p2p node, providing traditional access via http
--- a/api/datasets.go
+++ b/api/datasets.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/gorilla/schema"
 	"github.com/qri-io/dataset"
 	"github.com/qri-io/qri/api/util"
 	"github.com/qri-io/qri/base/archive"
@@ -142,20 +143,6 @@ func (h *DatasetHandlers) RenameHandler(w http.ResponseWriter, r *http.Request) 
 	}
 }
 
-// BodyHandler gets the contents of a dataset
-func (h *DatasetHandlers) BodyHandler(w http.ResponseWriter, r *http.Request) {
-	switch r.Method {
-	case http.MethodGet:
-		if h.ReadOnly {
-			readOnlyResponse(w, "/body/")
-			return
-		}
-		h.bodyHandler(w, r)
-	default:
-		util.NotFoundHandler(w, r)
-	}
-}
-
 // StatsHandler gets stats about the dataset
 func (h *DatasetHandlers) StatsHandler(w http.ResponseWriter, r *http.Request) {
 	switch r.Method {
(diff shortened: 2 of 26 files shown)


api/api.go
```diff
@@ -28,7 +28,7 @@ const (
 )
 
 func init() {
-	golog.SetLogLevel("qriapi", "info")
+	golog.SetLogLevel("qriapi", "debug")
```
Arqu, line 31: Ignore this

api/datasets.go
```diff
@@ -42,7 +43,7 @@ func NewDatasetHandlers(inst *lib.Instance, readOnly bool) *DatasetHandlers {
 // ListHandler is a dataset list endpoint
 func (h *DatasetHandlers) ListHandler(w http.ResponseWriter, r *http.Request) {
 	switch r.Method {
-	case http.MethodGet:
+	case http.MethodGet, http.MethodPost:
```
Arqu, line 46: Most methods will eventually support POST so we can pass around those parameters.

api/datasets.go
```diff
@@ -197,14 +198,69 @@ func extensionToMimeType(ext string) string {
 	}
 }
 
-func (h *DatasetHandlers) listHandler(w http.ResponseWriter, r *http.Request) {
-	args := lib.ListParamsFromRequest(r)
-	args.OrderBy = "created"
+// snoop reads from an io.ReadCloser and restores it so it can be read again
+func snoop(body *io.ReadCloser) (string, error) {
```
Arqu, line 202: reading the body also consumes it so this is here to restore it back.

api/datasets.go
```diff
@@ -197,14 +198,69 @@ func extensionToMimeType(ext string) string {
...
 
-	res := []dsref.VersionInfo{}
-	if err := h.List(&args, &res); err != nil {
+		*body = ioutil.NopCloser(bytes.NewReader(result))
+		return string(result), nil
+	}
+	return "", nil
+}
+
+func parseProxyParams(r *http.Request, res interface{}) error {
```
Arqu, line 217: This is a bit overcomplicated but the point is getting to a generic path to parse the forwarded parameters.

api/datasets.go
```diff
@@ -220,31 +276,69 @@ func (h *DatasetHandlers) listHandler(w http.ResponseWriter, r *http.Request) {
...
-		util.RespondWithError(w, err)
+	params := lib.GetParams{}
+	args := &GetReqArgs{}
+	err := parseProxyParams(r, &params)
+	if err != nil && err.Error() == "bad request" {
+		util.WriteErrResponse(w, http.StatusBadRequest, err)
 		return
 	}
+	if err != nil {
+		args, err = parseGetReqArgs(r, strings.TrimPrefix(r.URL.Path, "/get/"))
```
Arqu, line 287: The gist of the approach is to try to parse the full params as they would be passed down into lib, if that doesn't work revert to the default flow handling.

api/datasets.go
```diff
@@ -436,6 +529,33 @@ func (h *DatasetHandlers) pullHandler(w http.ResponseWriter, r *http.Request) {
...
 
 func (h *DatasetHandlers) saveHandler(w http.ResponseWriter, r *http.Request) {
+	params := lib.SaveParams{}
+	err := parseProxyParams(r, &params)
+	if err != nil && err.Error() == "bad request" {
+		util.WriteErrResponse(w, http.StatusBadRequest, err)
+		return
+	}
+	if err == nil {
+		res, err := h.Save(r.Context(), &params)
```
Arqu, line 539: When we pass the full pre-computed parameters, we want to just consume them as is on the receiver end.

api/datasets.go
```diff
@@ -220,31 +276,69 @@ func (h *DatasetHandlers) listHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
+	if err != nil {
+		args, err = parseGetReqArgs(r, strings.TrimPrefix(r.URL.Path, "/get/"))
+		if err != nil {
+			util.RespondWithError(w, err)
+			return
+		}
+		params = args.Params
+	} else {
+		// TODO(arqu): get rid of this once we move the proxy call above lib
```
Arqu, line 294: This is a bit ugly but have to do it to populate the "lost" bits as we don't pass the `GetReqArgs` but only hte `GetParams`

lib/params.go
```diff
@@ -11,11 +11,16 @@ import (
 // Limit param is provided to a paginated method
 const DefaultPageSize = 100
 
+// BaseParams defines the basic HTTP params for all requests
+type BaseParams interface {
+	Proxied() bool
```
b5, line 16: I'll admit to being confused about what we intend to do with this. Do we need to know the request is proxied?

api/datasets.go
b5: ok, I needed to think about this a bunch, and I think I'm answering a different question than this proxying project, but I wanted to deal with the issue of different forms of request parameter serialization:

https://github.com/qri-io/qri/commit/c8ff420471e05745222b1edbd6bfb5c352b7124b

The more I look at these new `Proxied` fields, the more I like the way we currently deal with RPC _within_ lib calls, and think part of this issue is just that the API is too thick

b5:
Ok, I'll admit to not loving the new `Proxy` fields in request params.

I understand this is building toward a "Client" type of struct that would rely on such a field, but I don't think we've put enough time into how that API will work. This API is showing early signs of fatigue under handle the growing complexity of asynchronous requests, but before we design this Client pattern, I think we need to clean up the API.

API cleanup will require a bunch of breaking changes, but we'll get context passing in the process, which is the real short-term roadblock

Arqu:
@b5 @dustmop @ramfox 
I've updated the PR comment to introduce some context to the changes.

PR #1634, opened 2021-02-09
refactor(api): introduce UnmarshalFromRequest, update List API

Arqu:
Based on the work in https://github.com/qri-io/qri/pull/1633 this is to land the baseline refactoring and and example with List.
Subsequent PRs will convert rest of the handlers/endpoints to this format and move to the HTTPClient. 
Lastly the "not clear cut" endpoints like `/get/` and `/save` will be refactored and/or split.

### DIFF
--- a/api/datasets.go
+++ b/api/datasets.go
@@ -42,7 +42,7 @@ func NewDatasetHandlers(inst *lib.Instance, readOnly bool) *DatasetHandlers {
 // ListHandler is a dataset list endpoint
 func (h *DatasetHandlers) ListHandler(w http.ResponseWriter, r *http.Request) {
 	switch r.Method {
-	case http.MethodGet:
+	case http.MethodGet, http.MethodPost:
 		if h.ReadOnly {
 			readOnlyResponse(w, "/list")
 			return
@@ -198,13 +198,14 @@ func extensionToMimeType(ext string) string {
 }
 
 func (h *DatasetHandlers) listHandler(w http.ResponseWriter, r *http.Request) {
-	args := lib.ListParamsFromRequest(r)
-	args.OrderBy = "created"
+	params := &lib.ListParams{}
+	if err := UnmarshalParams(r, params); err != nil {
+		util.WriteErrResponse(w, http.StatusBadRequest, err)
+		return
+	}
 
-	args.Term = r.FormValue("term")
-
-	res := []dsref.VersionInfo{}
-	if err := h.List(&args, &res); err != nil {
+	res, err := h.List(r.Context(), params)
+	if err != nil {
 		if errors.Is(err, lib.ErrListWarning) {
 			log.Error(err)
 			err = nil
@@ -214,7 +215,7 @@ func (h *DatasetHandlers) listHandler(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	}
-	if err := util.WritePageResponse(w, res, r, args.Page()); err != nil {
+	if err := util.WritePageResponse(w, res, r, params.Page()); err != nil {
 		log.Infof("error list datasests response: %s", err.Error())
 	}
 }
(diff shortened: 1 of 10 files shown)


api/datasets.go
```diff
@@ -197,14 +199,72 @@ func extensionToMimeType(ext string) string {
 	}
 }
 
-func (h *DatasetHandlers) listHandler(w http.ResponseWriter, r *http.Request) {
-	args := lib.ListParamsFromRequest(r)
-	args.OrderBy = "created"
+// snoop reads from an io.ReadCloser and restores it so it can be read again
```
Arqu, line 202: This should probably live elsewhere.

api/datasets.go
```diff
@@ -197,14 +199,72 @@ func extensionToMimeType(ext string) string {
...
+		return ioutil.NopCloser(bytes.NewReader(result)), nil
+	}
+	return nil, io.EOF
+}
+
+var decoder = schema.NewDecoder()
+
+// UnmarshalParams deserialzes a lib req params stuct pointer from an HTTP
+// request
+func UnmarshalParams(r *http.Request, p interface{}) error {
```
Arqu, line 225: Same as above.

api/datasets.go
```diff
@@ -197,14 +199,72 @@ func extensionToMimeType(ext string) string {
...
+	// TODO(arqu): this should be set on the global decoder
+	// probably means the decoder should live in the API struct or somewhere similar
+	decoder.IgnoreUnknownKeys(true)
```
Arqu, line 228: Any suggestions on this, I want to avoid having it living in a global instance like object and passing everything around and just be a standalone method.

api/datasets.go
```diff
@@ -197,14 +199,72 @@ func extensionToMimeType(ext string) string {
...
+	if err := r.ParseForm(); err != nil {
+		return err
+	}
+	return decoder.Decode(p, r.Form)
+}
 
-	args.Term = r.FormValue("term")
+func (h *DatasetHandlers) listHandler(w http.ResponseWriter, r *http.Request) {
+	params := &lib.ListParams{}
+	if err := UnmarshalParams(r, params); err != nil {
```
Arqu, line 261: Defaults are now handled in the lib.Params through `UnmarshalFromRequest` and `SetNonZeroDefaults`

api/datasets.go
b5: Do we _need_ to ignore unknown keys? I'm guessing this is to deal with the fact that we have non-matching params allowed like `page` and `pageSize`, but I'd like to have unknown keys produce an error. If we need this for now, let's add a `TODO` to remove it in a follow up

In this case I do think it's fine to have as a global. It's just for type caching to get a perf bump at runtime. I'd suggest we move both `UmarshalParams` and the decoder declaration to `api/api.go`

api/datasets.go
b5: I'd generally like to avoid needing this snoop method, but if it needs to exist for now to keep things moving, then so be it

b5 (approved):
some nits, then I think this is workable. Tests would be great, but let's keep moving & do a follow up with tests before we ship a release

api/datasets.go
Arqu: For now it's required but agree it should go away. It's a code smell we have to carry around until the refactor of the endpoints is done.

api/datasets.go
Arqu: Currently there are lots of endpoints that either don't have those generic params, or we ignored them anyways like `fsi=true` on many endpoints.

Arqu: Also moving to api is a bit harder as we don't have the `api/api.go` `Server` struct available in `lib` nor should we. I'll move it to `lib/api.go` and add the `TODOs`.

PR #1637, opened 2021-02-10
refactor(api): RPC > HTTP /save/

Arqu:


lib/api.go
```diff
@@ -133,3 +134,22 @@ const (
 	// AEWebUI serves the remote WebUI
 	AEWebUI = APIEndpoint("/webui")
 )
+
+// DatasetRefFromPath parses a path and returns a datasetRef
+func DatasetRefFromPath(path string) (reporef.DatasetRef, error) {
+	refstr := HTTPPathToQriPath(path)
+	return repo.ParseDatasetRef(refstr)
+}
```
b5, line 142: don't add imports to `repo/ref`, use `dsref.Parse` instead

api/context.go
```diff
@@ -3,9 +3,8 @@ package api
 import (
 	"context"
 	"net/http"
-	"strings"
 
-	"github.com/qri-io/qri/repo"
+	"github.com/qri-io/qri/lib"
```
b5, line 7: well shoot, we should just move this entire file into lib in a future PR

api/serialization.go
```diff
@@ -1,18 +1,9 @@
...
-	"path/filepath"
-	"strings"
 
 	"github.com/qri-io/dataset"
-	"github.com/qri-io/dataset/detect"
-	"github.com/qri-io/qri/base/fill"
-	"gopkg.in/yaml.v2"
 )
 
 // inlineScriptsToBytes consumes all open script files for dataset components
```
b5, line 9: very little left of this file after moving the `formFileDataset` func, let's move `inlineScriptToBytes` into `api/dataset.go`, the only place it's currently used

lib/api.go
b5: That said, I'm a big fan of moving these functions down into lib

b5 (approved):
Looking _great!_. I'm a fan of the few bits you've moved down into lib, and this `UnmarshalFromRequest` pattern seems to be showing it's value.

api/context.go
Arqu: Lol, yes, just added a todo for that. Should happen rather soon.

PR #1640, opened 2021-02-11
feat(logbook): write run operations

b5:
I've added lots of details to commit messages, please read commits while reviewing. I've kept this as a draft as it's not fully tested quite yet, but things are mostly working.

On save, transform components are recalled from history if no transform component is provided to save. This used to be accomplished with a "recall: tf" param, but I've kept it as a transform-specific feature this go-round. I think it ties in a little better with the new --apply semantics.

When saving with a trasnform we now build up a *run.State with the RunID event subscription. I'd initally experimented with designs that had runState built and returned within `transform.Apply`, but found it too easy to create a data race when transform.Apply's wait argument is false. In that case transform.Apply doesn't have a clear moment of when to return a run.State value without giving back a channel, or adding mutex locks & accessor methods to run.State. Seems
it's easier to just build run.State from an external event listener.

This commit introduces automation failure logging, but creates a new issue in the process: When a user "manually" invokes save with the apply flag and the transform fails, a record of that failure will be recorded, which might feel counter-intuitive. The problem here is we're not passing any details to save that lets us disambiguate automation-driven `save` calls from user-driven ones.

To smoke test **please make a backup of your logbook.qfb** repo file before running this branch. The only way to see the result of these changes is to:
1. Create at least one dataset version using the `--apply` flag.
2. Run `qri connect` & `curl http://localhost:2503/history/your/dataset | jq`. You should see fields like `RunID` and `RunDuration` in the JSON response.

### DIFF
--- a/transform/apply.go
+++ b/transform/apply.go
@@ -112,8 +112,8 @@ func Apply(
 		go func() {
 			for {
 				select {
-				case event := <-eventsCh:
-					pub.PublishID(ctx, event.Type, runID, event.Payload)
+				case e := <-eventsCh:
+					pub.PublishID(ctx, e.Type, runID, e.Payload)
 				case <-ctx.Done():
 					return
 				}
@@ -176,10 +176,10 @@ func Apply(
 
 			switch step.Syntax {
 			case SyntaxStarlark:
-				log.Debugw("runnning starlark step", "runID", runID, "category", step.Category, "name", step.Name, "scriptLen", scriptLen(step))
+				log.Debugw("running starlark transform step", "runID", runID, "category", step.Category, "name", step.Name, "scriptLen", scriptLen(step))
 				runErr = stepRunner.RunStep(ctx, target, step)
 				if runErr != nil {
-					log.Debugw("running transform step", "index", i, "err", runErr)
+					log.Debugw("error running starlark transform step", "runID", runID, "index", i, "err", runErr)
 					eventsCh <- event.Event{
 						Type: event.ETTransformError,
 						Payload: event.TransformMessage{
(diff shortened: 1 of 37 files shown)


transform/apply.go
```diff
@@ -102,6 +119,9 @@ func Apply(
 	// until doneCh gets signaled.
 	go func() {
 		if !wait {
+			// if we're running this script async, bind to the background context
+			// note that we lose any values attached to the given context
+			ctx = svc.bgCtx
```
Arqu, line 124: Hmmm, I had a hunc these things would start cropping up with all the context passing. 
For this specific issue, I think it is warranted to actually create a `context.Background()` and re-attach values onto it. Mainly because sharing a global context in this sense will prevent us from multiplexing the values for different simultaneous requests/users without some gross name spacing or introducing a global lock which kills half the multi tenancy benefits.

b5: yeah I'm thinking a very similar thing, but instead of using `context.Background`, use the `bgCtx` here as the root. That way we get to a place we want: a context that will be cancelled when the process tries to exit gracefully (by closing the root) that also has request-scoped values on the child context. Pseudocode:

```go
// package identity
const identityKeyIDKey = "keyID"

func NewUserContext(ctx, newRoot context.Context) (context.Context, context.CancelFunc) {
  // only keyID value added for now, but this func lets us add more later if needed
  key, _ := keyFromContext(ctx)
  return newRoot.WithValue(identityKeyIDKey, key)
}

// at this call site:
go func() {
  if !wait {
    ctx = identity.NewUserContext(ctx, svc.bgCtx)
    // ...
```

Arqu: Yes, forgot that this is still request based and the parent has a scope as well.

b5:
Oook, more to do here, but I've done a self-review & we need this guy to keep moving on qrimatic stuff. I'm going to merge but leave this open to feedback

Issue #1089, opened 2020-01-30
Upcoming circular dependency requires switching to dsref


dustmop:
Dscache wants to know about dataset refs, so it can for example `ListRefs`, which should return references, and also be built from the old `dsrefs.fbs` which is going to use dataset refs. At the same time, the `repo.Repo` interface wants to be able to return `dscache.Dscache` so that the codebase can access the cache. Since the old DatasetRef lives in `repo`, this will create a circular dependency.

The most reasonable way I see to fix this is to use `dsref` everywhere, since it's just plain-old-data and has basically no dependencies.

dustmop:
This might be too large of a change, as `repo.DatasetRef` is still using fields like `Dataset *dataset.Dataset`, and `Published bool`, and `Foreign bool` which don't exist on `dsref`. A stop gap solution may be to move `repo.DatasetRef`'s definition to `reporef.DatasetRef` inside a new package "qri/repo/ref".

b5:
Love the idea of getting this done. We introduced the notion of transient fields to `dataset.Dataset` Some time ago to make dataset a superset of dsref.

Lots of places that return repo.DatasetRef should instead return A Dataset, with the ref fields set on the dataset struct itself.

My hope is we end up in a place where we’re progressively enriching reference strings into datasets, passing through dsrefs as a lightweight layer for name resolution and listing

dustmop:
This is going to be a multi-step process. For now, I'm going to move the repo.DatasetRef definition to "qri/repo/ref/dataset_ref.go" (package reporef) as mentioned above, move ref parsing to "qri/repo/parse.go", and try to switch external users, like `List`, to start returning `dsref` instead of `Dataset.Ref` over the near future. There's a bit of messiness with `profile.ID` also mixed into places that makes the refactoring a bit complicated, so one thing at a time.

dustmop:
First part, big ol' refactoring: https://github.com/qri-io/qri/pull/1090

dustmop:
Part 2, adding `DetailedRef`, which is a ref with additional info from the dataset like `metaTitle` and `commitTime`: https://github.com/qri-io/qri/pull/1091

dustmop:
Part 3, Rename `DetailedRef` to `VersionInfo`, use in more places https://github.com/qri-io/qri/pull/1102

dustmop:
Part 4, Parse dsrefs from strings. Use for the rename and init commands: https://github.com/qri-io/qri/pull/1122

PR #1641, opened 2021-02-12
refactor(dsref.VersionInfo): merge logbook.DatasetLogItem into VersionInfo

b5:
Remove logbook.DatasetLogItem completely by adding `CommitTitle` and `CommitMessage` fields to `dsref.VersionInfo`. While we're here, add fields that commit's executed with the `--apply` flag will need to reflect transform run info.

`logbook.DatasetLogItem` is creating more trouble than it's worth. While working with @ramfox on qrimatic code it became clear that adding fields that qrimatic needs fields added to dsref.VersionInfo, and I had been planning on adding them to logbook.DatasetLogItem. By factoring these into VersionInfo, we remove a stray data structure that `lib` can return and drop a bunch of aliasing cruft.

I've also adjusted the documentation of a VersionInfo to reflect that a VersionInfo can model a failed run as well as a commit (this is important for upcoming work on adding run operations to logbook). A VersionInfo that describes a failed commit *must* have a `Path` field equal to the empty string.

One important thing to note here: the `RunDuration` and `RunStatus` fields are not stored on a `*dataset.Dataset`, which violates the principle of `VersionInfo` being a "strict subset" of a dataset.

~~BREAKING CHANGE:
fields from the http API `/history` endpoint `versionInfo` key have moved up one level~~
_my bad, this isn't a breaking change b/c `DatasetLogItem` embedded VersionInfo_


dsref/version_info.go
```diff
@@ -7,15 +7,18 @@ import (
...
+// datasets or a list of dataset versions ("qri list" and "qri log").
+//
+// VersionInfos can also describe dataset versions that are being created or
+// failed to create. In these cases the calculated VersionInfo.Path value must
+// always equal the empty string.
+//
 // If any fields are added to this struct, keep it in sync with:
 //   dscache/def.fbs       dscache
 //   dscache/fill_info.go  func fillInfoForDatasets
 //   repo/ref/convert.go   func ConvertToVersionInfo
```
b5, line 21: This was a _very_ helpful comment while refactoring 😄

lib/log.go
```diff
@@ -36,11 +37,8 @@ type LogParams struct {
 	Source string
 }
 
-// DatasetLogItem is a line item in a dataset logbook
-type DatasetLogItem = logbook.DatasetLogItem
-
 // Log returns the history of changes for a given dataset
-func (m *LogMethods) Log(params *LogParams, res *[]DatasetLogItem) error {
+func (m *LogMethods) Log(params *LogParams, res *[]dsref.VersionInfo) error {
```
b5, line 41: The big win of this PR is both `list` and `log` now actually return the same struct, which based on the comment above `VersionInfo` has always been the goal. 

Having a list of datasets & a dataset history be describable by the same thing without a lot of bending over backwards is very helpful IMHO

dsref/version_info.go
```diff
@@ -7,15 +7,18 @@ import (
...
-// VersionInfo is the way that everything talks about information about datasets in the cases
-// that an entire Dataset is not being returned. Used as an output value for many methods, such
-// as List, and Log.
+// VersionInfo is an aggregation of fields from a dataset version for caching &
+// listing purposes. VersionInfos are typically used when showing a list of
+// datasets or a list of dataset versions ("qri list" and "qri log").
+//
+// VersionInfos can also describe dataset versions that are being created or
+// failed to create. In these cases the calculated VersionInfo.Path value must
+// always equal the empty string.
```
b5, line 16: The addition of this caveat is kinda irritating. I'm of the mind that we'd clarify that subsytems like dscache and  the refstore only hold `VersionInfo`s that represent HEAD (the latest valid commit).

dustmop:
So, for a long time the intention was that LogItem was *not* the same VersionInfo. The two were combined at some point early them, then separated because VersionInfo was taking on too many roles and seemed like it might become a "ball of stuff" . That being said, if the separation is getting in the way of future development, I'm fine with combining them again.

Just want to document the history here, since we seem to be going back and forth and it would be good to capture the reasoning:

Originally, what was called DetailedRef was replacing a lot of sites that returned a repo.Ref with a Dataset pointer but only a small number of fields filled in. Background on that in this issue (which I badly need to update and/or close):
https://github.com/qri-io/qri/issues/1089

My original version, motivated by dscache, was to make a structure that contained only what was needed for the `list` command. Also, this PR mentions that `log` does some similar with its `DatasetLogItem`.
https://github.com/qri-io/qri/pull/1091

Next, changed DetailedRef to VersionInfo, which combined DatasetLogItem into VersionInfo.
https://github.com/qri-io/qri/pull/1102

More development, which includes a comment that VersionInfo is used as output for `list` and `log`.
https://github.com/qri-io/qri/pull/1107

Then, my suggestion was to bring back DatasetLogItem, by separating it from VersionInfo, since it filled a different role.
https://github.com/qri-io/qri/issues/1120

That was accomplished here:
https://github.com/qri-io/qri/pull/1187

Finally, it's unfortunate that dscache's completion has been blocked by logbook needing fixes. However, that shouldn't be a blocker on changing VersionInfo. But once we get back to dscache it will probably be useful to have this history here for reference.

dsref/version_info.go
```diff
@@ -64,13 +67,31 @@ type VersionInfo struct {
...
+	// Path value (version is present)
+	RunID string `json:"runID,omitempty"`
+	// RunStatus is a string version of the run.Status enumeration. This value will
+	// always be one of:
+	//    ""|"waiting"|"running"|"succeeded"|"failed"|"unchanged"|"skipped"
+	// it's a type a string to follow the "plain old data" pattern
+	RunStatus string `json:"runStatus,omitempty"`
+	// RunDuration is how long the run took/has currently taken in nanoseconds
+	// default value of 0 means no duration data is available
+	RunDuration int `json:"runDuration,omitempty"`
```
dustmop, line 94: Use int64 for compatibility on 32-bit systems. Otherwise any run longer than 4 seconds will break them.

dsref/version_info.go
```diff
@@ -64,13 +67,31 @@ type VersionInfo struct {
...
 	//
 	// Number of versions that the dataset has
 	NumVersions int `json:"numVersions,omitempty"`
 	// FSIPath is this dataset's link to the local filesystem if one exists
 	FSIPath string `json:"fsiPath,omitempty"`
+	//
+	// Run Fields come from applying a script during version creation
+	//
+	// RunID from either Commit.RunID, or the ID of a failed run when no
+	// Path value (version is present)
```
dustmop, line 85: I don't understand the meaning of this. Is this comment one sentence?

dsref/version_info.go
```diff
@@ -64,13 +67,31 @@ type VersionInfo struct {
...
+	RunID string `json:"runID,omitempty"`
+	// RunStatus is a string version of the run.Status enumeration. This value will
+	// always be one of:
+	//    ""|"waiting"|"running"|"succeeded"|"failed"|"unchanged"|"skipped"
+	// it's a type a string to follow the "plain old data" pattern
+	RunStatus string `json:"runStatus,omitempty"`
```
dustmop, line 91: The conversion functions copy RunID from the Commit component. Where do RunStatus and RunDuration come from? Are they not on the commit? Either way should noted on the definition here.

logbook/logbook.go
```diff
@@ -1018,22 +1018,20 @@ func (book *Book) ConstructDatasetLog(ctx context.Context, ref dsref.Ref, histor
...
-		},
+func versionInfoFromOp(ref dsref.Ref, op oplog.Op) dsref.VersionInfo {
+	return dsref.VersionInfo{
+		Username:    ref.Username,
+		ProfileID:   ref.ProfileID,
+		Name:        ref.Name,
+		Path:        op.Ref,
+		CommitTime:  time.Unix(0, op.Timestamp),
+		BodySize:    int(op.Size),
 		CommitTitle: op.Note,
```
dustmop, line 1029: Is logbook also going to be adding RunID here? I'm assuming it will, but that it requires work in some other PR.

dustmop (approved):
LGTM, except for a few comments in the VersionInfo definition.

I'm still a bit worried about VersionInfo taking on too much stuff that isn't always applicable. Specifically, it's easy to run into situations where high-level code is using a VersionInfo, but only some fields are filled in based upon where lower-down the struct was generated, which leads to lots of dynamic-type behavior that's hard to debug. Like, for example, log returns VersionInfos that don't have the same values that list has. This happened with Ref's Dataset pointer a lot back before VersionInfo. However, I'm not saying we shouldn't do this, just that it's something we should keep an eye on.

logbook/logbook.go
b5: yep!

b5:
wow thanks for taking the time to walk back through these decisions @dustmop, I absolutely hadn't noticed just how much we've gone back & fourth on this struct. Hopefully with some spec work we can arrive at a little more clarity on how it should work in the long run

PR #1642, opened 2021-02-17
fix(prove): Prove command updates local logbook as needed

dustmop:
When a user runs `prove`, the response they get back contains the original profileID and logbook data. Assign that data to their profile.


b5 (approved):
Looks great! I think we should merge as-is.

One other thing we should chat about in a separate change is adding protocol version headers to requests, which would have helped us get out of this situation in the first place, but let's save that for another round of work

PR #1646, opened 2021-02-19
refactor(dsref): replace ProfileID with InitID in refstring serialization

b5:
This is a backwards-compatible spec change to the way we parse dataset reference strings, specifically the concrete part of a ref (the bit after the `@`). The grammar change from this:
```
<concreteRef> = '@' [ <profileID> ] '/' <network> '/' <commitHash>
```
to this:
```
<concreteRef> = '@' [ <datasetID> | <profileID> ] '/' <network> '/' <commitHash>
```

The only change is replacing `profileID` with `datasetID`, where `datasetID` is a lowercase-base32 dataset identifier (initID). This change makes it possible to reference datasets by ID directly. For example:

    @jzk63pfkfefgwub7wvhl54s4inqr25ncwhmi7svahjjxzwuknrlq
    @jzk63pfkfefgwub7wvhl54s4inqr25ncwhmi7svahjjxzwuknrlq/ipfs/QmHaash

We can make this change backwards-compatible because dataset identifiers & profileIDs use different character sets. Replacing the profileID is the right decision, as the use of profileIDs in dsrefs has been inconsitent to date, and
will only get less useful as we transition toward additional collaborators on the same dataset.

### DIFF
--- a/dsref/parse.go
+++ b/dsref/parse.go
@@ -8,7 +8,7 @@ import (
 
 // These functions parse a string to create a dsref. We refer to a "human-friendly reference"
 // as one with only a username and dataset name, such as "my_user/my_dataset". A "full reference"
-// can also contain a "concrete reference", which includes an optional profileID plus a network
+// can also contain a "concrete reference", which includes an optional datasetID plus a network
 // and a "commit hash".
 //
 // Parse will parse a human-friendly reference, or only a concrete reference, or a full reference.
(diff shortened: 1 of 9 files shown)


dsref/parse.go
```diff
@@ -34,13 +34,15 @@ const (
...
 	alphaNumericDsname = `[a-zA-Z][\w-]{0,143}`
 	b58Id              = `Qm[0-9a-zA-Z]{0,44}`
+	b32LogbookID       = `[a-z2-7]{0,52}`
 )
 
 var (
 	validName      = regexp.MustCompile(`^` + alphaNumeric)
 	dsNameCheck    = regexp.MustCompile(`^` + alphaNumericDsname + `$`)
-	concretePath   = regexp.MustCompile(`^@(` + b58Id + `)?\/(` + alphaNumeric + `)\/(` + b58Id + `)`)
+	concretePath   = regexp.MustCompile(`^@(` + b32LogbookID + `|` + b58Id + `)?\/(` + alphaNumeric + `)\/(` + b58Id + `)`)
```
dustmop, line 43: Change `concretePath` to `concreteRef` to match other uses

logbook/logsync/logsync.go
```diff
@@ -25,6 +25,10 @@ var (
 	log = golog.Logger("logsync")
 )
 
+func init() {
```
dustmop, line 28: Is this empty init needed?

dustmop (approved):
LGTM.

I'm not totally convinced we should deprecate profileID just yet, but I see your point (discussed earlier today) that any resolver should have a mapping from initID to profileID that it can use to retrieve one from the other. I might suggest softening the language (such as `TODO: Possibly remove profileID from this struct once there are no more users`), but otherwise it's fine. Agree that parsing and stringifying with `initID` instead of `profileID` is completely doable now.

logbook/logsync/logsync.go
b5: nope. will remove

PR #1647, opened 2021-02-21
refactor package hierarchy

b5:
rearrange a few packages & some general housekeeping

* split all "token" logic out of the access package into a new `token` package, nested in an `auth` directory
* Move `key` into the `auth` directory
* Move `access` under the `remote` package. With tokens factored out the only remaining code is an access-control-list tool that should only be used by remotes
* move `watchfs` under the `fsi` package


dustmop (approved):
LGTM, just realized this is a subset of https://github.com/qri-io/qri/pull/1648 and should have been reviewed first.

PR #1648, opened 2021-02-21
remove profile methods from repo

b5:
builds on #1647

centralize all repo access around the `Profiles()` method that returns a profile store


dustmop (approved):
LGTM, fine with all the movements and name changes.

PR #1649, opened 2021-02-22
refactor(repo): remove all external callers to CanonicalizeDatasetRef & CanonicalizeProfile, un-export

b5:


cmd/test_runner_test.go
```diff
@@ -421,25 +420,30 @@ func (run *TestRunner) LookupVersionInfo(t *testing.T, refStr string) *dsref.Ver
 }
 
 // ClearFSIPath clears the FSIPath for a reference in the refstore
+// Note: ClearFSIPAth doesn't do reference resoultion, cannot use "me" in
```
dustmop, line 423: ClearFSIPAth -> ClearFSIPath, resoultion -> resolution. Also, big fan of this change and the clarifying comment.

lib/datasets.go
```diff
@@ -1100,31 +1094,36 @@ func (m *DatasetMethods) Remove(ctx context.Context, p *RemoveParams) (*RemoveRe
...
 	if err != nil {
-		return nil, err
-	}
-
-	if canonErr := repo.CanonicalizeDatasetRef(ctx, m.inst.repo, &ref); canonErr != nil && canonErr != repo.ErrNoHistory {
-		log.Debugf("Remove, repo.CanonicalizeDatasetRef failed, error: %s", canonErr)
-		if p.Force {
-			didRemove, _ := base.RemoveEntireDataset(ctx, m.inst.repo, reporef.ConvertToDsref(ref), []dsref.VersionInfo{})
+		log.Debugw("Remove, repo.ParseAndResolveRefWithWorkingDir failed", "err", err)
+		// TODO(b5): this "logbook.ErrNotFound" is needed to get cmd.TestRemoveEvenIfLogbookGone
```
dustmop, line 1100: This behavior is pretty old, it was implemented in https://github.com/qri-io/qri/commit/fe4598abaa9f4427f05ae8a046e9dcc1fc9a318e. Back then, adding a dataset using `qri add` would not write anything to the logbook, which made it impossible to remove datasets from the refstore. That's not the case anymore, so I think it's no longer necessary to bend over backwards to support this out-of-date functionality.

lib/datasets_test.go
```diff
@@ -797,7 +797,7 @@ func TestRenameNoHistory(t *testing.T) {
 	workDir := tr.CreateAndChdirToWorkDir("remove_no_history")
```
dustmop, line 797: Change this directory name as well, to match.

lib/fsi.go
```diff
@@ -240,32 +239,32 @@ func (m *FSIMethods) Write(p *FSIWriteParams, res *[]StatusItem) (err error) {
...
-	if err != nil {
-		return err
-	}
 	if p.Ds == nil {
 		return fmt.Errorf("dataset is required")
 	}
 
-	datasetRef := reporef.RefFromDsref(ref)
-	err = repo.CanonicalizeDatasetRef(ctx, m.inst.node.Repo, &datasetRef)
+	ref, _, err := m.inst.ParseAndResolveRef(ctx, p.Ref, "local")
```
dustmop, line 246: Before, this code was using `ParseHumanFriendly`, which guarantees that the ref is only the username/dsname, like "dustmop/my_dataset" and has no path. I don't think `ParseAndResolveRef` does the same? Meaning that path would be silently ignored instead of producing an error.

dustmop (approved):
LGTM!

PR #1652, opened 2021-03-01
refactor(key): merge gen package into key

b5:
Starting into a bunch of cleanup of profile/ key/ token packages. This moves the `CryptoGenerator` & marks up the things that should be changed in follow-ups


auth/key/key.go
```diff
@@ -2,9 +2,19 @@ package key
...
 
 var log = logger.Logger("key")
 
 // ID is the key identifier
-type ID string
+type ID = peer.ID
+
+// DecodeID parses an ID string
+func DecodeID(s string) (ID, error) {
+	pid, err := peer.Decode(s)
```
Arqu, line 15: Quick note that this is not strictly a peerID elsewhere and we already use pseudo peerIDs in some places. The initial implementation deliberately left out peerIDs and hence us having our own keybook implementation among other reasons.

Arqu (approved):
Remove `lib/logbook.qfb` seems like we have a leaky test somewhere since recently.
Other than that a quick note and a big LGTM things are getting real tidy these days :)

auth/key/key.go
b5: Totally, I'm mainly just piggybacking off the peer.Decode method, and explicitly coercing it to another type so we can refactor under the hood later. I'll add a note

PR #1667, opened 2021-03-04
feat: "access token" command, dispatch session support

b5:
working in bringing users to dispatch scopes


b5:
@Arqu I'd love a mid-way review of this when you get the chance!

api/middleware.go
```diff
@@ -96,3 +98,33 @@ func stripServerSideQueryParams(r *http.Request) {
...
+const (
+	bearerPrefix        = "Bearer "
+	authorizationHeader = "authorization"
+)
+
+// OAuthTokenMiddleware parses any "authorization" header containing a Bearer
+// token & adds it to the request context
+func OAuthTokenMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		reqToken := r.Header.Get(authorizationHeader)
```
Arqu, line 111: Can we add a TODO to extract this functionality and re-use where needed, by now we have several code bases that are doing 99% the same as here.

lib/scope.go
```diff
@@ -20,9 +22,46 @@ import (
...
+				return scope{}, err
+			}
+
+			if claims, ok := tok.Claims.(*token.Claims); ok {
+				// TODO(b5): at this point we have a valid signature of a profileID string
+				// but no proof that this profile is owned by the key that signed the
+				// token. We either need ProfileID == KeyID, or we need a UCAN. we need to
+				// check for those, ideally in a method within the profile package that
+				// abstracts over profile & key agreement
+				pro, err = inst.profiles.GetProfile(profile.IDB58DecodeOrEmpty(claims.ProfileID))
```
Arqu, line 46: 👍 UCAN would be great as a longer term expansion but think we can get by with ProfileID == KeyID for the time being, given that the keys need to be added from the owner side similar to a whitelisting process.

auth/token/token.go
```diff
@@ -36,14 +38,91 @@ type Token = jwt.Token
 // Claims is a JWT Claims object
 type Claims struct {
 	*jwt.StandardClaims
-	Username string `json:"username"`
+	ProfileID string `json:"profileID"`
```
Arqu, line 41: 👍

lib/access.go
```diff
@@ -0,0 +1,93 @@
...
+			return "", err
+		}
+	}
+
+	pk := grantee.PrivKey
+	if pk == nil {
+		return "", fmt.Errorf("cannot create token for %q (id: %s), private key is required", grantee.Peername, grantee.ID.String())
+	}
+
+	return token.NewPrivKeyAuthToken(pk, p.TTL)
```
Arqu, line 92: We don't really have a way to override the TTL here. Also if we move to UCAN this becomes a lot more complex but that's a discussion for another day.

lib/dispatch.go
```diff
@@ -30,6 +39,22 @@ func (inst *Instance) Dispatch(ctx context.Context, method string, param interfa
 	// If the http rpc layer is engaged, use it to dispatch methods
 	// This happens when another process is running `qri connect`
 	if inst.http != nil {
+		if tok := token.FromCtx(ctx); tok == "" {
+			// If no token exists, create one from configured profile private key &
+			// add it to the request context
+			// TODO(b5): we're falling back to the configured user to make requests,
+			// is this the right default?
+			p, err := profile.NewProfile(inst.cfg.Profile)
```
Arqu, line 47: I'm starting to really like this dispatch setup we have now.
As for the default I'd say yes, the owner is responsible for it's co-profiles in most normal use cases. 
The only alternative being to fail hard and fast here to prevent unintended user switcheroos.

auth/token/token.go
```diff
@@ -36,14 +38,91 @@ type Token = jwt.Token
...
+
+	var exp int64
+	if ttl != time.Duration(0) {
+		exp = Timestamp().Add(ttl).In(time.UTC).Unix()
+	}
+
+	// set our claims
+	t.Claims = &Claims{
+		StandardClaims: &jwt.StandardClaims{
+			Issuer:  id,
```
Arqu, line 81: These are not completely compatible with the cloud side so far, but I'm happy if we have this work for node to node comms before aligning those two.

Arqu (commented):
Left some commentary and notes.
Also `lib/logbook.qfb`...

cmd/access.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"context"
+
+	"github.com/qri-io/ioes"
+	"github.com/qri-io/qri/lib"
+	"github.com/spf13/cobra"
+)
+
+// NewAccessCommand creates a new `qri apply` cobra command for applying transformations
```
dustmop, line 11: Change to "NewAccessCommand creates a new `qri access` cobra command for managing permissions"

cmd/access.go
```diff
@@ -0,0 +1,79 @@
...
+	return nil
+}
+
+// CreateAccessToken constructs an access token suitable for making
+// authenticated requests
+func (o *AccessOptions) CreateAccessToken(ctx context.Context) error {
+	p := &lib.CreateAuthTokenParams{
+		GranteeUsername: o.GranteeUsername,
+	}
+	token, err := o.Instance.Access().CreateAuthToken(ctx, p)
```
dustmop, line 72: 👍

cmd/access_test.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"testing"
+)
+
+func TestAccessCreateToken(t *testing.T) {
+	run := NewTestRunner(t, "peer", "cmd_test_create_access_token")
+	defer run.Delete()
+
+	run.MustExecCombinedOutErr(t, "qri access token --for me")
```
dustmop, line 11: These aren't using the result of stdout or stderr. Just use `run.MustExec` instead.

lib/access.go
```diff
@@ -0,0 +1,93 @@
...
+
+// SetNonZeroDefaults uses default token time-to-live if one isn't set
+func (p *CreateAuthTokenParams) SetNonZeroDefaults() {
+	if p.TTL == 0 {
+		p.TTL = token.DefaultTokenTTL
+	}
+}
+
+// Valid checks if the profile in question is valid
+func (p *CreateAuthTokenParams) Valid() error {
```
dustmop, line 42: Small nitpick: `Valid` reads to me like a yes/no question, ie something that would return a `bool`. Could we rename this method to `Validate` instead? There's a bunch of other places in the codebase that use this name for methods that return an error on invalid structs, which feels like a good pattern to keep.

lib/scope.go
```diff
@@ -20,9 +22,46 @@ import (
...
 	ctx  context.Context
 	inst *Instance
+	pro  *profile.Profile
 	// TODO(dustmop): Additional information, such as user identity, their profile, keys
 }
 
+func newScope(ctx context.Context, inst *Instance) (scope, error) {
+	var pro *profile.Profile
+	// TODO(b5): shouldn't need this nil checking
+	if inst != nil && inst.profiles != nil {
```
dustmop, line 32: Could we move this part into its own helper function? Would suggest `pro := activeProfile(inst)`. It'll be nice to keep the `newScope` constructor itself short and clear, as I can only imagine it's going to be adding much more functionality soon.

lib/scope.go
dustmop: I'll test out the qri-logbook-integration tests and see how this interacts with `prove` changing the profileID. Last I left off cloud was *not* sending back UCAN, but that may be necessary to keep this sane.

profile/profile.go
```diff
@@ -198,7 +198,7 @@ func (p *Profile) ValidOwnerProfile() error {
 // to the profile ID if none is present
 func (p *Profile) GetKeyID() key.ID {
 	if p.KeyID == key.ID("") {
-		p.KeyID = key.ID(p.ID.String())
+		p.KeyID = key.ID(string(p.ID))
```
dustmop, line 201: Am I wrong or is this a change in functionality? Looking at https://github.com/libp2p/go-libp2p-core/blob/master/peer/peer.go#L56, `peer.ID.String()` calls `Pretty()` which calls `IDB58Encode`. However, I'm pretty sure `string(p.ID)` just coerces the bytes of the id to be a string without any additional encoding. Would love to have a unit test specifically on this function to prove me wrong/right.

profile/store_test.go
```diff
@@ -45,7 +46,12 @@ func TestPutProfileWithAddresses(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	ps, err := NewLocalStore(filepath.Join(path, "profiles.json"), &Profile{PrivKey: kd0.PrivKey, Peername: "user"}, ks)
+	owner := &Profile{
+		ID:       ID(kd0.PeerID),
```
dustmop, line 50: As per the description of https://github.com/qri-io/qri/pull/1121, don't call `profile.ID` directly. Instead, use `IDFromPeerID`.

profile/store_test.go
```diff
@@ -84,7 +100,12 @@ func TestProfilesWithKeys(t *testing.T) {
 		t.Errorf("error creating tmp directory: %s", err.Error())
 	}
 
-	ps, err := NewLocalStore(filepath.Join(path, "profiles.json"), &Profile{PrivKey: kd0.PrivKey, Peername: "user"}, ks)
+	owner := &Profile{
+		ID:       ID(kd0.PeerID),
```
dustmop, line 104: Same comment here.

dscache/dscache_test.go
```diff
@@ -97,7 +98,7 @@ func TestResolveRef(t *testing.T) {
 
 	dsrefspec.AssertResolverSpec(t, dsc, func(r dsref.Ref, author profile.Author, _ *oplog.Log) error {
 		builder := NewBuilder()
-		pid, err := profile.KeyIDFromPub(author.AuthorPubKey())
+		pid, err := key.IDFromPubKey(author.AuthorPubKey())
```
dustmop, line 101: Since this is explicitly now a keyID and not related to profileIDs, let's rename `pid` to `kid`.

dsref/mem_resolver_test.go
```diff
@@ -19,7 +20,7 @@ func TestMemResolver(t *testing.T) {
 	}
 
 	dsrefspec.AssertResolverSpec(t, m, func(ref dsref.Ref, author profile.Author, log *oplog.Log) error {
-		pid, err := profile.KeyIDFromPub(author.AuthorPubKey())
+		pid, err := key.IDFromPubKey(author.AuthorPubKey())
```
dustmop, line 23: Same here, `pid` -> `kid`.

p2p/resolve_ref_test.go
```diff
@@ -125,7 +126,7 @@ func TestResolveRef(t *testing.T) {
 
 	dsrefspec.AssertResolverSpec(t, p2pRefResolver, func(r dsref.Ref, author profile.Author, _ *oplog.Log) error {
 		builder := dscache.NewBuilder()
-		pid, err := profile.KeyIDFromPub(author.AuthorPubKey())
+		pid, err := key.IDFromPubKey(author.AuthorPubKey())
```
dustmop, line 129: And here as well, `pid` -> `kid`.

lib/fsi.go
```diff
@@ -19,7 +19,7 @@ import (
 
 // FSIMethods groups together methods for FSI
 type FSIMethods struct {
-	inst *Instance
+	d dispatcher
```
dustmop, line 22: Sure, seems fine. All these Methods should be doing is just dispatch.

cmd/access.go
b5: yeah, it's much better

lib/access.go
b5: ah totally agreed. Will change

profile/store_test.go
b5: lovely. There are a bunch of direct `ID() type-coercions in this file (looks like #1121 grepped for only calls _outside_ the  profile package) will fix.

profile/profile.go
b5: Yes I totally agree a proper set of tests are needed. This is actually fixing a bug introduced earlier, as key.ID has become an alias for peer.ID. I plan to fix _all_ of this & add tests in a follow-up

b5: I should add that this PR does help somewhat on this frontier by requiring that keys added adding a key to the keystore now **requires** a match between the ID & public key, so if the id is malformed, it'll fail.

dustmop: Ah, you're saying this *was* using encode before, but that was a bug, and it should only be coercing the bytes? Ok, that makes sense. Unfortunate how it's hard to tell at the moment.

lib/access.go
```diff
@@ -66,6 +66,12 @@ func (accessImpl) CreateAuthToken(scp scope, p *CreateAuthTokenParams) (string,
 		grantee *profile.Profile
 		err     error
 	)
+	// TODO(b5): it'd be great if dispatch checked if input params implemented
+	// a "validator" interface & auto-called validate *before* dispatching.
+	// would save an HTTP call in RPC-style contexts.
```
dustmop, line 71: This is a wonderful idea.

dustmop (approved):
LGTM!

Issue #1657, opened 2021-03-02
http rpc error handling


dustmop:
The http rpc implementation in lib/http.go, specifically the `do` function, could use some improvement in its error handling.

1) If the rpc endpoint doesn't exist, the http call gets a 404, with the body being "404 page not found". Then the body tries to Unmarshal, which gets the error "invalid character 'p' after top-level value", swallowing the fact that a 404 was hit. The code never reaches `checkError`.
2) It's hard to tell right now if an error is caused by our outgoing http request failing versus an error happening on the foreign handler.
3) If an error comes back from the rpc call, we wrap it in text starting with "HTTPClient req error". It could be nice to instead unwrap the error from the request, such that rpc calls display errors exactly the same as local calls.

dustmop:
Fixed by #1668

PR #1668, opened 2021-03-04
fix(http): fix httpClient error checking

Arqu:
Addresses https://github.com/qri-io/qri/issues/1657


dustmop (approved):
Tried it out with `qri checkout` and it works great! Really happy to see standard looking errors being unwrapped from the http rpc.

PR #1669, opened 2021-03-04
refactor(changes): RPC > HTTP & API for change reports

Arqu:
The breaking change is in the rename of the json fields.


lib/changes.go
```diff
@@ -9,41 +9,39 @@ import (
 
 // ChangeReportParams defines parameters for diffing two sources
 type ChangeReportParams struct {
-	LeftRefstr  string `json:"left"`
-	RightRefstr string `json:"right"`
+	LeftRefstr  string `schema:"left_ref",json:"left_ref"`
+	RightRefstr string `schema:"right_ref",json:"right_ref"`
```
b5, line 13: hrm, I'm torn on this. URL query params want `under_scores`, our JSON convention uses `lowerSnakeCase`. I know these are query params, but I'd prefer they be consistent.

Possible to harmonize around lower snake case? `leftRef` and `rightRef`?

dustmop: Just to be clear on the standard terminology, `this_is_snake_case` and `thisIsCamelCase`. I have no strong feelings on the matter, I'm ok with whatever we end up on.

dustmop (approved):
LGTM

lib/changes.go
dustmop: Also, circleci is failing between it looks like there needs to be a space after the comma and before `json:`:

```
lib/changes.go:12:2: struct field tag `schema:"left_ref",json:"left_ref"` not compatible with reflect.StructTag.Get: key:"value" pairs not separated by spaces
lib/changes.go:13:2: struct field tag `schema:"right_ref",json:"right_ref"` not compatible with reflect.StructTag.Get: key:"value" pairs not separated by spaces
```

PR #1671, opened 2021-03-04
refactor(config): RPC > HTTP & API for config

Arqu:

### DIFF
--- a/lib/config.go
+++ b/lib/config.go
@@ -2,6 +2,7 @@ package lib
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"strings"
@@ -34,14 +35,15 @@ type GetConfigParams struct {
 
 // GetConfig returns the Config, or one of the specified fields of the Config,
 // as a slice of bytes the bytes can be formatted as json, concise json, or yaml
-func (m *ConfigMethods) GetConfig(p *GetConfigParams, res *[]byte) (err error) {
-	if m.inst.rpc != nil {
-		return checkRPCError(m.inst.rpc.Call("ConfigMethods.GetConfig", p, res))
+func (m *ConfigMethods) GetConfig(ctx context.Context, p *GetConfigParams) ([]byte, error) {
+	if m.inst.http != nil {
+		return nil, ErrUnsupportedRPC
 	}
 
 	var (
 		cfg    = m.inst.cfg
 		encode interface{}
+		err    error
 	)
 
 	if !p.WithPrivateKey {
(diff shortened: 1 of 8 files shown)


lib/config.go
```diff
@@ -55,37 +57,40 @@ func (m *ConfigMethods) GetConfig(p *GetConfigParams, res *[]byte) (err error) {
 	if p.Field != "" {
 		encode, err = cfg.Get(p.Field)
 		if err != nil {
-			return fmt.Errorf("error getting %s from config: %s", p.Field, err)
+			return nil, fmt.Errorf("error getting %s from config: %s", p.Field, err)
```
b5, line 60: gotta wrap them errors:

```suggestion
			return nil, fmt.Errorf("error getting %s from config: %w", p.Field, err)
```

b5 (approved):
one nit then LGTM

lib/config.go
```diff
@@ -34,14 +35,15 @@ type GetConfigParams struct {
 
 // GetConfig returns the Config, or one of the specified fields of the Config,
 // as a slice of bytes the bytes can be formatted as json, concise json, or yaml
-func (m *ConfigMethods) GetConfig(p *GetConfigParams, res *[]byte) (err error) {
-	if m.inst.rpc != nil {
-		return checkRPCError(m.inst.rpc.Call("ConfigMethods.GetConfig", p, res))
+func (m *ConfigMethods) GetConfig(ctx context.Context, p *GetConfigParams) ([]byte, error) {
+	if m.inst.http != nil {
+		return nil, fmt.Errorf("config is not supported across nodes")
```
b5, line 40: on second thought, this error is the one users will when they try to `qri config get` while running `qri connect` somewhere else. We should make this a standard error `ErrUnsupportedRPC` so we can capture it with `errors.Is` & present a clearer error for CLI users ("this could mean you're running qri connect in another terminal")

b5 (changes requested):

lib/lib_test.go
```diff
@@ -187,7 +187,7 @@ func TestReceivers(t *testing.T) {
 	inst := &Instance{node: node, cfg: cfg}
 
 	reqs := Receivers(inst)
-	expect := 10
+	expect := 9
```
Arqu, line 190: Before merging, need to rebase and drop this to 8.

dustmop (approved):
LGTM. We don't need config for Desktop, is that right?

Arqu:
> LGTM. We don't need config for Desktop, is that right?

AFAIK No just checked the codebase real quick, also Desktop communicates API only.

PR #1672, opened 2021-03-04
refactor(diff): RPC > HTTP & API for diff

Arqu:


lib/diff.go
```diff
@@ -26,7 +26,8 @@ type DiffStat = deepdiff.Stats
 // LeftSide set with the UseLeftPrevVersion flag.
 type DiffParams struct {
 	// File paths or reference to datasets
-	LeftSide, RightSide string
+	LeftSide  string `schema:"left_path",json:"left_path"`
+	RightSide string `schema:"right_path",json:"right_path"`
```
dustmop, line 30: Need spaces between the comma and "json" to make `go vet` happy. Also, sounds like we're aiming for camelCase instead of snake_case.

dustmop (approved):
LGTM, just need to fix the schema/json struct tags.

PR #1682, opened 2021-03-05
lib.NewRequestHandler removes boilerplate from api package

b5:
building on ideas from #1650, this adds the first method where the HTTP handler is generated by package lib, but mapped to an endpoint by package api.

This PR is mainly to establish the patten & talk through repercussions. Right now this only works for formats where the response format is well known / singular.

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -251,7 +251,6 @@ func TestServerReadOnlyRoutes(t *testing.T) {
 		{"POST", "/registry/profile/new", 403},
 		{"POST", "/registry/profile/prove", 403},
 		{"GET", "/checkout/", 403},
-		{"GET", "/status/", 403},
 		{"GET", "/init/", 403},
 
 		// active endpoints:
--- a/lib/api.go
+++ b/lib/api.go
@@ -107,7 +107,7 @@ const (
 	// fsi endpoints
 
 	// AEStatus returns the filesystem dataset status
-	AEStatus = APIEndpoint("/status/{path:.*}")
+	AEStatus = APIEndpoint("/status")
 	// AEWhatChanged returns what changed for a specific commit
 	AEWhatChanged = APIEndpoint("/whatchanged/{path:.*}")
 	// AEInit invokes a dataset initialization on the filesystem
--- a/lib/http.go
+++ b/lib/http.go
@@ -200,11 +200,11 @@ func (c HTTPClient) checkError(res *http.Response, body []byte, raw bool) error 
 
 // NewHTTPRequestHandler creates a JSON-API endpoint for a registered dispatch
 // method
-func NewHTTPRequestHandler(inst *Instance, method string, ae APIEndpoint) http.HandlerFunc {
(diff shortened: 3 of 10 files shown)


api/api_test.go
```diff
@@ -251,7 +251,7 @@ func TestServerReadOnlyRoutes(t *testing.T) {
 		{"POST", "/registry/profile/new", 403},
 		{"POST", "/registry/profile/prove", 403},
 		{"GET", "/checkout/", 403},
-		{"GET", "/status/", 403},
+		// {"GET", "/status/", 403},
```
Arqu, line 254: This should work with `{"GET", "/status", 403},` ie remove the trailing `/` as gorilla is strict about slashes by default.

lib/http.go
```diff
@@ -182,3 +196,92 @@ func (c HTTPClient) checkError(res *http.Response, body []byte, raw bool) error
...
+		}
+
+		apiutil.WriteResponse(w, res)
+	}
+}
+
+// UnmarshalParams deserialzes a lib req params stuct pointer from an HTTP
+// request
+//
+// Deprecated: This is only exported for one-off handlers in the API package
```
Arqu, line 234: 👍

lib/http.go
```diff
@@ -182,3 +196,92 @@ func (c HTTPClient) checkError(res *http.Response, body []byte, raw bool) error
...
+func NewHTTPRequestHandler(inst *Instance, libMethod string) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		p := inst.NewInputParam(libMethod)
+		if p == nil {
+			apiutil.WriteErrResponse(w, http.StatusBadRequest, fmt.Errorf("no params for method %s", libMethod))
+			return
+		}
+
+		if err := UnmarshalParams(r, p); err != nil {
+			log.Debugw("unmarshal dataset save error", "err", err)
```
Arqu, line 211: This seems like an odd message, maybe `unmarshal request params error`

Arqu (commented):
Aside from some nits, LGTM but will leave to @dustmop to take a look at parts relevant to him.

But very smooth 🍸

dustmop (approved):
Implementation seems totally agreeable to me.

I have a suggestion for how to handle some of the outstanding questions, such as how to correlate endpoints to methods, and how to determine which http verb to use: https://github.com/qri-io/qri/commit/963e650910fbad360460e359b0c56613c730c5d2

lib/api.go
```diff
@@ -107,7 +107,7 @@ const (
 	// fsi endpoints
 
 	// AEStatus returns the filesystem dataset status
-	AEStatus = APIEndpoint("/status/{path:.*}")
+	AEStatus = APIEndpoint("/status")
```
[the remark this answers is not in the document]
Arqu: It can, but the route expansion is handled by `handleRefRouteMethods` in `api/api.go` where it registers all the variants, this basically just represents the "route prefix" as we used to carry around.

Issue #1675, opened 2021-03-04
RPC > HTTP Refactoring


Arqu:
This is a tracker issue to follow the progress on the refactoring of our lib methods to move over from RPC to HTTP.

Method Chunks (`lib/___.go`):
- [x] config
- [x] changes
- [x] datasets
- [x] diff
- [x] log
- [x] peers
- [x] profile
- [x] registry
- [x] remote
- [x] render
- [x] search
- [x] sql
- [x] remove rpc

Arqu:
🎉

Issue #1679, opened 2021-03-05
use dispatch for lib.ProfileMethods


b5:
convert profile methods to use dispatch

Arqu:
caught up on converting them to RPC first

b5:
ah good point. I'll wait on that!

ramfox:
closed by #1736

PR #1684, opened 2021-03-05
refactor(profile): RPC > HTTP & API for profile

Arqu:

Removed the setter paths as we don't want to allow API/RPC access to any methods that change config.
On the other hand `cmd` doesn't seem to have profile methods except through config to get/set the values.


api/profile.go
```diff
@@ -54,123 +49,60 @@ func (h *ProfileHandlers) getProfileHandler(w http.ResponseWriter, r *http.Reque
...
-		util.WriteErrResponse(w, http.StatusInternalServerError, err)
+	res, err := h.ProfilePhoto(r.Context(), params)
+	if err != nil {
+		log.Infof("error getting profile photo: %s", err.Error())
+		util.WriteErrResponse(w, http.StatusBadRequest, err)
 		return
 	}
-	util.WriteResponse(w, res)
+
+	w.Header().Set("Content-Type", "image/jpeg")
```
dustmop, line 77: I wonder if we're going to soon want some like UnmarshalParams but for outgoing results. Some kind of hook so that methods can add additional outgoing information.

dustmop (approved):
LGTM

api/profile.go
Arqu: Yeah, had similar thought, but think we're still need to refactor on step deeper as several methods conflate different types of output and that feels like we would just move bloat around. If we were to go that route, I'd first like to separate purely JSON APIs from raw data endpoints and work from there.

Arqu:
@b5 I'm going to take a second and wait to see if you have any comments on this before merging. Other than that https://github.com/qri-io/qri/issues/1679 should be unblocked now.

b5 (approved):
LGTM!

PR #1685, opened 2021-03-08
replace FSIHandler with lib.HTTPRequestHandler

b5:


api/api.go
```diff
@@ -221,16 +222,16 @@ func NewServerRoutes(s Server) *mux.Router {
...
-	m.Handle(lib.AEInit.String(), s.Middleware(fsih.InitHandler(lib.AEInit.NoTrailingSlash())))
-	m.Handle(lib.AECanInitDatasetWorkDir.String(), s.Middleware(fsih.CanInitDatasetWorkDirHandler(lib.AECanInitDatasetWorkDir.NoTrailingSlash())))
-	m.Handle(lib.AECheckout.String(), s.Middleware(fsih.CheckoutHandler(lib.AECheckout.NoTrailingSlash())))
-	m.Handle(lib.AERestore.String(), s.Middleware(fsih.RestoreHandler(lib.AERestore.NoTrailingSlash())))
-	m.Handle(lib.AEFSIWrite.String(), s.Middleware(fsih.WriteHandler(lib.AEFSIWrite.NoTrailingSlash())))
-	m.Handle(lib.AEFSICreateLink.String(), s.Middleware(fsih.CreateLinkHandler(lib.AEFSICreateLink.NoTrailingSlash())))
-	m.Handle(lib.AEFSIUnlink.String(), s.Middleware(fsih.UnlinkHandler(lib.AEFSIUnlink.NoTrailingSlash())))
+	handleRefRouteMethods(m, lib.AEWhatChanged, s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "fsi.whatchanged")), http.MethodGet, http.MethodPost)
+	m.Handle(lib.AEInit.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "fsi.init"))).Methods(http.MethodPost)
+	m.Handle(fmt.Sprintf("%s/{peername}/{name}", lib.AEInit), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "fsi.init"))).Methods(http.MethodPost)
```
Arqu, line 228: Tbh we abused that handleRef route a bit more than needed as we tack on that {selector} which is not really used in all that many endpoints. Yet we used it several times just to fill this usecase.
Think we should have this formatting extracted as a function and  have the inverse be the exception.
Happy to do that in one of my PRs, fine with it now (also getting that here specifically we don't want a full path passed on)

api/middleware.go
```diff
@@ -58,6 +58,24 @@ func (s *Server) addCORSHeaders(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// muxVarsToQueryParamMiddleware moves all mux variables to query parameter
+// values, failing with an error if a name collision with user-provided query
+// params occurs
+func muxVarsToQueryParamMiddleware(next http.Handler) http.Handler {
```
Arqu, line 64: Am not a huge fan of this. I know this gets us around having to `POST` everything. However, encoding everything and anything into the URL has it's own issues. Aside from some fringe formatting stuff that might come up I've personally hit issues with URLs being too long for servers/browsers/libs etc. Usually up to 1-2k chars is fine above that it highly depends on the implementation.

On the flip side I don't really have a cleaner way to avoid having to POST every request or add weird paths on verb/path collisions.

Arqu (approved):
I have some opinions, but don't have real solutions on some points.
LGTM

api/middleware.go
Arqu: After more consideration I have an observation and a consideration.
Given that this only encodes `mux` vars into the query path, length shouldn't be an issue unless the path in the first place was ridiculously long. As such I'm fine with keeping it as is.

On the other hand, I'm not sure what the utility of this is, because accessing those requires access to `http.Request` anyways, so might as well parse it out with `mux.Vars(r)` as the reverse doesn't hold (ie we dont map query params into mux vars) and have to do it anyways. The only case where I see the benefit of this if we use those same params in the `lib/http.go` client and send them over BUT in that case we need to know whether to parse from one or the other or have a param encoding that has some prefix and be able to reconstruct the original path. 

So unless we can also reverse query params back to mux vars this does not absolve us from having to `POST` params for typical `GET` routes.

b5: I'm a fan of this approach because it takes us from 3 places to supply param fields to 2. `mux variables, r.FormValue(), and r.Body` ->  `r.FormValue(), r.Body`. I see the mux.Router as an implementation detail the `api` package should care about, but the `lib` package shouldn't know about.

mapping to query params also gets us around `POST` only when used in conjunction with `lib.UnmarshalParams`, which uses the gorilla schema decoder. eg for 
* `GET http://localhost:2503/get/b5/world_bank_population` 
* muxVarsToQueryParamMiddleware will turn this into  `GET http://localhost:2503/save/b5/world_bank_population?peername=b5&name=world_bank_population`
* `api.refStringMiddleware` will map this to `refstring`, effectively giving us ``GET http://localhost:2503/save/b5/world_bank_population?peername=b5&name=world_bank_population&refstring=b5/world_bank_population`
* the gorilla schema decoder will map this to `lib.GetParams`, which should have a `refstring` field, making the GET request work

I think the approach is right, but the execution needs work. My suggestion would be to work on combining all middleware that maps mux variables to query params into one function, but I do think the approach of keeping the whole mux variable conversation in the `api` package is the right approach

PR #1687, opened 2021-03-08
wip(dispatch): Add Mapping() method which holds endpoint and verb

dustmop:


dustmop:
Incomplete, so you'll have to use some imagination, but the basic idea is that MethodSets declare their own "metadata" so that dispatch knows how to map each method. TODO: The registration sequence builds up the list of endpoints and registers them as http handlers.

dustmop:
Looking for feedback on the plan here, especially the `Mapping` method on each methodSet as a way to map functions to api endpoints.

lib/dispatch.go
```diff
@@ -304,13 +312,33 @@ func (inst *Instance) registerOne(ourName string, methods MethodSet, impl interf
...
+			panic(fmt.Sprintf("%s: not found in Mapping()", funcName))
+		}
+		if len(methodAttrs) != 2 {
+			panic(fmt.Sprintf("%s: incorrect number of attributes: %d", len(methodAttrs)))
+		}
+		endpointURL := methodAttrs[0]
+		if !strings.HasPrefix(endpointURL, "/") {
+			panic(fmt.Sprintf("%s: endpoint URL must start with /, got %q", endpointURL))
+		}
+		httpVerb := methodAttrs[1]
```
Arqu, line 328: PUT should be allowed to as we do use it.

lib/fsi.go
```diff
@@ -27,6 +27,22 @@ func (m *FSIMethods) Name() string {
 	return "fsi"
 }
 
+// Mapping is FIXME
+func (m *FSIMethods) Mapping() map[string][]string {
+	return map[string][]string{
```
Arqu, line 32: As hard coded as it is, I don't see a better way around it. Hence having it centralized is great for the future 👍

Arqu (commented):
Definitely in favor of a central location to manage these.

lib/fsi.go
dustmop: Sounds good. One thing to keep in mind, thinking far into the future: I've been envisioning this mapping more as a way to keep "metadata" about the methods in question. For example, maybe at some point we also add some kind of permissions information, or multi-processing info (is this method same to call in parallel with others?), so it may expand to be about more than just HTTP. Totally possible that I'm planning too far ahead with that idea, so I'll switch to your suggestion, but we may need to reevaluate in the future.

b5: ah ok, then yes let's drop the `HTTP`, but that's also  more reason to use a struct instead of a string slice

b5:
superseded by #1711

PR #1689, opened 2021-03-08
refactor(changes): use `Dispatch` with `lib.ChangeReport` function

ramfox:
Lays groundwork for other upcoming `DatasetMethods` method conversions by creating `datasetImpl` struct, `Name` method (needed to satisfy the `MethodSet` interface), and all corresponding `datasetImpl` methods

Fixes a bug in `changes` that occurred when only `rightRef` was present.

### DIFF
--- a/api/datasets.go
+++ b/api/datasets.go
@@ -23,10 +23,6 @@ import (
 
 // DatasetHandlers wraps a requests struct to interface with http.HandlerFunc
 type DatasetHandlers struct {
-	// TODO (ramfox): start here
-	// should have a dsm for dataset methods
-	// and an inst
-	// future refactors should have handlers rely on inst for node and repo
 	lib.DatasetMethods
 	inst     *lib.Instance
 	remote   *lib.RemoteMethods
--- a/api/datasets.go
+++ b/api/datasets.go
@@ -15,9 +15,7 @@ import (
 	"github.com/qri-io/qri/base/archive"
 	"github.com/qri-io/qri/dsref"
 	"github.com/qri-io/qri/lib"
-	"github.com/qri-io/qri/p2p"
 	"github.com/qri-io/qri/profile"
-	"github.com/qri-io/qri/repo"
 	reporef "github.com/qri-io/qri/repo/ref"
 )
 
@@ -26,8 +24,6 @@ type DatasetHandlers struct {
 	lib.DatasetMethods
 	inst     *lib.Instance
 	remote   *lib.RemoteMethods
-	node     *p2p.QriNode
-	repo     repo.Repo
 	ReadOnly bool
 }
 
@@ -35,7 +31,7 @@ type DatasetHandlers struct {
 func NewDatasetHandlers(inst *lib.Instance, readOnly bool) *DatasetHandlers {
 	dsm := lib.NewDatasetMethods(inst)
 	rm := lib.NewRemoteMethods(inst)
-	h := DatasetHandlers{*dsm, inst, rm, inst.Node(), inst.Node().Repo, readOnly}
+	h := DatasetHandlers{*dsm, inst, rm, readOnly}
 	return &h
 }
 
--- a/api/datasets.go
+++ b/api/datasets.go
@@ -23,7 +23,12 @@ import (
 
 // DatasetHandlers wraps a requests struct to interface with http.HandlerFunc
 type DatasetHandlers struct {
+	// TODO (ramfox): start here
+	// should have a dsm for dataset methods
+	// and an inst
+	// future refactors should have handlers rely on inst for node and repo
 	lib.DatasetMethods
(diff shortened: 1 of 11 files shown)


api/datasets.go
```diff
@@ -24,6 +24,7 @@ import (
 // DatasetHandlers wraps a requests struct to interface with http.HandlerFunc
 type DatasetHandlers struct {
 	lib.DatasetMethods
+	inst     *lib.Instance
 	remote   *lib.RemoteMethods
 	node     *p2p.QriNode
 	repo     repo.Repo
```
ramfox, line 30: right now `DatasetHandlers` needs to work with "Dispatch-ified" and non "Dispatch-ified" `DatasetMethods`. When each is converted, we can remove the `lib.DatasetMethods`.

changes/changes.go
```diff
@@ -386,14 +386,18 @@ func (svc *service) columnStatsDelta(left, right interface{}, lCol, rCol *tabula
...
+			return nil, fmt.Errorf("dataset has only one version")
+		}
+		leftRef.Path = rightDs.PreviousPath
 	}
-	rightDs, err := svc.loader.LoadDataset(ctx, rightRef, loadSource)
+
+	leftDs, err := svc.loader.LoadDataset(ctx, leftRef, loadSource)
 	if err != nil {
 		return nil, err
 	}
```
ramfox, line 403: This switches the `leftRef` & `rightRef` expectations. Above this point, we require at least a `right` side of the change report, and expect the `left` side to be derived from the `right`. This refactor matches those expectations.

lib/datasets.go
```diff
@@ -1723,3 +1730,67 @@ func formFileDataset(r *http.Request, ds *dataset.Dataset) (err error) {
...
 }
+
+// datasetImpl holds the method implementations for DatasetMethods
+type datasetImpl struct{}
+
+// List gets the reflist for either the local repo or a peer
+func (datasetImpl) List(scope scope, p *ListParams) ([]dsref.VersionInfo, error) {
+	return nil, fmt.Errorf("not yet implemented")
+}
+
```
ramfox, line 1741: we need to have some implementation of each of the `DatasetMethods` on the `datasetImpl` struct, or the `buildMethodMap` we use when we build up the `Dispatch` sigfaults.

api/datasets.go
```diff
@@ -34,7 +35,7 @@ type DatasetHandlers struct {
 func NewDatasetHandlers(inst *lib.Instance, readOnly bool) *DatasetHandlers {
 	dsm := lib.NewDatasetMethods(inst)
 	rm := lib.NewRemoteMethods(inst)
-	h := DatasetHandlers{*dsm, rm, inst.Node(), inst.Node().Repo, readOnly}
+	h := DatasetHandlers{*dsm, inst, rm, inst.Node(), inst.Node().Repo, readOnly}
```
dustmop, line 38: While you're here, could you also remove the node and repo fields? I was going to say you could just get them from the instance, but it turns out they're not used at all.

lib/changes.go
```diff
@@ -18,30 +18,31 @@ type ChangeReport = changes.ChangeReportResponse
...
-		if left, _, err = m.inst.ParseAndResolveRef(ctx, p.LeftRefstr, reportSource); err != nil {
+		if left, _, err = scope.ParseAndResolveRef(ctx, p.LeftRefstr, reportSource); err != nil {
 			return nil, err
 		}
 	} else {
 		left = dsref.Ref{Username: right.Username, Name: right.Name}
 	}
 
-	return changes.New(m.inst, m.inst.stats).Report(ctx, left, right, reportSource)
+	return changes.New(scope.inst, scope.inst.stats).Report(ctx, left, right, reportSource)
```
dustmop, line 47: I'd strongly prefer that users of `scope` don't directly access the `Instance` it holds onto.

For the first parameter, could you add a method to `scope` called `Loader` that returns `scope.inst`. There's already a method on `scope` with that name (my fault, sorry), please rename that older method to `ParseResolveFunc()` and update it's caller - I believe there is only one for `apply`.

For the second parameter, please add a method to `scope` called `Stats` that simply returns the stats component.

lib/datasets.go
dustmop: This is a good pattern to put in place, it acts as a TODO list for methods that aren't switched to dispatch yet, and makes it easier to start converting functions one at a time.

dustmop (approved):
A request for one small modification, then it's LGTM and good to merge

api/datasets.go
ramfox: wooo def, that's going to feel v good to remove.

lib/changes.go
ramfox: *chefs kiss*

Issue #1661, opened 2021-03-02
internal data structures should serialize to disk after dispatch completes


dustmop:
```
TODO from dispatch PR: https://github.com/qri-io/qri/pull/1653/files
// TODO(dustmop): If the method wrote to our internal data structures, like
// refstore, logbook, etc, serialize and commit those changes here
```

As dispatch is the synchronization mechanism for requests that may be running in parallel, it should also be handling changes to our internal data structures (logbook, refstore, dscache, profileStore, keyStore, etc). This allows changing data structures safely while read-only requests are running, without those requests having their data change underneath them while they are still executing, which would lead to incomprehensible bugs.

PR #1694, opened 2021-03-09
refactor(config): adjust `ConfigMethods` to use dispatch

### LINKED ISSUE #1693: use dispatch for config functions

ramfox:
- adds `Name` method to satisfy `MethodSet` interface
- adds `ConfigMethods` method to `Instance` so that instance is responsible for creating the `ConfigMethods` & removes `NewConfigMethods` func
- creates `configImpl` struct to use when registering the config methods for use in dispatch
- creates `GetConfig`, `GetConfigKeys` & `SetConfig` methods for `configImpl` that copies the bulk of the functionality from the original `ConfigMethod` methods
- adjusts `GetConfig`, `GetConfigKeys` & `SetConfig` to call `Dispatch`
- adjust `qri config` calls to use `inst.ConfigMethods()` rather than creating their own `ConfigMethods` struct using `NewConfigMethods`

closes #1693


lib/config.go
```diff
@@ -16,13 +16,15 @@ type ConfigMethods struct {
...
+func (m *ConfigMethods) Name() string {
+	return "config"
 }
 
-// CoreRequestsName specifies this is a configuration handle
-func (m ConfigMethods) CoreRequestsName() string { return "config" }
+// ConfigMethods returns the `ConfigMethods` that the instance has registered
+func (inst *Instance) ConfigMethods() *ConfigMethods {
+	return &ConfigMethods{inst: inst}
+}
```
ramfox, line 27: `Instance` already has a method called `Config`. So this was given the name `ConfigMethods`.

Our other `MethodSet`s have the names `Access`, `Dataset`, and `Filesys`. So `ConfigMethods` is way less pretty. Any alternative names are welcome :)

dustmop: I was afraid we'd eventually run into this situation. As more and more functionality is moved from `Instance` into `scope`, it should be unnecessary to have so many accessors and methods publicly accessible on `Instance`; implementations should be getting what they need from `scope` instead. Over time the only public methods on `Instance` should be these method set accessors and whatever is needed to those using qri as a library. It might be the case that the current `Config()` method becomes unneeded by that point (it seems to only be used by `lib/remote.go`, `lib/resolve.go` and `api/` right now) but we shouldn't require fixing that before getting this PR merged.

I'm fine with using the slightly less pretty `ConfigMethods` for now. Could you just add a TODO to reevaluate this name as the dispatch refactor nears completion.

lib/config.go
```diff
@@ -80,15 +109,10 @@ func (m *ConfigMethods) GetConfig(ctx context.Context, p *GetConfigParams) ([]by
 	return res, nil
 }
 
-// GetConfigKeys returns the Config key fields, or sub keys of the specified
-// fields of the Config, as a slice of bytes to be used for auto completion
-func (m *ConfigMethods) GetConfigKeys(ctx context.Context, p *GetConfigParams) ([]byte, error) {
-	if m.inst.http != nil {
-		return nil, ErrUnsupportedRPC
```
dustmop, line 87: I think we want to keep this behavior for these config methods? We don't want config to be doable over rpc. The check for `http != nil` should be moved into the method definition above, not the impl. I know it breaks the regularity of these method definitions, but I'm think it's the best we can do for now. Maybe the `Mapping` [metadata](https://github.com/qri-io/qri/pull/1687) can determine whether it's allowed to call something over rpc?

lib/config.go
```diff
@@ -155,18 +179,14 @@ func parseKeys(cfg interface{}, prefix, parentKey string) ([]byte, error) {
...
-	if m.inst.http != nil {
-		return &res, ErrUnsupportedRPC
-	}
-
 	if err := update.Validate(); err != nil {
 		return &res, fmt.Errorf("validating config: %w", err)
 	}
 
-	if err := m.inst.ChangeConfig(update); err != nil {
+	if err := scope.Loader().ChangeConfig(update); err != nil {
```
dustmop, line 189: This is okay, just wanted to mention that this will be changing soon as per this TODO: https://github.com/qri-io/qri/issues/1661. Writing any of our data structures should be pulled up into `dispatch` such that it can be safely done in a multi-process environment.

dustmop (commented):
One request for a change, to keep around the RPC restriction.

lib/config.go
ramfox: ahh, yes I misunderstood. I somehow thought this check happened in Dispatch for some reason. Adding back in. And yes, having a centralized place for this doesn't sound too shabby.

lib/config.go
```diff
@@ -16,13 +16,17 @@ type ConfigMethods struct {
...
 }
 
-// CoreRequestsName specifies this is a configuration handle
-func (m ConfigMethods) CoreRequestsName() string { return "config" }
+// ConfigMethods returns the `ConfigMethods` that the instance has registered
+// TODO (ramfox): as the dispatch refactoring nears completion, we should
+// re-evaluate this name
+func (inst *Instance) ConfigMethods() *ConfigMethods {
+	return &ConfigMethods{inst: inst}
+}
```
b5, line 29: getting caught up on this PR, I think we should take the other approach, calling _this_ config, and switching the existing `inst.Config()` method to `inst.GetConfig()` and marking `inst.GetConfig` as deprecated at the same time.

We're trying to get away from `*lib.Instance` having any non-method accessors, and this is a chance to move forward on that goal

lib/config.go
```diff
@@ -16,13 +16,17 @@ type ConfigMethods struct {
...
+// Name returns the name of this method group
+func (m *ConfigMethods) Name() string {
+	return "config"
 }
 
-// CoreRequestsName specifies this is a configuration handle
-func (m ConfigMethods) CoreRequestsName() string { return "config" }
+// Config returns the `Config` that the instance has registered
+// TODO (ramfox): as the dispatch refactoring nears completion, we should
+// re-evaluate this name
```
b5, line 26: comment is no longer relevant

lib/lib.go
```diff
@@ -791,8 +791,10 @@ func (inst *Instance) Connect(ctx context.Context) (err error) {
 	return nil
 }
 
-// Config provides methods for manipulating Qri configuration
-func (inst *Instance) Config() *config.Config {
+// GetConfig provides methods for manipulating Qri configuration
+// TODO (ramfox): DEPRECATED - we are moving to a paradigm where the only methods
+// on inst are the accessors
```
b5, line 796: Use the [golang deprecation format](https://github.com/golang/go/wiki/Deprecated). In addition, depreciation notices should always tell a user what to use instead.  Should look like this:

```go
// GetConfig provides methods for manipulating Qri configuration
// 
// Deprecated: this method will be removed in a future release.
// Use inst.Config().GetConfig instead
func (inst *Instance) GetConfig() *config.Config {
```

The fact that `inst.Config().GetConfig` isn't the exact same means we should consider moving `GetConfig` to config methods (`inst.Config().Get`), but let's save that for a follow-up PR

b5 (commented):
two comment nits that you should be fine to amend on your latest commit, then LGTM!

b5 (approved):

PR #1697, opened 2021-03-09
updated API spec

b5:
Now that we have/soon will have a bunch of really big refactoring under our belts, I think we should be doing a _massive_ overhaul of our API endpoint routing before cutting a release. This PR proposes a bunch of spec changes, while also aiming to explicitly document the connection between API endpoints & lib methods by way of the `operationId` field in the open api spec.

If we get this right & fill in the doc we'll leave with spec, accurate documentation, and punchiest for getting our API into release-ready shape.

_Lots_ of work to do here to fill in possible responses & query params based on input param structs. I think we should start with a high-level audit of the proposed endpoint changes

PR #1698, opened 2021-03-09
test(api): json file driven spec tests for API

b5:
meant as a compliment to #1697, specifically how we're going to test an updated API contract.

add framing & a spec test using JSON files for our HTTP api. I'm hoping we can use this to construct a portable set of tests that make our API more reliable. I expect these tests to evolve over time, but the main purpose is to have all test configuration done via JSON files with a higher signal-to-noise ratio. The spec test runs strictly via HTTP req/res.

I'd considered trying to do this via the open_api spec, which apparently has tooling for turning an OpenAPI doc into a set of contract tests, but after some research I think it's better to have these spec tests in conjunction, which can be ordered to create tests that amount to integrations via successive calls.

tests don't pass yet, just getting the framework in place

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -19,6 +19,7 @@ import (
 	golog "github.com/ipfs/go-log"
 	ma "github.com/multiformats/go-multiaddr"
 	manet "github.com/multiformats/go-multiaddr/net"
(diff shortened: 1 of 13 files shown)


Arqu:
Superseded by https://github.com/qri-io/qri/pull/1762

PR #1701, opened 2021-03-10
refactor(dataset): dataset methods rely on `Dispatch`, for both lib & api

### LINKED ISSUE #1700: use dispatch in dataset lib methods

ramfox:
Refactors dataset functions to use `Dispatch` at both the lib and api level
This pr also removes all references (the cmd level) to `NewDatasetMethods` in favor of using  the new `inst.Dataset` accessor. 

* [x] list
* [x] get
* [x] save

continued in #1707

### DIFF
--- a/api/datasets.go
+++ b/api/datasets.go
@@ -469,10 +469,11 @@ func (h *DatasetHandlers) saveHandler(w http.ResponseWriter, r *http.Request) {
 	scriptOutput := &bytes.Buffer{}
 	params.ScriptOutput = scriptOutput
 
-	res, err := h.Save(r.Context(), params)
-	if err != nil {
+	got, _, err := h.inst.Dispatch(r.Context(), "dataset.save", params)
+	res, ok := got.(*dataset.Dataset)
+	if err != nil || !ok {
 		log.Debugw("save dataset error", "err", err)
-		util.WriteErrResponse(w, http.StatusInternalServerError, err)
+		util.RespondWithError(w, err)
 		return
 	}
 	// Don't leak paths across the API, it's possible they contain absolute paths or tmp dirs.
--- a/api/datasets.go
+++ b/api/datasets.go
@@ -229,7 +229,14 @@ func (h *DatasetHandlers) listHandler(w http.ResponseWriter, r *http.Request) {
 	if params.Raw {
 		resRaw, err = h.ListRawRefs(r.Context(), params)
 	} else {
-		res, err = h.List(r.Context(), params)
+		method := "dataset.list"
+		got, _, err := h.inst.Dispatch(r.Context(), method, params)
+		ok := false
+		res, ok = got.([]dsref.VersionInfo)
+		if err != nil || !ok {
+			util.RespondWithError(w, err)
+			return
+		}
 	}
 
 	if err != nil {
--- a/api/datasets.go
+++ b/api/datasets.go
@@ -227,10 +227,15 @@ func (h *DatasetHandlers) listHandler(w http.ResponseWriter, r *http.Request) {
 	res := []dsref.VersionInfo{}
 	var err error
 	if params.Raw {
-		resRaw, err = h.ListRawRefs(r.Context(), params)
+		got, _, err := h.inst.Dispatch(r.Context(), "dataset.listrawrefs", params)
+		ok := false
+		resRaw, ok = got.(string)
+		if err != nil || !ok {
+			util.RespondWithError(w, err)
+			return
+		}
 	} else {
-		method := "dataset.list"
-		got, _, err := h.inst.Dispatch(r.Context(), method, params)
+		got, _, err := h.inst.Dispatch(r.Context(), "dataset.list", params)
 		ok := false
 		res, ok = got.([]dsref.VersionInfo)
 		if err != nil || !ok {
--- a/api/datasets.go
+++ b/api/datasets.go
@@ -267,20 +267,21 @@ func (h *DatasetHandlers) listHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func (h *DatasetHandlers) getHandler(w http.ResponseWriter, r *http.Request) {
-	params := lib.GetParams{}
+	params := &lib.GetParams{}
 
-	err := lib.UnmarshalParams(r, &params)
+	err := lib.UnmarshalParams(r, params)
 	if err != nil {
 		util.WriteErrResponse(w, http.StatusBadRequest, err)
 		return
 	}
 
-	result, err := h.Get(r.Context(), &params)
-	if err != nil {
+	got, _, err := h.inst.Dispatch(r.Context(), "dataset.get", params)
+	res, ok := got.(*lib.GetResult)
+	if err != nil || !ok {
 		util.RespondWithError(w, err)
 		return
 	}
-	h.replyWithGetResponse(w, r, &params, result)
+	h.replyWithGetResponse(w, r, params, res)
 }
 
 // inlineScriptsToBytes consumes all open script files for dataset components
--- a/api/datasets.go
+++ b/api/datasets.go
@@ -228,18 +228,26 @@ func (h *DatasetHandlers) listHandler(w http.ResponseWriter, r *http.Request) {
 	var err error
 	if params.Raw {
 		got, _, err := h.inst.Dispatch(r.Context(), "dataset.listrawrefs", params)
+		if err != nil {
+			util.RespondWithError(w, err)
+			return
+		}
 		ok := false
 		resRaw, ok = got.(string)
-		if err != nil || !ok {
-			util.RespondWithError(w, err)
+		if !ok {
+			util.RespondWithDispatchTypeError(w, got)
 			return
 		}
 	} else {
 		got, _, err := h.inst.Dispatch(r.Context(), "dataset.list", params)
+		if err != nil {
+			util.RespondWithError(w, err)
+			return
+		}
 		ok := false
 		res, ok = got.([]dsref.VersionInfo)
-		if err != nil || !ok {
-			util.RespondWithError(w, err)
+		if !ok {
+			util.RespondWithDispatchTypeError(w, got)
 			return
 		}
 	}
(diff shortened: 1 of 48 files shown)


cmd/connect.go
```diff
@@ -67,18 +67,18 @@ func (o *ConnectOptions) Complete(f Factory, args []string) (err error) {
...
 		} else if repoErr != nil {
-			return err
+			return
 		}
 	}
 
 	if err = f.Init(); err != nil {
-		return err
+		return
 	}
```
ramfox, line 83: unifying the way we do these returns in each `cmd.Complete`, since I was there while refactoring `DatasetMethods` and `Instance`

lib/scope.go
```diff
@@ -44,54 +47,65 @@ func (s *scope) ActiveProfile() *profile.Profile {
...
+func (s *scope) Bus() event.Bus {
+	// TODO(dustmop): Filter only events for this scope.
+	return s.inst.bus
+}
+
+// Config returns the config
+func (s *scope) Config() *config.Config {
+	return s.inst.cfg
+}
+
```
ramfox, line 60: alphabetized these methods, was having trouble quickly figuring out if a method existed on scope at a glance, so this was a solution.

api/datasets.go
```diff
@@ -227,9 +227,19 @@ func (h *DatasetHandlers) listHandler(w http.ResponseWriter, r *http.Request) {
 	res := []dsref.VersionInfo{}
 	var err error
 	if params.Raw {
-		resRaw, err = h.ListRawRefs(r.Context(), params)
+		got, _, err := h.inst.Dispatch(r.Context(), "dataset.listrawrefs", params)
+		ok := false
+		resRaw, ok = got.(string)
+		if err != nil || !ok {
```
dustmop, line 233: If `err` is nil, but `ok` is false, then conditional will be executed with a nil err, which won't do anything helpful. The state of `ok` needs to be checked separately.

There's a few places this is happening, please fix.

api/datasets.go
```diff
@@ -227,9 +227,19 @@ func (h *DatasetHandlers) listHandler(w http.ResponseWriter, r *http.Request) {
...
+			util.RespondWithError(w, err)
```
dustmop, line 234: Calls to `RespondWithError`, same as `WriteErrResponse`, should be followed by a return. We don't want to do anything else in a handler after sending a response.

lib/datasets.go
```diff
@@ -741,270 +413,41 @@ func (p *SaveParams) SetNonZeroDefaults() {
 // Save adds a history entry, updating a dataset
 func (m *DatasetMethods) Save(ctx context.Context, p *SaveParams) (*dataset.Dataset, error) {
 	log.Debugw("DatasetMethods.Save", "ref", p.Ref, "apply", p.Apply)
```
dustmop, line 415: This log statement should move into the implementation.

lib/scope.go
```diff
@@ -44,54 +47,65 @@ func (s *scope) ActiveProfile() *profile.Profile {
...
+func (s *scope) GetVersionInfoShim(ref dsref.Ref) (*dsref.VersionInfo, error) {
+	r := s.inst.Repo()
+	return repo.GetVersionInfoShim(r, ref)
 }
 
-// Dscache returns the dscache
-func (s *scope) Dscache() *dscache.Dscache {
-	return s.inst.Dscache()
+// Loader returns the instance
+func (s *scope) Loader() *Instance {
```
dustmop, line 92: Loader's return value should be `dsref.Loader`, just the interface that is able to load datasets. Returning the instance itself go against the desire to migrate state management over to scope.

For situations where `Instance` is needed for functionality to work, add a new function to `scope` with the same parameters and have that function call the `instance` version. This isolates the exposure of `instance`'s state and let's us know exactly where it is being used.

For example, I know `lib/config.go` is using `scope.Loader().ChangeConfig(...)`. Instead, `ChangeConfig` should be added to `scope` and that should be called instead.

lib/datasets.go
```diff
@@ -1736,22 +1179,534 @@ type datasetImpl struct{}
...
+		err = maybeWriteOutfile(p, res)
+		if err != nil {
+			return nil, err
+		}
+		return res, nil
+	} else if p.Selector == "stats" {
+		statsParams := &StatsParams{
+			Dataset: res.Dataset,
+		}
+		sa, err := scope.Loader().Dataset().Stats(scope.Context(), statsParams)
```
dustmop, line 1421: No need to get the Instance and go through another `Dispatch` invocation here. All `lib.Stats` is doing is resolving and loading the dataset, but we've already done that by this point.

Instead, add a method `Stats` to scope:

```
func (s *scope) Stats() *stats.Service {
  return s.inst.stats
}
```

and call that, passing in `res.Dataset`.

lib/datasets.go
```diff
@@ -1736,22 +1179,534 @@ type datasetImpl struct{}
...
+		}
+		dsf.Assign(ds)
+		ds = dsf
+	}
+
+	if p.Ref == "" && ds.Name != "" {
+		p.Ref = fmt.Sprintf("me/%s", ds.Name)
+	}
+
+	resolver, err := scope.Loader().resolverForMode("local")
```
dustmop, line 1529: Add `ResolverForMode` to scope, whose body calls the `Instance` function, and use the scope version here.

lib/datasets.go
```diff
@@ -1736,22 +1179,534 @@ type datasetImpl struct{}
...
+		// allocate an ID for the transform, subscribe to print output & build up
+		// runState
+		runID := transform.NewRunID()
+		runState = run.NewState(runID)
+		// create a loader so transforms can call `load_dataset`
+		// TODO(b5) - add a ResolverMode save parameter and call m.inst.resolverForMode
+		// on the passed in mode string instead of just using the default resolver
+		// cmd can then define "remote" and "offline" flags, that set the ResolverMode
+		// string and control how transform functions
+		loader := NewParseResolveLoadFunc("", scope.Loader().defaultResolver(), scope.Loader())
```
dustmop, line 1623:
```suggestion
		loader := scope.ParseResolveFunc()
```

lib/datasets.go
```diff
@@ -1736,22 +1179,534 @@ type datasetImpl struct{}
...
+		}
+		text = c.VerboseString(true)
+		return text, nil
+	}
+	return base.RawDatasetRefs(scope.Context(), scope.Repo())
+}
+
+// Get retrieves datasets and components for a given reference.t
+func (datasetImpl) Get(scope scope, p *GetParams) (*GetResult, error) {
+	if err := qfs.AbsPath(&p.Outfile); err != nil {
```
dustmop, line 1303: Ah, another caveat of `Dispatch`: this call has to happen in the method, not the implementation. The reason is that, when `qri connect` is running, the caller may pass a relative path which needs to be turned into an absolute path before being sent over RPC. Otherwise, the caller and receiver may be in running in different directories and the relative path won't work.

See `Status` in `lib/fsi.go` for an example of how this is done.

I'm of the opinion that `Dispatch` should be made aware of which parameter fields are filepaths, and it should be calling `AbsPath` as needed. But the code to do this doesn't exist yet.

lib/scope.go
```diff
@@ -104,18 +118,27 @@ func (s *scope) ParseAndResolveRefWithWorkingDir(ctx context.Context, refstr, so
...
+}
+
+// Stats returns the stats service
+func (s *scope) Stats() *stats.Service {
+	return s.inst.stats
+}
+
+// Transform returns the transform service
+func (s *scope) Transform() *transform.Service {
+	return s.inst.transform
```
dustmop, line 143: This has changed due to https://github.com/qri-io/qri/pull/1691. The `TransformService` no longer exists on `lib`.

lib/datasets.go
```diff
@@ -1736,22 +1179,534 @@ type datasetImpl struct{}
...
+						io.WriteString(scriptOut, "\n")
+					}
+				}
+			}
+			return nil
+		}, runID)
+
+		// apply the transform
+		shouldWait := true
+		err := scope.Transform().Apply(scope.Context(), ds, loader, runID, scope.Bus(), shouldWait, str, scriptOut, secrets)
```
dustmop, line 1640: Check out how this is called in `lib/transform.go` for the new version.

api/datasets.go
```diff
@@ -227,9 +227,19 @@ func (h *DatasetHandlers) listHandler(w http.ResponseWriter, r *http.Request) {
 	res := []dsref.VersionInfo{}
 	var err error
 	if params.Raw {
-		resRaw, err = h.ListRawRefs(r.Context(), params)
+		got, _, err := h.inst.Dispatch(r.Context(), "dataset.listrawrefs", params)
```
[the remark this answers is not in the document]
Arqu: The statement that only `cmd` relies on listrawrefs is correct, I just introduced the API for completion sake as it was hanging out alone. I'm fine with de-cluttering all these "multi" endpoints.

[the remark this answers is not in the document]
ramfox: made this refactor in follow up pr #1707!

api/util/errors.go
```diff
@@ -71,3 +73,9 @@ func RespondWithError(w http.ResponseWriter, err error) {
 	WriteErrResponse(w, http.StatusInternalServerError, err)
 	return
 }
+
+// RespondWithDispatchTypeError(w http.ResponseWriter, got
```
dustmop, line 77: lint error:
```
api/util/errors.go:77:1: comment on exported function RespondWithDispatchTypeError should be of the form "RespondWithDispatchTypeError ..."
```

api/util/errors.go
```diff
@@ -71,3 +73,9 @@ func RespondWithError(w http.ResponseWriter, err error) {
...
+func RespondWithDispatchTypeError(w http.ResponseWriter, got interface{}) {
+	log.Errorf("type mismatch: %v of type %s", got, reflect.TypeOf(got))
+	WriteErrResponse(w, http.StatusInternalServerError, fmt.Errorf("type mismatch: %v of type %s", got, reflect.TypeOf(got)))
```
dustmop, line 80: Great to have this utility function! Just wanted to point out that this is an appropriate use of `StatusInternalServerError`, since this error should only be hit if there's a logical / programming error in qri core - some method returning the wrong type from its declared method - rather than bad user input or a losing a network connection or something like that.

dustmop (approved):
LGTM, excellent! One fix for the lint error please, then good to merge.

PR #1702, opened 2021-03-11
refactor(remote): RPC > HTTP & API for remote

Arqu:
Still have to manually test this before actually marking as done.
Another interesting bit is the modification to `lib/http.go` where it now encodes some query params if you pass in a `map[string]string` which was needed for the `DELETE` request on the `AEPush` endpoint

### DIFF
--- a/lib/http.go
+++ b/lib/http.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 
 	"github.com/gorilla/schema"
 	ma "github.com/multiformats/go-multiaddr"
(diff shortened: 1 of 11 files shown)


lib/http.go
```diff
@@ -112,7 +112,19 @@ func (c HTTPClient) do(ctx context.Context, addr string, httpMethod string, mime
 	var err error
 
 	if httpMethod == http.MethodGet || httpMethod == http.MethodDelete {
-		req, err = http.NewRequest(httpMethod, addr, nil)
+		qvars := ""
+		if params != nil {
+			if pm, ok := params.(map[string]string); ok {
+				for k, v := range pm {
+					if qvars == "" {
+						qvars = fmt.Sprintf("?%s=%s", k, v)
```
dustmop, line 120: I think `k` and `v` need to be URL escaped here?

b5:
```go
               u, err := url.Parse(addr)
               if err != nil {
                 return err
               }
               
		if pm, ok := params.(map[string]string); ok {
			qvars := u.Query()
			for k, v := range pm {
				qvars.Set(k, v)
			}
			u.RawQuery = qvars.Encode()
		}
		req, err = http.NewRequest(httpMethod, u, nil)
```

lib/remote.go
```diff
@@ -29,155 +30,191 @@ func NewRemoteMethods(inst *Instance) *RemoteMethods {
...
-		return checkRPCError(r.inst.rpc.Call("RemoteMethods.Push", p, res))
+// UnmarshalFromRequest implements a custom deserialization-from-HTTP request
+func (p *PushParams) UnmarshalFromRequest(r *http.Request) error {
+	if p == nil {
+		p = &PushParams{}
+	}
+
+	params := *p
+	if params.Ref == "" {
+		params.Ref = r.FormValue("refstr")
```
b5, line 132: this seems to be the only thing standing in the way of being able to deserialize without the `UnmarshalFromRequest` override. Would love to get there in a follow-up PR

b5 (approved):
Just one request for query-encoding refactor, then LGTM

PR #1705, opened 2021-03-12
refactor(registry): RPC > HTTP & API for registry

Arqu:

### DIFF
--- a/lib/registry.go
+++ b/lib/registry.go
@@ -33,9 +33,9 @@ func (RegistryClientMethods) CoreRequestsName() string { return "registry" }
 type RegistryProfile = registry.Profile
 
 // CreateProfile creates a profile
-func (m RegistryClientMethods) CreateProfile(p *RegistryProfile, ok *bool) (err error) {
-	if m.inst.rpc != nil {
-		return checkRPCError(m.inst.rpc.Call("RegistryClientMethods.CreateProfile", p, ok))
+func (m RegistryClientMethods) CreateProfile(ctx context.Context, p *RegistryProfile) error {
+	if m.inst.http != nil {
+		return ErrUnsupportedRPC
 	}
 
 	// TODO(arqu): this should take the profile PK instead of active PK once multi tenancy is supported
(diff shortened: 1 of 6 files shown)


lib/registry.go
```diff
@@ -33,45 +33,47 @@ func (RegistryClientMethods) CoreRequestsName() string { return "registry" }
 type RegistryProfile = registry.Profile
 
 // CreateProfile creates a profile
-func (m RegistryClientMethods) CreateProfile(p *RegistryProfile, ok *bool) (err error) {
-	if m.inst.rpc != nil {
-		return checkRPCError(m.inst.rpc.Call("RegistryClientMethods.CreateProfile", p, ok))
+func (m RegistryClientMethods) CreateProfile(ctx context.Context, p *RegistryProfile) (bool, error) {
```
dustmop, line 36: Can we keep this as only returning the error? In my opinion, it's an anti-pattern to return both a `bool` and `error` if the bool's only purpose is to be equivalent to `err == nil`. Rather, it's perfectly fine to just use the error's presence or absence to determine success or failure. Especially since all the callers in this PR are ignoring the `bool` part of the return; a sure sign that it's not needed.

lib/registry.go
```diff
@@ -33,45 +33,47 @@ func (RegistryClientMethods) CoreRequestsName() string { return "registry" }
...
+	return true, nil
 }
 
 // ProveProfileKey sends proof to the registry that this user has control of a
 // specified private key, and modifies the user's config in order to reconcile
 // it with any already existing identity the registry knows about
-func (m RegistryClientMethods) ProveProfileKey(p *RegistryProfile, ok *bool) error {
-	if m.inst.rpc != nil {
-		return checkRPCError(m.inst.rpc.Call("RegistryClientMethods.CreateProfile", p, ok))
+func (m RegistryClientMethods) ProveProfileKey(ctx context.Context, p *RegistryProfile) (bool, error) {
```
dustmop, line 61: Same thing here, this can stay as returning a lone error.

lib/registry.go
Arqu: The only purpose is to conform to the original dispatch method format. Has that changed? Looking at the code we still require 2 output params.

Would love to get rid of it though

dustmop: Ah, I see, yes sorry for the confusion. The original dispatch required 2 return values, but it's valid now to return 1-3, with 1 being just an error.

dustmop (approved):
LGTM, good to merge!

PR #1706, opened 2021-03-12
starlark ds.set_body infers missing structure component

b5:
relies on qri-io/dataset#252


b5:
after a self-review, I'm going to merge this to keep moving

PR #1707, opened 2021-03-13
refactor(dataset): finish dataset dispatch refactor

ramfox:
Finish dataset dispatch refactor
* [x] Rename
* [x] Remove
* [x] Pull
* [x] Validate
* [x] Manifest
* [x] ManifestMisssing
* [x] DAGInfo
* [x] Stats 

depends on #1701

### DIFF
--- a/api/util/errors.go
+++ b/api/util/errors.go
@@ -50,6 +50,10 @@ func RespondWithError(w http.ResponseWriter, err error) {
 		WriteErrResponse(w, http.StatusUnprocessableEntity, err)
 		return
 	}
+	if errors.Is(err, dsref.ErrBadCaseShouldRename) || errors.Is(err, dsref.ErrDescribeValidName) || errors.Is(err, dsref.ErrDescribeValidUsername) {
+		WriteErrResponse(w, http.StatusBadRequest, err)
+		return
+	}
 	var perr *dsref.ParseError
 	if errors.As(err, &perr) {
 		WriteErrResponse(w, http.StatusBadRequest, err)
--- a/cmd/access.go
+++ b/cmd/access.go
@@ -61,7 +61,7 @@ type AccessOptions struct {
 // Complete adds any missing configuration that can only be added just before calling Run
 func (o *AccessOptions) Complete(f Factory, args []string) (err error) {
 	o.Instance, err = f.Instance()
-	return
+	return err
 }
 
 // CreateAccessToken constructs an access token suitable for making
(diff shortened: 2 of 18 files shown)


lib/api.go
```diff
@@ -65,6 +65,9 @@ const (
 
 	// dataset endpoints
 
+	// AEListRawl lists all datasets in your collection as a well formatted string
+	AEListRaw = APIEndpoint("/listraw")
+
```
Arqu, line 70: `AEListRawl > AEListRaw` in the comment
Also don't need the empty line.

api/util/errors.go
```diff
@@ -71,3 +77,9 @@ func RespondWithError(w http.ResponseWriter, err error) {
 	WriteErrResponse(w, http.StatusInternalServerError, err)
 	return
 }
+
+// RespondWithDispatchTypeError(w http.ResponseWriter, got
+func RespondWithDispatchTypeError(w http.ResponseWriter, got interface{}) {
+	log.Errorf("type mismatch: %v of type %s", got, reflect.TypeOf(got))
+	WriteErrResponse(w, http.StatusInternalServerError, fmt.Errorf("type mismatch: %v of type %s", got, reflect.TypeOf(got)))
```
Arqu, line 84: 👍

cmd/access.go
```diff
@@ -60,8 +60,8 @@ type AccessOptions struct {
 
 // Complete adds any missing configuration that can only be added just before calling Run
 func (o *AccessOptions) Complete(f Factory, args []string) (err error) {
-	o.Instance = f.Instance()
-	return nil
+	o.Instance, err = f.Instance()
+	return
```
Arqu, line 64: can drop the return here

cmd/checkout.go
```diff
@@ -49,24 +49,21 @@ type CheckoutOptions struct {
...
 	}
 
 	if len(args) == 2 {
 		o.Dir = args[1]
 	} else {
 		o.Dir = ""
 	}
-
-	return nil
+	return
```
Arqu, line 66: Can drop the return, but in general I'm not a huge fan of named return values

Arqu (approved):
LGTM. Some nits, but great stuff. `lib/datasets.go` is quite a mouthful.

cmd/access.go
b5: always use an explicit return value. It's fine to use named return values as documentation, but not as default returns

Issue #1704, opened 2021-03-12
want to clean up the dsref.Loader and lib/load.go situation


dustmop:
Turning a dataset from a string like "dustmop/my_dataset" into an actual dataset object involves these steps:

1) `parse`: "dustmop/my_dataset" -> dsref.Ref{Username: "dustmop", Name: "my_dataset"}, simple textual parsing
2) `resolve`: dsref.Ref{...} -> dsref.Ref{..., Path: "/ipfs/QmWhatever"}, resolve the name to get the storage location, as well as initID, et al. Might use the local refstore, or might ask the registry
3) `load`: dsref.Ref{...} -> *dataset.Dataset, load the object from the storage layer (such as IPFS)

We achieve dependency inversion by defining an interface `Loader` in `dsref/load.go` (very low-level), which is implemented by `lib.Instance` (very high-level). This way, packages such as `transform` and `sql` can load datasets without knowing anything about the details of parsing and resolving references. This interface defines `LoadDataset` which only performs step 3.

However, there's also a more powerful function defined in `lib/load.go` called `NewParseResolveLoadFunc` which does all 3 steps. It does not live on an interface, but is passed down as a function value into subsystems such as `transform` and `sql`. This has been especially awkward during the conversion onto `scope`, as there's lots of places that talk about a `Loader` and sometimes they mean the former while other times they mean the latter.

I would like to fix this.

My motivation comes from the fact that starlark transforms have a function `load_dataset` which runs all 3 steps, contradicting the semantics of the `Loader` interface. I feel strongly that this public facing function should inform the terminology used throughout our codebase.

I would propose the following changes:
```
type Loader interface {
  LoadDataset(ctx context.Context, refstr string) (*dataset.Dataset, error)
  LoadFullyResolved(ctx context.Context, ref dsref.Ref, source string) (*dataset.Dataset, error)
}
```

Rename the old method `LoadDataset` to `LoadFullyResolved`. It *must* fail if the ref does not have a `Path` value, since that means it has not been fully resolved.

At the same time, add new method called `LoadDataset`, with the same signature as the old `NewParseResolveLoadFunc`. This method should match the semantics of the starlark `load_dataset` function, meaning it performs all 3 steps.

This change will allow us to greatly simplify the codepath from `lib` down to `transform` and `sql`, while also lining up our terminology much better through the codebase.

[...]

PR #1708, opened 2021-03-15
refactor(diff): use `Dispatch` for `Diff` and `/diff`

### LINKED ISSUE #1700: use dispatch in dataset lib methods

ramfox:
closes #1700 

There were subtle bugs in the `NewInstance` and `buildrepo.New` paths that made it so that certain s    ubsystems would be created multiple times (once in at the lib level and once (or twice) at the repo     level), even if the systems were explicitly passed into the `buildrepo.New` function.

The major culprit was `NewMemRepo`, which is why these bugs only surfaced when trying to create a le    ss cumbersome way to create an in memory instance for testing.

This removes one major annoyance in our lib testing, every system has the same profile information!     This commit also adjusts any tests that get broken because of the different keys/peername.

One major subtly remains: you still need a cogent `repoPath` that you have full permission to edit.     The only subsystem this seems to break right now is `Stats`. We could pass configuration to force a     `nilCache` `Stats`, but if we want to use the test runner to do any testing around stats in lib in the future, we should probably support it.


config/test/config.go
```diff
@@ -19,7 +20,23 @@ func DefaultConfigForTesting() *config.Config {
 	return cfg
 }
 
-// DefaultProfileForTesting constructs a profile with precomputed keys, only used for testing.
+// DefaultMemConfigForTesting constructs a config with precomputed keys & settings for an in memory repo & fs, only used for testing.
+func DefaultMemConfigForTesting() *config.Config {
```
ramfox, line 24: It's possible we want this to be the `DefaultConfigForTesting`, but separated it out in case we wanted to split the concerns.

lib/lib.go
```diff
@@ -294,6 +296,22 @@ func OptProfiles(pros profile.Store) Option {
...
+}
+
+// OptDscache supplies a Dscache for the instance
+func OptDscache(dscache *dscache.Dscache) Option {
+	return func(o *InstanceOptions) error {
+		o.dscache = dscache
+		return nil
+	}
+}
+
```
ramfox, line 314: It's possible these are too... low level for us to reasonably want someone to add their own, however they were extremely helpful in debugging this issue!

lib/lib.go
```diff
@@ -663,11 +687,12 @@ func NewInstanceFromConfigAndNodeAndBus(ctx context.Context, cfg *config.Config,
...
-		node:    node,
-		dscache: dc,
-		logbook: r.Logbook(),
-		appCtx:  ctx,
+		cfg:      cfg,
+		node:     node,
+		dscache:  dc,
+		logbook:  r.Logbook(),
+		profiles: r.Profiles(),
+		appCtx:   ctx,
```
ramfox, line 695: Here is where I realized I shouldn't really be cleaning up `NewInstanceFromConfigAndNode` & if it's going to take an obnoxiously long time to understand what is going wrong, I might as well do it "right"

lib/scope.go
```diff
@@ -130,7 +130,7 @@ func (s *scope) ParseAndResolveRefWithWorkingDir(ctx context.Context, refstr, so
 
 // ParseResolveFunc returns a function that can parse a ref, then resolve and load it
 func (s *scope) ParseResolveFunc() dsref.ParseResolveLoad {
-	return NewParseResolveLoadFunc("", s.inst.defaultResolver(), s.inst)
+	return NewParseResolveLoadFunc(s.ActiveProfile().Peername, s.inst.defaultResolver(), s.inst)
```
ramfox, line 133: Allows us to use "me" as a prefix, which I'm sure we will have some other way to proxy when we have multi tenancy, but this preserve the behavior while we convert to `Dispatch` & `Scope`.

lib/test_runner_test.go
```diff
@@ -54,34 +50,35 @@ func newTestRunner(t *testing.T) *testRunner {
...
-	if err != nil {
-		t.Fatalf("error allocating test repo: %s", err.Error())
-	}
-	localResolver := dsref.SequentialResolver(mr.Dscache(), mr)
-	node, err := p2p.NewQriNode(mr, testcfg.DefaultP2PForTesting(), bus, localResolver)
-	if err != nil {
-		t.Fatal(err.Error())
+	qriPath := filepath.Join(tmpDir, "qri")
+	if err := os.Mkdir(qriPath, os.ModePerm); err != nil {
+		t.Fatal(err)
```
ramfox, line 61: Ensures we have write permissions, `ioutil.TempDir` doesn't guarantee anything.

lib/test_runner_test.go
```diff
@@ -54,34 +50,35 @@ func newTestRunner(t *testing.T) *testRunner {
...
+	cfg := testcfg.DefaultMemConfigForTesting()
 
-	inst := NewInstanceFromConfigAndNodeAndBus(ctx, testcfg.DefaultConfigForTesting(), node, bus)
-	// Assign the repoPath, as some tests inspect it in order to save files in the repository.
-	// Not assigning a path will cause them to use the current directory, and then save
-	// into the sourcetree.
-	inst.repoPath = tmpDir
+	// ensure we create a mock registry client for testing
+	key := InstanceContextKey("RemoteClient")
+	ctx = context.WithValue(ctx, key, "mock")
```
ramfox, line 67: This is gross, and the only thing I think needs to be actively fixed. Perhaps an addition to the `config.Remotes` settings that allows you to explicitly use a "mock" client?

config/test/config.go
b5: We now have a `test` package for Config details. Let's move this into `github.com/qri-io/qri/config/test`

See 1aeba849ee7b88c4e44cfbf7c1d4eb065c505cc3 for the commit that makes the change, and the rational behind a separate test configuration package

I'm a fan of the `Mem` differentiator. We do have tests that need to be stuff other than mem

config/test/config.go
```diff
@@ -19,7 +20,24 @@ func DefaultConfigForTesting() *config.Config {
...
+	}
+
+	// add settings for in mem repo
+	cfg.Repo.Type = "mem"
+
+	return cfg
+}
+
+// DefaultProfileForTesting constructs a profile with precompted keys, only used for testing.
 func DefaultProfileForTesting() *config.ProfilePod {
```
b5, line 41: I'd love this to move into a new `profile/test` package, but let's save that for a later PR. I'd love a 

```
TODO(b5): move this into a new profile/test package
```

while you're here 😄. Assign yourself if you're feeling bold

lib/diff.go
```diff
@@ -117,16 +117,63 @@ func (m *DatasetMethods) Diff(ctx context.Context, p *DiffParams) (*DiffResponse
...
+						titles[i] = title
+					}
+				}
+			}
+			return titles, nil
+		}
+	}
+	log.Debug("that terrible hack to detect header row & types just failed")
+	return nil, fmt.Errorf("nope")
+}
```
b5, line 164: remove this & replace all callers with [`tabular.ColumnsFromJSONSchema`](https://github.com/qri-io/dataset/blob/3e9020e68b0e6b789bbcc2b38cc90224b22a1058/tabular/tabular.go#L125)

lib/lib.go
b5: I think it's fine to have these options, but they should have documenting comments that clarifies these options _override defaults_.

lib/lib.go
b5: yeah, I'm glad this func has an ugly name. Can't wait to start ripping it out

lib/lib.go
```diff
@@ -936,6 +961,10 @@ func (inst *Instance) activeProfile(ctx context.Context) (pro *profile.Profile,
 		return inst.profiles.Owner(), nil
 	}
 
+	if pro == nil {
+		return nil, fmt.Errorf("no active profile")
+	}
```
b5, line 966: 👍

lib/scope.go
b5: into it. closing #1704 will help this situation

lib/test_runner_test.go
b5: _faaaaaancy_

lib/test_runner_test.go
b5: yep I was thinking the same thing. I'd love an issue for this that we can fix in a follow up

repo/buildrepo/build.go
```diff
@@ -43,26 +46,31 @@ func New(ctx context.Context, path string, cfg *config.Config, opts ...func(o *O
 
 	var err error
 	if o.Keystore == nil {
+		log.Debugf("buildrepo.New: creating keystore")
```
b5, line 49: flip all these `Debugf`'s to `Debug`

b5 (commented):
Looking great

config/test/config.go
ramfox: this is in the config test package! :)

lib/test_runner_test.go
dustmop: This code is my fault, and was added because I needed a way for tests in the `cmd` package (specifically pull_test.go) to use the MockRemoteClient. See: https://github.com/qri-io/qri/blob/05fc38a4cd33506695f28730d3f942c03f880cca/lib/lib.go#L517

It's not possible to inject dependencies into cobra commands, which is the start of this trouble. Adding an option to the configuration means then we need to document it and it becomes "public", which has other problems. What we really want is a way for tests that pull to be able to use this mockRemoteClient on all commands they execute. Maybe an environment variable is better?

Regardless, I'm all for coming up with a better solution than this hack.

config/test/config.go
ramfox: got it #1715

lib/test_runner_test.go
ramfox: okay good to know, added your comment to the issue #1716

b5 (approved):
LGTM

PR #1709, opened 2021-03-15
feat(validate): Parameters to methods will Validate automatically

dustmop:


b5 (approved):
lovely!

dustmop:
The approach I'm going with here is that Param structs passed into methods / Dispatch can optionally define a `Validate` method, which will ensure the parameters meet some requirements. The alternative would be to define an interface such as `InputParams` which would *require* parameters to always define a `Validate` method. However:
1) Most of our structs don't have a Validate method, I only found these 3 cases. Not worth it to define empty `Validate` methods all over the place.
2) When Validate methods *do* exist, we can just then test the method call using a unit test, as done here.

PR #1711, opened 2021-03-16
feat(dispatch): Methods attributes to hold metadata about methodSets

dustmop:
Originally prototyped by https://github.com/qri-io/qri/pull/1687, this change adds an "Attributes" method to each MethodSet. For now, this is only defining the HTTP endpoint and verb, but in the future can include things like permissions, data structures that need locking (logbook and profile store etc), and more.

We can then delete some of the earlier hacks put in place for Dispatch, such as methodEndpoint function.

For now, endpoints only allow a single HTTP verb. We may find we need to allow a list of verbs instead, but if so, need a way to determine which verb to use when calling a method over HTTP RPC.

Open question on which of these directions to go:
1) should the endpoint urls should remain in lib/api.go, with Attributes using the constants
2) should we keep the endpoint urls in Attributes, and remove them from lib/api.go
I'm leaning a bit towards (2), because it keeps the urls closer to their owner, but am not completely sure

### DIFF
--- a/lib/dispatch.go
+++ b/lib/dispatch.go
@@ -3,6 +3,7 @@ package lib
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"reflect"
 	"strings"
 	"time"
@@ -29,6 +30,14 @@ type Cursor interface{}
 // with the context.Context replaced by a scope.
 type MethodSet interface {
 	Name() string
+	Attributes() map[string]AttributeSet
+}
+
+// AttributeSet is extra information about each method, such as: http endpoint,
+// http verb, (TODO) permissions, and (TODO) other metadata
+type AttributeSet struct {
+	endpoint APIEndpoint
+	verb     string
 }
 
 // Dispatch is a system for handling calls to lib. Should only be called by top-level lib methods.
@@ -80,14 +89,14 @@ func (inst *Instance) Dispatch(ctx context.Context, method string, param interfa
 		}
 
 		if c, ok := inst.regMethods.lookup(method); ok {
-			// TODO(dustmop): This is always using the "POST" verb currently. We need some
-			// mechanism of tagging methods as being read-only and "GET"-able. Once that
-			// exists, use it here to lookup the verb that should be used to invoke the rpc.
+			if c.Endpoint == "" {
+				return nil, nil, ErrUnsupportedRPC
+			}
 			if c.OutType != nil {
 				out := reflect.New(c.OutType)
 				res = out.Interface()
 			}
-			err = inst.http.Call(ctx, methodEndpoint(method), param, res)
+			err = inst.http.CallMethod(ctx, c.Endpoint, c.Verb, param, res)
 			if err != nil {
 				return nil, nil, err
 			}
@@ -193,6 +202,8 @@ type callable struct {
 	InType    reflect.Type
 	OutType   reflect.Type
 	RetCursor bool
+	Endpoint  APIEndpoint
+	Verb      string
 }
 
 // RegisterMethods iterates the methods provided by the lib API, and makes them visible to dispatch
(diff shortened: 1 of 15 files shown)


b5:
> should the endpoint urls should remain in lib/api.go, with Attributes using the constants
should we keep the endpoint urls in Attributes, and remove them from lib/api.go
I'm leaning a bit towards (2), because it keeps the urls closer to their owner, but am not completely sure

A common thing I do whenever I'm looking at source code for an API is ask "where's the router"? The first place I was exposed to this pattern was with node/express.js, but the thing I'm after is a slew of router method calls that map string patterns to endpoints. Our version of that is now useless from a developer-getting-oriented standpoint:

https://github.com/qri-io/qri/blob/590ebdf5a06aca3b93a83124482c84f1f17e85dc/api/api.go#L197-L308

But this file is getting closer:

https://github.com/qri-io/qri/blob/590ebdf5a06aca3b93a83124482c84f1f17e85dc/lib/api.go#L25-L155

For that reason I'm a fan of option 1. Long-term I'd like to see these constants defined in one place & bound to their method set via the Attributes method you've described here. That way as a dev I still have a file I can use to orient to all defined API endpoints.

Long term, these constants should be unexported and refactored to be as legible as possible for this visual scan:

```go
const (
	// base endpoints
	aeHome   APIEndpoint = "/"
	aeHealth APIEndpoint = "/health"
	aeIPFS   APIEndpoint = "/ipfs/{path:.*}"

	// profile endpoints
	aeMe            APIEndpoint = "/me"
	aeProfile       APIEndpoint = "/profile"
	aeProfilePhoto  APIEndpoint = "/profile/photo"
	aeProfilePoster APIEndpoint = "/profile/poster"

	// ...
)
```

lib/dispatch.go
```diff
@@ -29,6 +30,14 @@ type Cursor interface{}
 // with the context.Context replaced by a scope.
 type MethodSet interface {
 	Name() string
+	Attributes() map[string]AttributeSet
+}
+
+// AttributeSet is extra information about each method, such as: http endpoint,
+// http verb, (TODO) permissions, and (TODO) other metadata
+type AttributeSet struct {
```
b5, line 38: are attributes _required_ for all methods defined on `MethodSet` implementers? This commend should clarify.

lib/dispatch.go
```diff
@@ -259,92 +274,118 @@ func (inst *Instance) registerOne(ourName string, methods MethodSet, impl interf
...
+			Endpoint:  endpoint,
+			Verb:      httpVerb,
 		}
 		log.Debugf("%d: registered %s(*%s) %v", k, funcName, inType, outType)
 	}
 
 	for k := range methodMap {
-		if k != "Name" {
-			panic(fmt.Sprintf("%s: did not find implementation for method %s", msetType, k))
+		if k != "Name" && k != "Attributes" {
```
b5, line 379: I think this check means that any `MethodSet` can only have "method functions" and the `MethodSet` interface implementation (formally: `Name` and `Attributes` methods). Can we get a comment on the `MethodSet` interface definition:

```
MethodSet implementations cannot have extra methods, only MethodSet interface methods and methods to be registered
```

lib/lib.go
```diff
@@ -790,6 +790,31 @@ func (inst *Instance) Connect(ctx context.Context) (err error) {
...
+
+// Filesys returns the FSIMethods that Instance has registered
+func (inst *Instance) Filesys() *FSIMethods {
+	return &FSIMethods{d: inst}
+}
+
+// Transform returns the TransformMethods that Instance has registered
+func (inst *Instance) Transform() *TransformMethods {
+	return &TransformMethods{d: inst}
+}
```
b5, line 816: do these need to return pointers? I'd prefer they not

lib/dispatch.go
dustmop: Yes, the reasoning being that it's easier to require all methods to have attributes, rather than allow misspellings to go unnoticed without error.

lib/lib.go
dustmop: It was done to be consistent with the old Method style structs, but you're right that there's no need for that.

lib/dispatch.go
dustmop: Added clarifying comment

b5 (approved):
LGTM!

PR #1713, opened 2021-03-16
refactor(rpc): remove rpc

Arqu:
Last part of https://github.com/qri-io/qri/issues/1675


cmd/factory_test.go
```diff
@@ -127,8 +127,8 @@ func (t TestFactory) ConnectionNode() (*p2p.QriNode, error) {
 	return t.node, nil
 }
 
-// RPC returns from internal state
-func (t TestFactory) RPC() *rpc.Client {
+// HTTPClient returns from internal state
```
dustmop, line 130: This old comment seems like it was stale, or something. How about `HTTPClient returns nil for tests`

cmd/qri.go
```diff
@@ -216,12 +215,12 @@ func (o *QriOptions) CryptoGenerator() key.CryptoGenerator {
 	return o.generator
 }
 
-// RPC returns from internal state
-func (o *QriOptions) RPC() *rpc.Client {
+// HTTPClient returns from internal state
```
dustmop, line 218:
```suggestion
// HTTPClient returns a client for performing RPC over HTTP
```

dustmop (approved):
Some suggestions for comments, but LGTM aside from that

PR #1718, opened 2021-03-23
fix(api, token): dataset lib methods now working with http rpc

### LINKED ISSUE #1717: fix(api): dataset lib methods not working with http rpc
Two main errors occuring:
- segfault / Nil pointer dereference error
- "path not found"

### nil pointer dereference error
- Issue: We call now call `o.Instance()` to get access to the `dataset methods`, so we can call the lib method for the cmd
- `o.Instance()` calls `o.Init()` to ensure we have an instance, `o.Init` calls `NewInstance`
- if we are working over rpc in line 429-446 of `lib/lib.go` in `NewInstance`, we create a `NewHTTPClient`, add it to the instance (`inst.http`), call `inst.waitForAllDone()`, and return the instance, this all happens before the full instance is set up, because it assumes the work will be done over the http call
- however, when we return back to `o.Init`, we continue on assuming we have a full instance. In line, 170 of `cmd/qri.go`, inside `o.Init`, we call `PrintProgressBarOnEvents`, which takes a bus, we supply `o.inst.Bus()`, which, because it was never set up in `NewInstance`, returns `nil`. When, in `PrintProgressBarOnEvent`, we call `bus.SubscribeTypes`, we get a nil pointer dereference error.

Way to solve:
- we will probably want progress bars over http one day, so it doesn't seem right to not call `PrintProgressBarsOnEvents`, just because we are over rpc http, so it seems like the sanest choice that won't mess up future work is to just check we don't have a `nil` bus before attempting to subscribe to it.

Clearing up that error leads us to....

### `404 path not found` error:
- THIS ERROR MESSAGE WAS SUPER MISLEADING. 
- we actually make it to the list endpoint/listHandler just fine
- we know this because we are getting an info level log "error listing datasets: path not found" and that log is being emitted by `listHandler`! So it's not the uri path that is messing with us!
- so it's actually part of dispatch that's not letting us move forward
- following the error we go through `Dispatch`, to `newScope`, to `activeProfile`, 
- this context has a token (assuming the http client issues one), looks like the issue is actually that we cannot find the profile in question based on the `claims.ProfileID` and the `profile.IDB58DecodeOrEmpty`
- we parse the token from the context, convert it to a `claim`, but this claim does not contain a `ProfileID`. So, since this profileID is the empty string, we cannot find it when we `GetProfile` (in `inst.activeProfile()`). `GetProfile` returns a `qfs.ErrNotFound` error aka `Path Not Found`.
- hmm, yes, in `HTTPClient.do`, line 148 of `lib/http.go`, we call `token.AddContextTokenToRequest`, but this function does not add a profile id to the context. Since we assume in `inst.activeProfile` that if you have a token, you must have a profile id set, we are getting the `Path not found` error.
- Turns out `token.NewPrivAuthToken` is not adding `ProfileID` into the token!

Way to solve:
- add `ProfileID` to `token.NewPrivAuthToken` and pass `ProfileID` into `claim` struct
- adjust `GetProfile` to return `profile.ErrNotFound` rather than `qfs.ErrNotFound`

ramfox:
closes #1717 

read above issue for context!

### DIFF
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -244,6 +244,10 @@ func PrintProgressBarsOnEvents(w io.Writer, bus event.Bus) {
 	p := mpb.New(mpb.WithWidth(80), mpb.WithOutput(w))
 	progress := map[string]*mpb.Bar{}
 
+	if bus == nil {
+		log.Errorf("event bus is nil")
+		return
+	}
 	// wire up a subscription to print download progress to streams
 	bus.SubscribeTypes(func(_ context.Context, e event.Event) error {
 		lock.Lock()
--- a/profile/store.go
+++ b/profile/store.go
@@ -11,7 +11,6 @@ import (
 	"sync"
 
 	"github.com/libp2p/go-libp2p-core/peer"
-	"github.com/qri-io/qfs"
 	"github.com/qri-io/qri/auth/key"
 	"github.com/qri-io/qri/config"
 	qerr "github.com/qri-io/qri/errors"
@@ -441,7 +440,7 @@ func (r *LocalStore) PeernameID(peername string) (ID, error) {
 			return IDB58Decode(id)
 		}
 	}
-	return "", qfs.ErrNotFound
+	return "", ErrNotFound
 }
 
 // GetProfile fetches a profile from the store
@@ -469,7 +468,7 @@ func (r *LocalStore) GetProfile(id ID) (*Profile, error) {
 		}
 	}
 
-	return nil, qfs.ErrNotFound
+	return nil, ErrNotFound
 }
 
 // ProfilesForUsername fetches all profile that match a username (Peername)
@@ -524,7 +523,7 @@ func (r *LocalStore) PeerProfile(id peer.ID) (*Profile, error) {
 		}
 	}
 
-	return nil, qfs.ErrNotFound
+	return nil, ErrNotFound
 }
 
 // DeleteProfile removes a profile from the store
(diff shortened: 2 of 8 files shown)


profile/store.go
```diff
@@ -469,7 +468,7 @@ func (r *LocalStore) GetProfile(id ID) (*Profile, error) {
 		}
 	}
 
-	return nil, qfs.ErrNotFound
+	return nil, ErrNotFound
```
ramfox, line 471: Somewhere above this in the stack (maybe @ `inst.activeProfile`) the `profile.ErrorNotFound` error should instead be changed to an "unauthorized request" error or something similar.

Arqu: We had a chat a while ago, one of the things dispatch chould help centralize the error processing and we could potentially translate "internal error states" into something human readable. For now maybe just a more specific error can help - "request profile not set" or something along the lines.

cmd/print.go
```diff
@@ -244,84 +244,86 @@ func PrintProgressBarsOnEvents(w io.Writer, bus event.Bus) {
...
-			evtID := fmt.Sprintf("%s/%s", evt.Username, evt.Name)
-			cpl := int64(math.Ceil(evt.Completion * 100))
-
-			switch e.Type {
-			case event.ETDatasetSaveStarted:
-				bar, exists := progress[evtID]
-				if !exists {
-					bar = addElapsedBar(p, 100, "saving")
-					progress[evtID] = bar
+	if bus != nil {
```
Arqu, line 247: I would rather have the guard on the entry point next to `shouldColorOutput` and make it the caller responsibility to ensure the bus is not nil. Or at least make this log out an `event bus is nil` message so we don't swallow this error state in the future.

Arqu (approved):
LGTM as a solution but would introduce logging or move the guard so we don't have this from another angle soon again.

PR #1719, opened 2021-03-23
refactor(api): adjust dataset request handlers to more up-to-date api patterns

### LINKED ISSUE #1712: some API endpoints seem to be broken after conversion to dispatch
With our recent wave of refactoring we moved from RPC to an HTTP layer for communicating with a running node. We also moved to a new dispatch paradigm. Since then, some of our endpoints no longer work which might come from several different issues. 

Upon quick inspection 1 issue that comes up is that routes are automatically generated by dispatch and they usually include a trailing slash. Gorrila.Mux is strict about slashes and we do differentiate routes based on whether they have a trailing `/` and path params.

ramfox:
## dataset api change
list of handlers that were wholesale replaced by `NewHTTPRequestHandler`:
- ListHandler POST GET
- ListRawHandler  POSt GET
- ValidateHandler POST GET
- ManifestHandler POST
- ManifestMissingHandler POST
- DAGInfoHandler POST
- ChangesHandle POST GET
- DiffHandler POST GET
- RenameHandler POST PUT

This means that any tests associated with the above handlers have also been removed.

All other handlers were adjusted to the most current pattern (... that isn't `NewHTTPRequestHandler`)

Known RPC bugs:
Get - rpc doesn't error but no response - The http api is expecting a `GetResponse` over the wire (since that is the return type of the `dataset.get` method). However, our api handler does further adjustments to the `GetResponse` depending on what the user requested in their inital request. Eg, if "meta" is selected, only the meta section is returned. There is pagination if the body was requested, etc. So, unlike most of our endpoints, our api response is not the same as our lib response params. When Dispatch tries to decode the response into a `GetResponse`, it can't because the response isn't shaped like `GetResponse`. We may need to unify the possible get responses & figure out how to represent all variations in a `GetResponse`, and have the `lib.Get` method return only what is needed, rather than returning the entire dataset and having the api and cmd layers narrow down what needs to be shown.
list of RPC bugs created here: https://github.com/qri-io/qri/issues/1721

closes #1712


cmd/print.go
```diff
@@ -245,7 +245,7 @@ func PrintProgressBarsOnEvents(w io.Writer, bus event.Bus) {
 	progress := map[string]*mpb.Bar{}
 
 	if bus == nil {
-		log.Errorf("event bus is nil")
+		log.Debugf("event bus is nil")
```
ramfox, line 248: this was logging everytime we ran an rpc call... moved it to debug for better user experience

lib/api.go
```diff
@@ -108,36 +108,36 @@ const (
...
-	AEPreview = APIEndpoint("/preview/{path:.*}")
+	AEPreview = APIEndpoint("/preview")
 
 	// fsi endpoints
 
 	// AEStatus returns the filesystem dataset status
 	AEStatus = APIEndpoint("/status")
 	// AEWhatChanged returns what changed for a specific commit
-	AEWhatChanged = APIEndpoint("/whatchanged/{path:.*}")
+	AEWhatChanged = APIEndpoint("/whatchanged")
```
ramfox, line 124: between the AttributesSet changes & the way we now register routes with ref handlers, we no longer need this path suffix.

lib/http.go
```diff
@@ -270,9 +270,7 @@ func UnmarshalParams(r *http.Request, p interface{}) error {
 			// this avoids resolving on empty body requests
 			// and tries to handle it almost like a GET
 			if err != io.EOF {
-				if err := json.NewDecoder(body).Decode(p); err != nil {
-					return err
-				}
+				return json.NewDecoder(body).Decode(p)
```
ramfox, line 273: If we POST w/ a json body, expect the params to contain everything we need. Necessary for the dispatch/http-rpc flow.

api/api_test.go
```diff
@@ -239,15 +239,18 @@ func TestServerReadOnlyRoutes(t *testing.T) {
...
 		{"PUT", "/save", 403},
 		{"POST", "/remove", 403},
 		{"DELETE", "/remove", 403},
 		{"POST", "/rename", 403},
 		{"PUT", "/rename", 403},
-		{"POST", "/diff", 403},
-		{"GET", "/diff", 403},
+		// TODO(ramfox): when `AEDiff` endpoint was refactored to use `NewHTTPHandlerFunc`, we lost
+		// {"POST", "/diff", 403},
+		// {"GET", "/diff", 403},
```
b5, line 253: I don't think we need to TODO these. I've filed #1723 to give a little direction to the `readOnly` situation

cmd/print.go
b5: yep. 99% of logging should be `Debugf`

lib/dispatch.go
```diff
@@ -99,7 +99,12 @@ func (inst *Instance) Dispatch(ctx context.Context, method string, param interfa
 				out := reflect.New(c.OutType)
 				res = out.Interface()
 			}
-			err = inst.http.CallMethod(ctx, c.Endpoint, c.Verb, param, res)
+			// TODO(ramfox): dispatch is still unable to give enough details to the url
+			// (because it doesn't know how or what param information to put into the url or query)
+			// for it to reliably use GET. All POSTs w/ content type application json work, however.
+			// we may want to just flat out say that as an RPC layer, dispatch will only ever use
+			// json POST to communicate.
+			err = inst.http.CallMethod(ctx, c.Endpoint, "POST", param, res)
```
b5, line 107: Based on this comment, I'm tempted to hold on approving this PR until we get @dustmop to weigh in on #1720 and the shift to thinking of our API as RPC + REST. If we agree on the approach we can remove the `"POST"` argument from `inst.http.CallMethod` entirely.

api/datasets.go
```diff
@@ -34,179 +33,87 @@ func NewDatasetHandlers(inst *lib.Instance, readOnly bool) *DatasetHandlers {
...
-			return
-		}
-		h.listHandler(w, r)
-	default:
-		util.NotFoundHandler(w, r)
-	}
-}
+// SaveHandler is a dataset save/update endpoint
+func (h *DatasetHandlers) SaveHandler(routePrefix string) http.HandlerFunc {
+	handleChanges := h.saveHandler(routePrefix)
```
b5, line 38: `handleChanges` -> `handleSave`, but instead of doing that, I'd prefer we refactor this `SaveHandler` + `saveHandler` into a single method that inlines the `saveHandler` method body into the exported `SaveHandler`.

Instead of a `switch` on the HTTP method, `if r.Method != http.MethodPut && r.Method != http.MethodPost` early-return an error. with bad HTTP verbs handled the rest of the function can be the contents of `saveHandler`. Whenever we have these wrapping funds that just check for one or two HTTP verbs & then call an unexported method, we should combine them

api/datasets.go
```diff
@@ -34,179 +33,87 @@ func NewDatasetHandlers(inst *lib.Instance, readOnly bool) *DatasetHandlers {
...
-	default:
-		util.NotFoundHandler(w, r)
 	}
 }
 
-// ChangesHandler is the endpoint for showing the changes between two datasets
-func (h *DatasetHandlers) ChangesHandler(routePrefix string) http.HandlerFunc {
-	handleChanges := h.changesHandler(routePrefix)
+// GetHandler is a dataset single endpoint
+func (h *DatasetHandlers) GetHandler(routePrefix string) http.HandlerFunc {
```
b5, line 65: Same comment about refactoring `getHandler` into `GetHandler`

api/datasets.go
```diff
@@ -34,179 +33,87 @@ func NewDatasetHandlers(inst *lib.Instance, readOnly bool) *DatasetHandlers {
...
-func (h *DatasetHandlers) RenameHandler(w http.ResponseWriter, r *http.Request) {
-	switch r.Method {
-	case http.MethodPost, http.MethodPut:
-		h.renameHandler(w, r)
-	default:
-		util.NotFoundHandler(w, r)
-	}
-}
+// PullHandler is an endpoint for creating new datasets
+func (h *DatasetHandlers) PullHandler(routePrefix string) http.HandlerFunc {
```
b5, line 93: ditto. Refactor `pullHandler` into `PullHandler`

api/datasets.go
```diff
@@ -34,179 +33,87 @@ func NewDatasetHandlers(inst *lib.Instance, readOnly bool) *DatasetHandlers {
...
-	switch r.Method {
-	case http.MethodPost:
-		h.manifestMissingHandler(w, r)
-	default:
-		util.NotFoundHandler(w, r)
-	}
-}
+// UnpackHandler unpacks a zip file and sends it back as json
+func (h *DatasetHandlers) UnpackHandler(routePrefix string) http.HandlerFunc {
+	handleChanges := h.unpackHandler(routePrefix)
```
b5, line 108: ditto, inline `unbackHandler` into `UnpackHandler`

api/datasets.go
```diff
@@ -372,217 +238,146 @@ func (h *DatasetHandlers) replyWithGetResponse(w http.ResponseWriter, r *http.Re
...
-func (h *DatasetHandlers) diffHandler(w http.ResponseWriter, r *http.Request) {
-	params := &lib.DiffParams{}
-	err := lib.UnmarshalParams(r, params)
-	if err != nil {
-		util.WriteErrResponse(w, http.StatusBadRequest, err)
-		return
-	}
+func (h *DatasetHandlers) peerListHandler(routePrefix string) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		log.Info(r.URL.Path)
```
b5, line 243: Stray comment, delete.

b5 (commented):
looking good. I have a few nits, and want to get some consensus before we push forward with the POST-only RPC assumption we're discussing

lib/dispatch.go
ramfox: Refactoring on top of this will be very easy, this isn't forcing anything on future decisions this just gets us working rpc right now & documents the problem. I don't think it should keep us from merging.

Either we go w/ dispatch always POSTing json, or dispatch (or callable or the params themselves, depending on how we implement) having more details on each endpoint and we have some mechanism for dispatch adding refs & form vals into the url. The former is how we are set up right now, the later is a larger refactor regardless.

api/datasets.go
ramfox: omg I thought `handleChanges` was some pattern we were using, because I saw it outside of the `ChangeHandler`... but now that seems really obvious hahah.

b5 (approved):
LGTM!

Issue #1723, opened 2021-03-24
Remove Support for `readOnly` API mode, use auth tokens for privileged access


b5:
In the next release we should remove support for `api.ReadOnly: true` entirely, marking the configuration field as deprecated. qri commands should fail if `api.ReadOnly` is set to true with an error message about configuration.

ReadOnly mode was a half-baked effort to get around not having a proper authorization scheme. Once we have a mechanism for authentication in place, we should remove it entirely.

cc @Arqu, as this impacts cloud more than anything.

PR #1724, opened 2021-03-25
refactor(lib): convert peer methods to use Dispatch pattern

### LINKED ISSUE #1722: refactor(lib, api): convert peers lib methods to use dispatch, adjust handler methods accordingly
- [ ] List (PeersHandler)
- [ ] ConnectedIPFSPeers (ConnectionsHandler)
- [ ] ConnectedQriProfiles (QriConnectionsHandler)
- [ ] ConnectToPeer (ConnectToPeer)
- [ ] DisconnectFromPeer (ConnectToPeer)
- [ ] Info (PeerHandler)

- added `inst.Peer()` for returning `PeerMethods`
- added `PeerMethod.Name` & `PeerMethod.Attributes`
- adding "/disconnect" & "/disconnect/{path:.**}" endpoints
- renamed some functions for clarity:
    - "ConnectToPeer" -> "Connect" (so it will read inst.Peer().Connect())
    - "DisconnectFromPeer" -> "Disconnect" (so it will read inst.Peer().Disconnect())
    - "ConnectedIPFSPeers" -> "Connections" (so it will read inst.Peer().Connections())
- converted lib methods to dispatch
- added `ConnectionsParams`, but limit and offset either don't do anything or aren't both used, should document in the code both in the params and at the lib function

to do:
- remove NewPeerMethods in favor of inst.Peers()
- remove api handlers & just use `NewHTTPRequestMethods`
- test over cmd api & rpc

ramfox:
closes #1722


lib/lib.go
```diff
@@ -823,6 +823,11 @@ func (inst *Instance) Dataset() DatasetMethods {
 	return DatasetMethods{inst: inst}
 }
 
+// Peer returns the PeerMethods that Instance has registered
```
dustmop, line 826: Please keep these *Methods functions alphabetized.

lib/peers.go
```diff
@@ -23,8 +23,20 @@ type PeerMethods struct {
 	inst *Instance
```
dustmop, line 23: Change `inst *Instance` to `d dispatcher`. Now that only Dispatch is called, there's no need to keep a reference to the entire Instance.

api/api.go
```diff
@@ -227,13 +227,12 @@ func NewServerRoutes(s Server) *mux.Router {
...
-	m.Handle(lib.AEConnect.String(), s.Middleware(ph.ConnectToPeerHandler))
-	m.Handle(lib.AEConnectAlt.String(), s.Middleware(ph.ConnectToPeerHandler))
-	m.Handle(lib.AEConnections.String(), s.Middleware(ph.ConnectionsHandler))
-	m.Handle(lib.AEConnectionsQri.String(), s.Middleware(ph.QriConnectionsHandler))
+	m.Handle(lib.AEPeers.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "peer.list"))).Methods(http.MethodPost)
+	m.Handle(lib.AEPeer.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "peer.info"))).Methods(http.MethodPost)
+	m.Handle(lib.AEConnect.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "peer.connect"))).Methods(http.MethodPost)
+	m.Handle(lib.AEDisconnect.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "peer.disconnect"))).Methods(http.MethodPost)
+	m.Handle(lib.AEConnections.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "peer.connections"))).Methods(http.MethodPost)
+	m.Handle(lib.AEConnectedQriProfiles.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "peer.connectedqriprofiles"))).Methods(http.MethodPost)
```
ramfox, line 235: using `NewHTTPRequestHandler` for each peer method!

api/api_test.go
```diff
@@ -236,9 +236,6 @@ func TestServerReadOnlyRoutes(t *testing.T) {
 		{"PUT", "/profile/photo", 403},
 		{"POST", "/profile/poster", 403},
 		{"PUT", "/profile/poster", 403},
-		{"GET", "/peers", 403},
-		{"GET", "/peers/test_peer", 403},
-		{"GET", "/connections", 403},
```
ramfox, line 241: `NewHTTPRequestHandler` does not support read only & we are refactoring away from this pattern as per https://github.com/qri-io/qri/issues/1723 anyhow

lib/api.go
```diff
@@ -54,15 +54,15 @@ const (
...
-	// AEConnectAlt initiates an explicit connection to a peer
-	AEConnectAlt = APIEndpoint("/connect/{path:.*}")
+	// AEDisconnect closes an explicit connection to a peer
+	AEDisconnect = APIEndpoint("/disconnect")
 	// AEConnections lists qri & IPFS connections
 	AEConnections = APIEndpoint("/connections")
-	// AEConnectionsQri lists qri profile connections
-	AEConnectionsQri = APIEndpoint("/connections/qri")
+	// AEConnectedQriProfiles lists qri profile connections
+	AEConnectedQriProfiles = APIEndpoint("/connections/qri")
```
ramfox, line 65: This `APIEndpoint` var name doesn't follow the other pattern we seem to be following. The var endpoint name usually matches the lib method name, while the actual endpoint string varies.

Arqu (approved):
Nothing really concrete to add. Agree on the `AEConnectedQriProfiles` being a bit unfriendly. We already have a discussion I think around this and the consensus was that we have `/connections` and then we had the flavors with `/ipfs`, `/qri`... as sub routes.
The rename is good though for less mental gymnastics in the code.

PR #1725, opened 2021-03-25
fix(dispatch): Use dispatcher interface for DatasetMethods

dustmop:
New DatasetMethods don't use anything on the Instance aside from Dispatch. Switch them all to use the dispatcher interface, in order to guarantee that this fact holds.


b5 (approved):
lovely

PR #1726, opened 2021-03-25
feat(dispatch): Abs paths on inputs to dispatch methods

dustmop:
Background: when `qri connect` is running, it holds an exclusive lock, and other invocations of qri will send their commands to that connected process using RPC. The two processes may be running in different directories, so it is necessary to change file paths to be absolute before sending the params across RPC. Before, we would do this explicitly, which was error-prone and
easy to forget about, since the broken behavior would not happen in the common case.

The proposed solution implemented here adds tags to structs that are passed to methods. Most use `qri:"fspath"` to specify fields that are file system paths. Diff is special and uses `qri:"dsrefOrFspath"` because its inputs can be either dsrefs or file paths. Adding this behavior for new commands is simple; it just requires adding this tag.

An alternative would be to add a type for qfs.FilePath that does this behavior itself. However, that requires users of qri to know about this type, and FilePath types are typically quite tricky to fully implement. This solution is simpler and adds less maintenance vs creating our own path type,
though it is admittedly a bit magical. Personally, I don't think it's that bad, as it matches the standard use case of tags for serialization, if we view fully absolute paths being sent over RPC as a type of data marshalling.

Plus, the Diff inputs act quite special. Simply creating a qfs.FilePath would not be sufficient to handle Diff's parameters, and fixing Diff in some way would probably add more complication than its worth.

### DIFF
--- a/lib/dispatch.go
+++ b/lib/dispatch.go
@@ -8,7 +8,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/qri-io/qfs"
 	"github.com/qri-io/qri/auth/token"
+	"github.com/qri-io/qri/dsref"
 	"github.com/qri-io/qri/profile"
 )
 
@@ -134,6 +136,9 @@ func (inst *Instance) Dispatch(ctx context.Context, method string, param interfa
 		if err != nil {
 			return nil, nil, err
 		}
+
+		// Handle filepaths in the params by calling qfs.Abs on each of them
+		param = normalizeInputParams(param)
 
 		// Construct the parameter list for the function call, then call it
 		args := make([]reflect.Value, 3)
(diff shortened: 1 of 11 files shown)


lib/dispatch.go
```diff
@@ -411,3 +416,50 @@ func dispatchReturnError(got interface{}, err error) error {
 	}
 	return err
 }
+
+// normalizeInputParams will look at each field of the params, and modify filepaths so that
+// they are absolute paths, making them safe to send across RPC to another process
+func normalizeInputParams(param interface{}) interface{} {
```
b5, line 422: What do you think about moving these to `params.go` in this same package? While this is only called within dispatch, it feels like a separate concept.

lib/dispatch.go
```diff
@@ -411,3 +416,50 @@ func dispatchReturnError(got interface{}, err error) error {
...
+		typ = typ.Elem()
+		val = val.Elem()
+	}
+	if typ.Kind() == reflect.Struct {
+		num := typ.NumField()
+		for i := 0; i < num; i++ {
+			tfield := typ.Field(i)
+			vfield := val.Field(i)
+			qriTag := tfield.Tag.Get("qri")
+			if qriTag == "fspath" || qriTag == "dsrefOrFspath" {
```
b5, line 435: flip these into constants & add a comment describing the `qri` struct tag:

```
qri struct tags augment the way the qri runtime interprets a field
```

then add a comment on each constant describing how it will augment normalization. Your PR description lays both of these out nicely 😄

b5 (approved):
Seeing qri struct tags in practice has me sold.

lib/dispatch.go
dustmop: Sounds good to me, agreed it's somewhat removed from dispatch.

lib/dispatch.go
dustmop: 👍

PR #1729, opened 2021-03-30
refactor(sql): Dispatch for sql. Move resolver source to scope

dustmop:
Use the Dispatch pattern for sql methods.

Instead of source being a parameter that is passed down the callstack, until code calls LoadDataset or ParseAndResolveRef, attach source to the scope. This will let us remove the source from the interface, since the callsite of LoadDataset doesn't care about source, the concrete implementation of the Loader does. Clients of lib can configure the source just-in-time by using WithSource on the Instance.

Also, discovered that the --offline flag for sql was broken, so resolution would never be local-only. The cause was ParseResolveFunc always using the defaultResolver. This type of cleanup should make these sorts of bugs much harder to cause.

Finally, remove the ResolverForMode function from scope. Only lib.Save was using it, but in actually it only needs a localResolver, so it uses that instead now.

### DIFF
--- a/lib/integration_test.go
+++ b/lib/integration_test.go
@@ -20,8 +20,6 @@ import (
 )
 
 func TestTwoActorRegistryIntegration(t *testing.T) {
-	t.Skip("TODO(dustmop): meaning of source changed, this needs to be fixed")
-
 	tr := NewNetworkIntegrationTestRunner(t, "integration_two_actor_registry")
 	defer tr.Cleanup()
 
@@ -70,7 +68,7 @@ func TestTwoActorRegistryIntegration(t *testing.T) {
 	PushToRegistry(tr.Ctx, t, nasim, ref.Alias())
 
 	// 7. hinshun logsyncs with the registry for world bank dataset, sees multiple versions
-	_, err = hinshun.Dataset().Pull(tr.Ctx, &PullParams{LogsOnly: true, Ref: ref.String()})
+	_, err = hinshun.WithSource("network").Dataset().Pull(tr.Ctx, &PullParams{LogsOnly: true, Ref: ref.String()})
 	if err != nil {
 		t.Errorf("cloning logs: %s", err)
 	}
@@ -437,7 +435,7 @@ func SearchFor(ctx context.Context, t *testing.T, inst *Instance, term string) [
 
 func Pull(ctx context.Context, t *testing.T, inst *Instance, refstr string) *dataset.Dataset {
 	t.Helper()
-	res, err := inst.Dataset().Pull(ctx, &PullParams{Ref: refstr})
+	res, err := inst.WithSource("network").Dataset().Pull(ctx, &PullParams{Ref: refstr})
 	if err != nil {
 		t.Fatalf("cloning dataset %s: %s", refstr, err)
 	}
--- a/lib/integration_test.go
+++ b/lib/integration_test.go
@@ -20,6 +20,8 @@ import (
 )
 
 func TestTwoActorRegistryIntegration(t *testing.T) {
+	t.Skip("TODO(dustmop): meaning of source changed, this needs to be fixed")
+
 	tr := NewNetworkIntegrationTestRunner(t, "integration_two_actor_registry")
 	defer tr.Cleanup()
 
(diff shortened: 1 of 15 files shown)


lib/sql.go
```diff
@@ -3,79 +3,59 @@ package lib
...
 
-// SQLQueryParams defines paremeters for the exec Method
-// ExecParams provides parameters to the execute command
+// SQLQueryParams defines paremeters for running a SQL query
 type SQLQueryParams struct {
-	Query        string
-	OutputFormat string
-	ResolverMode string
+	Query  string
+	Format string
```
b5, line 28: Not a massive fan of the  `OutputFormat`  -> `Format` switch. To me the test in `lib/integration_test.go` now reads kinda like the format of the _query_ is JSON.

That said, this might be in the name of aligning field names. Care to clarify why you made the change?

lib/integration_test.go
```diff
@@ -20,6 +20,8 @@ import (
 )
 
 func TestTwoActorRegistryIntegration(t *testing.T) {
+	t.Skip("TODO(dustmop): meaning of source changed, this needs to be fixed")
```
b5, line 23: This is an important test. Can't skip. let's instead figure out what's going on before merging

b5 (commented):
Big fan of these changes. I'd like your input on a `WithConfig` method, and we need to un-skip an integration test

lib/sql.go
dustmop: Yes, done to align the field name with other commands. For example `list` and `diff` both use `Format` to talk out about their output format, and all three of these fall into the "aggregation" category of commands. It also matches the command-line flag `--format` already in use for the command.

I sort of see what you mean about the ambiguity of input/output format, but the context makes it clearer.

lib/integration_test.go
dustmop: The spike at https://github.com/qri-io/qri/commit/c5c65b84afca781587251da5cb21a04350b05c36 was super helpful. Pushed another commit that fixes the test.

What was happening is that the implementation of `dataset.pull` overrides the source with "network" if a blank source is passed in. However, this change is moving to a world where the `scope` locks-in the source, and implementations should not be allowed to change it.

I see 3 options going forward:

1) Require users of `pull` to always do this:
```
instance.WithSource("network").Pull(...)
```
This is the change I went with because it seems easiest. Now `cmd` calls WithSource('network'). In the future we can add a method to scope to get (but not set) the source, and pull can fail if it's not "network".

2) Add a method to scope to override the source. pull can call this to set the source. I'm not a fan of this option because it goes against what this PR is trying to do, but if it's only needed in a few places it's not that bad.

3) Add a field to Attributes to set the default source for commands. This will usually be empty, but can be "network" for pull. I don't think we should do this until we find more places that need it.

b5: Very much agree this is the right approach for now. This PR makes the need for the override explicit, which I'm a huge fan of. Option 3 (adding a field to a method set's returned Attributes) seems like a reasonable design choice, but I'm with you that it's too soon to make such a change

lib/lib.go
```diff
@@ -833,11 +833,40 @@ func (inst *Instance) Peer() PeerMethods {
...
+		source: source,
+		inst:   inst,
+	}
+}
+
+// InstanceSourceWrap is a wrapped instance with an explicit resolver source added
+type InstanceSourceWrap struct {
+	source string
+	inst   *Instance
+}
```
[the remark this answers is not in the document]
b5: Yeah in my head `WithConfig` would construct shallow-copy `*Instance` . Specifically a method like `WithConfig` should  _not_ construct new copies of subsystems. With access to subsystems governed by scope I think that's a safe approach to take.

As for the config change, I vey much see your point on adding a `Source` field to config, and think your mental model of it being an "Option" according to the parlance laid out here: https://github.com/qri-io/qri/blob/9f36fc56b1bba27d3875f4418ec73020c4efad40/lib/lib.go#L63-L70

To capture setting both "Config" and "Options", we could flip to a `WithOptions` method instead of a `WithConfig`, and use the `OptConfig` function to override configurations.

But now we're actively modeling a different issue, so I'm good to merge this as-is & think we should move this discussion to an issue.

b5 (approved):
LGTM!

PR #1730, opened 2021-03-31
refactor(lib, api, cmd): adjust log methods to use dispatch pattern

### LINKED ISSUE #1728: refactor(lib): convert log methods to use dispatch pattern
- [ ] add `Name` & `Attributes` methods on `LogMethods`
- [ ] add `Log` method to Instance
- [ ] add `logImpl`
- [ ] add `Log` `Logbook`, `PlainLogs`, `LogbookSummary` methods to logImpl, that take a scope
- [ ] dispatch-ify `Log` `Logbook`, `PlainLogs`, and `LogbookSummary` methods
- [ ] register `Log` with dipatch
- [ ] replace `NewLogMethods` with `inst.Log`
- [ ] replace inst in `LogMethod` with `d dispatcher`
- [ ] figure out if `LogHandler` can be replaced with `NewHTTPRequestMethod` (specifically look at error responses)
- [ ] convert `LogbookHandler`, `PlainLogs`, and `LogbookSummary` methods with `NewHTTPRequestMethods`
- [ ] rename `lib.Log` to `lib.History` `/history`
- [ ] rename `lib.Logbook` to `lib.Logs` `/log`
- [ ] rename `lib.PlainLogs` to `lib.Logbook` `/logbook`

ramfox:
closes #1728 

The most controversial thing in this pr is I did some cleaning/renaming of log method names and endpoints. The endpoints are less important, since most of them are going away when we refactor to the 0.10.0 qri api standard.

However, the lib method names I can see being a spicy topic.

Why rename? Because we had a few wires crossed and used the same term to mean different things in different cases. This rename ensure that when we talk about `Logbook` at the `LogMethod` and `api` level, we are referring to the entire logbook. When we talk about `Log` or `Entries`, we are talking about a list of log entries. We no longer cross wires and refer to a commit history as a `Log`, as that is reserved for the list of log entries. 

### lib method name changes
From `lib.Log` to `lib.History` - this function does not actually return a list of logs, it returns a list of version infos, aka the history rather than the list of log entries

From `lib.Logbook` to `lib.Entries` - this function does not actually return a Logbook, it returns a list of log entries

From `lib.PlainLogs` to `lib.RawLogbook` - this function returns an entire logbook as json. The `cmd` layer referred to this as `raw`. `Raw` seemed more intuitive (to me) and like something someone using `lib` as a library would grok easier. But this one is definitely up for debate. Regardless if we use `Plain` or `Raw` as a prefix, we should use `Logbook` as the suffix, to distinguish this method as returning the entire logbook (rather than a list of log entries).

### api changes
The `/logbook` endpoint (which is on the slate for deprecation), now returns a logbook, rather than a list of log entries.

The `/log` now returns a list of log entries, which is consistent with the  0.10.0 qri api


lib/api.go
```diff
@@ -148,12 +148,12 @@ const (
 
 	// AEHistory returns dataset logs
 	AEHistory = APIEndpoint("/history")
-	// AELogbook lists log entries for actions taken on a given dataset
-	AELogbook = APIEndpoint("/logbook")
+	// AEEntries lists log entries for actions taken on a given dataset
+	AEEntries = APIEndpoint("/log")
+	// AERawLogbook returns the full logbook encoded as human-oriented json
+	AERawLogbook = APIEndpoint("/logbook")
```
ramfox, line 154: read the main pr comment why I think these are good changes:
https://github.com/qri-io/qri/pull/1730#issue-604927842

lib/api.go
```diff
@@ -148,12 +148,12 @@ const (
...
 	// AELogbookSummary returns a string overview of the logbook
 	AELogbookSummary = APIEndpoint("/logbook/summary")
-	// AELogs returns the full logbook encoded as human-oriented json
-	AELogs = APIEndpoint("/logs")
```
ramfox, line 156: read the main pr comment why I think these are good changes:
https://github.com/qri-io/qri/pull/1730#issue-604927842

lib/log.go
```diff
@@ -17,22 +16,26 @@ import (
...
-func NewLogMethods(inst *Instance) *LogMethods {
-	return &LogMethods{
-		inst: inst,
+// Attributes defines attributes for each method
+func (m LogMethods) Attributes() map[string]AttributeSet {
+	return map[string]AttributeSet{
+		"history":        {AEHistory, "POST"},
+		"entries":        {AEEntries, "POST"},
+		"rawlogbook":     {denyRPC, ""},
+		"logbooksummary": {denyRPC, ""},
```
ramfox, line 33: read the main pr comment for why I think these are good method names:
https://github.com/qri-io/qri/pull/1730#issue-604927842

api/testdata/api.snapshot
```diff
@@ -5062,12 +5062,12 @@ HTTP/1.1 200 OK
 Connection: close
 
 /* snapshot: TestFSIHandlers init case 0: POST / */
-HTTP/1.1 500 Internal Server Error
+HTTP/1.1 400 Bad Request
```
Arqu, line 5065: This case means we don't properly parse/unmarshal the structs or the test is set up wrong. Would be nice to actually fix it as these will be useful for keeping record/testing of how things work until we transition to the new JSON based test framework.

lib/log.go
```diff
@@ -82,16 +85,93 @@ func (p *LogParams) UnmarshalFromRequest(r *http.Request) error {
 }
 
 // Log returns the history of changes for a given dataset
```
Arqu, line 87: Comment needs updating.

lib/log.go
```diff
@@ -82,16 +85,93 @@ func (p *LogParams) UnmarshalFromRequest(r *http.Request) error {
...
+	}
+	return nil, dispatchReturnError(got, err)
+}
+
+// RawLogbookParams enapsulates parameters for the RawLogbook methods
+type RawLogbookParams struct {
+	// no options yet
+}
+
+// RawLogbook is an alias for a human representation of a plain-old-data logbook
```
Arqu, line 149: `RawLogbook` > `RawLogs`

Arqu (approved):
IMHO, don't think this is spicy, this is RIGHT. I had troubles following the full logic on my own on the previous pass.

Left some nitpicks but LGTM.

b5 (approved):
Agree that these changes are either not spicy, or good spice. Well seasoned PR

api/testdata/api.snapshot
ramfox: In `InitDataset`, the first place we check the Ref is `dsref.IsValidName`, and return `dsref.ErrDescribeValidName`. This test is supposed to be checking that we error when there is no Ref. The status code change occurred because `NewHTTPRequestMethod` now uses `RespondWithErr` rather than `WriteErrResponse`, with every error returning `500`.

We should probably have a `dsref.ErrEmtpyRef` error and `InitDataset` should be responding with that instead (and a `422` error probably?).

In `RespondWithErr` we have `dsref.ErrBadCaseShouldRename`, `dsref.ErrDescribeValidName`, and `dsref.ErrDescribeValidUsername`. Do those make sense as `400 Bad Request` errors, or should they be `422 Unprocessable Entity`?

dustmop:
Missed the review on this yesterday, but wanted to double down that all these renames are good! 👍

PR #1733, opened 2021-04-01
fix(search): Dispatch for search

dustmop:


api/search.go
```diff
@@ -35,12 +34,15 @@ func (h *SearchHandlers) searchHandler(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
-		log.Infof("search error: %s", err.Error())
-		util.WriteErrResponse(w, http.StatusBadRequest, err)
+		util.RespondWithError(w, err)
+		return
+	}
+	res, ok := got.([]lib.SearchResult)
+	if !ok {
+		util.RespondWithDispatchTypeError(w, got)
 		return
```
ramfox, line 45: This handler can be replaced w/ `NewHTTPRequestHandler` in `api/api.go` & then we can remove this file! (Any handlers that don't adjust the return from Dispatch (and can be expressed w/ "Content-Type application/json", so no "text/csv") can be replaced by `NewHTTPRequestHandler`

cmd/log.go
```diff
@@ -211,7 +211,7 @@ func (o *LogbookOptions) Complete(f Factory, args []string) (err error) {
 	return err
 }
 
-// Logbook executes the Logbook command
+// LogEntries gets entries from the logbook
```
ramfox, line 214: 🙌

ramfox (approved):
Just take a look at implementing `SearchHandler` as a `NewHTTPRequestHandler` directly in `api/api.go` & removing the `api/search.go` file (an adjust/remove any associated tests), otherwise this looks great!

Issue #1731, opened 2021-03-31
Qri API v0.10.0 tracker


ramfox:
# Qri 0.10.0 API Tracker

### TODO:
- [x] Dispatch-ify project
	- [x] profile methods
	- [x] registry methods
	- [x] remote methods
	- [x] render methods
	- [x] search methods
	- [x] sql methods
	- [x] transform methods
	- [x] remaining dataset methods
- [x] unify params <sup>[1](#params)</sup> #1732
	- [x] `ListParams` - struct contains `Limit`, `Offset`, `All`
	- [x] `RefParams` - struct contains `Ref`, `Filter`<sup>[2](#filter)</sup>
	- [x] `RefListParams` - struct contains `ListParams`, `RefParams`
	- [x] Audit ALL lib method params, have them extend the above if they can
- [x] Simplify api & Remove Lingering code that does not support `RPC` style api
	- [x] remove `UnmarshalFromRequest`?
	- [x] convert all `api` handler methods to `lib.NewHTTPRequestHandlers`
	- [x] align the API with the spec #1734 
	- [x] add `Instance.GiveAPIServer()` that returns a mux with all the appropriate endpoints handled, aka only the endpoints that are expected in `0.10.0` <sup>[3](#denyRPC)</sup>
- [x] Adding back RESTful api "Sugar"
	- [x] add `GetHandler` that implements RESTful api expectations
	- [x] /ds/get/{username}/{name}/at/{path}/{component}.{encoding} w/ `RefListParams`
```
	# REST-style:
	/ds/get/{username}/{name}
	$ curl http://localhost:2503/ds/get/b5/world_bank_population
	$ curl http://localhost:2503/ds/get/b5/world_bank_population/meta
	$ curl http://localhost:2503/ds/get/b5/world_bank_population/body.csv?all=true
	$ curl http://localhost:2503/ds/get/b5/world_bank_population/body.csv?limit=25&offset=0
	$ curl -H "Accept: text/csv" http://localhost:2503/ds/get/b5/world_bank_population/body
	$ curl -H "Accept: application/zip" http://localhost:2503/ds/get/b5/world_bank_population
```
- [x] Tests!
The tests will transition to a JSON based test framework where we have a central set of expected JSON inputs and outputs. This should make the maintenance of the tests easier as well as more portable in terms of other clients relying on them or devs navigating/reusing them.

- [x] Automated api docs
- [ ] release notes
[...]

PR #1734, opened 2021-04-01
refactor(api): aligning our API with spec

Arqu:
Aiming to get us closer to the API spec contained in https://github.com/qri-io/qri/issues/1731, I've refactored some of the code to move to POST only, cleaned up the route definitions and rearranged most of the routes in the same order as in our tables. The main reason for rearranging is to make the follow up steps more explicit:
- Which handler methods need to move into which handler
- Highlight unfinished or missing routes
- Highlight routes that are no longer needed or otherwise misplaced

I'll run a round of self commenting on the PR just to add some color to it. Looking for some input before diving into the code move.


api/api.go
```diff
@@ -215,103 +215,113 @@ func NewServerRoutes(s Server) *mux.Router {
 	m.Use(refStringMiddleware)
 	m.Use(token.OAuthTokenMiddleware)
 
+	dsh := NewDatasetHandlers(s.Instance, cfg.API.ReadOnly)
```
Arqu, line 218: Moved up to de-clutter as the rows are hard to read as they are.

api/api.go
```diff
@@ -215,103 +215,113 @@ func NewServerRoutes(s Server) *mux.Router {
...
-		m.Handle(lib.AERemoteLogSync.String(), s.Middleware(remh.LogsyncHandler))
-		m.Handle(lib.AERemoteRefs.String(), s.Middleware(remh.RefsHandler))
+	if !cfg.API.DisableWebui {
+		m.Handle(lib.AEWebUI.String(), s.Middleware(WebuiHandler))
 	}
 
-	dsh := NewDatasetHandlers(s.Instance, cfg.API.ReadOnly)
+	// aggregate endpoints
 	m.Handle(lib.AEList.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.list"))).Methods(http.MethodPost, http.MethodGet)
 	m.Handle(lib.AEListRaw.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.listrawrefs"))).Methods(http.MethodPost, http.MethodGet)
```
Arqu, line 239: This needs to merge into `AEList`

api/api.go
```diff
@@ -215,103 +215,113 @@ func NewServerRoutes(s Server) *mux.Router {
...
 	m.Handle(lib.AEPeerList.String(), s.Middleware(dsh.PeerListHandler(lib.AEPeerList.String())))
```
Arqu, line 240: This should also merge into `AEList`

api/api.go
```diff
@@ -215,103 +215,113 @@ func NewServerRoutes(s Server) *mux.Router {
...
+	m.Handle(lib.AEChanges.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.changereport"))).Methods(http.MethodPost, http.MethodGet)
+
+	// access endpoints
+
+	// automation endpoints
+	m.Handle(lib.AEApply.String(), s.Middleware(tfh.ApplyHandler(lib.AEApply.NoTrailingSlash())))
+
+	// dataset endpoints
+	routeParams = newrefRouteParams(lib.AEWhatChanged, false, false, http.MethodGet, http.MethodPost)
+	handleRefRoute(m, routeParams, s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "fsi.whatchanged")))
```
Arqu, line 252: All the `handleRefRoute` mean that these routes include all the *sugar* which we should drop and lock to `POST` only. Only `/ds/get` should be an exception for now.

api/api.go
```diff
@@ -215,103 +215,113 @@ func NewServerRoutes(s Server) *mux.Router {
...
-	routeParams = newrefRouteParams(lib.AESave, false, false, http.MethodPost, http.MethodPut)
-	handleRefRoute(m, routeParams, s.Middleware(dsh.SaveHandler(lib.AESave.String())))
+	m.Handle(lib.AESQL.String(), s.Middleware(sqlh.QueryHandler))
+	m.Handle(lib.AEDiff.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.diff"))).Methods(http.MethodPost, http.MethodGet)
+	m.Handle(lib.AEChanges.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.changereport"))).Methods(http.MethodPost, http.MethodGet)
+
+	// access endpoints
+
+	// automation endpoints
+	m.Handle(lib.AEApply.String(), s.Middleware(tfh.ApplyHandler(lib.AEApply.NoTrailingSlash())))
```
Arqu, line 248: Move handler from `apply/transform` > `automation`.

api/api.go
```diff
@@ -215,103 +215,113 @@ func NewServerRoutes(s Server) *mux.Router {
...
 
-	dsh := NewDatasetHandlers(s.Instance, cfg.API.ReadOnly)
+	// aggregate endpoints
 	m.Handle(lib.AEList.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.list"))).Methods(http.MethodPost, http.MethodGet)
 	m.Handle(lib.AEListRaw.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.listrawrefs"))).Methods(http.MethodPost, http.MethodGet)
 	m.Handle(lib.AEPeerList.String(), s.Middleware(dsh.PeerListHandler(lib.AEPeerList.String())))
-	routeParams = newrefRouteParams(lib.AESave, false, false, http.MethodPost, http.MethodPut)
-	handleRefRoute(m, routeParams, s.Middleware(dsh.SaveHandler(lib.AESave.String())))
+	m.Handle(lib.AESQL.String(), s.Middleware(sqlh.QueryHandler))
+	m.Handle(lib.AEDiff.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.diff"))).Methods(http.MethodPost, http.MethodGet)
```
Arqu, line 242: `/diff` and `/changes` should move into it's own handler set `diff`

api/api.go
```diff
@@ -215,103 +215,113 @@ func NewServerRoutes(s Server) *mux.Router {
...
+	m.Handle(lib.AEChanges.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.changereport"))).Methods(http.MethodPost, http.MethodGet)
+
+	// access endpoints
+
+	// automation endpoints
+	m.Handle(lib.AEApply.String(), s.Middleware(tfh.ApplyHandler(lib.AEApply.NoTrailingSlash())))
+
+	// dataset endpoints
+	routeParams = newrefRouteParams(lib.AEWhatChanged, false, false, http.MethodGet, http.MethodPost)
```
Arqu, line 251: This should move `fsi` > `dataset`?

api/api.go
```diff
@@ -215,103 +215,113 @@ func NewServerRoutes(s Server) *mux.Router {
...
+	handleRefRoute(m, routeParams, s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "fsi.whatchanged")))
 	routeParams = newrefRouteParams(lib.AEGet, false, true, http.MethodGet, http.MethodPost)
 	handleRefRoute(m, routeParams, s.Middleware(dsh.GetHandler(lib.AEGet.String())))
+	m.Handle(lib.AERename.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.rename"))).Methods(http.MethodPost, http.MethodPut)
+	routeParams = newrefRouteParams(lib.AESave, false, false, http.MethodPost, http.MethodPut)
+	handleRefRoute(m, routeParams, s.Middleware(dsh.SaveHandler(lib.AESave.String())))
+	routeParams = newrefRouteParams(lib.AEPull, false, false, http.MethodPost, http.MethodPut)
+	handleRefRoute(m, routeParams, s.Middleware(dsh.PullHandler(lib.AEPull.NoTrailingSlash())))
+	routeParams = newrefRouteParams(lib.AEPush, false, false, http.MethodGet, http.MethodPost, http.MethodDelete)
```
Arqu, line 260: Push should also move into `dataset`?

api/api.go
```diff
@@ -215,103 +215,113 @@ func NewServerRoutes(s Server) *mux.Router {
...
+	// access endpoints
+
+	// automation endpoints
+	m.Handle(lib.AEApply.String(), s.Middleware(tfh.ApplyHandler(lib.AEApply.NoTrailingSlash())))
+
+	// dataset endpoints
+	routeParams = newrefRouteParams(lib.AEWhatChanged, false, false, http.MethodGet, http.MethodPost)
+	handleRefRoute(m, routeParams, s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "fsi.whatchanged")))
 	routeParams = newrefRouteParams(lib.AEGet, false, true, http.MethodGet, http.MethodPost)
 	handleRefRoute(m, routeParams, s.Middleware(dsh.GetHandler(lib.AEGet.String())))
```
Arqu, line 254: `/ds/log` is still TODO

api/api.go
```diff
@@ -215,103 +215,113 @@ func NewServerRoutes(s Server) *mux.Router {
...
+		remh := NewRemoteHandlers(s.Instance)
+		m.Handle(lib.AERemoteDSync.String(), s.Middleware(remh.DsyncHandler))
+		m.Handle(lib.AERemoteLogSync.String(), s.Middleware(remh.LogsyncHandler))
+		m.Handle(lib.AERemoteRefs.String(), s.Middleware(remh.RefsHandler))
+	}
 
-	m.Handle(lib.AEHistory.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "log.history"))).Methods(http.MethodPost)
+	// TODO(aqru): clear up these endpoints up to spec
+
+	m.Handle(lib.AELog.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "log.history"))).Methods(http.MethodPost)
```
Arqu, line 321: We've had a discussion around these, but I don't have all the details to spell them all out on how we will merge these/how they should behave/which we will just drop from the HTTP API.

api/api_test.go
```diff
@@ -173,111 +173,6 @@ func TestHealthCheck(t *testing.T) {
 	runHandlerTestCases(t, "health check", HealthCheckHandler, healthCheckCases, true)
 }
 
-func TestServerReadOnlyRoutes(t *testing.T) {
```
Arqu, line 176: This became useless and since we're going to remove readOnly anyways I simply moved it out of the way completely.

lib/fsi.go
```diff
@@ -30,16 +30,18 @@ func (m FSIMethods) Name() string {
...
-		"init":                  {AEInit, "POST"},
-		"caninitdatasetworkdir": {AECanInitDatasetWorkDir, "GET"},
-		"ensureref":             {AEEnsureRef, "POST"},
+		"write":                 {AEFSIWrite, "POST"},
+		"createlink":            {AEFSICreateLink, "POST"},
+		"unlink":                {AEFSIUnlink, "POST"},
+
+		// TODO(arqu): reorganize these handlers into appropriate packages
+		"whatchanged": {AEWhatChanged, "POST"}, // dataset.componentstatus
+		"ensureref":   {AEEnsureRef, "POST"},   // ??
```
Arqu, line 44: Unsure what the purpose of `ensureref` is.

lib/datasets.go
```diff
@@ -53,22 +53,29 @@ func (m DatasetMethods) Name() string {
 // Attributes defines attributes for each method
 func (m DatasetMethods) Attributes() map[string]AttributeSet {
 	return map[string]AttributeSet{
-		"changereport": {AEChanges, "POST"},
```
dustmop, line 56: So, others may disagree, but I don't think we should be changing the Attributes values for GET/POST, even though we're switching (possibly temporarily) to all POST requests all the time. The reason is both aspiration (where we want to get to, if we add more RESTful sugar), and also acts as documentation for whether the method is read-only or read-write. It'll be easier to adapt code in the future if we don't need to undo this modification.

Rather, the function `dispatchMethodCall` in `lib/dispatch.go` is already ignoring this value, and always calling `inst.http.CallMethod` with the "POST" verb. So we've already switched to a POST-only world, and these attributes reflect who the methods *could* act.

api/api.go
dustmop: Isn't this "/peer/list"? It's still there in https://github.com/qri-io/qri/pull/1697.

api/api.go
dustmop: I thought that we decided that `dataset.listrawrefs` isn't being combined into `dataset.list`, but rather it is not going to have an endpoint at all. Same different in this file, but the method `dataset.listrawrefs` should still exist in `lib/datasets.go` and use `denyRPC` in its `Attributes` defintion.

api/api.go
dustmop: Yes, and it is changing name from `fsi.whatchanged` to `dataset.componentstatus`.

api/api.go
dustmop: Correct, yes.

api/api.go
dustmop: Yes, I believe that is correct.

api/api.go
dustmop: According to https://github.com/qri-io/qri/pull/1730, `rawlogbook` and `logbookssummary` should be dropped from the API, because they're "hard-mode" only. Not totally sure about the other 2.

cmd/log.go
```diff
@@ -211,7 +211,7 @@ func (o *LogbookOptions) Complete(f Factory, args []string) (err error) {
 	return err
 }
 
-// Logbook executes the Logbook command
+// LogEntries executes the Logbook command
```
dustmop, line 214: I have the same fix in https://github.com/qri-io/qri/pull/1733/files, but wrote "LogEntries gets entries from the logbook" which more closely describes what's going on.

lib/api.go
```diff
@@ -36,44 +36,11 @@ const (
...
-	AERemoteRefs = APIEndpoint("/remote/refs")
+	AEIPFS = APIEndpoint("/qfs/ipfs/{path:.*}")
+	// AEWebUI serves the remote WebUI
+	AEWebUI = APIEndpoint("/webui")
 
-	// dataset endpoints
+	// aggregate endpoints
 
 	// AEListRaw lists all datasets in your collection as a well formatted string
 	AEListRaw = APIEndpoint("/listraw")
```
dustmop, line 46: As mentioned, this is going away.

lib/datasets.go
```diff
@@ -53,22 +53,29 @@ func (m DatasetMethods) Name() string {
...
+		"get": {AEGet, "GET"},
+		// "log": {AE?, "POST"},
+		"rename": {AERename, "POST"},
+		"save":   {AESave, "POST"},
+		"pull":   {AEPull, "POST"},
+		// "push":            {AE??, "POST"},
+		// "render":            {AE??, "POST"},
+		"remove":   {AERemove, "POST"},
+		"validate": {AEValidate, "POST"},
+		// "unpack": {AE??, "POST"},
```
dustmop, line 66: AEUnpack

lib/fsi.go
dustmop: EnsureRef is used by FSI to fixup linked datasets that may have been modified by outside programs (or manual user intervention) between runs of qri. For example, if someone checks out a dataset to a working directory, and then deletes the `.qr-ref` file from the folder it was checked out to, when another command is run, EnsureRef will notice this and unlink the dataset. It needs to be available on RPC because most commands will call it via `cmd/ref_select.go`. It runs the method `fsi.EnsureRef`.

Ah, and I just noticed now it's not in the spec file. We should add it there.

dustmop (commented):
Comments so far

api/api.go
ramfox: I believe `history` will be the `/log` endpoint with `filter=history` (or `selector=history`, whatever language we choose)

lib/datasets.go
ramfox: I'm actually personally against adding the notion of anything except "POST" w/ application/json to anything at the lib level or lower. I think it will be cleaner if lib only understands one way of doing thing, with `api`'s new role to be conforming to any special cases. 

This is slightly off topic, but related. In order to use the callable/regMethods to programmatically allow the instance to add handlers to a server, I'm pitching that actually what we want for AttributeSet is really:

```
type AttributeSet struct {
   endpoint APIEndpoint
   denyRPC bool
}
```

Especially if we find that there are lib methods we are okay with folks accessing via the api, but not rpc. The only one I can think of right now would be something like `ProfilePhotoSet`, which we would need for any frontend work, but we may not want to have over rpc. I'm basing this off the fact that we don't want any config changes over rpc, and profile changes are technically also config changes. However, w/ multi tenancy this is probably no longer the case!

If there IS a one-to-one mapping of non-api endpoints & methods we don't want over rpc, then we can just proceed w/ only having the `APIEndpoint` in the `AttributeSet` & drop the http method type entirely.

api/api.go
ramfox: So this particular endpoint is not the list of peers, but instead the list of datasets you have from a specific peer. Which I believe would become `/list` w/ the `filter=ramfox` in the params

lib/fsi.go
ramfox: added!

lib/log.go
```diff
@@ -27,7 +27,7 @@ func (m LogMethods) Name() string {
 // Attributes defines attributes for each method
 func (m LogMethods) Attributes() map[string]AttributeSet {
 	return map[string]AttributeSet{
-		"history":        {AEHistory, "POST"},
+		"history":        {AELog, "POST"},
 		"entries":        {AEEntries, "POST"},
```
ramfox, line 31: To be clear here, I just made this name change from `/logbook` to `/entries` for clarify, but always intended it to be temporary until we adjust to the new spec. When we do our final refactor, `/entries` should be `/log` and `/history` should be removed!

api/api.go
Arqu: Ah yes, I keep forgetting what I told myself.

api/api.go
Arqu: Yes what @ramfox said.
- `/peer/list` lists your peers
- `/list/{peer}` lists datasets of a given peer and should be moved into list with the new filtering approach

cmd/log.go
Arqu: That's fine, I'll update mine as well and rebase later given that you will probably merge before I do on this. Sneaked in some other commit I guess.

lib/log.go
Arqu: Awesome, and yeah I'll be keeping rebasing this as I'm trailing all the other changes.

lib/datasets.go
Arqu: In this case I would side with the `POST` only option. While I understand that this might expand later and would technically be doubling the work, I'd like to remove as much cruft as possible and get as close as possible to the spec for a clean slate moving forward. Mostly to establish nice patterns which we can build on top rather than bending them here and there.

Unsure whether we should spend any time right now on handling the case where we support API but not RPC on a global level. But even if we did want to do that now, feels like a config thing that should be a switch in the dispatch not on the attribute/method level.

Arqu:
Majority of the refactor is done and I would like to land this before making this PR not just a refactor but a big change in the logic of some remaining methods.

A list of remaining things:
    - merge log <> history
    - merge list <> list/peer
    - clean up handleRefRoute 
    - align remaining routes in correct methods
    - one more run through routes vs spec

Arqu:
Quick observations from the current standing of the PR:
- /remote/remove vs /ds/remove we currently have both I think
- /peer/remove - this exists in the spec but not in code as far as I can see
- /access/token - need to bring this one back seems to be lost in the rebases - only the API handler/route binding, code is there

ramfox:
hmmm re `/peer/remove` I'm not sure what this would do? Remove the peer from your peerstore? Otherwise, we already have the ability to disconnect.

`/remote/remove` and `/ds/remove` is tricky & i think clashes with two thought processes we have. On one hand, we specifically decided that we wanted our api to have `qri remove --remote qri_remote` & `/remote?remote="qri_remote"` when we first established removing from a remote. However, the way it's set up we switch on local vs remote at the api/cmd level. This means we would need a custom api handler function to control the flow here. Another option is combining these at the lib level. But then this opens up more questions about remote behavior, resolution, and how we separate out concerns, that we will come up against in the future when we open up the ability to do more behavior on a remote (saving, triggering runs, changing permissions).

I think the easiest thing to do rn would be to remove the remote behavior from the `/ds/remove` endpoint, punt the convo, separate the concerns completely, and have this discussion when we open up more remote behaviors.

ramfox:
This is so lovely. Noticed that you refactored away from `sql.QueryHandler`, but that the `api/sql.go` & `api/sql_test.go` files wasn't removed.

b5 (approved):
@Arqu this is looking great. To avoid further rebasing I vote we merge this now & do follow-ups for any subsequent fixes

PR #1735, opened 2021-04-02
WIP(param): Change param struct fields to align more closely.

dustmop:
Does not compile, code aside from the Param fields has not be changed to match the new definitions (aside from some stuff I was hacking on). Want to finalize names (and finish some work this change depends on) first.

Big changes:
1) No remote `Remote` or `Source` fields. This is now handled by Scope.
2) `Refstr` -> `Ref` everywhere.
3) Use `Component` where it makes sense. Get still uses `Selector` because it allows things like `qri get meta.title`
4) Change some contractions like `Ds` to `Dataset`
5) `Username` instead of `Peername` when not in the p2p context.

I think there may be one or two places where structs can be further combined, if the endpoints have enough in common. I want to dig into this some more.

### DIFF
--- a/api/datasets.go
+++ b/api/datasets.go
@@ -94,10 +94,11 @@ func (h *DatasetHandlers) RemoveHandler(routePrefix string) http.HandlerFunc {
 			return
 		}
 
-		if params.Remote != "" {
+		source := lib.SourceFromRequest(r)
+		if source != "" {
+			// TODO(dustmop): h.remote.WithSource(source).Remove
 			res, err := h.remote.Remove(r.Context(), &lib.PushParams{
 				Ref:    params.Ref,
-				Remote: params.Remote,
 			})
 			if err != nil {
 				log.Error("deleting dataset from remote: %s", err.Error())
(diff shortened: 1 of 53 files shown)


b5:
Aside from using `Selector` instead of `Component`, looks great!

dustmop:
Code compiles now, but tests are not passing. Due to moving Source onto
the Scope, the remaining parts of lib (especially Remote) need to use
Dispatch before the tests will pass.

Open question:
* do we want to change Page & PageSize on cmd/log to be Limit & Offset?

Source:
* Source is not on *Params, instead it is handled through Instance and Scope
* for HTTP, Source is set by an HTTP Header
* lib.SourceFromRequest(http) is the getter from this

Renames:
* Peername -> Username (except for p2p)
* Refstr -> Ref
* RemoteName & Remote -> Source (cmd package only)
* Path -> Version (for restore, this parameter always refers to a historical version to restore)
* ComponentName & Component -> Selector

Removed:
* `UseFSI` removed, not used
* remove `RPC` from base.ListDatasets, as we fixed this bug a while ago
* `UseDscache` removed, should live in config, or on scope instead
* `All` _should_ be removed, but then functionality is lost until Cursors exist

Not working:
* cmd/preview
* cmd/push
* cmd/remove [remote] -> RemoteMethods.Remove
* lib/remote.go [preview,push,remove]
* lib/render.go
* stats

dustmop:
Superseded by https://github.com/qri-io/qri/pull/1754.

PR #1736, opened 2021-04-02
refactor(lib): profile follows the dispatch pattern

ramfox:
One thing to note here: there is a bit of a blurring between the repo config profile and the "active profile". I adjusted to differentiate where I was sure of the difference, but I'm 100% sure that I haven't drawn the line exactly where we will want it when we move toward multi-tenancy.

- [x] add `Name` & `Attributes` methods on `ProfileMethods`
- [x] add `Profile` method to Instance
- [x] add `profileImpl`
- [x] add `GetProfile`, `SaveProfile`, `ProfilePhoto`, `SetProfilePhoto`, `PosterPhoto`, `SetProfilePhoto` methods to `profileImpl` that take a scope
- [x] dispatch-ify `GetProfile` `SaveProfile`, `ProfilePhoto`, `SetProfilePhoto`, `PosterPhoto`, `SetProfilePhoto` methods
- [x] register `Profile` with dipatch
- [x] replace `NewProfileMethods` with `inst.Profile`
- [x] replace inst in `ProfileMethods` with `d dispatcher`
- [x] figure out if `ProfileHandlers` can be replaced with `NewHTTPRequestMethod` (specifically look at error responses), looks like `profile/photo` and `profile/poster` endpoint return w/ header type image/jpeg
- [x] add `/profile/photo/set` AESetProfilePhoto & /profile/poster/set AESetProfilePoster
- [x] remove `/me`
- [x] use `NewHTTPRequestMethods` for all api handlers
- [x] remove `ProfilePhoto` and `ProfilePoster`, if a user wants to get a photo or poster they should resolve the field from the profile

-----

One of our tests has brought up a funny interaction! In `TestProfileRequestSave`, we were pass in `nil` parameters. Then, `Dispatch` casts those `nil` parameters (a `nil` pointer to a `*config.ProfilePod`), to an `interface{}`.

So it's as if we have done this:
```
var param *config.ProfilePod = nil
var i interface{}
i = param
```

Here is the expectation:
```
param == nil // true
i == nil // true
```

BUT that's not the case, because of how interfaces are defined. An interface consists of a type and a value, both must be nil for the interface to be considered nil. But in the above we have set interface i to be `(*config.ProfilePod)(nil)`, so we can't just rely on `i == nil` to guard against nil calling.

Where is this causing us a problem? In `Dispatch`, we assert that the params are of type `Validator` in order to validate them before we pass them to the method. *config.ProfilePod IS a validator.  But! we then call `param.Validate()`, which tries to automatically dereference the pointer, and the pointer is nil! Which causes a nil pointer dereference panic. 

So we should guard against that! Using reflect we can ensure if we have a param that is is nil, or a param that is both a pointer & is nil, we return w/ an error.

At the end of the day... it turns out we specifically DON'T want to validate the `config.ProfilePod` that comes in via the Save path, because the user can just be updating a single part of the config & that config won't contains a private key, that the validator expects. So for this particular case, it's actually a moot point. But, it seems really easy for future us to pass in a `nil` params to `Dispatch` & cause a panic. This change guards future foot guns!


api/api.go
```diff
@@ -221,11 +221,10 @@ func NewServerRoutes(s Server) *mux.Router {
...
 
-	proh := NewProfileHandlers(s.Instance, cfg.API.ReadOnly)
-	m.Handle(lib.AEMe.String(), s.Middleware(proh.ProfileHandler))
-	m.Handle(lib.AEProfile.String(), s.Middleware(proh.ProfileHandler))
-	m.Handle(lib.AEProfilePhoto.String(), s.Middleware(proh.ProfilePhotoHandler))
-	m.Handle(lib.AEProfilePoster.String(), s.Middleware(proh.PosterHandler))
+	m.Handle(lib.AEGetProfile.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "profile.getprofile"))).Methods(http.MethodPost)
+	m.Handle(lib.AESetProfile.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "profile.setprofile"))).Methods(http.MethodPost)
+	m.Handle(lib.AESetProfilePhoto.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "profile.setprofilePhoto"))).Methods(http.MethodPost)
+	m.Handle(lib.AESetPosterPhoto.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "profile.setposterphoto"))).Methods(http.MethodPost)
```
ramfox, line 227: Removed the `ProfilePhoto` and `PosterPhoto` getter handlers in favor of their respective setters, as per our face to face convo today

lib/api.go
```diff
@@ -38,16 +38,16 @@ const (
...
+	// profile endpoints
+
+	// AEGetProfile is an alias for the me endpoint
+	AEGetProfile = APIEndpoint("/profile")
+	// AESetProfile is an endpoint to set the profile
+	AESetProfile = APIEndpoint("/profile/set")
+	// AESetProfilePhoto is an endpoint to set the profile photo
+	AESetProfilePhoto = APIEndpoint("/profile/photo/set")
+	// AESetPosterPhoto is an endpoint to set the profile poster
+	AESetPosterPhoto = APIEndpoint("/profile/poster/set")
```
ramfox, line 50: Renamed for clarity/uniformity `AEProfile` -> `AEGetProfile`
`AESaveProfile` -> `AESetProfile`

removed `AEProfilePhoto` and `AEPosterPhoto`, added `AESetProfilePhoto` & `AESetPosterPhoto`

lib/datasets.go
```diff
@@ -858,6 +859,24 @@ func (datasetImpl) List(scope scope, p *ListParams) ([]dsref.VersionInfo, error)
...
+	}
+
+	id, err := profile.IDB58Decode(idStr)
+	if err != nil {
+		log.Debugw("decoding profile ID", "err", err)
+		return nil, err
+	}
+	return pros.GetProfile(id)
+}
+
```
ramfox, line 879: This used to live in `lib/profile.go`, but now those methods are exclusive to the active user. Moved this closer to it's only caller.

lib/dispatch.go
```diff
@@ -71,7 +79,10 @@ func (isw *InstanceSourceWrap) Dispatch(ctx context.Context, method string, para
 
 func (inst *Instance) dispatchMethodCall(ctx context.Context, method string, param interface{}, source string) (res interface{}, cur Cursor, err error) {
 	if inst == nil {
-		return nil, nil, fmt.Errorf("instance is nil, cannot dispatch")
+		return nil, nil, ErrDispatchNilInstance
+	}
+	if param == nil || (reflect.ValueOf(param).Kind() == reflect.Ptr && reflect.ValueOf(param).IsNil()) {
+		return nil, nil, ErrDispatchNilParam
```
ramfox, line 85: Here is the MEGA fun part, detailed in the commit message and https://github.com/qri-io/qri/pull/1736#issue-608221386

lib/profile_test.go
```diff
@@ -63,12 +61,12 @@ func TestProfileRequestsSave(t *testing.T) {
...
 	cases := []struct {
-		p   *config.ProfilePod
+		p   *SetProfileParams
 		res *config.ProfilePod
 		err string
 	}{
-		{nil, nil, "profile required for update"},
-		{&config.ProfilePod{}, nil, ""},
+		{nil, nil, ErrDispatchNilParam.Error()},
+		{&SetProfileParams{Pro: &config.ProfilePod{}}, nil, ""},
```
ramfox, line 69: test that triggered the dispatch/nil pointer dereference bug!

lib/profile_test.go
```diff
@@ -95,20 +93,9 @@ func TestProfileRequestsSave(t *testing.T) {
...
-	// TODO: Remove the randomly built Profile that config.DefaultProfile creates.
-	mockID := "QmWu3MKx2B1xxphkSNWxw8TYt41HnXD8R85Kt2UzKzpGH9"
-	mockTime := time.Unix(1234567890, 0)
-	cfg.Profile.ID = mockID
-	cfg.Profile.Created = mockTime
-	cfg.Profile.Updated = mockTime
-	cfg.Profile.Peername = "test_mock_peer_name"
+func TestSetProfile(t *testing.T) {
+	tr := newTestRunner(t)
+	defer tr.Delete()
```
ramfox, line 98: using the old `NewInstanceFromConfigAndNode` was actually messing with our expectations, since `NewInstanceFromConfigAndNode` does it's own strange config magic. using `testRunner` instead.

lib/profile.go
```diff
@@ -20,46 +19,82 @@ import (
...
-		err = m.inst.http.CallMethod(ctx, AEProfile, http.MethodGet, nil, res)
-		if err != nil {
-			return nil, err
-		}
-		return res, nil
+// SetProfileParams defines parameters for setting parts of a profile
+// Cannot use this to set private keys, your peername, or peer id
+type SetProfileParams struct {
+	Pro *config.ProfilePod
+}
```
ramfox, line 55: If we just use `*config.ProfilePod` as our input params, `Dispatch` will attempt to validate the params when they first pass through. `*config.ProfilePod` does not consider a `ProfilePod` valid if it is missing its private key. However, as an input, we should not have a private key filled in at this point. 

My solution was to wrap the profilepod in a `SetProfileParams` struct, but would like feedback here as there may be other places where we need to abstract for a similar reason.

lib/profile.go
```diff
@@ -71,78 +106,65 @@ func (m *ProfileMethods) GetProfile(ctx context.Context, in *bool) (*config.Prof
...
+	prevPeername := cfg.Profile.Peername
+	if pro.Peername != "" && pro.Peername != cfg.Profile.Peername {
+		cfg.Set("profile.peername", pro.Peername)
+	}
 
-	cfg := m.inst.cfg
-	r := m.inst.repo
+	if err := cfg.Profile.Validate(); err != nil {
+		return nil, err
+	}
```
ramfox, line 144: ensure we have a valid profile before we attempt to save it.

api/api.go
Arqu: After the last convo, not really sure. Prior to this my mental image was that we only support getters not setters as they meddle with the config.

lib/api.go
Arqu: Im going to cry on my next rebase :D 
@b5 I share the same confusion here now.

lib/dispatch.go
Arqu: That is SUPER interesting. Great catch, so subtle. 👏

lib/api.go
ramfox: I KNOW I'M SORRYYYYYY

lib/profile.go
dustmop: I think this is good. The config validation is not necessarily the same as input parameters being valid.

lib/profile.go
```diff
@@ -20,46 +19,82 @@ import (
...
 	}
+	return nil, dispatchReturnError(got, err)
+}
 
-	var pro *profile.Profile
-	r := m.inst.repo
+// FileParams defines parameters for Files as arguments to lib methods
+type FileParams struct {
+	// Url      string    // url to download data from. either Url or Data is required
+	Filename string    // filename of data file. extension is used for filetype detection
```
dustmop, line 69: This needs the struct tag `qri:"fspath"`. Also, comment should move up above the field instead of to the right.

lib/profile.go
```diff
@@ -20,46 +19,82 @@ import (
...
+	Data     io.Reader // reader of structured data. either Url or Data is required
```
dustmop, line 70: I think this doesn't work. An `io.Reader` won't serialize across HTTP rpc.

On the other hand, it looks like `SetProfilePhoto` and `SetPosterPhoto` both ignore `Filename` and only try to `ReadAll` the `Data`. Might be necessary to switch over to a `[]byte` and have those functions actually try both, unless there's a good reason why the code was working this way before.

dustmop (commented):
Dispatch side looks good! Couple of comments about parameters.

api/api.go
ramfox: So I know we aren't doing config setters over rpc, but this is specifically for use w/ a frontend that needs an api for setting any profile photos.

@b5 specific questions: in our convo on friday we talked about how if a user wants to get a config photo, they should take the path from the profile (rather than fetching the actual file from the backend). The spec still has `/profile/photo` and `/profile/poster` endpoints. Should those be the "setter" endpoints or would something link `/profile/photo/set` and `/profile/poster/set` be preferred?

b5: `/profile/photo` & `/profile/poster` should be POST-only, RPC-style setter endpoints.

For this pass of the API we're sticking with the "RPC can't change config" until the dispatch pattern is worked into the entire codebase, so I'd be fine with either:

* wire up these endpoints but always have them return an error
* use `denyRPC` and just turn them off entirely

dustmop (approved):
LGTM!

lib/profile.go
```diff
@@ -20,46 +18,86 @@ import (
...
+func (m ProfileMethods) Attributes() map[string]AttributeSet {
+	return map[string]AttributeSet{
+		"getprofile":      {denyRPC, ""},
+		"setprofile":      {denyRPC, ""},
+		"setprofilephoto": {denyRPC, ""},
+		"setposterphoto":  {denyRPC, ""},
+	}
 }
 
+type ProfileParams struct{}
```
dustmop, line 39: Please add a comment for this struct, to make `go lint` happy.

ramfox: ooof ty

Issue #1737, opened 2021-04-05
Unify list aggregation fields


b5:
We have a number of places across where input parameters affect a listing result. While working on API refactoring (#1731) we've arrived at the following terms & fields for aggregation control:

| name | data type | description |
| --- | --- | --- |
| offset | int | how many elements to skip from the zeroith |
| limit | int | max number of resulting items |
| filter | ? |  reduce the resulting set according to one or more named filter functions |
| orderby | ? | order items according to a set of named ordering functions |

We've all agreed that `selector` is a separate concept that mapping _into_ a data structure via a string path. While a selector may be used on an aggregation, it is not the same as a filter. Selectors will generally be used with individual dataset references.

This issue can be closed when:
* [ ] we've defined how `filter` & `orderBy` will work in practice
* [ ] all aggregation methods are enumerated here
* [ ] all aggregation methods accept the above control fields

It'd be great if we could write one package that defines aggregation fields, and an interface that each aggregator function can define filter & orderly functions for.

b5:
I'm seeing a nice chance for harmony coming from recent code spikes

@dustmop has a great PR #1746 that examines out how _cursors_ could work in practice:

```go
package lib

// MakeCursor returns a cursor that is able to retrieve the next page of results
func (s *scope) MakeCursor(nextPage interface{}) Cursor {
 	return cursor{s.inst, s.method, nextPage}
}
```

In a [recent spec test](https://github.com/qri-io/qri/pull/1738/commits/7d6c7861c370a0587be3d8faa381d7ca6a092c6c) I sketched out a `list` package that defines a `list.Params` struct:

```go
package list

type Params struct {
 	Filter  []string
 	OrderBy []string
 	Limit   int
 	Offset  int
 }
```

if Params got a new method: `NextPageParams() list.Params`, we could define a new interface that `list.Params` would implement, and modify `MakeCursor` to accept it instead:

```
// Pager provides parameters for iterating an aggregation
type Pager interface {
  // NextPageParams returns a new list.Params that advances one page
  NextPageParams() list.Params
}

// elsewhere ...
func (s *scope) MakeCursor(pager list.Pager) Cursor {
 	return cursor{s.inst, s.method, pager}
}
```

Using an interface would let us do embedded-struct pagination:

```go
package lib

type ListDatasetParams struct {
  list.Params  // embedding gives ListDatasetParams a NextPageParams method

  // these are all currently needed by listing datasets, here mainly to prove that listing
[...]

PR #1738, opened 2021-04-05
collection spec test

b5:
The goal of this PR is to merge a clear definition of what our "collection" replacement package needs to do based on product requirements. We'll need _two_ implementations of this (one for local qri nodes, another for qri cloud)

In an effort to make engineering requirements easier, I'm trying my hand at writing a spec test to define clear requirements for code that doesn't exist. This PR is _not_ meant to prescribe how to get there (for example, we might get these tests to pass locally by refactoring on dscache)

Along the way, I ended up writing two helper packages:
* `github.com/qri-io/qri/profile/test`, which just contains a dictionary of test profiles, built atop the `keys/test` package. This'll become important later as we do more work mapping many keys to one profile
* a `list` package, that sets up a first crack at #1737 (The placement of the list package feels awkward. suggestions welcome)


collection/list/list.go
```diff
@@ -0,0 +1,43 @@
+package list
```
Arqu, line 1: I'd really like to have this in `qri/api/util` as there's already some of the required logic present and would be good to keep it together and source it from a central place for other endpoints besides just collection.

collection/collection.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/qri-io/qri/event"
+	"github.com/qri-io/qri/profile"
+)
+
+// RootUserKeyID is a special key identifer that lists *all* datasets in a
+// collection
+const RootUserKeyID = profile.ID("root")
+
+type Collection interface {
+	List(ctx context.Context, pid profile.ID, lp list.Params) ([]Item, error)
```
Arqu, line 18: I'm very happy with this interface and it flows nicely with the hosted version and the logic of the previews.

collection/collection.go
```diff
@@ -0,0 +1,78 @@
...
+
+type EditableCollection interface {
+	Collection
+	Put(ctx context.Context, items ...Item) error
+	Delete(ctx context.Context, profileID profile.ID, initIDs ...string) error
+}
+
+// Item is a dataset in a collection of datasets
+// it's a slightly-modified version of the flatbuffer defined in dscache.fbs
+type Item struct {
```
Arqu, line 29: On the flip-hand, I do feel a bit itchy with this one, from a store perspective, this is a lot of fields and a lot of those fields are duplicated / de-normalized. I have to look into how this would be generated from what we have at hand (ie `dataset.Dataset` or `VersionInfo`) and have it be combined with some `WorkflowInfo` which has yet to be added.

Arqu:
I'm happy to work off of this if @ramfox feels this looks good with her line of work.

collection/list/list.go
b5: I don't think this belongs in `api/util`. Trying out a new top-level `params` package, would love your thoughts

collection/collection.go
b5: agreed it requires a lot more thought. I've added clarifying comments to both the `Collection` interface and the `Item` struct def. I'm hoping they help justify what's owned by the collection and what is cached.

I think we still have a long way to go comparing `VersionInfo` to `Item`

collection/list/list.go
Arqu: Huh, honestly, no clue what should be best. This feels a lot like going full circle on some things.

At the start of the API refactor I went a similar route where I started condensing some of these "generic" params and some others into `lib/params.go` which we then decided not to do. We wanted to have each params struct have it's own in it's own file (and potentially even split wherever possible even at the expense of duplicating code; think Dustin was dealing with that part specifically).

I'd say we just pick one setup and stick with it, either has it's own mini tradeoffs none of which are worth fussing around too much. My vote would be in `api/util` or a standalone package, I'd hate to have very large imports (and potential circular deps) just because it lives in a large package. Other than that I'm fine with anything.

collection/collection.go
```diff
@@ -0,0 +1,247 @@
...
+				found = true
+				copy(col[i:], col[i+1:])              // Shift a[i+1:] left one index.
+				col[len(col)-1] = dsref.VersionInfo{} // Erase last element (write zero value).
+				col = col[:len(col)-1]                // Truncate slice.
+				break
+			}
+		}
+
+		if !found {
+			fmt.Println("can't find id", removeID)
```
dustmop, line 170: Is this print statement needed?

collection/collection.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func (c *collection) saveProfileCollection(pid profile.ID) error {
+	if c.basePath == "" {
+		return nil
+	}
+
+	items := c.collections[pid]
+	if items == nil {
+		return fmt.Errorf("cannot save empty collection")
```
dustmop, line 232: Error is a bit misleading. This isn't an empty collection (which a user may end up at if they Put a ref then Delete it), it's a user whose pid is not in the map, so a totally unknown user.

collection/collection.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	path := filepath.Join(c.basePath, fmt.Sprintf("%s.json", pid.String()))
+	f, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY, 0655)
```
dustmop, line 236: 0644 is preferable, the executable bit isn't needed for json files.

collection/collection.go
```diff
@@ -0,0 +1,247 @@
...
+// and must be kept in sync as subsystems mutate their state.
+type Collection interface {
+	List(ctx context.Context, pid profile.ID, lp params.List) ([]dsref.VersionInfo, error)
+}
+
+// Writable is an extension interface for collection that adds methods for
+// adding and removing items
+type Writable interface {
+	Collection
+	Put(ctx context.Context, profileID profile.ID, items ...dsref.VersionInfo) error
```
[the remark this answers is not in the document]
b5: definitely isn't a use case for putting zero refs. I've added a test just to ensure consistent behavior across implementations.

As for multiple refs, the first use case is migrations. We have the repo-to-collection migration locally, and another possible migration on cloud (which may or may not want to use this interface, just as likely a migration would happen on the SQL side).

Second possible use comes from pulling the dependencies of an automated dataset. If we `qri add` another dataset & automate it, the dependencies become part of our collection within a single operation.

I agree both examples may seem like a stretch, but I find the local code pairs with mutex locking & disk write nicely when the `Put` operation supports batching.

collection/collection.go
```diff
@@ -0,0 +1,247 @@
...
+	Delete(ctx context.Context, profileID profile.ID, initIDs ...string) error
+}
+
+const collectionsDirName = "collections"
+
+type collection struct {
+	basePath string
+
+	sync.Mutex  // collections map lock
+	collections map[profile.ID][]dsref.VersionInfo
```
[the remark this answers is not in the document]
b5: Totally. In my head those could be accomplished with a additional internal state that runs parallel to `VersionInfo` and a `params.ListParams.Filter = []string{ "folder:FOLDER_ID" }` or something.

collection/collection.go
```diff
@@ -0,0 +1,247 @@
...
+	return c, err
+}
+
+func (c *collection) List(ctx context.Context, pid profile.ID, lp params.List) ([]dsref.VersionInfo, error) {
+	c.Lock()
+	defer c.Unlock()
+
+	col, ok := c.collections[pid]
+	if !ok {
+		return []dsref.VersionInfo{}, nil
```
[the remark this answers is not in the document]
b5: yeah great point, will replace with a wrap of an `ErrNotFound` declaration

dsref/version_info.go
```diff
@@ -197,3 +201,78 @@ func ConvertVersionInfoToDataset(info *VersionInfo) *dataset.Dataset {
...
+		return func(a, b *VersionInfo) bool { return a.BodySize < b.BodySize }, nil
+	}
+
+	return nil, fmt.Errorf("unrecognized sorting field: %q", name)
+}
+
+// VersionInfoAggregator sorts slices of VersionInfos according to a provided
+// string configuration. Call its Sort method to sort the data
+// TODO(b5): add support for filtering
+type VersionInfoAggregator struct {
```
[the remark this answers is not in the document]
b5: > I imagined keeping the dsref package very leaf-level and light on dependencies

Me too. I don't think we're violating that principle here.

collection/collection.go
```diff
@@ -0,0 +1,247 @@
...
+	"github.com/qri-io/qri/event"
+	"github.com/qri-io/qri/params"
+	"github.com/qri-io/qri/profile"
+)
+
+// RootUserKeyID is a special key identifer that lists *all* datasets in a
+// collection
+const RootUserKeyID = profile.ID("root")
+
+// Collection is a set of datasets scoped to a user profile. A users's
```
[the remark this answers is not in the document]
b5: Glad you brought this up. I saw it too & me brain got lazy. I'm down for `type CollectionSet map[profile.ID][]dsref` and `type Collection []dsref`

collection/collection.go
```diff
@@ -0,0 +1,247 @@
...
+
+	"github.com/qri-io/qri/dsref"
+	"github.com/qri-io/qri/event"
+	"github.com/qri-io/qri/params"
+	"github.com/qri-io/qri/profile"
+)
+
+// RootUserKeyID is a special key identifer that lists *all* datasets in a
+// collection
+const RootUserKeyID = profile.ID("root")
```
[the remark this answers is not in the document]
b5: Yeah we're on the same wavelength. I didn't get far enough to answer these exact questions, so I'll remove this & we can write a PR later that answers these questions

params/list.go
```diff
@@ -0,0 +1,45 @@
+// Package params defines generic input parameter structures
+package params
```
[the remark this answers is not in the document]
b5: yeah, I just can't find a place where it feels right, so let's go with your suggestion, I'll move it under `base`

b5:
I'm going to merge this so we can keep moving. There are a few conversations started in review here that I'd love to finish out tho

PR #1739, opened 2021-04-05
SPIKE: move qrimatic into qri

b5:
DO NOT MERGE. This PR exists to discuss how to move Qrimatic into qri, with spike code attached so we can see what it'll look like.

If you pull & build this, running `qri connect` will spin up the same scheduler that qrimatic currently uses. Any automated jobs will be executed on their cron schedules.

#### new `/auto` endpoint, `automation` package reorg
I'd like to propose a new `automation` package that for now exists just to group automation-related packages:

```
/automation
  /transform
  /workflow
  /scheduler
  /run
```

This move `run` up out of `transform`, and `transform` moves into automation. `workflow` and `scheduler` move into qri from the qrimatic project.

Also establish a related method set at the lib level: move `lib.transform` methods into `lib.automation` methods, expand methods to the following set:

| method name | description | required on first pass? |
| --- | --- | --- |
| automation.apply | relocate transform.apply to here | yes |
| automation.workflows         | list local workflows   | yes |
| automation.workflow          | get a workflow   | yes
| automation.saveWorkflow      | "deploy" in qrimatic UI, create/update a workflow   | yes |
| automation.removeWorkflow    | "undeploy" in qrimatic UI   | yes |
| automation.runs              | list automation runs   | no |
| automation.run               | get automation run log   | no |

### Next Steps:
To make this spike mergeable the code we're porting from the qrimatic project should reach feature parity with qrimatic, so that's: scheduling, and workflow management. We'll move dataset collection / combined workflow listing as another issue.

* [ ] workflow package & store needs work, formalization, tests
* [x] add workflow create, update, delete methods, API endpoint
* [ ] test for feature parity between `qrimatic serve` and `qri connect`
* [x] we need a story around automation configuration: https://github.com/qri-io/qri/issues/1748
* [ ] define CLI commands for interacting with automation


dustmop:
So I'm not very familiar with the qrimatic codebase, but overall the plan sounds alright to me. It's a bit strange to me to move `apply` to `automation.apply` and especially `/auto/apply`, since `apply` can be run manually and interactively (TODO). But that's not a blocker by itself. Is the package name `automation` committed? If not, would something like `execution` or `code` be doable?

Something I'm wondering about, that may or not be related to the work outlined in this issue (so feel free to ignore if it's irrelevant): is the problem of how deploying is integrated into the save path worked out? Also, how much of the rest of the `qri` codebase is `automation` going to be built upon directly; I'm guessing mostly `apply` and `list`? I guess these kinds of questions inform how the code should be organized and integrated, which is why it's on my mind, but keeping all of `automation` basically by itself, with only a few external deps seems fine for now.

b5:
superseded by #1804

PR #1742, opened 2021-04-06
refactor(lib): adjust `RegistryClientMethods` to use dispatch pattern

ramfox:
- [x] add `Name` & `Attributes` methods on `RegistryMethods`
- [x] add `Registry` method to Instance
- [x] add `registryImpl`
- [x] add `CreateProfile`, `ProveProfileKey` to `registryImpl` that take a scope
- [x] dispatch-ify `CreateProfile` `ProveProfileKey` methods
- [x] register `Registry` with dipatch
- [x] replace `NewRegistryMethods` with `inst.Registry`
- [x] replace inst in `RegistryMethods` with `d dispatcher`
- [x] use `NewHTTPRequestMethods` for all api handlers


lib/registry.go
```diff
@@ -16,55 +16,72 @@ import (
...
 
-// CoreRequestsName implements the Requests interface
-func (RegistryClientMethods) CoreRequestsName() string { return "registry" }
+// Attributes defines attributes for each method
+func (m RegistryClientMethods) Attributes() map[string]AttributeSet {
+	return map[string]AttributeSet{
+		"createprofile":   {denyRPC, ""},
+		"proveprofilekey": {denyRPC, ""},
+	}
+}
```
ramfox, line 33: wasn't sure if these should be available over rpc!

lib/registry.go
```diff
@@ -16,55 +16,72 @@ import (
...
 
+// registryImpl holds the method implementations for RegistryMethods
+type registryImpl struct{}
+
+// CreateProfile creates a profile
+func (registryImpl) CreateProfile(scope scope, p *RegistryProfileParams) error {
 	// TODO(arqu): this should take the profile PK instead of active PK once multi tenancy is supported
-	ownerPk := m.inst.repo.Profiles().Owner().PrivKey
-	pro, err := m.inst.registry.CreateProfile(p, ownerPk)
+	ownerPk := scope.Profiles().Owner().PrivKey
```
ramfox, line 63: wasn't sure if this should actually be changed to `scope.ActiveProfile`

lib/registry.go
```diff
@@ -16,55 +16,72 @@ import (
...
+}
 
 // RegistryProfile is a user profile as stored on a registry
 type RegistryProfile = registry.Profile
 
+// RegistryProfileParams encapsulates arguments for creating or proving a registry profile
+type RegistryProfileParams struct {
+	Profile *RegistryProfile
+}
+
```
ramfox, line 42: again, `RegistryProfileParams` added b/c using a `RegistryProfile` triggers `Validate()` when dispatching, and this profile won't have every field filled in.

lib/scope.go
```diff
@@ -174,6 +179,11 @@ func (s *scope) LocalResolver() (dsref.Resolver, error) {
 	return s.inst.resolverForMode("local")
 }
 
+// SetLogbook allows you to replace the current logbook with the given logbook
+func (s *scope) SetLogbook(l *logbook.Book) {
+	s.inst.logbook = l
+}
+
```
ramfox, line 186: In the `ProveProfileKey` path, we have one scenario where we actually replace the given logbook (now line 153 in `lib/registry.go`). This feels... weird, so would like some eyes on it.

lib/registry.go
Arqu: think this is correct

lib/registry.go
Arqu: Per the new setup, scope should be the governing body for the profile in lib and bellow. Think owner is the correct choice by the latest discussions unless we specifically give a sub-profile a "superuser" token. Keeping owner for now makes more sense.

Arqu (commented):
I don't have a great mental overview of the `logbook`/`ProveProfileKey` stuff to be able to fully review this. Otherwise LGTM!

lib/scope.go
dustmop: Yes, that is correct. When running `prove` the user is trying to reestablish an existing account (registered on cloud) to the current machine they're using. This involves pulling data down from cloud, including their old profileID, and replacing it in their current repo. They also need to get their logbook data from cloud and replace what was there before, so that their profileID agrees with their logbook which lets `push` and `pull` continue to work. Not doing so causes the "split logbook identity" problem.

If you read the definition of the `ProveProfileKey` function, you'll see that the `prove` command will fail if the repository contains more than zero refs. This means it can only run on a repository that has run `setup` but not yet run `save`, which will have a logbook that is basically empty (it will have the user op for the newly created profile, but that's all). In this case it's safe to replace the old logbook data, since it didn't have any important data in it anyway. In the future, it would be better to merge this temporary logbook data, treating it as a separate profile, but getting to that point probably requires finishing the logbook redesign first.

ramfox: great thank you for the detailed response!

Arqu (approved):
:+1:

PR #1746, opened 2021-04-06
wip(cursor): Proof of concept for Cursor, used by List

dustmop:


dustmop:
While working on lib parameter cleanup, I ran into `All` which is used by `Get`. I don't think it should be on the lib parameters, but `cmd` and especially the `api` want it for things like `/body.csv`. We can fix this by having Cursors in place, and from the `cmd` package, just iterate a Cursor until it completes to get all. Whipped up a quick demonstration of how this would work for the `dataset.list` method.

dustmop:
Superseded by https://github.com/qri-io/qri/pull/1969

PR #1749, opened 2021-04-07
refactor(lib): fold `RenderMethods` into `DatasetMethods` & follow the dispatch pattern

### LINKED ISSUE #1745: Question about `/render` endpoints
We have two lib methods for rendering: `RenderReadme` and `RenderViz`

The api handler checks the params for a `Viz` field, if true it calls `RenderViz`, other wise it's assumed we are talking about `RenderReadme`

These are both off the same `/render` endpoint. 

One one hand, this doesn't necessarily violate our pattern of each endpoint returning one type of thing, because both return bytes of rendered html. But, we are technically returning something different depending on the input (a rendered `Readme` vs a rendered `Viz`), which does violate our 0.10.0 spec expectations.

Possible solutions, depending on if we feel this violates our principles:
- adding a `/render/viz` endpoint, and using `/render` for readme
- having both `/render/readme` & `/render/viz` & removing `render`
- use the `RenderParams` as they are now, but replace the `Viz` flag with the `Component` field of the `Ref`, if it says "viz", return a rendered viz, if it says "readme" or is empty, return a rendered readme

ramfox:
closes #1745 

Unifies `RenderViz` and `RenderReadme` to `Render`. Moves `Render` method to `DatasetMethods`


base/render_readme.go
```diff
@@ -10,12 +10,12 @@ import (
...
+func RenderReadme(ctx context.Context, file qfs.File) ([]byte, error) {
 	data, err := ioutil.ReadAll(file)
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 	unsafe := blackfriday.Run(data)
 	htmlBytes := bluemonday.UGCPolicy().SanitizeBytes(unsafe)
-	return string(htmlBytes), nil
+	return htmlBytes, nil
```
ramfox, line 20: This method is only called in one place, and it immediately converted the returned string back into a slice of bytes. Took out the middle man.

lib/datasets.go
```diff
@@ -1744,3 +1820,42 @@ func (datasetImpl) Stats(scope scope, p *StatsParams) (*dataset.Stats, error) {
...
+
+// Render renders a viz or readme component as html
+func (datasetImpl) Render(scope scope, p *RenderParams) (res []byte, err error) {
+	ds := p.Dataset
+	if ds == nil {
+		parseResolveLoad := scope.ParseResolveFunc()
+		ds, err = parseResolveLoad(scope.Context(), p.Ref)
+		if err != nil {
+			return nil, err
+		}
```
ramfox, line 1832: Okay so this is my major question mark here.

I took out the `Remote` field of the `RenderParams`, because it looks like (based on the scope refactors & https://github.com/qri-io/qri/issues/1704) that these lib/dispatchified methods are not supposed to be "resolve source" aware. 

It also turns out that the `qri render` command didn't have a place for a user to input a remote anyway.

The major question here, is that if we are relying on `scope.ParseResolveFunc`, it is expected that the we've already set the `scope.source` (looks like via the `inst.WithSource()` method). We are doing this in several places at the `cmd` package level, but we are not doing this at all at the `api` level. Specifically, wondering how we incorporate this into the `NewHTTPRequestMethod` func, since that is the path that through which all of our api handlers are passing.

Had trouble parsing if this issue would be solved when the `dsref.Loader` work was finished without having to make changes to the `NewHTTPRequestHandler` func, so wanted to make note of it.

b5: Great question. I don't think this is properly documented, but the answer we came up when specifying these changes boils down to this:

1. We have a category of "runtime options" that affect the way qri behaves (`source` is a prime example, the current user is another). They don't map well to configuration _or_ to input parameters.
2. When using the CLI, runtime options should map to global flags.
3. When using the HTTP api, runtime options should map to HTTP request headers.

So based on that we'd expect you to pass something like:
```
$ curl http://localhost:2503/render -H "Content-Type: application/json" -H "QriResolverSource: p2p" -d '{ "ref": "b5/world_bank_population/readme" }'
```

Where `QriResolverSource` overrides the default resolver source. This work is not yet implemented, but we can add it one time in the retuned http handler from lib & it will work with all methods.

ramfox: Okay great! So basically the way this is structured currently is correct, and the plans to put the resolver information in the http call hasn't been implemented yet.

dustmop: In the WIP https://github.com/qri-io/qri/pull/1735 I made a change to put the source into an HTTP Header, but's a demonstration for now and may need more discussion particularly on the name used.

PR #1751, opened 2021-04-09
refactor(load): dsref.Loader overhaul

dustmop:
Before, we had the Loader interface with the method LoadDataset, which expected to receive a resolved reference and loaded a dataset. In addition there was the ParseAndResolveLoadFunc type which would parse a ref and resolve it and then load it. This latter behavior more closes matches the `load_dataset` function in starlark.

This change renames the Loader interface's old method to LoadResolved, and moves the ParseAndResolveLoadFunc implementation to a second method on the interface, now named LoadDataset.

Remove the source parameter from scope.ParseAndResolveRef. Instead, the scope is assigned the source at construction time. Many commands require a certain source, so they check for it and return an error upon mismatch.


dustmop:
Solves https://github.com/qri-io/qri/issues/1704.

dsref/load.go
```diff
@@ -10,14 +10,15 @@ import (
 // ErrNoHistory indicates a resolved reference has no HEAD path
 var ErrNoHistory = fmt.Errorf("no history")
 
-// Loader loads and opens a dataset. The only useful implementation of the
-// loader interface is in github.com/qri-io/qri/lib.
-// TODO(b5) - This interface is a work-in-progress
+// ErrRefNotResolved is the error when LoadResolved is called on a non-resolved ref
+var ErrRefNotResolved = fmt.Errorf("ref has not been resolved, cannot load")
+
+// Loader loads datasets
```
b5, line 16: love straightforward comments like this. `Loader loads datasets` isn't going to go stale anytime soon 😄

dsref/load.go
```diff
@@ -10,14 +10,15 @@ import (
...
 type Loader interface {
-	LoadDataset(ctx context.Context, ref Ref, source string) (*dataset.Dataset, error)
+	// LoadDataset will parse the ref string, resolve it, and load the dataset
+	// from whatever store contains it
```
b5, line 19: I'd suggest:

```
LoadDataset turns a reference string into a *dataset.Dataset. 
LoadDataset composes together reference parsing, reference resolving, and dataset loading. 
The behavior of LoadDataset will vary based on external configuration, including  where 
to resolve  (resolver sources) and load from (loader locations).
```

dsref/load.go
```diff
@@ -10,14 +10,15 @@ import (
...
 var ErrNoHistory = fmt.Errorf("no history")
 
-// Loader loads and opens a dataset. The only useful implementation of the
-// loader interface is in github.com/qri-io/qri/lib.
-// TODO(b5) - This interface is a work-in-progress
+// ErrRefNotResolved is the error when LoadResolved is called on a non-resolved ref
+var ErrRefNotResolved = fmt.Errorf("ref has not been resolved, cannot load")
+
+// Loader loads datasets
 type Loader interface {
```
b5, line 17: Looking through the code, I don't think both of these methods shouldn't be on the same interface. I can't find an example looking through the this PR that accepts a `Loader` and uses both methods. (If that exists, then we should keep them combined).

I'd prefer to see:
```go
type Loader interface {
  LoadDataset(ctx context.Context, refstr string) (*dataset.Dataset, error)
}

type ResolvedLoader interface {
  LoadResolved(ctx context.Context, ref Ref, location string) (*dataset.Dataset, error)
}
```

To me this is important to keep the inversion-of-control pattern where subsystems accept a `Loader`. We shouldn't be giving subsystems an escape hatch where they can start dictating load locations

dsref/load.go
```diff
@@ -10,14 +10,15 @@ import (
...
-	LoadDataset(ctx context.Context, ref Ref, source string) (*dataset.Dataset, error)
+	// LoadDataset will parse the ref string, resolve it, and load the dataset
+	// from whatever store contains it
+	LoadDataset(ctx context.Context, refstr string) (*dataset.Dataset, error)
+	// LoadResolved loads a dataset that has already been resolved by finding it
+	// at the given location. Will fail if the Ref has no Path with ErrRefNotResolved
```
b5, line 22: I'd prefer dsref get a new `Resolved() bool` or `IsResolved() bool` method, and use that as the check. The formal definition of resolution is having an InitID. It's possible to have a resolved dataset with no path (empty history). Having that logic on dsref will avoid ad-hoc checks across the codebase like this

b5 (commented):
ok literally only had a minute to look at one file & I need to jump. Loving what I'm seeing. Will continue review in a bit

lib/load.go
```diff
@@ -13,45 +13,43 @@ import (
 	"github.com/qri-io/qri/fsi"
 )
 
+type datasetLoader struct {
+	inst   *Instance
+	source string
+}
```
b5, line 19: Moving this to it's own struct is lovely. I can see a future PR that moves loading into it's own package, pulling the `noHistoryLoader`, `base.NewLocalDatasetLoader` and `transform.Loader` into one place.

lib/load.go
```diff
@@ -13,45 +13,43 @@ import (
...
 // LoadDataset fetches, dereferences and opens a dataset from a reference
 // implements the dsfs.Loader interface
 // this function expects the passed in reference is fully resolved
-func (inst *Instance) LoadDataset(ctx context.Context, ref dsref.Ref, source string) (*dataset.Dataset, error) {
-	if inst == nil {
-		return nil, fmt.Errorf("no instance")
+func (d *datasetLoader) LoadResolved(ctx context.Context, ref dsref.Ref, location string) (*dataset.Dataset, error) {
+	if ref.Path == "" {
+		return nil, dsref.ErrRefNotResolved
 	}
```
b5, line 27: Another place where I'd prefer to see a new `ref.IsResolved()` method call

b5 (commented):
Looking really great. Let's work a few questions out & get this in

dsref/load.go
dustmop: I'd rather not introduce additional interfaces where they're not needed. My primary concern is the API surface of `scope` and how many similar but slightly different methods it provides. But, investigating more deeply, it looks like the second method `LoadResolved` can actually be completely removed, just by using `LoadDataset` in a few additional places. Currently only `dataset.get`, `fsi.checkout` and `changes.changereports` use it (ignoring tests), and they can all switch to the new `LoadDataset`. So that kinda sidesteps the issue.

> We shouldn't be giving subsystems an escape hatch where they can start dictating load locations

Very much agree that subsystems should not be aware of the load location.

dsref/load.go
dustmop: I also like that better, making it explicit is good.

base/dataset.go
```diff
@@ -27,11 +27,18 @@ type loader struct {
...
 
-// LoadDataset fetches, derefernces and opens a dataset from a reference
-// implements the dsfs.Loader interface
-func (l loader) LoadDataset(ctx context.Context, ref dsref.Ref, source string) (*dataset.Dataset, error) {
-	if source != "" {
-		return nil, fmt.Errorf("only local datasets can be loaded")
+// LoadDataset is not implemented for base
+func (l loader) LoadDataset(ctx context.Context, refstr string) (*dataset.Dataset, error) {
+	return nil, fmt.Errorf("not implemented")
+}
```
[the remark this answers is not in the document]
dustmop: This base implementation of Loader is only used for tests (sql and transform), which are then defining their own wrapper around it (what used to be ParseAndResolveFunc). It may be possible to remove it instead, but I haven't spent time yet to figure out how much test cleanup that requires.

cmd/validate.go
```diff
@@ -138,7 +138,7 @@ func (o *ValidateOptions) Run() (err error) {
 	}
 
 	ctx := context.TODO()
-	res, err := o.inst.Dataset().Validate(ctx, p)
+	res, err := o.inst.WithSource("local").Dataset().Validate(ctx, p)
```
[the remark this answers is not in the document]
dustmop: Right, this is here mostly just to maintain feature parity; validate already had this check.

Curious if you have any thoughts on: https://github.com/qri-io/qri/issues/1752

cmd/get_test.go
```diff
@@ -527,8 +527,7 @@ func TestGetRemoteDataset(t *testing.T) {
 		t.Errorf("response mismatch\nwant: %q\n got: %q", expect, err)
 	}
 
-	expect = dstest.Template(t, `pulling other_peer/their_dataset from registry ...
```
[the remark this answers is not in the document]
dustmop: 1. Not sure, but it only appeared here in the test due to https://github.com/qri-io/qri/commit/7d92e0c86dbc7e9b546870303d71c830210cdcaa, which makes me think it's a test-only problem. I'm fairly certain the output appears on the real command-line, but I'll make sure to try it out.
2. Yes, that will change in the parameter rename: https://github.com/qri-io/qri/pull/1735/files

cmd/remove.go
```diff
@@ -152,7 +152,7 @@ func (o *RemoveOptions) Run() (err error) {
 	}
 
 	ctx := context.TODO()
-	res, err := o.inst.Dataset().Remove(ctx, &params)
+	res, err := o.inst.WithSource("local").Dataset().Remove(ctx, &params)
```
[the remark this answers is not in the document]
dustmop: This lib method is only for removing local remotes (`/ds/remove`), the remote case it handled a few lines of code up by calling `o.RemoveRemote()`.

dsref/load.go
b5: Love it. The `Loader` interface  abstracts a common composition (Parse,Resolve,Load), dropping `LoadResolved` is a win. Double points if the code that replaces it presents a technique for "loading without a loader"

cmd/remove.go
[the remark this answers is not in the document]
b5: perfect. I was apparently too lazy to unfurl this during review

api/testdata/api.snapshot
```diff
@@ -3792,19 +3792,13 @@ Content-Type: application/json
 }
 
 /* snapshot: TestDatasetHandlers remove case 1: POST /remove/peer/cities */
-HTTP/1.1 200 OK
+HTTP/1.1 500 Internal Server Error
```
ramfox, line 3795: These tests are removed in #1755

dustmop:
Ready for review now, thanks to @b5 for providing insight into how to fix the `lib/load_test.go` test.

b5 (approved):
🥳

PR #1753, opened 2021-04-09
refactor(api): remove unneeded `DatasetHandlers`

### LINKED ISSUE #1721: Running list of http RPC bugs
Known RPC bugs:

#### Get - rpc doesn't error but no response
Problem:
The http api is expecting a `GetResponse` over the wire (since that is the return type of the `dataset.get` method). However, our api handler does further adjustments to the `GetResponse` depending on what the user requested in their inital request. Eg, if "meta" is selected, only the meta section is returned. There is pagination if the body was requested, etc. So, unlike most of our endpoints, our api response is not the same as our lib response params. When Dispatch tries to decode the response into a `GetResponse`, it can't because the response isn't shaped like `GetResponse`. 

Potential solution: We may need to unify the possible get responses & figure out how to represent all variations in a `GetResponse`, and have the `lib.Get` method return only what is needed, rather than returning the entire dataset and having the api and cmd layers narrow down what needs to be shown.

ramfox:
`AESave`, `AEPull`, & `AERemove` endpoints are all handled using `NewHTTPRequestHandler`
`AEPeerList` & `PeerListHandler` is removed
Removes associated tests.

closes #1721

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -215,8 +215,6 @@ func NewServerRoutes(s Server) *mux.Router {
 	m.Use(refStringMiddleware)
 	m.Use(token.OAuthTokenMiddleware)
 
-	dsh := NewDatasetHandlers(s.Instance, cfg.API.ReadOnly)
-
 	var routeParams refRouteParams
 
 	// misc endpoints
@@ -229,7 +227,6 @@ func NewServerRoutes(s Server) *mux.Router {
 
 	// aggregate endpoints
 	m.Handle(lib.AEList.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "collection.list"))).Methods(http.MethodPost)
-	m.Handle(lib.AEPeerList.String(), s.Middleware(dsh.PeerListHandler(lib.AEPeerList.String())))
 	m.Handle(lib.AESQL.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "sql.exec"))).Methods(http.MethodPost)
 	m.Handle(lib.AEDiff.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "diff.diff"))).Methods(http.MethodPost, http.MethodGet)
 	m.Handle(lib.AEChanges.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "diff.changes"))).Methods(http.MethodPost, http.MethodGet)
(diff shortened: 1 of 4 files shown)


api/api.go
```diff
@@ -241,24 +238,23 @@ func NewServerRoutes(s Server) *mux.Router {
...
 	m.Handle(lib.AERename.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.rename"))).Methods(http.MethodPost, http.MethodPut)
-	routeParams = newrefRouteParams(lib.AESave, false, false, http.MethodPost, http.MethodPut)
-	handleRefRoute(m, routeParams, s.Middleware(dsh.SaveHandler(lib.AESave.String())))
-	routeParams = newrefRouteParams(lib.AEPull, false, false, http.MethodPost, http.MethodPut)
-	handleRefRoute(m, routeParams, s.Middleware(dsh.PullHandler(lib.AEPull.NoTrailingSlash())))
+	m.Handle(lib.AESave.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.save"))).Methods(http.MethodPost)
+	m.Handle(lib.AEPull.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "remote.pull"))).Methods(http.MethodPost)
 	m.Handle(lib.AEPush.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "remote.push"))).Methods(http.MethodPost)
 	m.Handle(lib.AERender.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.render"))).Methods(http.MethodPost)
-	m.Handle(lib.AERemoteRemove.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "remote.remove"))).Methods(http.MethodPost)
```
Arqu, line 253: While not mentioned in the PR comment, `remote.remove` was supposed to go away too unless it was not part of that PR. I'll be doing another round of cleanup tomorrow I hope, happy to pick it up there.

api/api.go
```diff
@@ -241,24 +238,23 @@ func NewServerRoutes(s Server) *mux.Router {
...
 	m.Handle(lib.AEManifest.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.manifest"))).Methods(http.MethodPost)
 	m.Handle(lib.AEManifestMissing.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.manifestmissing"))).Methods(http.MethodPost)
 	m.Handle(lib.AEDAGInfo.String(), s.Middleware(lib.NewHTTPRequestHandler(s.Instance, "dataset.daginfo"))).Methods(http.MethodPost)
 
+	// non POST/json dataset endpoints
+	dsh := NewDatasetHandlers(s.Instance, cfg.API.ReadOnly)
+	routeParams = newrefRouteParams(lib.AEGet, false, true, http.MethodGet, http.MethodPost)
+	handleRefRoute(m, routeParams, s.Middleware(dsh.GetHandler(lib.AEGet.String())))
+	m.Handle(lib.AEUnpack.String(), s.Middleware(dsh.UnpackHandler(lib.AEUnpack.NoTrailingSlash())))
+
```
Arqu, line 257: Unsure if we should move unpack into the `misc` handlers as we mentioned on the call. I'm fine with both here.

api/datasets_test.go
```diff
@@ -20,183 +13,6 @@ import (
...
-	h := NewDatasetHandlers(run.Inst, false)
-
-	// TODO: Remove this case, update API snapshot.
-	saveCases := []handlerTestCase{
-		{"POST", "/?new=true", mustFile(t, "testdata/newRequestFromURL.json"), nil},
-		{"DELETE", "/", nil, nil},
-	}
-	runHandlerTestCases(t, "init", h.SaveHandler(""), saveCases, true)
-
-	getCases := []handlerTestCase{
```
Arqu, line 41: The few `get` tests are useful for reference for a short while longer, but happy with them going away, more power to the new tests I guess. 👍

Arqu (approved):
Some random comments, but 🚢  it

api/datasets_test.go
ramfox: Yea the issue here, was it felt like more work to get the tests working correctly, because they rely on the save handler saving datasets before "getting" the datasets, and the save handler doesn't exist anymore. Since we are planning a whole round of testing, it seemed like, exactly as you said, "more power to the new tests I guess".

PR #1754, opened 2021-04-09
fix(lib): Align input params across all lib methods

dustmop:
Modify names of all input param structs that are used for lib methods. This includes things like:

* Refstr -> Ref
* Component -> Selector
* Peername -> Username (except for p2p)
* Source moving to the scope (Remote remains for remote `pull` and `remove`)
* a few other small changes

Some outstanding work renames: the All flag from Get should be removed, but doing so will break functionality (such as /body.csv on the API) until Cursors exist. Also, there's some dscache specific tests that are being skipped, as I think we should be using some other mechanism to enable dscache instead of a field on the input params, such as an environmental variable or an option in the config file. Finally, ListParams need some additional work to combine them into a common struct.

### DIFF
--- a/base/dataset.go
+++ b/base/dataset.go
@@ -157,7 +157,7 @@ func CloseDataset(ds *dataset.Dataset) (err error) {
 }
 
 // ListDatasets lists datasets from a repo
-func ListDatasets(ctx context.Context, r repo.Repo, term, profileID string, offset, limit int, RPC, publishedOnly, showVersions bool) ([]dsref.VersionInfo, error) {
+func ListDatasets(ctx context.Context, r repo.Repo, term, profileID string, offset, limit int, publishedOnly, showVersions bool) ([]dsref.VersionInfo, error) {
 	fs := r.Filesystem()
 	num, err := r.RefCount()
 	if err != nil {
@@ -228,9 +228,6 @@ func ListDatasets(ctx context.Context, r repo.Repo, term, profileID string, offs
 			ds.Peername = ref.Peername
 			ds.Name = ref.Name
 			ref.Dataset = ds
-			if RPC {
-				ref.Dataset.Structure.Schema = nil
-			}
 
 			if showVersions {
 				dsVersions, err := DatasetLog(ctx, r, reporef.ConvertToDsref(ref), 1000000, 0, false)
--- a/cmd/push.go
+++ b/cmd/push.go
@@ -38,7 +38,7 @@ If no remote is specified, qri pushes to the registry.`,
 	}
 
 	cmd.Flags().BoolVarP(&o.Logs, "logs", "", false, "send only dataset history")
-	cmd.Flags().StringVarP(&o.RemoteName, "remote", "", "", "name of remote to push to")
+	cmd.Flags().StringVarP(&o.Remote, "remote", "", "", "name of remote to push to")
 
 	return cmd
 }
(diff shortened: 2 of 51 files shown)


base/dataset.go
```diff
@@ -228,9 +228,6 @@ func ListDatasets(ctx context.Context, r repo.Repo, term, profileID string, offs
 			ds.Peername = ref.Peername
 			ds.Name = ref.Name
 			ref.Dataset = ds
-			if RPC {
```
Arqu, line 231: I can't remember exactly what, but when converting to HTTP RPC I had to carry this over because removing it produced an odd bug.

lib/remote.go
```diff
@@ -52,25 +47,17 @@ func (m RemoteMethods) Feeds(ctx context.Context, p *FeedsParams) (map[string][]
...
 	}
-
-	params := *p
-	if params.Ref == "" {
-		params.Ref = r.FormValue("refstr")
-	}
-	if params.Remote == "" {
-		params.Remote = r.FormValue("remote")
+	if p.Ref == "" {
+		p.Ref = r.FormValue("refstr")
```
Arqu, line 59: Should we also change `r.FormValue("refstr")` > `r.FormValue("ref")` ~~and have the middleware reflect that too | I now checked and that part is done so we should just make sure all `refstr` are converted to `ref`~~?

Arqu (commented):
I did a first pass, but will have to have another go at it. Will also try to remember the `List w/RPC` issue.

cmd/push.go
```diff
@@ -71,10 +71,12 @@ func (o *PushOptions) Run() error {
...
 		p := lib.PushParams{
 			Ref:    ref,
-			Remote: o.RemoteName,
+			Remote: o.Remote,
 		}
 
-		res, err := o.inst.Remote().Push(ctx, &p)
+		// Though push is pushing to a remote, it has to resolve datasets
+		// from your local collection.
+		res, err := o.inst.WithSource("local").Remote().Push(ctx, &p)
```
ramfox, line 79: very helpful comment

lib/remote.go
dustmop: Ah, I missed this one, good call!

lib/http.go
```diff
@@ -232,7 +232,8 @@ func NewHTTPRequestHandler(inst *Instance, libMethod string) http.HandlerFunc {
 			return
 		}
 
-		res, cursor, err := inst.Dispatch(r.Context(), libMethod, p)
+		source := SourceFromRequest(r)
+		res, cursor, err := inst.WithSource(source).Dispatch(r.Context(), libMethod, p)
```
ramfox, line 236: 🙌

base/dataset.go
dustmop: I *think* the bug was that certain values wouldn't serialize properly over RPC because they didn't have json marshallers defined for their type. I'm thinking of this: https://github.com/qri-io/qri/commit/08322923674f999ad6ebc8e4e822ce4ac03dd291, but possibly for some other type of data. If we do some manual testing and find that it's still a problem, and can't be fixed by defining another serializer hook, we can add something to `scope` that knows if it's operating over RPC. Would be best if we could avoid having the parameter on the input param struct.

ramfox (approved):
Did a search for `Refstr`, `Peername` etc, couldn't find anything that should have been changed in params.

One thing I did find, but this is probably for a follow up, we use the term "peername" in some of the documentation of our cli commands. I've found it specifically in `cmd/render.go` & `cmd/fsi.go`.

PR #1756, opened 2021-04-13
wip(get): Fix http rpc return value for dataset.get

dustmop:


ramfox:
This works for a plain `qri get username/dataset`, but if you try to specify a component or body, you get an empty response. Also, over rpc, it can only print to the user as json, for some reason it is failing on parsing as yaml.

Popping in to revive the "maybe we need to change `GetResult`" convo, from our notes:
https://github.com/qri-io/qri/issues/1731#issuecomment-816093726

```
type GetResult struct {
    res interface{}
    type GetResultType
    format GetSerializationType 
}
```

dustmop:
You're totally correct, and it appears that changing the return type as you said is the correct solution. I wanted to put together this demo just to verify that the problem wasn't about reflection and dispatch, as was brought up in discord chat.

dustmop:
Closing this, the discussion in https://github.com/qri-io/qri/issues/1758 has the correct approach for how we handle this going forward.

Issue #1760, opened 2021-04-14
JSON API Test suite


Arqu:
Tracker issue for getting the first version of the JSON test suite up and running:

- [x] set up spec runner
- [x] definitions of expected responses (status code and content type for the first pass)
- [x] enumerate all spec `.json` files
- [x] enumerate all endpoints
- [x] cover core param variations for endpoints
- [x] make tests pass wherever possible
- [ ] fix remaining tests

Issue #1758, opened 2021-04-13
refactor(lib): Adjust `GetResponse` to be polymorphic


ramfox:
There are many different cases that our `lib.Get` method (and our `/ds/get` endpoint) are covering. In order for `lib.Get` to work over rpc, we need to adjust the `GetResponse` so that we can send over different types of return values using the http/json api.

```
type GetResponse struct {
    res interface{}
    type GetResultType
}
```

ramfox:
After some research & experimenting, I don't think just changing `GetResult` to be polymorphic is going to answer all our problems.

I'm proposing a few changes that should make Get & it's functionality easier to grok.

### taking stock of current expectations

|             | json | yaml | csv | zip | script |
| ----------- | ---- | ---- | --- | --- | ------ |
| --          | y    | y    | n   | y   | n      |
| commit      | y    | y    | n   | n   | n      |
| body        | y    | y    | y   | n   | n      | 
| meta        | y    | y    | n   | n   | n      |
| readme      | y    | y    | n   | n   | y      | 
| stats       | y    | y    | n   | n   | n      |
| structure   | y    | y    | n   | n   | n      | 
| transform   | y    | y    | n   | n   | y      | 
| viz         | y    | y    | n   | n   | y      |
| rendered    | n    | n    | n   | n   | y      |


one-off behavior that doesn't exist uniformly over api & cli:
  - /ds/get/username/name/dataset.zip returns zip contents of entire dataset, `lib.Get` returns data in `results.Bytes` & it's written to the response w/ a filename & content type set
  - /ds/get/username/name/body.csv returns body as a csv - same as above
  - /ds/get/username/name/body.json returns body as json - same as above
  - qri get readme.script username/name - at `lib.Get` , if `p.Selector` contains `script`, gets scriptfile from dataset & reads all into `results.Bytes`, cmd returns correct file type
  - qri get viz.script username/name - at `lib.Get` , if `p.Selector` contains `script`, gets scriptfile from dataset & reads all into `results.Bytes`, cmd returns correct file type
  - qri get rendered.script username/name - at `lib.Get` , if `p.Selector` contains `script`, gets scriptfile from dataset & reads all into `results.Bytes`, cmd returns correct file type
  - qri get transform.script username/name - at `lib.Get` , if `p.Selector` contains `script`, gets scriptfile from dataset & reads all into `results.Bytes`, cmd returns correct file type


### 3 main pain points:
[...]

PR #1762, opened 2021-04-15
tests(api): json test suite

Arqu:
Part of https://github.com/qri-io/qri/issues/1731
Closes https://github.com/qri-io/qri/issues/1760
Also does some clean up to bring the API closer to spec.


Arqu:
Some of the unresolved things:
- What was the final verdict on `/remote/remove` - that should go away?
- `peer.connections` & `peer.connectedqriprofiles` should go away?

All the routes in `api/spec/testdata/skip.json` should be fixed
- `Profile` requires a simple setup of the correct inputs
- `Remote` - I'm not familiar enough with this piece but generally it's missing a remote/registry against which to work, unsure how to set it up at the moment
- `Peers` - connect and disconnect are missing, not sure how to tackle the network aspect, need to dig into it
- `/ds/pull & push` missing the network component
- `/ds/render` need to set up a dataset with a readme
- `/ds/validate` unsure need to dig in
- `/ds/unpack` should we even test this here as it's a sugar route
- `/ds/manifest etc` failing on some IPFS dependency
- `/qfs/ipfs/...` - need to figure out what IPFS addresses we have in the tests

Some bugs/regressions I found:
- /wd/write - nil ptr exception
- /sql - something changed recently and now it has issue with the output that's being received
- /ds/get - can take the ref from the POST body, only as a query param

api/spec/spec.go
```diff
@@ -0,0 +1,236 @@
...
+		"testdata/misc.json",
+		"testdata/peer.json",
+		"testdata/profile.json",
+		"testdata/remote.json",
+		"testdata/working_directory.json",
+
+		// sync.json is intentionally left out
+		// as it's more a protocol that doesn't belong
+		// in the RPC API
+		// "testdata/sync.json",
```
Arqu, line 49: sync endpoints remain untested in this pass

api/spec/spec.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	sl := mustLoadSkipList(t, filepath.Join(specPackagePath, "testdata/skip.json"))
```
Arqu, line 52: added the skip mechanics

api/spec/spec.go
```diff
@@ -0,0 +1,236 @@
...
+	mockDataServer.Start()
+	return mockDataServer
+}
+
+// TestCase is a single request-response round trip to the API with parameters
+// for constructing the request & expectations for assessing the response.
+type TestCase struct {
+	Endpoint string            // API endpoint to test
+	Method   string            // HTTP request method, defaults to "GET"
+	Params   map[string]string // Request query or path parameters
```
Arqu, line 100: added query param support

api/spec/spec.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	return fmt.Sprintf("<unexpected response body. Content-Type: %q DataType: %T>", ct, bd)
+}
+
+// Response holds the expected HTTP response
+type Response struct {
+	Code       int
+	Headers    map[string]string
+	IgnoreBody bool
```
Arqu, line 203: Should remove this, leftover from some body checking experimenting

api/test_runner_test.go
```diff
@@ -57,6 +58,7 @@ func NewAPITestRunner(t *testing.T) *APITestRunner {
 
 func (r *APITestRunner) Delete() {
 	os.RemoveAll(r.TmpDir)
+	os.RemoveAll("/tmp/api_test")
```
Arqu, line 61: This is a bad hack to make tests work as the JSON suite is hard coded and the tmpdir is generated at runtime. Maybe invest a bit of time to make some sort of templating of values possible in the suite. I'm just afraid of introducing "special logic" to the test file contents which would make the tests much less portable and require other users to understand implementation details of our own.

Arqu (commented):
Highlighting some conversation points.

Other than that, done some of the API spec clean up as it went hand in hand with the test wrriting. Among other things:
- Removed the `log` endpoints and moved `history` to `activity` along with setting it in the propper methods
- Moved a few more of those around like `push` from `remote` into `dataset`
- Removed some unused endpoints
- Brought back in the token endpoint

api/test_runner_test.go
b5: totally, I'd ask that we make this directory `/tmp/qri_api_test` just to be super sure the dir doesn't collide with other software

b5 (approved):
I haven't had a chance to dig in on the tests themselves, but running the tests locally gives 40% coverage of the `api` package. Things are passing, no weird artifacts. Lovely. I'd like to dig in & do a proper review, but we could save that for a follow up PR

lib/http.go
```diff
@@ -226,7 +226,7 @@ func NewHTTPRequestHandler(inst *Instance, libMethod string) http.HandlerFunc {
 			return
 		}
 
-		if err := DecodeParams(r, &p); err != nil {
+		if err := DecodeParams(r, p); err != nil {
```
Arqu, line 229: @ramfox thanks for the hint in https://github.com/qri-io/qri/issues/1758. The fix for actually parsing the params was very needed here.

Arqu:
@b5 @dustmop @ramfox given your blessing I'd merge this and we can work through the `skip.json` list in subsequent PRs. This one already has too many minor fixes and cleanups that it's getting hard to review and maintain in a PR.

api/spec/spec.go
```diff
@@ -0,0 +1,234 @@
...
+)
+
+// AssertHTTPAPISpec runs a test suite of HTTP requests against the given base URL
+// to assert it conforms to the qri core API specification. Spec is defined in
+// the "open_api_3.yaml" file in the api package
+func AssertHTTPAPISpec(t *testing.T, baseURL, specPackagePath string) {
+	t.Helper()
+
+	// Create a mock data server. Can't move this into the testRunner, because we need to
+	// ensure only this test is using the server's port "55555".
```
dustmop, line 26: `newMockDataServer` and `api/spec/testdata/dataset.json` both use 55556, which disagree with the comment. I think the comment needs to be updated?

api/spec/spec.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	sl := mustLoadSkipList(t, filepath.Join(specPackagePath, "testdata/skip.json"))
+
+	for _, path := range testFiles {
+		t.Run(filepath.Base(path), func(t *testing.T) {
+			ts := mustLoadTestSuite(t, filepath.Join(specPackagePath, path))
+			for i, c := range ts {
+				skip := false
+				for _, s := range sl {
```
dustmop, line 59: How about extracting out the function `isInSkipList` so that this turns into a call `isInSkipList(sl, c.Endpoint)`?

dustmop (approved):
Two small fixes please, then looks good!

ramfox (approved):
fully agree with the general sentiment here that we should merge and build on top of this! Some questions around how we deal with network/peer/remote testing for the api in our future, dun dun DUN.

Arqu:
> fully agree with the general sentiment here that we should merge and build on top of this! Some questions around how we deal with network/peer/remote testing for the api in our future, dun dun DUN.

Yup that's the biggie which I wanted to isolate out.

Issue #1752, opened 2021-04-09
Consider adding a default source resolver to the Attributes of lib methods


dustmop:
Since I've been moving Source away from InputParams and over to scope, I've run into a lot of cases where certain commands only work if they're using a particular source.

For example:

* `pull` has to use `network` since by definition the ref is on another node
* `remove`, `rename`, `validate` all have to use `local`
* `push` has to use `local`, since it is something that starts on your own machine that goes to another

As of https://github.com/qri-io/qri/pull/1751 callers are required to always use `WithSource` to call these methods, and return an error if that isn't done. Instead, we can add a field to the Attributes map which will set a default source if the value is a non-empty string.

ramfox:
would appreciate something like this .Even if it's just a dev life improvement, I found myself searching through the code while reviewing #1754 to make sure I correctly understood default behavior (was trying to figure out if we would run into any problems if a source wasn't set for a particular lib method)

dustmop:
Totally, agreed that the current version is hard to read and keep straight in one's head. Thinking about it from the perspective of someone using `qri` as a library, they shouldn't be forced to set their own reasonable default values when we can do that instead.

b5:
Yep I'm all for it. I think it'll also make changing default behavior easier to do and review

Arqu:
+1 tests need sane defaults

dustmop:
Fixed by https://github.com/qri-io/qri/pull/1764 and https://github.com/qri-io/qri/pull/1765

PR #1764, opened 2021-04-16
fix(dispatch): Add default source resolver to Attributes

dustmop:
Attributes has a field for the default source for the method. This lets us remove the WithSource calls throughout the cmd package, and makes tests more readable.

PR #1765, opened 2021-04-16
feat(dispatch): add support for default source if none provided

Arqu:
closes https://github.com/qri-io/qri/issues/1752
required for https://github.com/qri-io/qri/pull/1762

### DIFF
--- a/lib/access.go
+++ b/lib/access.go
@@ -22,7 +22,7 @@ func (m AccessMethods) Name() string {
 // Attributes defines attributes for each method
 func (m AccessMethods) Attributes() map[string]AttributeSet {
 	return map[string]AttributeSet{
-		"createauthtoken": {AECreateAuthToken, "GET"},
+		"createauthtoken": {AECreateAuthToken, "GET", ""},
 	}
 }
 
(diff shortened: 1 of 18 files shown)


lib/access.go
```diff
@@ -22,7 +22,7 @@ func (m AccessMethods) Name() string {
 // Attributes defines attributes for each method
 func (m AccessMethods) Attributes() map[string]AttributeSet {
 	return map[string]AttributeSet{
-		"createauthtoken": {AECreateAuthToken, "GET"},
+		"createauthtoken": {AECreateAuthToken, "GET", ""},
```
b5, line 25: with this move from 2 -> fields, and the added combination of many field values being default zero values, I think we should switch all of these to named params:

```go
"createauthtoken": { endpoint: AECreateAuthToken, verb: get }
```

I know it's not as pretty, but I think it reads easier to see each field that actually uses a non-empty-string resolver default have the field show up & change the line length

Arqu:
Closing out in favor of https://github.com/qri-io/qri/pull/1764 and will move over missing bits there.

Issue #1507, opened 2020-09-03
commit component doesn't exist in FSI


dustmop:
Because FSI only reads component files in the working directory, and we have chosen to not create a "commit.json" file because it is not "user-editable", it is not possible to get the commit when a dataset is checked out. This is bad because it also means the user cannot look at (for example) the commit message for their previous commit using `qri get commit.message`. It has also been a cause of confusion in the development of the python client: https://github.com/qri-io/qri-python/commit/669e4ca260986a3895ca451ae1cd23b2343f6a00#diff-c94ef49167d684ce3fd9905de55a19e3R140

It is probably worth pulling the last existing commit from the qri repository when the FSI dataset is read, somewhere in `base/component`. Note that that code is a bit old and has a lot of TODOs; it may be in need of some general fixing up.

PR #1766, opened 2021-04-20
refactor(Get): adjust `lib.Get`, `qri get`, and `/get` handlers to cover all special cases as well as rpc

### LINKED ISSUE #1758: refactor(lib): Adjust `GetResponse` to be polymorphic

ramfox:
closes #1758


lib/datasets_test.go
```diff
@@ -439,21 +515,48 @@ func TestDatasetRequestsGetFSIPath(t *testing.T) {
...
+	cases := []struct {
+		description string
+		params      *GetParams
+		expect      interface{}
+	}{
+
+		{"ref without path",
+			&GetParams{Ref: "peer/movies"},
+			setDatasetName(&fsiExpectedDs, "peer/movies")},
+
```
ramfox, line 533: This test case is failing, and TBH I'm, not exactly sure what should be "correct" in this case.

I know when we "checkout" a dataset to fsi, we don't pull computer generated fields into the file, since these should not be editable.

However, for this get function, should we only be passing through the content of the files, or should we be passing through a completed look at the dataset? Basically, should fields like "commit", "stats", and "structure.checksum" be empty, or should they be filled in with the content from the latest committed version?

I don't understand the expectations/uses, so I'm having trouble answering this question.

Here is the diff:
```              &dataset.Dataset{
              	... // 1 ignored field
              	Body:      nil,
              	BodyBytes: nil,
            - 	BodyPath:  "/mem/QmQpAK5bPJKj65Bmzwa8WuoASiLUW4Hf4hR4h8ddPnEmAe",
            + 	BodyPath:  "/var/folders/4d/4pgbh8d1083g1f7skny7bwy00000gn/T/get_fsi_test120231655/movies/body.csv",
              	Commit:    nil,
              	ID:        "",
              	Meta: &dataset.Meta{
              		... // 1 ignored and 9 identical fields
              		Language:  nil,
              		License:   nil,
            - 		Path:      "/mem/QmZQNhYYVRx8LyMmPV9mqzVZVEeZKpso4Ywu7nwyWvT4X4",
            + 		Path:      "",
              		Qri:       "md:0",
              		ReadmeURL: "",
              		... // 3 identical fields
              	},
              	Name:         "movies",
            - 	Path:         "/mem/QmV6bPoBtAg2P3v1cTLjyXjcaYKodjwiYhywxzo8h36m9d",
            + 	Path:         "",
              	Peername:     "peer",
              	PreviousPath: "",
              	... // 2 identical fields
              	NumVersions: 0,
              	Qri:         "",
              	Structure: &dataset.Structure{
            - 		Checksum:     "/mem/QmQpAK5bPJKj65Bmzwa8WuoASiLUW4Hf4hR4h8ddPnEmAe",
            + 		Checksum:     "",
              		Compression:  "",
            - 		Depth:        2,
            + 		Depth:        0,
              		Encoding:     "",
            - 		ErrCount:     4,
            + 		ErrCount:     0,
            - 		Entries:      2335,
            + 		Entries:      0,
              		Format:       "csv",
              		FormatConfig: {"headerRow": bool(true)},
            - 		Length:       49753,
            + 		Length:       0,
            - 		Path:         "/mem/QmQqNmcUdpt9x6cUYp4YJNpppkQGhLEDzs1PsNtzQmi1RK",
            + 		Path:         "",
              		Qri:          "st:0",
              		Schema:       {"items": map[string]interface{}{"items": []interface{}{map[string]interface{}{"title": string("title"), "type": string("string")}, map[string]interface{}{"title": string("duration"), "type": string("integer")}}, "type": string("array")}, "type": string("array")},
              		Strict:       false,
              	},
              	Stats:     nil,
              	Transform: nil,
              	Viz:       nil,
              }
```

base/body.go
```diff
@@ -45,6 +45,30 @@ func ReadBody(ds *dataset.Dataset, format dataset.DataFormat, fcfg dataset.Forma
 	return data, nil
 }
 
+func ReadBodyAsInterface(ds *dataset.Dataset, limit, offset int, all bool) (interface{}, error) {
```
ramfox, line 48: Considering we should rename this to `GetBody` & rename the current `ReadBody` to `ReadBodyBytes`, or if we think it is self explanatory, keep it as `ReadBody`

fsi/body.go
```diff
@@ -81,3 +82,74 @@ func GetBody(dirPath string, format dataset.DataFormat, fcfg dataset.FormatConfi
 
 	return dsio.ConvertFile(file, structure, st, limit, offset, all)
 }
+
+// GetBodyAsInterface is an FSI version of base.ReadBodyAsInterface
+func GetBodyAsInterface(dirPath string, offset, limit int, all bool) (interface{}, error) {
```
ramfox, line 87: like the `base` versions, maybe this should be renamed to `GetBody` & the current `GetBody` should be `ReadBody` or `ReadBodyBytes`

fsi/body.go
```diff
@@ -81,3 +82,74 @@ func GetBody(dirPath string, format dataset.DataFormat, fcfg dataset.FormatConfi
...
+	err = component.ExpandListedComponents(components, nil)
+	if err != nil {
+		return nil, err
+	}
+	return components, nil
+}
+
+// prepareInStructure loads and fills the structure component. If no schema exists on the structure, if generates one
+// using the bodyComponent
+func prepareInStructure(components component.Component, f *os.File) (*dataset.Structure, error) {
```
ramfox, line 123: Better name?
`structureWithSchema`
`ensureStructureWithSchema`
`generateStructure`
`prepareStructureForReadingBody`

lib/datasets.go
```diff
@@ -666,101 +599,33 @@ type datasetImpl struct{}
 
 // Get retrieves datasets and components for a given reference.t
 func (datasetImpl) Get(scope scope, p *GetParams) (*GetResult, error) {
-	res := &GetResult{}
-
-	scope.EnableWorkingDir(true)
-	ds, err := scope.Loader().LoadDataset(scope.Context(), p.Ref)
+	ref, ds, err := openAndLoadFSIEnabledDataset(scope, p)
```
ramfox, line 602: I ultimately decided it was probably overkill for this pr, but this is where I was curious if loading & opening the dataset should happen in a lower package

lib/datasets.go
```diff
@@ -74,136 +74,62 @@ type GetParams struct {
...
+var validSelector = regexp.MustCompile(`^[\w-\.]*[\w]$`)
+
 // GetResult combines data with it's hashed path
 type GetResult struct {
-	Ref       *dsref.Ref       `json:"ref"`
-	Dataset   *dataset.Dataset `json:"data"`
-	Bytes     []byte           `json:"bytes"`
-	Message   string           `json:"message"`
-	FSIPath   string           `json:"fsipath"`
-	Published bool             `json:"published"`
```
ramfox, line 206: Want to note that we are losing these two pieces of information, and need an alternative way of getting the `FSIPath` and the `Published` status. Although, thinking back to the ui presentation this morning, neither piece of information was needed in those mockups.

lib/datasets_test.go
dustmop: So yes, the current implementation has been that `Get` on a dataset which has a working directory will only ever look at the files on disk, in that working directory. It won't load anything from the repository or IPFS. As you rightfully point out, this means that the `stats` and `commit` will always be nil, since we don't store any "machine-targeted" files for an FSI directory. The structure is similar; a structure in a working directory will only contain the `Format`, `FormatConfig` and `Schema` fields, since the rest of the fields are all values derived from the last save and so shouldn't be user-editable.

This is arguably a misfeature. It means that commands like `qri get commit.message` won't work in FSI; so it's not possible to view the most recent commit information. I would argue that we should fix that, but the fix is probably not appropriate for this refactoring, and if it's not too hard we should keep the current behavior. Fixing the larger issue probably needs a bit of discussion and should be its own separate PR. An earlier issue talking about this problem: https://github.com/qri-io/qri/issues/1507

This is also definitely a good example of behavior we should document with the long discussed spec.

ramfox: This is extremely helpful, thank you.

api/middleware.go
```diff
@@ -69,19 +69,26 @@ func (s *Server) readOnlyCheck(r *http.Request) bool {
...
+	q := r.URL.Query()
+	for varName, val := range mux.Vars(r) {
+		if q.Get(varName) != "" {
+			return fmt.Errorf("conflict in query param: %s = %s", varName, val)
+		}
+		q.Add(varName, val)
+	}
+	r.URL.RawQuery = q.Encode()
+	return nil
+}
```
ramfox, line 90: needed to pull this out so we can use it in tests

lib/datasets_test.go
```diff
@@ -439,21 +517,72 @@ func TestDatasetRequestsGetFSIPath(t *testing.T) {
...
 	}
 
-	getParams := &GetParams{
-		Ref: "peer/movies",
-	}
-	got, err := inst.Dataset().Get(ctx, getParams)
+	var nilCommit *dataset.Commit
+
+	moviesDsAtPath := &dataset.Dataset{}
+	copier.Copy(moviesDsAtPath, moviesDs)
```
ramfox, line 523: is there a better way to deep copy a dataset?

lib/datasets.go
```diff
@@ -74,136 +74,62 @@ type GetParams struct {
...
+	Value interface{} `json:"value,omitempty"`
+	Bytes []byte      `json:"bytes,omitempty"`
```
dustmop, line 132: Is a separate `Bytes` field still needed? Only script is using it, but since `Value` is an `interface` it could just be putting that `[]byte` into `res.Value`.

Dispatch prefers method output values to be structs, in order to simplify how their allocation is done, but it's okay to just have this:
```
type GetResult struct {
  Value interface{}
}
```

api/handlers.go
```diff
@@ -5,42 +5,193 @@ import (
...
+
+		switch returnCase {
+
+		case "zip":
+			outBytes, err := inst.Dataset().GetZip(r.Context(), params)
+			if err != nil {
+				util.RespondWithError(w, err)
+				return
+			}
+			writeFileResponse(w, outBytes, "dataset.zip", "zip")
```
dustmop, line 39: It was kinda nice before how downloading a zip of a dataset would have the dataset name (and commit timestamp, I believe?) in the filename that got downloaded. An unfortunate consequence of simplifying the GetResult is that we lose the ability to do that. I don't know if it's worth the effort to add that back in, just wanted to add it as a data point.

api/handlers.go
```diff
@@ -5,42 +5,193 @@ import (
...
+
+	case format == "pretty":
+		fallthrough
+	case arrayContains(r.Header["Accept"], "text/json"):
+		if !(format == "json" || format == "pretty" || format == "") {
+			return "", nil, fmt.Errorf("format %q conflicts with header %q", format, "Accept: text/json")
+		}
+		fallthrough
+	case strings.HasSuffix(p.Selector, ".json"):
+		// curl -H "Accept: text/json" http://localhost:2503/ds/get/b5/world_bank_population
```
dustmop, line 160: Same here as well, please: `// Examples:`

fsi/body.go
```diff
@@ -81,3 +82,74 @@ func GetBody(dirPath string, format dataset.DataFormat, fcfg dataset.FormatConfi
...
+	}
+
+	rr, err := dsio.NewEntryReader(structure, f)
+	if err != nil {
+		return nil, fmt.Errorf("error allocating reader: %s", err)
+	}
+	return base.ReadEntries(rr)
+}
+
+func getAllComponents(dirPath string) (component.Component, error) {
```
dustmop, line 108: This is totally fine, just wanted to point out that this is likely a result of `component.Component` having an unwieldy API. These two calls should probably be combined into one, which would remove the need for this `getAllComponents` function. Personal TODO for myself.

dustmop (commented):
Ok! Lots of comments for this. The one in `cmd/get.go` in particular raises some questions that I think should be considered first, as its about the recent design decisions made around `Get`'s polymorphism. It's possible some choice was already made that I missed at some point.

Also, great job on getting this PR up, `Get` is so especially hairy so it's awesome to see it getting a much needed cleanup.

lib/datasets.go
```diff
@@ -52,7 +50,9 @@ func (m DatasetMethods) Name() string {
 func (m DatasetMethods) Attributes() map[string]AttributeSet {
 	return map[string]AttributeSet{
 		"componentstatus": {endpoint: AEComponentStatus, httpVerb: "POST"},
-		"get":             {endpoint: AEGet, httpVerb: "GET"},
+		"get":             {endpoint: AEGet, httpVerb: "POST"},
+		"getcsv":          {endpoint: denyRPC},
+		"getzip":          {endpoint: denyRPC},
```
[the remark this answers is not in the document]
ramfox: In person, we decided that for now that needs to be supported over rpc!

Main points:
- not sure that there are many people who would expect to both be running a qri node somewhere & expect all functionality to still work the same way over cli
- desktop would still be able to fetch over the api via the sugar endpoints
- it would be simple to add in if someone requests it later

cmd/get.go
```diff
@@ -129,59 +130,85 @@ func (o *GetOptions) Complete(f Factory, args []string) (err error) {
...
+		}
+	default:
+		res, err := o.inst.WithSource(o.Remote).Dataset().Get(ctx, p)
+		if err != nil {
+			return err
+		}
+		switch {
+		case lib.IsSelectorScriptFile(o.Selector):
+			outBytes = res.Bytes
+		case o.Format == "json" || (o.Selector == "body" && o.Format == ""):
```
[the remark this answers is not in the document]
ramfox: Recap of our decision and discussion about this issue!

We want to enforce to the user that Qri interprets your data for you and has a default output format, and that format is json (for now). This is particularly useful if/when we want to change Qri's underlying storage format (optimized in some way, but not a format that makes sense for humans to read). Making this expectation clear now means users can already grok that how the data is persisted in Qri doesn't mean that's how it is presented to them. 

To that end, we are keeping the functionality that running a "plain" `get body me/my_dataset` will by default return in json format, and `get body me/my_dataset --format csv` will return the body as csv.

Discussion:
- This is a breaking change and would effect folks who have expectations for the CLI already. 
- In the future, if we want to give users the ability to return a `raw` type, the simplest way would be:
  - Remove `GetCSV`, make `getBodyBytes` an exported `GetBodyBytes`, that is dispatch-ified
  - create `GetBodyBytesParams` that pulls in the expected params from `getBodyBytes`
  - this is the default behavior in `getBodyByte` currently, but to reenforce: if `params.Format` is empty, return the body bytes encoded as the default format that we get when we examine the structure.
  - in `cmd/get.go`, if the selector is `body`, call `inst.Dataset().GetBodyBytes`

lib/datasets.go
ramfox: Ah, okay, so b5 and I had this discussion in chat and it never surfaced into an issue. If we don't have both fields, and value encompases all of them, this would break over rpc.`json.Marshal` would serialize the byte slice as a base 64 encoded string, but then `json.Unmarshal` would deserialize into a string. Unless it's in a field that is explicitly a byte slice, in which case `json.Unmarshal` knows to decode.

Here is part of the discussion from discord, and we decided to just go with the first option as being the least disruptive.

> Actually, right after the call I realized that .script calls will still have the issue of possibly not deserializing properly as bytes over rpc, but this can be solved by either 
> 1) keeping both GetResult.Value and GetResult.Bytes, like we have now
> 2) if we specifically don't want GetResult to only have have both fields, and instead have only one field of type interface{}, we can primarily rely on an (actual) helper function lib.GetScript that calls Get & can type assert to a byte slice OR decode from a string, as needed.
> 3) move GetScript into its own datasetMethod, which would be v clean but could be overkill
> 4) We resurrect that GetParams discussion... which does feel like overkill

api/handlers.go
ramfox: Ahh yes, this was another tacit change that I never surfaced to the group, thank you. 

One solution I thought of was `lib.GetZip` could return `ZipResults`
```
type ZipResults struct {
  Bytes []bytes
  Name string // generated name
}
```

lib/datasets.go
dustmop: Oh right, that totally makes sense. Could you leave a comment on the field explaining this, that it's required to preserve byte slices over serialization?

api/handlers.go
```diff
@@ -5,42 +5,193 @@ import (
...
+	}
+}
+
+// possible return cases:
+//  - "zip"
+//  - "csv"
+//  - "pretty"
+//  - "text/json"
+//  - ""
+func getCaseAndParamsFromRequest(r *http.Request) (string, *lib.GetParams, error) {
```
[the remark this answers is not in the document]
ramfox: ty, def going to create an enum for this & rename.

Pointing out the `text/json` discrepancy made me realize something, all of these "OutTypes" are actually file types that we are potentially returning. Otherwise we are returning `application/json`. I thought it would be too confusing to have a "json" as a `Case`, since it could be potentially ambigious (hence the odd "text/json"), but renaming that to `OutFileType` would make "json" as a type less confusing.

api/handlers.go
```diff
@@ -5,42 +5,193 @@ import (
...
+
+	switch {
+	case format == "csv":
+		fallthrough
+	case arrayContains(r.Header["Accept"], "text/csv"):
+		if !(format == "csv" || format == "") {
+			return "", nil, fmt.Errorf("format %q conflicts with header %q", format, "Accept: text/csv")
+		}
+		fallthrough
+	case p.Selector == "body.csv":
```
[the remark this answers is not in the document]
ramfox: In "api/api.go" line 163 we have `handleRefRoute` that ensures we handle: "/ds/get/{username}/{name}/{selector}"
We have `refStringMiddleware` that takes a ref string from the muxVars and adds a query param "ref" to the request. We have `setMuxVarsToQueryParams` that takes all the mux vars and sets them as query params on the request.

So if we ping url "/ds/get/my_username/my_dataset/body.csv", "body.csv" get's seen as the `selector`.

I agree this is ambiguous & muddled and I don't love it. Part of the issues is I initially had `getOutfileTypeAndParamsFromRequest` as two functions, one that parsed the request and set the `GetParams` and one that set the `outfileType`. However, as you can see from the section you pointed out, sometimes what ends up in the `Selector` field, actually represents something beyond just being a selector.

One thought I had was to use a different term than "selector" in `handleRefRoute` & `setMuxVarsToQueryParams` so it is clear that what we get from this value may not be a traditional "selector". I just didn't know what to call it.

Regardless of if we change the term in the route, I shouldn't be setting `p.Selector` until the value has been de-coupled from special cases:
```
func getOutfileTypeAndParamsFromRequest(r *http.Request) (outfileType, *lib.GetParams, error) {
  // parse request into params
  maybeSelector := r.FormValue("selector")
  // go through different outfileType options & adjust `maybeSelector` if it's one of our special cases
  p.Selector = maybeSelector
  return outfileType, p, nil
}
```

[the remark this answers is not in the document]
dustmop: Could we use this?

```
"/ds/get/{username}/{name}/{componentFile}"
```

Let's say we go with that. If there any valid value for `componentFile` that we need to support aside from "body.csv"? I know users specifically asked for that one, are there others? If not, I'd prefer just checking this `componentFile` field against the one allowed value, and disallowing anything else. That is, seeing `componentFile == "body.csv"` would set `selector="body"`, and `format="csv"`, etc.

If a user is requesting some specific component or field, they should still be able to do:

```
"/ds/get/{username}/{name}?selector={selector}"
```

such as:

```
"/ds/get/dustmp/my_ds?selector=structure.field"
```

Does that work?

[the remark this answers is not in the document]
ramfox: Yea, the issue is that one of the sugar cases we definitely want to support IS having the selector as part of the url:

```
"/ds/get/ramfox/my_dataset/meta" // returns just the meta component
```

Which is why `selector` was used in the first place, but... again... isn't always true.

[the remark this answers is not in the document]
dustmop: Ok, wasn't aware that also should be allowed. I thought part of the api break was removing this type of ref in the url, or at least requiring callers to use something like `"/ds/get/ramfox/my_dataset?selector=meta"`.

If `componentFile` is too specific, I'd go with something general like `path`.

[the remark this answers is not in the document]
dustmop: (Post-edit follow-up): I'm ok with handling it in whatever way is simplest. I guess my main concerns are:

1) Not adding features we don't need to, since there's an maintenance cost. I think /dataset.json are /meta.json etc are unnecessary, unless I missed some explicit need for them. Pretty sure they're just holdovers from the original api.
2) Keeping the term "selector" accurate in how it is used, and not allowing values like "body.csv", which are filenames not selectors.

With those in mind, my vote would be using either `path` or `componentFile`, whichever you think makes more sense.

lib/datasets_test.go
ramfox: So looking into it, looks like `reflect.Copy` is only for slices/arrays/or copying a string into a []UInt8

Leaving this as us & adding an issue in `dataset` pitching a `dataset.Copy`.

api/handlers.go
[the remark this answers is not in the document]
ramfox: After taking a step back, I think the best most logical way to handle this is for us to keep the "selector" field pure, keep it as "selector", and handle "/ds/get/{username}/{name}/body.csv" separately.

All other "special cases" should be removed from this pr. As you mentioned above, there is a maintenance cost for every additional case & it is easier to add a case when a user requests it then it is to take a case away/maintain an old case. After reviewing these the special cases besides "body.csv", they aren't in our 0.10.0 endpoint list, I just got excited while coding and didn't keep in mind the maintenance cost.

This also dramatically reduces the complexity of the `GetHandler`, since there are way fewer "gotchyas" and expected return types.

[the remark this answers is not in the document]
dustmop: Sounds good! Reducing complexity is excellent!

ramfox:
Major question: we have some notes throughout our code about special ways to handle the body & limit & offset. What is the default? Default in master is to return `All` by default or when `Limit == -1 && offset == 0` 

Just making sure that's still what we want.

dustmop:
> Major question: we have some notes throughout our code about special ways to handle the body & the list params.
> 
> I'm specifically looking at the `api` package, but this can extend to `cmd`.
> 
> The default in master is to return `All` if no `Limit` or `Offset` are set and to return `All` when `Limit == -1 && offset == 0`
> 
> Just making sure that's still what we want, or should we instead be defaulting to a sensible `Limit`? 25 rows? 50? 100?
> 
> Since we are going to have a follow up specifically surrounding the cursor and pagination, I'm thinking we keep `All` by default, unless someone explicitly sets `Limit` or `Offset` (aka, `Limit != 0 && Offset != 0`). Thoughts?

A little while back, the limit was always set to a default of a small value like 25, but that got lost during the api / lib / dispatch cleanup to just use `all` instead. The idea is to keep http requests relatively tiny so that they don't timeout. I'm ok with keeping `all` for now, as its hopefully going to be replaced soon with cursors that handle pagination by repeatedly calling the endpoint.

base/select.go
```diff
@@ -34,6 +34,9 @@ func Select(ctx context.Context, fs qfs.Filesystem, ref dsref.Ref, valuePath str
 // ApplyPath gets a dataset value by applying a case.Sensitve.dot.separated.path
 // ApplyPath cannot select file fields
 func ApplyPath(ds *dataset.Dataset, path string) (interface{}, error) {
```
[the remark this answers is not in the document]
ramfox: Do you mean remove this function and replace calls to it w/ calls to `base.GetPathValue`, or do you mean have `ApplyPath` use `base.GetPathValue`?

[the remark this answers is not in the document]
dustmop: Oh sorry, I don't think it has to happen now. Maybe just a TODO for now is fine, I'm not sure exactly how to fix it just yet.

api/handlers_test.go
```diff
@@ -86,6 +125,51 @@ func TestDatasetGet(t *testing.T) {
...
+		t.Errorf("%s", w.Body.String())
+		t.Fatal(fmt.Errorf("expected call to unpack handler with GET method to return status code 404, got %d", w.Result().StatusCode))
+	}
+
+	r = httptest.NewRequest("POST", "/unpack", nil)
+	w = httptest.NewRecorder()
+	hf(w, r)
+	if w.Result().StatusCode != 500 {
+		t.Errorf("%s", w.Body.String())
+		t.Fatal(fmt.Errorf("expected call to unpack handler with GET method to return status code 404, got %d", w.Result().StatusCode))
```
dustmop, line 169: The error message says 404, but the condition is checking 500. They should match.

dustmop (approved):
One comment about a tiny mistake I noticed in a test's error message, but aside from that approved! Awesome to see this, excited to get it merged!!

api/handlers_test.go
ramfox: EAGLE EYESSSSS ty

PR #1767, opened 2021-04-20
docs: generate API reference documentation from source

b5:
Long hill to climb here, but It'd be great to get started with a program that produces a usable swagger doc. Things we should get completed in the first pass:

* [x] rebase this branch & properly document changes to `lib` that expose things needed for docs package
* [x] map all lib data structures to the `components/schemas` section of the doc
* [x] use reflection to populate a new `Response` field in `LibMethod` based on return type
* [x] set a new `Paginated bool`  field to true on `LibMethod` if the method returns a cursor (3 return args)
* [x] hardcode types defined outside `github.com/qri-io/qri` like `*dataset.Dataset`
* [x] populate `components/responses` in the openAPI doc
* [x] ignore non API routes
* [x] handle raw byte responses
* [ ] ~~capture sugar routes too~~
* [x] populate responses of each endpoint
    * [x] handle normal response object/base types
    * [x] handle array response
    * [x] handle error response
    * [x] handle paginated responses
* [x] sanitize output a bit (remove empty lines etc)
* [x] take a read through the result. Is is usable? It should be usable.
* [x] clean up code
* [ ] ~~integrate with CI to generate and quickly validate the spec~~

I've been working on this by pasting the results of test output into https://editor.swagger.io & iterating on the results.


Arqu:
Ok, so this is a bit ugly in the current state, but the thing I wanted to bring up for discussion is how to handle the response types we wanted to hard code. In the PR there's an example of the spec generated at `docs/docs/open_api_spec.yaml` and an example of just the `dataset.Dataset` struct encoded in yaml at `docs/docs/dataset_sample.yaml`. After some quick cleanup it sits at ~850 LOC. I have yet to produce the remaining structures for: `VersionInfo`, `StatusItem`, `Profile`, `Ref`, `DAGManifest`, `DAGInfo` & `ChangeReport`.

The point is that it's very cumbersome to write out manually, but even worse maintaining long term. I've taken a stab at making a POC to get it automated. It currently spits out some nearly done JSONSchema which can then be converted to yaml. This needs some more work but can be done with some attention to finer points of our structures and cleanup functions.

The alternative is to be a bit more bland and just take the "top level" properties and describe them in the response object and have the user go in the weeds if they really need to understand the full structs. Which is kinda meh, as for example they'd have knowledge there's a `Commit` in the `Dataset` but not knowing what that looks like.

It's not the worst though as these are the responses not the input params which definitely need to be represented fully.

Anyways, would love input @dustmop  @b5 @ramfox before doubling down on further work.

b5:
> The point is that it's very cumbersome to write out manually, but even worse maintaining long term. I've taken a stab at making a POC to get it automated.

Looking at the output and boy, you aren't kidding! I think this is far smarter to do in the automated fashion you've laid out here. With the approach you've laid out here, it's easier to write code that generates a more concise OpenAPI document than through hand-maintenance. So I'd vote to keep the reflection-based approach, and use it as a base to build upon.

This PR is getting large, and we should focus this effort on spitting out a _valid_ OpenAPI spec (basically as it is now), and then focus on making the document itself _useful_ in follow-up PR(s). I'm ok if the document is huge, just needs to be valid, and a thing we can build upon both by modifying source code that we're generating from, and massaging the doc generating code.

One small change I'd ask for: adding the reflection based approach introduces new `gomod` dependencies that I'd prefer be kept to the `docs` package itself. To that end, I'd ask that you declare a `go.mod` for the docs package itself, [here's an example](https://github.com/go-modules-by-example/submodules) of the structure I'm imagining.

Arqu:
This should now be ready to go in per the latest comment.

Remaining things to do:
- generate real response structs instead of generic placeholders
- add sugar routes
- automate in CI

lib/dispatch.go
```diff
@@ -336,6 +357,12 @@ func (inst *Instance) registerOne(ourName string, methods MethodSet, impl interf
 			regFail("%s: last output val should be error, got %v", funcName, outErrType)
 		}
 
+		// if outType != nil {
+		// 	fmt.Printf("regOne: f(%s) outType - '%s'\n", funcName, outType.String())
+		// } else {
+		// 	fmt.Printf("regOne: f(%s) outType is nil\n", funcName)
+		// }
```
b5, line 364: dead code, plz remove

lib/dispatch.go
```diff
@@ -249,6 +250,26 @@ type callable struct {
 	DenyRPC   bool
 }
 
+// AllMethods returns a method set for documentation purposes,
+func (inst *Instance) AllMethods() []MethodSet {
```
b5, line 254: this method being used _only_ for documentation is a potential source of drift. I'd initially intended to combine this with the function body of `*lib.Instance.RegisterMethods` somehow, but haven't gotten to it. Can we complete this comment & add a `TODO`?

b5 (commented):
some small nits, but otherwise LGTM

Arqu (approved):
Comments implemented. Merging.

Issue #1757, opened 2021-04-13
ensure that prove still works after dispatchification


dustmop:
The following situation should work:

1. A user installs and uses qri on their machine, signs up for cloud, pushes a dataset
2. A user goes to a second machine and installs qri on it
3. On this second machine they run `qri registry prove ...`
4. This second machine now has their account connected to cloud, and they can continue to push and pull datasets

We want to make sure this behavior is not broken when we release 0.10.0 https://github.com/qri-io/qri/issues/1731

dustmop:
Fixed by https://github.com/qri-io/qri/pull/1768.

PR #1768, opened 2021-04-20
fix(prove): Prove command also sets profile.KeyID

dustmop:
When a user runs the `prove` command, the registry will send back their original profileID which gets assigned to their profile. We need to save the previous ID to the profile.KeyID field, otherwise the keyStore will no longer be able to initialize itself, since without a keyID its default behavior is to match the profileID to the keypair (see `GetKeyID` in `profile/profile.go`).


dustmop:
Fixes https://github.com/qri-io/qri/issues/1757

Arqu (approved):
Yup, this is correct, the profile is supposed to hold a Key reference which can later be used to retrieve the proper/active key but in the default setup it should be the profile ID.

b5 (approved):
yep, agree this is intended & correct behavior

Issue #575, opened 2018-10-10
Document the new /unpack/ and other sugar endpoints in open_api yamls


dustmop:
In https://github.com/qri-io/qri/pull/574 we add the new "/unpack/" endpoint. It also needs to be added to api/open_api_2.yaml and api/open_api_3.yaml.

Arqu:
Should be covered in https://github.com/qri-io/qri/pull/1767

PR #1772, opened 2021-04-29
docs(api): add non lib routes to spec

Arqu:
Closes: https://github.com/qri-io/qri/issues/575

~~Still missing: `get` routes~~

### DIFF
--- a/docs/go.mod
+++ b/docs/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/getkin/kin-openapi v0.55.0
 	github.com/iancoleman/orderedmap v0.2.0
 	github.com/qri-io/ioes v0.1.1
-	github.com/qri-io/qri v0.9.13
+	github.com/qri-io/qri v0.9.14-0.20210428083546-967b708545a5
 	github.com/spf13/cobra v1.1.3
 	gopkg.in/yaml.v2 v2.4.0
 )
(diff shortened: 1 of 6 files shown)


Arqu:
Added path param support for routes and included `/ipfs/...` & `/get` & `/get/...`. But shortened the `/get/{ref}` options to just `{ref}` instead of all the different ref structure options (full & short ref) to avoid generating too much cruft, and I dislike the vagueness of the response definitions for it, but get is... complicated.

Anyways, I've included the latest spec output for quick looks/review, but will remove before merging.

docs/go.mod
```diff
@@ -6,7 +6,9 @@ require (
 	github.com/getkin/kin-openapi v0.55.0
 	github.com/iancoleman/orderedmap v0.2.0
 	github.com/qri-io/ioes v0.1.1
-	github.com/qri-io/qri v0.9.13
+	github.com/qri-io/qri v0.9.14-0.20210428083546-967b708545a5
 	github.com/spf13/cobra v1.1.3
 	gopkg.in/yaml.v2 v2.4.0
 )
+
+replace github.com/qri-io/qri => /Users/arqu/dev/qri/qri
```
Arqu, line 14: Will remove this too

dustmop (approved):
LGTM

PR #1777, opened 2021-04-29
refactor(lib, api): move sugar endpoints out of `lib` into `api`

### LINKED ISSUE #1776: Move sugar endpoint definitions into `api` from `lib`
We have done a great job in our codebase separating the concerns of api endpoints that require access to the `lib` package, and the endpoints that only exist at the `api` package level. However, we are breaking this rule when it comes to endpoint definition. 

Defining all the endpoints in one place made perfect sense as we were refactoring the api, and it helped us better keep track of what the expectations were. However, we should now pull the definition of endpoints that only exist at the `api` level to the `api` package as well.

Those endpoints are:

```
AEHome
AEHealth
AEIPFs
AEWebUI
AEUnpack
```

Must also update any docs!

ramfox:
The `lib` package was defining endpoints that don't have any relation to
methods defined in lib. These endpoints have been moved into `api/endpoints.go`:

```
AEHome
AEHealth
AEIPFs
AEWebUI
AEUnpack
```

closes #1776

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -215,19 +215,19 @@ func NewServerRoutes(s Server) *mux.Router {
 	var routeParams refRouteParams
 
 	// misc endpoints
-	m.Handle(lib.AEHome.String(), s.NoLogMiddleware(s.HomeHandler))
-	m.Handle(lib.AEHealth.String(), s.NoLogMiddleware(HealthCheckHandler))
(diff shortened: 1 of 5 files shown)


Arqu (approved):
LGTM, I'll rebase and update the docs definitions.

Issue #1775, opened 2021-04-29
DenyHTTP should be handled explicitly


dustmop:
The loop in `lib/routes.go` should be explicitly checking for methods with endpoint == `DenyHTTP`, and then skipping them. In its current implementation, all methods with this endpoint get `Handle`d but since their endpoint is a blank string, it's not possible to call any of them. This change should be functionally the same, but more correct, less confusing, and less prone to potential weirdness down the line.

PR #1778, opened 2021-04-30
fix(routes): skip over endpoints that are DenyHTTP

Arqu:
Closes: https://github.com/qri-io/qri/issues/1775

@dustmop noticed that writing the issue was more work than this commit. Sorry.


dustmop:
LGTM!

PR #1783, opened 2021-05-05
fix(sql): Fix sql command for arm build

ramfox:
cherry picks single commit from branch `release-0.10.0` branch by @dustmop, so that only the version bump and change report is in the release commit


ramfox:
self reviewing, since this is really @dustmop's work, LGTM

Issue #1784, opened 2021-05-06
future docs package improvements


ramfox:
1) We have a few commands that are specific to devs and that we aren't necessarily trying to advertise to the public. The main one that comes to mind is `qri logbook`, which is great for debugging, but would just take up oxygen in the docs for folks that are actually looking for `qri log`. We should have a mechanism for specifically not adding certain commands into the documentation.

2) The "table of contents" portion of the cli commands page should be generated from our code. It not longer reflects the generated commands. Edited by hand for this latest release, but we should probably be generating these from code as well.

3) should we add to the `make api-spec` command, renaming and moving the `temp/open_api_spec.yaml` file into `api/open_api_3.yaml`?

dustmop:
The help text for each command declared in `cmd/` is used to generate the docs on the website. However, each command contains the repeated "Options inherited from parent commands" for flags such as `--no-prompt`. Might be worth changing the generator so that it only shows this text once, rather than always repeating it.

PR #1786, opened 2021-05-07
fix(docs): update table of contents

ramfox:
hmmm, these changes probably should have gone through before the release, but I updated the "table of contents" part of the docs generator. have an issue open about it (https://github.com/qri-io/qri/issues/1784), but in the future we should probably generate those as well.

### DIFF
--- a/docs/docs.go
+++ b/docs/docs.go
@@ -94,28 +94,35 @@ on the distributed web.
 
 |Command | Description |
 |--------|-------------|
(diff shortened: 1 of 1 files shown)


Arqu (approved):
Good catch. Changes look good to me, but the link to the issue above points to a merged PR. Would be better to have an open issue for it until it's auto generated.

ramfox:
Thanks, fixed the link!

PR #1795, opened 2021-05-19
feat(automation): add `Orchestrator` that manages workflow automation

### LINKED ISSUE #1789: feat(automation): implement inital `Orchestrator`
### Test:
The initial orchestrator project will be complete when this test passes:
```
// automation/orchestrator_test.go
func testDryRunFunc(ctx context.Context, streams ioes.IOStreams, w *workflow.Workflow) error {
    // emit test dryrun event
}

func testRunWorkflowFunc(ctx context.Context, streams ioes.IOStreams, w *workflow.Workflow, runID string) error {
    // emit test save event
}

func TestIntegration(t *testing.T) {
    // 
    // create handler that listens for events on the bus & records them
    // construct orchestrator w/ testDryRunFunc, testRunWorkflowFunc & bus
    // o.CreateWorkflow()
    // o.GetWorkflow()
    // o.RunWorkflow()
    // o.DryRun()
    // wait
    // check events
}
```
### options
We pass Option funcs into the `NewOrchestrator` method to construct a specific orchestrator.
```
type Options struct {
    workflowStore *workflow.Store
    saveFuncFactory SaveFuncFactory
    applyFuncFactory ApplyFuncFactory
}

// Option is a function that manipulates config details when fed to NewOrchestrator. Fields on the opts parameter may be null
type Option func(opts *Options) error

func OptWorkflowStore(ws workflow.Store) Option {
        o.workflowStore = ws
        return nil
    }
}

func OptSaveFuncFactory(f SaveFuncFactory) Option {
        o.saveFuncFactory = f
        return nil
    }
}

func OptApplyFuncFactory(f ApplyFuncFactory) Option {
        o.applyFuncFactory = f
        return nil
    }
}
```

### SaveFuncFactory, ApplyFuncFactory interface
```
type SaveFunc func(ctx context.Context, streams ioes.IOStreams, w *workflow.Workflow, runID string) error
type SaveFuncFactory func(ctx context.Context) (saver SaveFunc)
type ApplyFunc func(ctx context.Context, streams ioes.IOStreams, w *workflow.Workflow) error
type ApplyFuncFactory func(ctx context.Context) (applier ApplyFunc)
```

### Orchestrator struct
The Orchestrator is responsible for managing the pieces needed for automation. It constructs a workflow store, run store, launches any trigger listener operations, handles trigger events, calls a SaveFunc or ApplyFunc at the correct moments, launches any hook completor operations, and emits the proper hook events when necessary. It is also the entry point for outside systems to interact with the workflow and run stores.

This first iteration should create a workflow store & handle trigger events.
```
type Orchestrator struct {
    workflows *workflow.Store
    save SaveFunc
    apply ApplyFunc
    bus event.Bus
}

func (o *Orchestrator) handleTrigger() // event handler func, listens for event trigger & fires off RunWorkflow, blocking
func (o *Orchestrator) RunWorkflow() error // runs Save
func (o *Orchestrator) DryRun() error // runs Apply, blocking
func (o *Orchestrator) CreateWorkflow(did dataset.ID, pid peer.ID) (*workflow.State, error)
func (o *Orchestartor) GetWorkflow(id workflow.ID) (*workflow.State, error)

// Constructs an orchestrator, whose only responsibility, right now, is to create a workflow store & listen for trigger events
func NewOrchestrator(ctx context.Context, bus event.Bus, opts ...Option) (*Orchestrator, error)
```

### barebones workflow store

Need to implement bare-bones workflow struct and workflow store
```
package workflow

type State struct {
    ID workflow.ID
    DatasetID dataset.ID
    OwnerID peer.ID
    Deployed bool
}

var (
    _ json.Marshaler = (*State)(nil)
    _ json.Unmarshaler = (*State)(nil)
)

func (wf *State) MarshalJSON() ([]byte, error)
func (wf *State) UnmarshalJSON(data []byte) error

type Store interface {
    Updater
}

type Updater interface {
    Create(did dataset.ID, pid peer.ID) (*State, error)
    Get(wid ID) (*State, error)
}

type MemStore struct { /* ... */ }

```

ramfox:
closes #1789

Can I merge this to master, or should I create a `feat_automation` branch that I do progressive prs on & merge everything once it's "finished"?


automation/orchestrator.go
```diff
@@ -0,0 +1,165 @@
...
+	for _, opt := range opts {
+		if err := opt(o); err != nil {
+			return nil, err
+		}
+	}
+
+	orch.bus = bus
+	orch.saveFactory = saveFactory
+	orch.applyFactory = applyFactory
+	orch.workflows = o.workflowStore
```
b5, line 151: process options before allocating the Orchestrator pointer & do assignments while allocating:

```go
orch := &Orchestrator{
  bus: bus,
  saveFactory: saveFactory,
  // ...
}
```

b5 (commented):
first batch of review

automation/orchestrator.go
```diff
@@ -0,0 +1,165 @@
...
+// OptWorkflowStore allows you to pass in a custom workstore to the Orchestrator
+func OptWorkflowStore(ws workflow.Store) Option {
+	return func(opts *Options) error {
+		opts.workflowStore = ws
+		return nil
+	}
+}
+
+// Save is what the Orchestrator will call to run a transform in during a workflow
+type Save func(ctx context.Context, streams ioes.IOStreams, w *workflow.State, runID string) error
```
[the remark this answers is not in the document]
ramfox: took this pattern from your spike branch (moving the responsibility of creating a runID to the Orchestrator):

https://github.com/qri-io/qri/blob/e3fda04c6fd3403ba237dbbace5b91282934d54e/lib/automation.go#L293-L310

automation/orchestrator.go
```diff
@@ -0,0 +1,165 @@
...
+var (
+	log = golog.Logger("automation")
+)
+
+// Options provides details to NewOrchestrator
+// NewOrchestrator will alter Options by applying any provied Option functions
+type Options struct {
+	workflowStore workflow.Store
+	saveFactory   SaveFactory
+	applyFactory  ApplyFactory
```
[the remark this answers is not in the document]
ramfox: thank you for catching this, this is hold over before I realized these didn't make sense as options since they were, as you stated, required

automation/orchestrator.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// Option is a function that manipulates config details when fed to NewOrchestrator. Fields on the opts parameter may be null
+type Option func(opts *Options) error
```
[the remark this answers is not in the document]
ramfox: This is helpful and fully agree.

The only cases I can see where a user can "foot gun" themselves can happen using the first approach any way (if somehow the workflow store and run store that are loaded don't reference the same data).

automation/orchestrator.go
ramfox, line 39: Thing I've been considering: should I change `Save` to `Run` and `Apply` to `DryRun`, since that is the language we use in the rest of this package?

b5: I'm a fan changing `Save` to `Run`. I think it solves the confusion both of us encountered with this function, expecting save to get a `*dataset.Dataset` somewhere.

I'd like to avoid renaming `Apply` to `DryRun`. @dustmop's [RFC](https://github.com/qri-io/rfcs/blob/master/text/0027-transform_application.md) formalized our switch to using `Apply`.

automation/workflow/workflow.go
```diff
@@ -0,0 +1,122 @@
...
+type Lister interface {
+	// List
+	// ListDeployed
+}
+
+// NewDefaultStore returns the default expected store
+// TODO (ramfox): replace this with a file store asap, currently a MemStore
+func NewDefaultStore(ctx context.Context) (*MemStore, error) {
+	return NewMemStore(), nil
+}
```
b5, line 75: I don't think we need this function. I'd be better to just force users to pick the store type, and define a configuration-driven constructor once we have config in place

automation/workflow/workflow.go
```diff
@@ -0,0 +1,122 @@
...
+	m.mu.Lock()
+	wf, ok := m.workflows[wid]
+	m.mu.Unlock()
+	if !ok {
+		return nil, ErrNotFound
+	}
+	return wf, nil
+}
+
+var _ Store = (*MemStore)(nil)
```
b5, line 122: move this to immodestly follow the `type MemStore struct` definition. Order I've come to prefer:

1. type definition
2. interface assertion(s)
3. constructors
4. Exported Methods
5. unexported methods

automation/workflow/workflow.go
```diff
@@ -0,0 +1,122 @@
...
+	workflows map[ID]*State
+}
+
+// NewMemStore return a MemStore
+func NewMemStore() *MemStore {
+	return &MemStore{
+		mu:        &sync.Mutex{},
+		workflows: map[ID]*State{},
+	}
+}
```
b5, line 89: We should have a spec test for workflow stores. I'll write one up

event/trigger.go
```diff
@@ -0,0 +1,8 @@
+package event
+
+const (
+	// ETTriggerWorkflow signals that a workflow has been triggered
+	// Payload will be the `workflow.ID` of the workflow that should be run
+	// This event should block until the workflow has completed its run
+	ETTriggerWorkflow = Type("trigger:Workflow")
```
b5, line 7: categorical nesting error here. this should be `workflow:trigger` so we can add more `workflow`-namespaced events

automation/orchestrator_test.go
```diff
@@ -0,0 +1,92 @@
...
+		select {
+		case msg := <-c:
+			t.Log(msg)
+		case <-time.After(200 * time.Millisecond):
+			t.Errorf(errMsg)
+		}
+		done <- struct{}{}
+	}()
+	return done
+}
```
b5, line 26: move this below the test func

automation/orchestrator.go
```diff
@@ -0,0 +1,157 @@
...
+			return fmt.Errorf("handleTrigger: expected event.Payload to be a workflow.ID: %v", e.Payload)
+		}
+		return o.RunWorkflow(ctx, wid)
+	}
+	return nil
+}
+
+// RunWorkflow runs the given workflow
+func (o *Orchestrator) RunWorkflow(ctx context.Context, wid workflow.ID) error {
+	log.Debugf("RunWorkflow, workflow ID: %s", wid)
```
b5, line 66:
```suggestion
	log.Debugw("RunWorkflow, workflow", "id", wid)
```

automation/orchestrator.go
```diff
@@ -0,0 +1,157 @@
...
+	save := o.saveFactory(ctx)
+	wf, err := o.GetWorkflow(wid)
+	if err != nil {
+		log.Debugf("RunWorkflow: error getting workflow from store: %s", err)
+		return err
```
b5, line 71: wrap errors to add context:

```suggestion
		log.Debugw("RunWorkflow: getting workflow from store", "err", err)
		return fmt.Errorf("getting stored workflow: %w", err)
```

automation/orchestrator.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// TODO (ramfox): when hooks/completors are added, this should wait for the err, iterate through the hooks
+	// for this workflow, and emit the events for hooks that this orchestrator understands
+	runID := run.NewID()
+	return save(ctx, streams, wf, runID)
+}
+
+// DryRunWorkflow runs the given workflow, but does not record the output
+func (o *Orchestrator) DryRunWorkflow(ctx context.Context, wid workflow.ID) error {
+	log.Debugf("DryRunWorkflow, workflow ID: %s", wid)
```
b5, line 84:
```suggestion
	log.Debugw("DryRunWorkflow", "workflowID", wid)
```

automation/orchestrator.go
```diff
@@ -0,0 +1,157 @@
...
+	apply := o.applyFactory(ctx)
+	wf, err := o.GetWorkflow(wid)
+	if err != nil {
+		log.Debugf("DryRunWorkflow: error getting workflow from store: %s", err)
+		return err
```
b5, line 89:
```suggestion
		log.Debugw("DryRunWorkflow: getting workflow from store", "err", err)
		return fmt.Errorf("getting workflow from store: %w", err)
```

automation/workflow/workflow.go
```diff
@@ -0,0 +1,122 @@
...
+	// Update
+	// Deploy
+	// Undeploy
+}
+
+// A Lister lists entries from a workflow store
+type Lister interface {
+	// List
+	// ListDeployed
+}
```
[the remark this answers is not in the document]
ramfox: There are! I don't want the `trigger.Watcher` to have access to the full array of workflow store things, I only want it to be able to get lists of workflows, which is why I added `Lister`

automation/workflow/workflow.go
ramfox: Ahh, yes I just wrote an `AssertStore` func in the follow up workflows pr, if that is what you mean!

automation/workflow/workflow.go
```diff
@@ -0,0 +1,115 @@
...
+//    myString := "SomeRandomStringThatIsLong-SoYouCanCallItAsMuchAsNeeded..."
+//    run.SetIDRand(strings.NewReader(myString))
+//    a := NewID()
+//    run.SetIDRand(strings.NewReader(myString))
+//    b := NewID()
+func SetIDRand(r io.Reader) {
+	uuid.SetRand(r)
+}
+
+// Workflow holds all the information necessary to run a workflow
```
b5, line 40: Describe what a workflow is:

```
Workflow attaches automation to a dataset
```

b5 (approved):
One more round of naming nits. This is looking _really_ good

automation/workflow/workflow.go
```diff
@@ -0,0 +1,115 @@
...
+	// Update
+	// Deploy
+	// Undeploy
+}
+
+// A Lister lists entries from a workflow store
+type Lister interface {
+	// List
+	// ListDeployed
+}
```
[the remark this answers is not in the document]
ramfox: Yes ty... I forgot that was an option...

PR #1797, opened 2021-05-26
feat(workflow): fleshes out `workflow.Store`, `Trigger` and `Hook` interfaces, expands `Workflow` struct

### LINKED ISSUE #1790: feat(automation): implement `workflow.State`, `workflow.Store`, `Trigger`, and `Hook`
### Test
```
// workflow_test.go
package workflow

func TestWorkflowAdvance(t *testing.T)
func TestWorkflowComplete(t *testing.T)

func TestStoreIntegration(t *testing.T) {
    // create store
    // declare a workflow
    // declare a second workflow
    // s.Create
    // s.Get
    // s.GetDatasetWorkflow
    // s.Create
    // s.List // should have 2
    // s.Update
    // s.Get
    // s.Deploy
    // s.ListDeployed // should be one
    // s.Undeploy
    // s.ListDeployed// should be zero
    // s.Remove
    // s.Get
}
```
### Trigger interface & TriggerConstructor interface
```
package workflow

type Store interface {
    Updater
    Lister
}

type Updater interface {
    Create(did dataset.ID, pid peer.ID, triggers []trigger.Trigger) (*State, error)
}

type Lister interface {
    // List lists all the workflows in the store
    List(ctx context.Context, offset, limit int) ([]*Workflow, error)
    // ListDeployed lists all the deployed workflows in the store
    ListDeployed(ctx context.Context, offset limit int) ([]*Workflow, error)
}

type State struct {
    ...
    triggers []trigger.Trigger
}

func (m *MemStore) Create(did dataset.ID, pid peer.ID, triggers []trigger.Trigger) (*State, error)
func (m *MemStore) List(ctx context.Context, offset, limit int) ([]*Workflow, error)
func (m *MemStore) ListDeployed(ctx context.Context, offset, limit int) ([]*Workflow, error)
```
```
// workflow/trigger.go
package workflow

// Type describes the type of trigger the system listens for
type TriggerType string

// Trigger has all the information needed to determine under what circumstances an `event.ETTriggerWorkflow`
// should be emitted. A Trigger should also know how to "advance" itself once a trigger was... triggered.
type Trigger interface {
    json.Marshaler
    json.Unmarshaler
    ID() string
    Enabled() boolean
    Type() TriggerType
    ToMap() map[string]interface{}
    Advance() error
}

```
### Trigger tests
```
// trigger_test.go
package workflow

type TestTrigger struct {
}

```

### Hook & HookConstructor interface
```
// workflow/hook.go
package workflow

// Type describes the type of hook the system listens for
type HookType string

// Hook has all the information needed to determine how to react to a transform completed event
type Hook interface {
    json.Marshaler
    json.Unmarshaler
    ID() string
    Enabled() boolean
    Type() HookType
    ToMap() map[string]interface{}
    Advance() error
}
```
### Hook tests
```
// hook_test.go
package workflow

type TestHook struct {
}
```

### expand Workflow methods
```
package workflow

// Workflow holds both the instructions on how to run an automation
type Workflow struct { /* ... */ }

// Advance sets the workflow up for a new run
func (w *Workflow) Advance() error

// Complete records any information needed to note at the end of the run
func (w *Workflow) Complete() error

// Store handles the persistence and listing of Workflows. Store implementations must be safe for concurrent use
type Store interface {
	Lister
	// Creates a new undeployed workflow
	Create(wf *Workflow) (*Workflow, error)
	// Updates a workflow 
	Update(wf *Workflow) error
	// Remove removes a workflow from the store
	Remove(id string) error
	// Get a workflow from a workflow id
	Get(id string) (*Workflow, error)
	// Deploy enables the workflow to be run by an orchestrator
	Deploy(id string) error
	// Undeploy disables the workflow from being run by an orchestrator
	Undeploy(id string) error
        // GetDatasetWorkflow fetches a workflow using the dataset id
        // GetDatasetWorkflow(did string) (*Workflow, error)
}

// Lister lists workflows
type Lister interface {
	// List lists all the workflows in the store
	List(ctx context.Context, offset, limit int) ([]*Workflow, error)
	// ListDeployed lists all the deployed workflows
	ListDeployed(ctx context.Context, status Status, offset, limit int) ([]*Workflow, error)
}

```

ramfox:
closes #1790


automation/workflow/store.go
```diff
@@ -0,0 +1,304 @@
...
+	wf, err := m.Get(id)
+	if err != nil {
+		return err
+	}
+	wf.Deployed = false
+	return m.Update(wf)
+}
+
+// List lists all the workflows in the store, by decending order from time of
+// creation
```
ramfox, line 153: Want to expand the description in the interface definition for `List` and `ListDeployed` to include expectations for limit and offset. These are the assumptions I'm going off of, but would like feedback:
- limit = -1 && offset = 0 means `fetchAll`
- negative limit or negative offset returns an error

Main question is around the offset being greater than the size of the list. Should that be an error or an empty list? I have an error coded up here, but looking at it an empty list is probably the way to go.

automation/workflow/trigger.go
```diff
@@ -0,0 +1,23 @@
...
+
+// TriggerType is the type of trigger
+type TriggerType string
+
+// A Trigger determines under what circumstances an `event.ETWorkflowTrigger`
+// should be emitted on the given event.Bus. It knows how to `Advance` itself.
+type Trigger interface {
+	// json.Marshaler
+	// json.Unmarshaler
+	ID() ID
```
ramfox, line 15: is it weird to have the Trigger and Workflow (and Hook) take the same kind of `ID`? Should I be defining `TriggerID` and `HookID` as types as well?

automation/workflow/workflow_test.go
```diff
@@ -0,0 +1,165 @@
...
+		Triggers:  []Trigger{aliceTestTrigger},
+		// Hooks:     []Hook{aliceTestHook},
+	}
+	// got, err := store.Create(aliceDatasetID, aliceProID, []Trigger{aliceTestTrigger}, []Hook{aliceTestHook})
+	SetIDRand(strings.NewReader(aliceSeedStr))
+	got, err := store.Create(aliceDatasetID, aliceProID, []Trigger{aliceTestTrigger})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if diff := cmp.Diff(alice, got, cmpopts.IgnoreFields(Workflow{}, "Created"), cmp.AllowUnexported(TestTrigger{})); diff != "" {
```
ramfox, line 43: I'm already mocking IDs, should I also be mocking time? Chose to just ignore the field for now, but it doesn't seem logical to be mocking some things and not others.

automation/workflow/hook.go
```diff
@@ -0,0 +1,24 @@
...
+// HookType is the type of hook
+type HookType string
+
+// A Hook determines under what circumstances its `hook.Event()` should be
+// emitted, and what the event payload should be.
+type Hook interface {
+	// json.Marshaler
+	// json.Unmarshaler
+	ID() ID
+	IsEnabled() bool
```
b5, line 16: rename this to `Enabled() bool`

automation/workflow/hook.go
```diff
@@ -0,0 +1,24 @@
...
+	Enable() error
+	Disable() error
```
b5, line 18: Combine these into a setter: `SetEnabled(enabled bool) error`

automation/workflow/hook.go
```diff
@@ -0,0 +1,24 @@
...
+import "github.com/qri-io/qri/event"
+
+// "encoding/json"
+
+// HookType is the type of hook
+type HookType string
+
+// A Hook determines under what circumstances its `hook.Event()` should be
+// emitted, and what the event payload should be.
+type Hook interface {
```
b5, line 12: we have a notion of `Hook` in the remote package that is a more general form of these: remote hooks can be fired at lots of different procedure lifecycle points. These can only fire when a workflow Succeeds. 

What do you think about calling this interface `CompletionHook` & using this definition:

```go
// CompletionHook is a typed configuration to execute when a workflow transform succeeds
type CompletionHook interface {
```

I'd welcome pushback on this, just trying to capture a succinct, accurate definition.

automation/workflow/trigger.go
b5: nah, I think it's fine for them to all just use `ID` types

automation/workflow/hook.go
```diff
@@ -0,0 +1,24 @@
...
+	// json.Marshaler
+	// json.Unmarshaler
+	ID() ID
```
b5, line 15: do these need an ID at all? If we're only planning on persisting these in a slice within `workflow.Workflow` then we shouldn't need to worry too much about separate storage.

On the other hand, if the plan is to use these in events to connect an event to which workflow fired, that might be a case for having an ID, but even there, I'm not entirely sure.

automation/workflow/hook.go
```diff
@@ -0,0 +1,24 @@
...
+	IsEnabled() bool
+	Enable() error
+	Disable() error
+	Type() HookType
+	ToMap() map[string]interface{}
+	Advance() error
```
b5, line 21: I'd love a comment here:

```
Advance is called on each execution, providing hooks a chance to update any internal state
```

automation/workflow/hook.go
```diff
@@ -0,0 +1,24 @@
...
+	Event() event.Type
+	Payload() interface{}
```
b5, line 23: Is there any reason not to combine these?

```go
Event() (t event.Type, payload interface{})
```

automation/workflow/hook.go
```diff
@@ -0,0 +1,24 @@
...
+// emitted, and what the event payload should be.
+type Hook interface {
+	// json.Marshaler
+	// json.Unmarshaler
+	ID() ID
+	IsEnabled() bool
+	Enable() error
+	Disable() error
+	Type() HookType
+	ToMap() map[string]interface{}
```
b5, line 20: I'd love a comment here:
```go
// ToMap serializes all hook state
ToMap() map[string]interface{}
```

I'm confused about the need for this method, given the presence of the `Payload`  method and the planned requirement for json serialization. If I can get a hook to go to-and-from JSON, and I'm not using `ToMap` as the payload of an event, what do I need `ToMap` for?

automation/workflow/store.go
b5: totally agree on `limit: -1, offset: 0` as the fetchAll sentinel value. I'm working on a `params.ListAll` that does exactly the same thing: https://github.com/qri-io/qri/blob/3133469fe112b8590ee8ddb5a9ba19f2adbe9c11/params/list.go#L4-L6

very much agree that offsets greater than the length of the list should return an empty slice & no error

b5 (commented):
ok just feedback on the design of `Hook` (some of this will translate onto triggers)

automation/workflow/hook.go
ramfox: Yea, I was thinking more that they might be needed to disambiguate different hooks/triggers in the same workflow, but that was the only reason. 

Specifically if we allow cases where a Trigger or Hook can be "updated" without the user handing us the full context of the workflow.

I'm thinking about the difference between the frontend vs the cli. In the frontend, we already have the entire workflow. If a user wants to update a specific hook, we have all the workflow information to send over the api to the backend for us to construct a `*Workflow` and use it in a `store.Update` call.

For the CLI, a similar flow would be maybe first running `qri workflow get` to get a json representation of the `Workflow`. The user could save that to a file, make edits, and then use that file to update the store: `qri workflow update --file`.

In either of these cases, we don't need IDs for the triggers or hooks, because we have the full context of the workflow and the other triggers and hooks.

However, if we give users the ability to 'piecemeal' update a single trigger or hook in the workflow, we may need to keep IDs, especially if we can have multiple triggers or hooks for the same type.
`qri workflow update --id workflow_id --trigger-id trigger_id --periodicity R3/P1W`
Or something like that.

I personally favor the `--file` options because it seems easier to me, to both code and work with, but I haven't done a deep dive on `qri workflow create` and `qri workflow update` possible options.

automation/workflow/hook.go
ramfox: Not sure yet, lol. This was holdover from our previous iteration, and I held onto it while planning in case I was missing something. But I agree, with the plans for serialization what's the point?

automation/workflow/hook.go
ramfox: None, thank you, this is great!

automation/workflow/hook.go
ramfox: This is great ty

automation/workflow/hook.go
ramfox: Fab

automation/workflow/hook.go
ramfox: Not necessarily push back, but I don't think we ever fully committed to the idea that hooks only fire when the transform succeeds. Could be that a user wants an alert when it runs, regardless of success.

`Hook` is such a great word, I kept coming back to it even thought I tried to get rid of it when planning. I left off with having the system that listens for these `Hooks` be called `Completors` and the actual emitted "events" be called hooks. I think we eventually want what we call `Hooks` be more what you traditionally think of as a hook and be a place where users can have custom responses to a transform finishing.

Other words I thought could maybe work were `Connectors`, `Linkers`... `DoneTriggers` and then call what we call `Triggers` right now `StartTriggers`... or `KickOffTriggers`

`Kickoffs` and `Connectors`? ... spitballing here

Maybe they should just be called `DoneEvents`? They essentially just describe under what circumstances certain events that signal the end of the workflow be called.

automation/workflow/store.go
```diff
@@ -0,0 +1,305 @@
...
+func (m *MemStore) ListDeployed(ctx context.Context, limit, offset int) ([]*Workflow, error) {
+	wfs := NewWorkflowSet()
+	fetchAll := false
+	switch {
+	case limit == -1 && offset == 0:
+		fetchAll = true
+	case limit < 0:
+		return nil, fmt.Errorf("limit of %d is out of bounds", limit)
+	case offset < 0:
+		return nil, fmt.Errorf("offset of %d is out of bounds", offset)
```
ramfox, line 198: Since I there is work happening on listing elsewhere in qri, do we have any standard on what should be returned if the limit and/or offset are negative?

automation/workflow/hook.go
b5: > However, if we give users the ability to 'piecemeal' update a single trigger or hook in the workflow, we may need to keep IDs, especially if we can have multiple triggers or hooks for the same type.

I can't see us going this route. once we allow multiple triggers of the same type, I think users will want to have the store save the order they're in, which requires passing the entire workflow around. Based on this, I vote we drop the ID field on triggers and hooks. If we need disambiguation we should use `workflowID` concatenated with slice index on the workflow struct

automation/workflow/hook.go
b5: hook it is!

automation/workflow/store.go
```diff
@@ -0,0 +1,307 @@
...
+
+// Store manages & stores workflows, allowing listing and updating of workflows
+type Store interface {
+	Lister
+	// Create creates a new Workflow and adds it to the Store. If a workflow
+	// with that dataset id already exists, it emits an
+	// `ErrWorkflowForDatasetExists` error
+	Create(did string, pid profile.ID, triggers []Trigger, hooks []Hook) (*Workflow, error)
+	Get(wid ID) (*Workflow, error)
+	GetDatasetWorkflow(did string) (*Workflow, error)
```
b5, line 23: I'd rename this to `GetByDatasetID`

automation/workflow/store.go
```diff
@@ -0,0 +1,307 @@
...
+	"github.com/qri-io/qri/profile"
+)
+
+// Store manages & stores workflows, allowing listing and updating of workflows
+type Store interface {
+	Lister
+	// Create creates a new Workflow and adds it to the Store. If a workflow
+	// with that dataset id already exists, it emits an
+	// `ErrWorkflowForDatasetExists` error
+	Create(did string, pid profile.ID, triggers []Trigger, hooks []Hook) (*Workflow, error)
```
b5, line 21: When describing interface methods, I think it's a good idea to avoid implementation-specific details like what error is returned, but focus on behavior instead.

```
// Create add a Workflow to the store. Datasets can have at most one workflow
```

Regarding the method itselfI think this should be refactored to:

```go
func Put(*Workflow) (error)
```

Breaking the fields out & constructing the structure internally requires breaking this interface whenever the Workflow struct changes. Using the workflow struct as a parameter means the store just needs to validate the structure before saving.

automation/workflow/store.go
```diff
@@ -0,0 +1,307 @@
...
+	Get(wid ID) (*Workflow, error)
+	GetDatasetWorkflow(did string) (*Workflow, error)
+	Remove(id ID) error
+	Update(wf *Workflow) error
```
b5, line 25: If you're ok with the `Put` change, we can merge update into it & branch off weather the provided workflow has an ID or not (workflow with an ID this store recognizes == update)

automation/workflow/store.go
```diff
@@ -0,0 +1,307 @@
...
+	Deploy(id ID) error
+	Undeploy(id ID) error
```
b5, line 27: Does the store do anything special when we deploy? In my head all the specialness of deploying happens at the orchestrator?

If so we can get away with just `Put`ting a workflow with `wf.Deployed` set to the desired value. This would help keep the store dumb

automation/workflow/hook.go
ramfox: Sounds good

automation/workflow/store.go
ramfox: Great, totally on board w/ following the usual data store pattern.

automation/workflow/store.go
ramfox: 👍

ramfox:
Having a bit of weird discovery that I wasn't thinking about before.

We know that the trigger listener will emit a `ETWorkflowTrigger` event with the `workflow.ID` as the payload.

When we `Put` the workflow into the store the first time, we don't have a workflow ID. It gets generated in `Put` and we can get the ID from the returned workflow. 

But if `Triggers` need to know the `workflow.ID` then this becomes very odd, because we can potentially be `Put`-ting a workflow whose triggers have empty `WorkflowID` fields or mismatched `WorkflowID` fields, because we can't create the triggers before knowing what the ID is.

Some ideas I've had to combat this odd behavior:
1) we assumed the trigger listener would get a list of `Triggers`, rather than a list of `Workflows`, which is why the `Trigger` interface needs a `WorkflowID` method. If the trigger listener subsystem gets a list of `Workflows` instead, it would have everything it needs to emit the correct event/payload.
2) OR creating a workflow ends up becoming a two step process in the `Orchestrator`. We first `Put` a workflow to get back an ID, use that ID to create the Triggers, and then `Put` the workflow again, this time with the triggers attached. This coupled with a workflow validator that ensures the workflow id is correct in all places, would ensure we are creating valid workflow/trigger combinations.

Thoughts? It does look a bit odd to have `Triggers` that exist in the `Workflow` that also track the workflow ID, when `Triggers` are stored right along the `Workflow` itself, so I'm... sort of leaning toward the first option?

b5:
I'll try to spit this back to you as pseudocode to make sure we're on the same page. I'm operating under the assumption that Triggers & Hooks are persisted in a `workflow.Store` by saving them in fields on the `Workflow`  struct so the following should be true:

```go
wf := &workflow.Workflow{
  DatasetID: "foo",
  Triggers: []Trigger{
    NewCronTrigger(/* ... */),
  },
}

if err  := store.Put(wf); err != nil {
  panic(err)
}

// workflow now has an ID,
// problem: workflow.Triggers[0] doesn't have that ID stored
```

I think this block of code should be happening as early as possible when constructing new workflows. We want to fail right away if the store won't accept / can't save the workflow. I think your second option is getting at this need to `Put` early.

A third option would be to add a `SetID` method to the `Trigger` and `Hook` interfaces, and have the store call that when an ID is assigned to the `Workflow`.

That said, I think your first option gets at the root of the problem: Triggers and Hooks are tight-coupled with a workflow, and that's been by design. In review I've been advocating for removing IDs from Triggers and Hooks, another way of framing that same request is "please make Triggers and Hooks rely on the Workflow struct more".

But this tight coupling means we're passing a lot of stuff to separate subsystems. I think an extension off your first approach would be to add methods to `Workflow` and define new interfaces:

```go

type TriggerSource interface {
  WorkflowID() workflow.ID
  ActiveTriggers() []Trigger
}

type HookSource interface {
  WorkflowID() workflow.ID
  ActiveHooks() []Hook
}

func (w Workflow) WorkflowID() ID {
  return w.ID
}

func (w Workflow) ActiveTriggers() []Trigger {
  if !w.Deployed {
    return nil
  }
  active := make([]Trigger)
  for _, t := range w.Triggers {
    if !t.Disabled() {
      active = append(active, t)
    }
  }

  return active
}
```

Then have each subsystem accept `[]HookSource` and `[]TriggerSource`

ramfox:
Yea love this. Feels like a good follow through of what I was trying to do with the `Lister` interface, and only give the other subsystems just what they need and no more. Makes me re-think it as well. Maybe removing the `Lister` interface and instead having `trigger` define an interface for the kind of store it accepts

```
type TriggerSourceLister interface {
    List(ctx context.Context, limit, offset int) []TriggerSource 
}
```

regardless, the approach you outlined above sounds like a great solution for the current issue.

b5 (approved):
Looking _very_ good. I don't have time to do an extensive review ATM, but given we've already been through a number of design reviews, I vote we merge & uncover what needs improvement on the next steps of building out the automation feature. Let's merge!

PR #1798, opened 2021-05-26
fix(remote): update handlers with gorilla mux

Arqu:
Updating remaining routes to `gorilla.mux` and bringing it in line with the rest.


b5 (approved):
LGTM

PR #1803, opened 2021-06-02
feat(run): add `run.Store` interface, spec tests, and implement `run.MemStore`

### LINKED ISSUE #1791: feat(automation): RunStore
## 3. RunStore
A run.State is a record of events from an executing or executed workflow, as well as containing metadata about a run such as the Duration, Status, Start and Stop times of a run.

The run.Store is responsible for storing and updating the run.State. It is also responsible for tracking the number of times workflow has been run.

The run.Store knows when to update the run.State by subscribing to transform events & updating accordingly.

Much of the run.go file stays the same, including the `run.State` and `run.Status`, except that we should add `WorkflowID` as a field on `run.State`.

Do we need any special tear down for the subscriber?

### testing
```go
func TestMemStore(t *testing.T) {
    bus := bus.NewBus()
    store := NewMemStore(bus)
    AssertStoreSpec(t, bus)
}

func AssertStoreSpec(t *testing.T, store *run.Store) {
    testIntegration(t *testing.T, store)
    testSubscribeID(t *testing.T, store)
}

func testIntegration(t *testing.T, store *run.Store)) {
    // add new run
    // add new run w/ same workflow id
    // add new run w/ same workflow id
    // check s.Count == 3
    // get s.List, should return 3
    // s.Get()
    // s.Update one of the runs, change the status
    // s.Get
    // s.ListByStatus(new status)	
}
func testSubscribeID(t *testing.T, store *run.Store) {
    // get bus using store.Bus()
    // emit ETTransformStart
    // use s.Get() to check update
    // emit ETTransformStepStart
    // emit ETTransformPrint
    // emit ETTransformError
    // emit ETTransformDatasetPreview
    // emit ETTransformStepStop
    // emit ETTransformStepSkip
    // emit ETTransformStop
    // create a new bus, use the same events & initial run state
    // ensure the end runState is the same in both cases
}

```
### interface

run.Store methods:
```go
// Put adds a run state to the store. If the given RunID is empty, it creates a new run.State. If RunID is not empty, it updates the associated run. 
func (s *Store) Put (r *State) error
// Count returns the number of runs of given id
func (s *Store) Count(id workflow.ID) (int, error)
// List lists all the runs associated with the id
func (s *Store) List(id workflow.ID) ([]*State, error)
// GetLatest returns the most recent run associated with the workflow id
func (s *Store) GetLatest(id workflow.ID) (*State, error)
// Get gets the assocaited run.State
func (s *Store) Get(id run.ID) (*State, error)
// SubscribeID subscribes to events emitted w/ the given run id. Uses `AddTransformEvent` to interpret/change the runState, and calls Update on the store
func (s *Store) SubscribeID(id run.ID) error 
// GetStatus returns the status of the latest run based on the workflow.ID
func (s *Store) GetStatus(id workflow.ID) (Status, error)
// ListByStatus returns a list of the most recent run.State entries with a given status.
func (s *Store) ListByStatus(s Status) []*State
// Bus returns the bus that the store subscribes to
func (s *Store) Bus() event.Bus

```
run.State methods:
```go
// alters state based on a given event
// TODO ramfox: should this be pure?
func (rs *State) AddTransformEvent(e event.Event) error
```

constructor:
```go
type MemStore struct {
    bus event.Subscriber
    counts map[workflow.ID] int
    runs map[workflow.ID][]*State
}

func NewMemStore(ctx context.Context, bus event.Subscriber) (*Store, error)

```

### adjust save path accordingly
`lib.Save` & `automation.Orchestrator` both track the `runState` & update their own internal `runState` separately. However, they are reacting to the same events using the same updater function, so their outputs are identical. This is important because the `lib.Save` `runState` goes on to inform the entry written in the `log`, and the `automation.Orchestrator` `runSave` gets saved in the `run.Store`. It's important that they are the same. 

We will have a test that ensures this behavior.

Changes needed in `lib.Save`:
- save needs to get the runID from the dataset. If it doesn't exist, then it can create its own.
- the runID needs to be propagated to the `Commit.RunID` field in `lib.Save`

### adjust Orchestrator
```go
// automation/orchestrator.go
package automation

type Options struct {
    ...
    runstore *run.Store
}

func OptRunStore(rs *run.Store) Option {
        o.runstore = rs
        return nil
    }
}

type Orchestrator struct {
    ...
    runs *run.Store
}

// Constructs an orchestrator, whose only responsibility, right now, is to create a workflow store, listen for trigger events, and create & start a run store
func NewOrchestrator(ctx context.Context, bus event.Bus, opts ...Option) (*Orchestrator, error)

```

ramfox:
closes #1791

### DIFF
--- a/automation/orchestrator.go
+++ b/automation/orchestrator.go
@@ -100,7 +100,7 @@ func NewOrchestrator(ctx context.Context, bus event.Bus, runFactory RunFactory, 
 	}
 
 	if o.runs == nil {
-		// TODO(ramfox): once we ahve a `config.Automation` specified, we will have a
+		// TODO(ramfox): once we have a `config.Automation` specified, we will have a
 		// specific `run.NewStore` function that takes a `config.RunStore` & will
 		// return a specified `run.Store`
 		return nil, fmt.Errorf("no run store specified")
--- a/automation/orchestrator.go
+++ b/automation/orchestrator.go
@@ -154,7 +154,7 @@ func (o *Orchestrator) RunWorkflow(ctx context.Context, wid workflow.ID) error {
 	runID := run.NewID()
 	if o.runs != nil {
 		r := &run.State{ID: runID, WorkflowID: wid}
-		if _, err := o.runs.Put(r); err != nil {
+		if _, err := o.runs.Create(r); err != nil {
 			return err
 		}
 
--- a/automation/orchestrator.go
+++ b/automation/orchestrator.go
@@ -3,6 +3,7 @@ package automation
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	golog "github.com/ipfs/go-log"
@@ -27,6 +28,7 @@ var NowFunc = func() *time.Time {
 // OrchestratorOptions encapsulate runtime configuration for NewOrchestrator
 type OrchestratorOptions struct {
 	WorkflowStore workflow.Store
+	RunStore      run.Store
 }
 
 // Run persists the dataset that results from executing a workflow transform
@@ -43,15 +45,22 @@ type ApplyFactory func(ctx context.Context) Apply
 
 // Orchestrator manages automation in qri
 type Orchestrator struct {
+	// TODO(ramfox): this runLock is the current shim to ensure only one workflow runs at a time
+	// we should probably have a run queue subsystem that ensure the orchestrator is running
+	// the workflows in the expected order, running only as many at once as configured, and
+	// allows communication back to the user about where they are in the run queue, allows for
+	// cancelling runs that haven't happened yet
+	runLock      sync.Mutex
 	workflows    workflow.Store
+	runs         run.Store
 	runFactory   RunFactory
 	applyFactory ApplyFactory
 	bus          event.Bus
 	cancel       context.CancelFunc
 }
 
-// NewOrchestrator constructs an orchestrator, whose only responsibility,
-// right now, is to create a workflow store & listen for trigger events
+// NewOrchestrator constructs an orchestrator, whose only responsibility, right
+// now, is to create a workflow store, run store, & listen for trigger events
 func NewOrchestrator(ctx context.Context, bus event.Bus, runFactory RunFactory, applyFactory ApplyFactory, opts OrchestratorOptions) (*Orchestrator, error) {
 	log.Debugw("NewOrchestrator", "opts", opts)
 
@@ -80,6 +89,7 @@ func NewOrchestrator(ctx context.Context, bus event.Bus, runFactory RunFactory, 
 		runFactory:   runFactory,
 		applyFactory: applyFactory,
 		workflows:    opts.WorkflowStore,
+		runs:         opts.RunStore,
 	}
 
 	if o.workflows == nil {
@@ -87,6 +97,13 @@ func NewOrchestrator(ctx context.Context, bus event.Bus, runFactory RunFactory, 
 		// specific `workflow.NewStore` function that takes a `config.Workflow` & will
 		// return a specified `workflow.Store`
 		return nil, fmt.Errorf("no workflow store specified")
+	}
+
+	if o.runs == nil {
+		// TODO(ramfox): once we ahve a `config.Automation` specified, we will have a
+		// specific `run.NewStore` function that takes a `config.RunStore` & will
+		// return a specified `run.Store`
+		return nil, fmt.Errorf("no run store specified")
 	}
 
 	o.bus.SubscribeTypes(o.handleTrigger, event.ETWorkflowTrigger)
@@ -109,13 +126,19 @@ func (o *Orchestrator) handleTrigger(ctx context.Context, e event.Event) error {
 		if !ok {
 			return fmt.Errorf("handleTrigger: expected event.Payload to be a workflow.ID: %v", e.Payload)
 		}
-		return o.RunWorkflow(ctx, wid)
+		go func() {
+			if err := o.RunWorkflow(ctx, wid); err != nil {
+				log.Errorf("%s", err)
+			}
+		}()
 	}
 	return nil
 }
 
 // RunWorkflow runs the given workflow
 func (o *Orchestrator) RunWorkflow(ctx context.Context, wid workflow.ID) error {
+	o.runLock.Lock()
+	defer o.runLock.Unlock()
 	log.Debugw("RunWorkflow, workflow", "id", wid)
 	runFunc := o.runFactory(ctx)
 	wf, err := o.GetWorkflow(wid)
@@ -129,11 +152,19 @@ func (o *Orchestrator) RunWorkflow(ctx context.Context, wid workflow.ID) error {
 	// TODO (ramfox): when hooks/completors are added, this should wait for the err, iterate through the hooks
 	// for this workflow, and emit the events for hooks that this orchestrator understands
 	runID := run.NewID()
+	if err := o.runs.SubscribeID(runID); err != nil {
+		log.Debugw("RunWorkflow: subscribing the run store to events", "err", err)
+		return fmt.Errorf("subscribing run store to events for run id %q: %w", runID, err)
+	}
+	defer o.runs.UnsubscribeID(runID)
+
 	return runFunc(ctx, streams, wf, runID)
 }
 
 // ApplyWorkflow runs the given workflow, but does not record the output
 func (o *Orchestrator) ApplyWorkflow(ctx context.Context, wid workflow.ID) error {
+	o.runLock.Lock()
+	defer o.runLock.Unlock()
 	log.Debugw("ApplyWorkflow, workflow", "id", wid)
 	apply := o.applyFactory(ctx)
 	wf, err := o.GetWorkflow(wid)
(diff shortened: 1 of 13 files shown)


automation/orchestrator.go
```diff
@@ -43,15 +45,22 @@ type ApplyFactory func(ctx context.Context) Apply
 
 // Orchestrator manages automation in qri
 type Orchestrator struct {
+	// TODO(ramfox): this runLock is the current shim to ensure only one workflow runs at a time
+	// we should probably have a run queue subsystem that ensure the orchestrator is running
+	// the workflows in the expected order, running only as many at once as configured, and
+	// allows communication back to the user about where they are in the run queue, allows for
+	// cancelling runs that haven't happened yet
+	runLock      sync.Mutex
```
ramfox, line 53: So there was a bug in the code caught by the orchestrator integration test. Previously, we discussed that as a simple way to keep workflows from not running simultaneously, we should have the handlers for the `event.ETTriggerEvent` be blocking.

However, this meant that we couldn't use the bus for any functionality during the run... which we need to do for the run store, since we need to collect transform events off the bus in order to keep the run state up to date.

Instead, using a mutex lock for now to ensure runs are sequential, and changed the `handleTrigger` event handler to be non-blocking.

automation/orchestrator.go
```diff
@@ -109,13 +126,19 @@ func (o *Orchestrator) handleTrigger(ctx context.Context, e event.Event) error {
 		if !ok {
 			return fmt.Errorf("handleTrigger: expected event.Payload to be a workflow.ID: %v", e.Payload)
 		}
-		return o.RunWorkflow(ctx, wid)
+		go func() {
+			if err := o.RunWorkflow(ctx, wid); err != nil {
+				log.Errorf("%s", err)
```
ramfox, line 131: not sure how to surface this error, is logging it good enough? probably needs more context then, right?

automation/run/run.go
```diff
@@ -2,19 +2,40 @@
...
+// type ID string
+
+// // NewID creates a run identifier
+// func NewID() ID {
+// 	return ID(uuid.New().String())
+// }
+
+// // String returns the ID as the underlying string
+// func (id ID) String() string { return string(id) }
+
```
ramfox, line 38: I'm weirdly attached to the idea of having a separate run id type. Maybe because I assume that it'll be easy to switch the run id with some other identifying string?

lib/datasets.go
```diff
@@ -982,10 +982,13 @@ func (datasetImpl) Save(scope scope, p *SaveParams) (*dataset.Dataset, error) {
...
-		// runState
-		runID := run.NewID()
-		runState = run.NewState(runID)
+		runID := ds.Commit.RunID
+		if runID == "" {
+			// if there is no given runID, allocate an ID for the transform,
+			// subscribe to print output & build up the run.State
+			runID = run.NewID()
+		}
+		runState = &run.State{ID: runID}
```
ramfox, line 991: Okay, big question here. In the spike automation pr, we pass in a runID in the `dataset.Commit.RunID` field. Wondering if instead we should expand the `SaveParams` to have a `RunID` field, and pass it through that way.

automation/orchestrator.go
dustmop: Very interesting. I agree that the best solution would be a queue with a lock, but this is a fine stop-gap for now.

automation/orchestrator.go
```diff
@@ -89,6 +99,13 @@ func NewOrchestrator(ctx context.Context, bus event.Bus, runFactory RunFactory,
 		return nil, fmt.Errorf("no workflow store specified")
 	}
 
+	if o.runs == nil {
+		// TODO(ramfox): once we ahve a `config.Automation` specified, we will have a
```
dustmop, line 103: ahve -> have

automation/orchestrator.go
dustmop: Oof, yeah this is tough. I'm not really sure how all the parts fit together, but ideally we'd have something that gets error events off the bus and does something reasonable with them. This is ok for now, maybe there should also be a TODO about handling asynchronous errors better.

automation/run/run_test.go
```diff
@@ -116,7 +116,7 @@ func TestStateAddTransformEvent(t *testing.T) {
 
 	for i, s := range states {
 		t.Run(fmt.Sprintf("after_event_%d", i), func(t *testing.T) {
-			got := NewState(runID)
+			got := &State{ID: runID}
```
dustmop, line 119: Could we keep `NewState` instead of constructing `State` directly? It makes it easier to update all call-sites if another field becomes a required value for construction.

automation/run/store.go
```diff
@@ -0,0 +1,306 @@
...
+
+// Store stores and updates run.States. It is also responsible for tracking the
+// number of times a workflow has run, as well as listening for transform
+// events and updating the associated the run.State accordingly
+type Store interface {
+	// Put adds a run state to the Store. If the given RunID is empty, it
+	// creates a new run.State. If RunID is not empty, it updates the
+	// associated run.
+	Put(r *State) (*State, error)
+	// Get gets the assocaited run.State
```
dustmop, line 35: assocaited -> associated

automation/run/run.go
```diff
@@ -54,23 +75,42 @@ const (
...
+		return ErrNoID
+	}
+	if rs.WorkflowID.String() == "" {
+		return ErrNoWorkflowID
+	}
+	return nil
+}
+
+// Copy shallowly copies the contents of run parameter into the receiver
+func (rs *State) Copy(run *State) {
```
dustmop, line 101: The function names feels off, usually `Copy` makes a copy of the current thing and returns it, in order to be used like this:

```
a := b.Copy()
```

Could we rename this to `CopyFrom`?

automation/run/store.go
```diff
@@ -0,0 +1,306 @@
...
+	for i := len(runIDs) - 1; i >= 0; i-- {
+		id := runIDs[i]
+		run, ok := s.runs[id]
+		if !ok {
+			return nil, fmt.Errorf("run %q missing from the store", id)
+		}
+		runs = append(runs, run)
+	}
+
+	if offset > len(runs) {
```
dustmop, line 185: Use `offset >= len(runs)`, since indexing an array at an index equal to its length is invalid.

automation/run/store.go
```diff
@@ -0,0 +1,306 @@
...
+		run, ok := s.runs[rid]
+		if !ok {
+			return nil, fmt.Errorf("run %q missing from the store", rid)
+		}
+		if run.Status == status {
+			set.Add(run)
+		}
+	}
+
+	if offset > set.Len() {
```
dustmop, line 255: Same here, use `>=` instead of `>`.

automation/workflow/workflow.go
```diff
@@ -82,6 +82,20 @@ func (w *Workflow) Validate() error {
 	return nil
 }
 
+// Copy shallowly copies the contents of the workflow parameter into the receiver
+func (w *Workflow) Copy(workflow *Workflow) {
```
dustmop, line 86: Could this also change to `CopyFrom`?

automation/orchestrator.go
ramfox: Talked to @b5 about this. The conclusion was to actually be logging errors in `log.Debug` unless they are a critical failure. In the case of a run or dry run, the user is going to get output about the state of the run when the transform events are caught by the `Save` command.

automation/run/run.go
ramfox: I'd rather just change it to be the more correct version of `func (rs *State) Copy() State`!

ty

automation/run/run_test.go
ramfox: That's a great reason! Yes!

automation/run/store.go
ramfox: thank youuuuuuuuu

lib/datasets.go
ramfox: Dope, want to get @b5 & @Arqu 's eyes on this before I make the change cause these lib params feel settled since our api refactoring and this would technically be a change to the api as well.

ramfox: so at @b5 request, we are leaving as is!

automation/run/store.go
```diff
@@ -0,0 +1,274 @@
...
+		s.workflows[run.WorkflowID] = wf
+	}
+	_, ok := s.runs[run.ID]
+	if !ok {
+		s.workflows[run.WorkflowID].count++
+		runIDs := s.workflows[run.WorkflowID].runIDs
+		s.workflows[run.WorkflowID].runIDs = append(runIDs, run.ID)
+	}
+	s.runs[run.ID] = run
+	return run, nil
```
ramfox, line 129: @b5 Okay, so started implementing pulling out adjusting the run count and workflow in a separate method and adjusting the store Interface... it felt a bit like overkill & kind of awkward.

Basically, felt like the orchestrator was having opinions about the way that `run.Store` should be implemented.

EG, in a future SQL implementation, having a second call to something like `UpdateCount` or `UpdateOtherThings` would make no sense. If we really want to control that something different happens the first time you initiate a run, I'd rather separate the concerns by having `runStore.Create` method, and use `Put` for updates, rather than a second call after `Put` for special behavior. 

I think the only reason we had this bug was because I made a mistake while coding, not because there is something inherently wrong in how we've designed it.

Thoughts?

b5: yeah, looking at this implementation, and how we're working with the store up in the orchestrator, I think a `Create` method makes sense. I feel like we're relying too heavily on an empty `ID` field to determine user intent, where a `Create` method would make it explicit that you expect to, well, create a thing.

So, I'd vote for the former: let's add `run.Store.Create`. Let's reposition existing spec tests that work the empty-id-Put call to test the create method. I think it'll clean a bunch of stuff up

b5 (approved):
LGtM

PR #1807, opened 2021-06-14
fix(api): don't error when running connect with disabled p2p

b5:
I'm in a place with really spotty internet, and need to work on qri frontend without blowing up the router. This change makes it possible to run qri connect with p2p features disabled. I've renamed lib.Instance.Connect -> lib.Instance.ConnectP2P to clarify that it's really just about p2p connections, and defined a canonical error for trying to connect with disabled p2p via configuration.

Output when running connect without p2p doesn't make allusions to the d.web, and skips listing libp2p addresses (because there are none).

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -3,6 +3,7 @@ package api
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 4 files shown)


api/api.go
```diff
@@ -60,19 +61,30 @@ func (s Server) Serve(ctx context.Context) (err error) {
...
+	}
 	info += cfg.SummaryString()
-	info += "IPFS Addresses:"
-	for _, a := range node.EncapsulatedAddresses() {
-		info = fmt.Sprintf("%s\n  %s", info, a.String())
+	if p2pConnected {
+		info += "IPFS Addresses:"
+		for _, a := range node.EncapsulatedAddresses() {
+			info = fmt.Sprintf("%s\n  %s", info, a.String())
+		}
```
b5, line 87: I fully admit this is terrible, but I don't want to disturb the output that I'm sure will break tests in cmd. A follow up PR that cleans up this output & makes tests less sensitive would be great, but not a massive priority IMHO

dustmop (approved):
LGTM

PR #1808, opened 2021-06-15
feat(base): support progress in save, shorter stats keys, support NDJSON

b5:
upgrade dataset & qfs dependencies, bringing in support for NDJSON & compression settings in the process. Changes to stats frequency keys have the unfortunate side effect of breaking hashes & blowing up the size of this commit.

On the qfs side we gain a SizeFile interface, and use it in the save path to show import progress


base/dsfs/compute_fields.go
```diff
@@ -20,9 +20,10 @@ import (
 type computeFieldsFile struct {
 	*sync.Mutex
 
-	fs qfs.Filesystem
-	pk crypto.PrivKey
-	sw SaveSwitches
+	fs  qfs.Filesystem  // filesystem we're writing to
+	pub event.Publisher // bus to publish progress events to
```
Arqu, line 24: Can we rename this to `bus` or `publisher`, seeing it next to `pk crypto.PrivKey` threw me off a bit.

Arqu (approved):
Left a small nitpick.

LGTM!

Issue #1806, opened 2021-06-08
Unify Local & Cloud authentication with a common interface


b5:
We need to harmonize the story of authentication across qri.cloud & local qri instances. The most practical thing to focus on is getting responses for two authentication routes (`/token` & `/refresh`) to be backed by a common interface with separate local & cloud implementations.

on the cloud side this interface should provide a full-fledged oauth implementation, and the local side is basically "just pick from the key pairs you have".

There needs to be come alignment between cloud & core about _how_ they use oauth. For this to work any frontend/external code needs to be able to disambiguate what type of authentication mechanism a qri node expects.

While we're here, it'd be great to also eliminate the two different flows for authentication (one for RPC, and another for API). 

The authentication interface should be declared in the auth package, and should be providable via a `lib.NewInstance` option. Ideally all auth is delegated to this interface.

Arqu:
Drawing some basic lines around the common interface
```
package auth

import (
	...
)

type AuthProvider interface {
	// Token handles the auth token flow
	Token(ctx context.Context, req *TokenRequest) (*TokenResponse, error)
	// ParseTokenRequest extracts the TokenRequest from the incoming http request
	ParseTokenRequest(r *http.Request) (*TokenRequest, error)
	// TokenMiddleware parses any "authorization" header containing a Bearer
	// token & adds it to the request context
	TokenMiddleware(next http.Handler) http.Handler
}

// TokenRequest is a wrapper for incoming token requests
type TokenRequest struct {
	GrantType    GrantType
	Code         string
	Username     string
	Password     string
	RefreshToken string
	RedirectURI  string
}

// TokenResponse wraps the token response object
type TokenResponse struct {
	AccessToken  string `json:"access_token"`
	TokenType    string `json:"token_type"`
	ExpiresIn    int64  `json:"expires_in"`
	RefreshToken string `json:"refresh_token,omitempty"`
}
```

The interface is quite simple in that the intersection between local & cloud and to make it work we only need to support getting a token (along with refreshing it). The middleware is only forced on here because down the line ideally most cloud request would flow through the core implementation in some shape or form, so it's important to be able to inject a custom middleware for each which process the token slightly differently. Standalone `qri` is meant to set the `scope.ActiveProfile` while on cloud it's intentionally kept separate as we have a global shared state via the `datastore` and will not be ready to move to scope in the near term.

[...]

PR #1809, opened 2021-06-15
feat(token): add and implement token provider interface

Arqu:
Part of: https://github.com/qri-io/qri/issues/1806

There will be a follow up PR that deals with aligning the token claims with what is expected for interop with cloud. 
Also it's still outstanding to do: https://github.com/qri-io/qri/issues/1810, https://github.com/qri-io/qri/issues/1811, https://github.com/qri-io/qri/issues/1812

### DIFF
--- a/auth/token/token_provider.go
+++ b/auth/token/token_provider.go
@@ -1,0 +1,175 @@
+package token
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/qri-io/qri/auth/key"
+	"github.com/qri-io/qri/profile"
+)
+
+const (
+	// RTCode signals the token response type is 'code'
+	RTCode ResponseType = "code"
+	// RTToken signals the token response type is 'token'
+	RTToken ResponseType = "token"
+
+	// AccessTokenTTL is the lifespan of an access token
+	AccessTokenTTL = time.Hour * 2
+	// RefreshTokenTTL is the lifespan of a refresh token
+	RefreshTokenTTL = time.Hour * 24 * 30
+)
+
+var (
+	// ErrInvalidRequest is returned on any parse or void output error
+	ErrInvalidRequest = fmt.Errorf("invalid request")
+	// ErrInvalidCredentials signals a bad username/password/key error
+	ErrInvalidCredentials = fmt.Errorf("invalid user credentials")
+	// ErrNotFound is returned when no matching results exist for the provided credentials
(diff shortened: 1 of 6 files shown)


auth/token/token_provider.go
```diff
@@ -0,0 +1,167 @@
...
+		resp.RefreshToken = refreshToken
+	case Refreshing:
+		tok, err := ParseAuthToken(req.RefreshToken, p.keys)
+		if err != nil {
+			log.Debugf("token.Provider error parsing refresh token: %q", err.Error())
+			return nil, ErrInvalidRequest
+		}
+
+		if claims, ok := tok.Claims.(*Claims); ok {
+			pro, err := p.profiles.GetProfile(profile.IDB58DecodeOrEmpty(claims.ProfileID))
```
dustmop, line 142: Since we already have a bunch of error handling here, and the `claims.ProfileID` is coming from external user input, and passing an empty string to `profiles.GetProfile` will mask decoding errors, it would be better here to use `profile.IDB58Decode` and return an error rather than using the `...OrEmpty` version.

auth/token/token_provider.go
```diff
@@ -0,0 +1,167 @@
...
+			}
+			resp.AccessToken = accessToken
+		}
+	default:
+		return nil, ErrInvalidRequest
+	}
+	return resp, nil
+}
+
+var (
```
dustmop, line 160: Typically we put custom error definitions closer to the top of the file. Could you move this to after the `var` definitions and before `type Provider interface`.

auth/token/token_provider.go
```diff
@@ -0,0 +1,167 @@
...
+		pros, err := p.profiles.ProfilesForUsername(req.Username)
+		if err != nil {
+			log.Debugf("token.Provider failed to fetch profiles: %q", err.Error())
+			return nil, ErrInvalidRequest
+		}
+		if len(pros) == 0 {
+			log.Debugf("token.Provider no matching profiles found")
+			return nil, ErrNotFound
+		}
+		pro := pros[0]
```
dustmop, line 117: How about adding a `log.Info` when there is more than 1 profile.

dustmop (approved):
LGTM, just a few small comments.

auth/token/token_provider.go
Arqu: did the same fix on `lib/lib.activeProfile`

PR #1813, opened 2021-06-17
refactor(token): extend provider generics for other implementations

Arqu:
This sets up a good baseline for depending project like cloud and qrimatic to leverage the same infrastructure as in core.


auth/token/token_provider.go
```diff
@@ -29,6 +31,18 @@ var (
 	ErrInvalidCredentials = fmt.Errorf("invalid user credentials")
 	// ErrNotFound is returned when no matching results exist for the provided credentials
 	ErrNotFound = fmt.Errorf("user not found")
+	// ErrServerError is returned on unexpeted errors
```
dustmop, line 34: unexpeted -> unexpected

dustmop (approved):
LGTM

PR #1814, opened 2021-06-23
feat(trigger): `Trigger`, `Listener`, and `Source` interfaces, as well as `AssertTriggerListener` spec tests

### LINKED ISSUE #1792: feat(automation): implement runtime triggers & add trigger listeners to orchestrator
### Test
```go
// orchestrator_test.go
package automation

// create test listeners w/ test triggers?

func TestListeners(t *testing.T) {
    // create bus
    // create handler that records when `event.ETTriggerWorkflow` is emitted
    // create orchestrator with 2 Listeners
    // emit triggers on bus
    // should have no `event.ETTriggerWorkflow` emitted
    // o.Start
    // emit triggers on bus
    // should have `event.ETTriggerWorkflow` emitted
    // o.Stop
    // emit triggers on bus
    // should not have more `event.ETTriggerWorkflow` events emitted
}
```

```go
// trigger_test.go
package trigger

// figure out how to create an AssertListenerSpec?

func TestCronListener(t *testing.T) {
    // create store
    // create bus
    // create handler that listens for `event.ETTriggerWorkflow`
    // create CronListener
    // ensure correct Type
    // ConstructTrigger from map, ensure expected
    // Start
    // Wait long enough to trigger
    // Stop
    // ensure we have emitted event
}

func TestCronTrigger(t *testing.T) {
    // create CronListener
    // great CronTrigger from `cl.ConstructTrigger` & options map
    // check enabled is expected
    // check ToMap creates map identical to options Map
    // Advance
    // check we have changed everything we need to for advancing
}
```

### declare Trigger interface & adjust workflow 
```go
package workflow

type Store interface {
    Updater
    Lister
}

type Updater interface {
    Create(did dataset.ID, pid peer.ID, triggers []trigger.Trigger) (*State, error)
}

type Lister interface {
    // List lists all the workflows in the store
    List(ctx context.Context, offset, limit int) ([]*Workflow, error)
    // ListEnabled lists all the enabled workflows in the store
    ListEnabled(ctx context.Context, offset limit int) ([]*Workflow, error)
}

type State struct {
    ...
    triggers []trigger.Trigger
}

func (m *MemStore) Create(did dataset.ID, pid peer.ID, triggers []trigger.Trigger) (*State, error)
func (m *MemStore) List(ctx context.Context, offset, limit int) ([]*Workflow, error)
func (m *MemStore) ListEnabled(ctx context.Context, offset, limit int) ([]*Workflow, error)
```
```go
// workflow/trigger.go
package workflow

// Type describes the type of trigger the system listens for
type TriggerType string

// Trigger has all the information needed to determine under what circumstances an `event.ETTriggerWorkflow`
// should be emitted. A Trigger should also know how to "advance" itself once a trigger was... triggered.
type Trigger interface {
    json.Marshaler
    json.Unmarshaler
    ID() string
    Enabled() boolean
    Type() TriggerType
    ToMap() map[string]interface{}
    Advance() error
}

type TriggerOptions struct {
    Type TriggerType
    Options map[string]interface{}
}
```

### define Listener interface & implement CronListener & CronTrigger
```go
// cron.go
package trigger

// Listener emits a `event.ETTriggerWorkflow` when a specific stimulus is triggered
// It knows how to start and stop itself, as well as how to create new triggers for its specific stimulus
type Listener interface {
    ConstructTrigger(m map[string]interface{}) (*workflow.Trigger, error)
    Type() workflow.TriggerType
    Start(ctx context.Context) error
    Stop() error
}

type CronListener struct { 
    pub event.Publisher
    store workflow.Lister
    interval time.Duration
}

func NewCronListener(store workflow.Lister, bus event.Bus) (*CronListener, error)
func (c *CronListener) ConstructTrigger(m map[string]interface{}) (*workflow.Trigger, error)
func (c *CronListener) Type() workflow.TriggerType { return CronTriggerType }
func (c *CronListener) Start(ctx context.Context) error {
    // start up loop to periodically check for triggers that has passed their run start times
    // basically just adapt existing Cron.Start function
}
func (c *CronListener) Stop() error

var _ Listener = (*CronListener)(nil)


const (
    CronTriggerType = workflow.TriggerType("CronTriggerType")
)

type CronTrigger struct {
    ID
    Type
    Enabled
    Start time.Time
    Perodicity iso8601.RepeatingInterval
    NextRunStart *time.Time
}
// much of this can be lifted from workflow/trigger.go
func (ct *CronTrigger) MarshalJSON() ([]byte, error)
func (ct *CronTrigger) UnmarshalJSON([]byte) error
func (ct *CronTrigger) ID() string
func (ct *CronTrigger) Enabled() boolean
func (ct *CronTrigger) Type() workflow.TriggerType
func (ct *CronTrigger) ToMap() map[string]interface{}
func (ct *CronTrigger) Advance() error

var _ workflow.Trigger = (*CronTrigger)(nil)
    
```

### adjust Orchestrator
```go
package automation

var (
    // ErrNoTriggerType indicates that no "type" information was given in the trigger option map
    ErrNoTriggerType = fmt.Errorf("")
    // ErrTriggerTypeNotFound indicates that the given trigger "type" given in the trigger option map
    // is not a type we support in the orchestrator
    ErrTriggerTypeNotFound = fmt.Errorf("")
)

type Orchestrator struct {
    ...
    listeners []*trigger.Listeners
}

func (o *Orchestrator) CreateWorkflow(did dataset.ID, pid peer.ID, triggerOpts []map[string]interface{}) (*workflow.State, error) {
    // iterate over trigger Opts
    triggers := []*workflow.Trigger
    for _, opt := range triggerOpts {
        type, ok := opt["type"]
        if !ok { return nil, ErrNoTriggerType }
        listener, err := o.listenerForTriggerType(type)
        if err != nil { return nil, err }    
        trigger, err := listener.ConstructTrigger(opt)
        triggers := append(triggers, trigger)
    }
    o.workflows.Create(did, pid, triggers)    
}

func (o *Orchestrator) listenerForTriggerType(type string) (*trigger.Listener, error) {
    // iterates over listeners for if they match the type
}

func (o *Orchestrator) StartListeners(ctx context.Context) error {
    // iterate over listeners & call listener.Start(ctx)
}

func (o *Orchestrator) StopListeners() error {
    // iterate over listeners & call listener.Stop(), only exiting when the listeners have completed stopping
}

func (o *Orchestrator) Shutdown() error {
    // call o.StopListeners()
}
```

### adjust orchestrator options
```go
type Options struct {
    ...
    listeners []*trigger.Listener
}

func OptListeners(l []*trigger.Listener) Option {
        o.listeners = l
        return nil
    }
}
```

ramfox:
closes #1792


ramfox:
@b5 last convo we had about the listeners keeping internal stores of active triggers, we decided to add `listener.UpdateTriggers(source Source) error` to the `trigger.Listener` interface. Looking through your june demo spike surrounding the `CronTrigger`, I see a `CronListener.Listen(source ...Source)` method.

What do you think about removing `UpdateTriggers` in favor of `Listen`, with the expectation that `Listen` is called, providing a list of active sources when the `Listener` is mounted, and `Listen` is called whenever a workflow is created or updated (the way `UpdateTriggers` is used now).

b5:
yep, I think renaming `UpdateTriggers` to `Listen` & making the argument variadic are both wins. Repeated calls to listen makes sense to me

automation/trigger/runtime.go
```diff
@@ -0,0 +1,288 @@
...
+		return nil, fmt.Errorf("expected \"active\" field to be a boolean value")
+	}
+	c, ok := opt.Config["advanceCount"]
+	if !ok {
+		c = 0
+	}
+	count, ok := c.(int)
+	if !ok {
+		return nil, fmt.Errorf("expected \"advanceCount\" field to be an int value")
+	}
```
ramfox, line 146: This should be re-written using json.Marshal and Unmarshal. can do now or in follow up when I get cron trigger in.

automation/trigger/source.go
```diff
@@ -0,0 +1,10 @@
+package trigger
+
+import "github.com/qri-io/qri/profile"
+
+// Source is an abstraction for a `workflow.Workflow`
+type Source interface {
+	WorkflowID() string
+	ActiveTriggers(t Type) []Trigger
+	ScopeID() profile.ID
```
b5, line 9: let's rename this to `Owner() profile.ID`. Scopes are a runtime thing.

automation/workflow/workflow.go
```diff
@@ -99,6 +99,27 @@ func (w *Workflow) Copy() *Workflow {
 	return workflow
 }
 
+// ScopeID returns the owner id
+func (w *Workflow) ScopeID() profile.ID {
```
b5, line 103: `ScopeID` -> `Owner`

automation/trigger/listener.go
```diff
@@ -0,0 +1,36 @@
...
+	Listen(source ...Source) error
+	// Type returns the Type of Trigger that this Listener listens for
+	Type() Type
+	// Start puts the Listener in an active state of listening for triggers
+	Start(ctx context.Context) error
+	// Stop stops the Listener from listening for triggers
+	Stop() error
+	// Bus returns the underlying Bus on which the Listener will emit the
+	// `event.ETTriggerWorkflow` event
+	Bus() event.Bus
```
b5, line 35: would love to see this method removed. When external systems can access & retain the bus it can lead to a rats nest of "how is this event.Bus nil!?".

automation/spec/trigger.go
```diff
@@ -54,3 +58,80 @@ func AssertTrigger(t *testing.T, trig trigger.Trigger) {
 		t.Fatalf("json.Unmarshal should emit a `trigger.ErrUnexpectedType` error if the given type does not match the trigger.Type of the Trigger")
 	}
 }
+
+// AssertListener confirms the expected behavior of a trigger.Listener
+// NOTE: this does not confirm behavior of the `Listen` functionality
+// beyond the basic usage of adding a trigger using a `trigger.Source`
+func AssertListener(t *testing.T, listener trigger.Listener, source trigger.Source, activateTrigger func()) {
```
b5, line 65: To remove the need for the `Bus` method, pass a constructor function as an arg that returns a `Listener` and an `activate` function instead:

```go
// ListenerConstructor creates a trigger listener and function that fires the listener when called
type ListenerConstructor func() (listener trigger.Listener, activate func()

func AssertListener(t *testing.T, constructor ListenerConstructor) {
```

you should also be able to internalize the creation of the workflow being passed as a `source` arg to this test

b5 (commented):
Looking great!

automation/trigger/listener.go
ramfox: Ahh, yes, I thought I got rid of all of these in a previous iteration!

automation/hook/runtime_hook.go
```diff
@@ -18,7 +18,7 @@ type RuntimeHook struct {
 var _ Hook = (*RuntimeHook)(nil)
 
 // RuntimeType denotes a `RuntimeHook`
-var RuntimeType = Type("Runtime Hook")
+var RuntimeType = "Runtime Hook"
```
b5, line 21: use `const` instead

automation/orchestrator.go
```diff
@@ -92,6 +95,16 @@ func NewOrchestrator(ctx context.Context, bus event.Bus, runFactory RunFactory,
 		runs:         opts.RunStore,
 	}
 
+	for _, l := range opts.Listeners {
+		if o.listeners == nil {
+			o.listeners = map[string]trigger.Listener{}
+		}
+		_, ok := o.listeners[l.Type()]
+		if ok {
```
b5, line 103: combine these lines:

```go
if _, ok o.listeners[l.Type()]; ok {
```

b5 (approved):
LGTM!

automation/orchestrator.go
ramfox: brain fart, ty

PR #1815, opened 2021-06-24
refactor(connect): re-work connection detail printout for qri connect

b5:
Small tweaks to improve connection UX from the command line.

Remove websockt & RPC values, which are no longer in use. Stagger printout while initial ports are being acquired. Print version number first.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 
 default: build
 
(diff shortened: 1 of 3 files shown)


Arqu (approved):
LGTM

PR #1816, opened 2021-06-25
fix(search): use consistent naming and output format for search

Arqu:
Aligning the search response format and registry response parsing.


b5 (approved):
LGTM!

PR #1819, opened 2021-06-29
update dependencies, migrate configuration from revision 3 to 4

### LINKED ISSUE #1618: Remove config.API.WebsocketAddress
config.API.WebsocketAddress is now deprecated as we use the same port as the http API. 
It shoud be removed on the next config migration.

### LINKED ISSUE #1619: Rename config.API.DisableWebui
As per code TODO:
```
// when we next have a config migration, we should probably rename this to
// EnableWebui and default to true. the double negative here can be confusing.
```

### LINKED ISSUE #1716: refactor(`config`): add `config.Remote` field `Type`, that will allow us to pass in options for a "mock"
Currently, the way we are handling `remoteClient` in a testing environment involves adding a `InstanceContextKey` "RemoteClient" field with value "mock".

We should instead add an optional field `config.Remote` `Type` that takes an string. If that `cfg.Remote.Type == "mock"`, then we load a `remote.NewMockClient`.

There are some other subtleties in the code, can you have a "mock" remoteClient & "operate" as `Remote`? Basically need to brush up on what exactly a mock remote client does.

Do we need to migrate the config to add this?

### LINKED ISSUE #1723: Remove Support for `readOnly` API mode, use auth tokens for privileged access

### LINKED ISSUE #1774: upgrade to go-ipfs v0.8.0 or higher
moving from v0.6.0 requires an IPFS repo migration

b5:
Warning: This PR adds a repo migration. Please back up your repo before running.

This is a big PR that does a bunch of stuff at once, most of which is centered around a repo migration:

* upgrades to IPFS v0.9.0
* ups our new minimum go version to go 1.15
* adds a new qri configuration revision: `4` and associated migration code
* prepares configuration for our new `automation` subsystem
* rename remote configuration fields to make more sense
* rename remote package structs to match changes to remote configuration
* re-work `lib.NewInstance` Options to match remote changes
* add a `lib.OptRemoteClientConstructor` option to address #1716
* move all mock remote client code into it's own package & out of production binary
* removes a  bunch of dead configuration fields

Known issues / things that could be better:
* we don't have a good strategy for doing "historical" migrations, so this PR changes the way migrating from config revision 1 to 2 will work. Specifically: `confg.RPC` and `cfg.API.WebsocketAddress` won't be upgraded. While this is bad, I'm confident we could work around it in practice. Most folks will be running a number of migrations in succession to get to the latest, which should be ok.
* There's still more renaming to do in the `remote` package

closes #1618, closes #1619, closes #1716. closes #1723, closes #1774.


config/migrate/migrate.go
```diff
@@ -145,21 +150,25 @@ func oneToTwoConfig(cfg *config.Config) error {
 		if apiCfg.Address == "" {
 			apiCfg.Address = defaultAPICfg.Address
 		}
-		if apiCfg.WebsocketAddress == "" {
-			apiCfg.WebsocketAddress = defaultAPICfg.WebsocketAddress
-		}
+		// TODO(b5): need a strategy for setting config now that this field is removed
```
Arqu, line 153: Think we should drop the code and just create and issue for the next config migration to work out how we maintain older migrations intact if future migrations make a "breaking change".

config/api.go
```diff
@@ -93,15 +75,12 @@ func (a API) Validate() error {
...
+		Enabled: true,
+		Address: DefaultAPIAddress,
 		AllowedOrigins: []string{
-			"electron://local.qri.io",
-			"http://app.qri.io",
-			"https://app.qri.io",
+			fmt.Sprintf("http://localhost:%s", DefaultAPIPort),
 		},
-		DisableWebui: false,
+		EnableWebui: true,
```
Arqu, line 83: Unsure about the default state of this. The rename to me implies that the default is `false` and then you enable it by config. I'm fine with it to stay as is, but think `DisableWebui` reflected the basic assumptions better.

Arqu (approved):
Some nitpicks, but looks good to me. Sets a great stage to work with remotes as a plug and play system. I've tested the migrations out locally and runs smooth on my setup (which is probably fairly broken, so yay for passing).

On a side note, was testing some basic usage metrics. Performance seems to be a bit better on save/get, memory usage while still accumulating on new peer connections seems to be at a slightly lower pace so I'm happy with that. Haven't tested the memory usage thoroughly though.

config/api.go
dustmop: I am ok with the rename to `EnableWebui`. I think generally, defaults for bool fields should be false, but we already break that by having `Enabled: true` field up a little bit higher. With that in mind, at least getting rid of the double negative is being more consistent.

b5: Yeah I'm generally with @dustmop's view on this one. How y'all feel about just `Webui` instead of `EnableWebui`?
`Webui: true` feels like "would you like webui? yes"

dustmop: Yeah, I'm ok with `Webui` as well.

Arqu: +1

PR #1820, opened 2021-06-29
feat(feature_flags): added support for basic feature flag management

Arqu:
Adding feature flags management to qri core. Cloud should reuse this with a different store backing it, and frontend should source flags from the `/config/flags` endpoint.

### DIFF
--- a/config/feature/flag.go
+++ b/config/feature/flag.go
@@ -1,0 +1,22 @@
+package feature
+
+import ()
+
+// ID is a string identifier for a workflow
+type ID string
+
+// Flag represents a feature and its state
+type Flag struct {
+	ID     ID   `json:"id"`
+	Active bool `json:"active"`
+}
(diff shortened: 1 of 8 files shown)


Arqu:
Just distilling come notes from our discussions around this:
- we should rename this to `capabilities`
- it's intentionally loose compared to config
- we're prioritizing cloud/frontend usage with this currently
- we should be weary of the code AND business complexity this adds long term
- the client/user of the flags should hold the responsibility to look and understand the flags that are relevant to it
- the overall response shape should look something like
```json
{
    "core.p2p": true,
    "core.automation": true,
    "core.email": false,

    "automation.emailCompletion": true,
    
    "experimental.issue.comments": false
}
``` 

@b5 @ramfox @dustmop @chriswhong tagging you to add in feedback and help shape this

PR #1822, opened 2021-06-30
feat(automation): expand Orchestrator methods

### LINKED ISSUE #1794: feat(automation): expand orchestrator methods
We need to add methods to the orchestrator so that other systems can use it to manipulate the `workflow.Store`. The `Orchestrator` understands what `trigger.Listeners` and `hook.Completors` we have access to. We can use this knowledge to convert []map[string]interface{} fields into specific triggers and hooks of the specified opt["type"]. If there is no "type" field, then the `CreateWorkflow` and `UpdateWorkflow` functions will error.

The other functions are more or less pass throughs of `o.workflows` functions.

```go
func (o *Orchestrator) UpdateWorkflow(wid workflow.ID, triggersOpts []map[string]interface{}, hookOpts []map[string]interface{}) (*workflow.State, error)
func (o *Orchestrator) GetWorkflow(wid workflow.ID) (*workflow.State, error)
func (o *Orchestrator) RemoveWorkflow(wid workflow.ID) error
func (o *Orchestrator) DeployWorkflow(wid workflow.ID) error
func (o *Orchestrator) UndeployWorkflow(wid workflow.ID) error
func (o *Orchestrator) ListWorkflows() ([]*workflow.State, error)
func (o *Orchestrator) ListDeployedWorkflow() ([]*workflow.State, error)
```

Tests
```go
// automation_test.go
func TestWorkflowIntegration(t *testing.T) error {
    // create orchestrator with a test listener and test completor
    // create trigger options
    // create hook options
    // CreateWorkflow
    // GetWorkflow
    // ensure correct trigger and hooks
    // CreateWorkflow
    // ListWorkflow
    // UpdateWorkflow
    // GetWorkflow
    // DeployWorkflow
    // ListDeployed // one
    // UndeployWorkflow
    // ListDeployed // zero
    // RemoveWorkflow
    // GetWorkflow // should fail
}
```

ramfox:
closes #1794 

This PR fills out the orchestrator methods and add the orchestrator to the instance

it also:

- Changes `workflow.Workflow` struct - Triggers and Hooks are now `map[string]interface{}`s. `Workflow.Deployed` has been changed to `Workflow.Active`  
- Adds IDs to triggers
- Adds ToMap method to `Trigger` interface
- Adds events `ETWorkflowStarted` `ETWorkflowStopped`
- Adds `Deploy` method to AuthomationMethods
- Adds `profile.IDFromCtx` and `profile.AddToContext`, so we can get the profile owner from the context


automation/orchestrator.go
```diff
@@ -295,28 +318,31 @@ func (o *Orchestrator) CreateWorkflow(did string, pid profile.ID, triggerOpts []
 	return wf, nil
 }
 
+// SaveWorkflow updates a workflow in the workflow.Store
+func (o *Orchestrator) SaveWorkflow(wf *workflow.Workflow) (*workflow.Workflow, error) {
+	return o.workflows.Put(wf)
+}
```
ramfox, line 324: Thinking I should move the checks for empty fields and mismatched fields into this function to allow the workflow store methods to be even simpler, just dumbly putting the entry into the store.

automation/orchestrator.go
```diff
@@ -196,13 +200,28 @@ func (o *Orchestrator) Shutdown() {
...
+					trig.Advance()
+					break
+				}
+			}
+			wf, err = o.SaveWorkflow(wf)
+			if err != nil {
+				log.Debugw("handleTrigger: error saving workflow", "id", wtp.WorkflowID, "err", err)
+			}
+			if err := o.runWorkflow(ctx, wf); err != nil {
+				log.Debugw("handleTrigger: error running workflow", "err", err)
```
ramfox, line 224: I foresee some potential bugs where a user can edit a `Workflow` after it was triggered but before it is run. This worries me less in a 

I think we need to add "locking" workflows for editing to the responsibilities of the future "sync" system.

automation/orchestrator.go
```diff
@@ -125,22 +129,36 @@ func NewOrchestrator(ctx context.Context, bus event.Bus, runFactory RunFactory,
...
+// DefaultOrchestratorOptions is a temporary solution to supplying options to the orchestrator
+// TODO (ramfox): remove this in favor of using the automation configuration to
+// determing what the orchestrator should be configured as
+func DefaultOrchestratorOptions() OrchestratorOptions {
+	return OrchestratorOptions{
+		WorkflowStore: workflow.NewMemStore(),
+		RunStore:      run.NewMemStore(),
+	}
+}
+
```
ramfox, line 149: this is temporary, so we can iterate

automation/orchestrator.go
```diff
@@ -191,41 +209,91 @@ func (o *Orchestrator) Shutdown() {
...
+				return
+			}
+			wf = o.advanceTrigger(wf, wtp.TriggerID)
+			wf, err = o.SaveWorkflow(wf)
+			if err != nil {
+				log.Debugw("handleTrigger: error saving workflow", "id", wtp.WorkflowID, "err", err)
+			}
+			runID := run.NewID()
+			if err := o.runWorkflow(ctx, wf, runID); err != nil {
+				log.Debugw("handleTrigger: error running workflow", "err", err)
```
ramfox, line 258: potential bug here, where a user updates a workflow as we are prepping for a run

automation/orchestrator.go
```diff
@@ -238,92 +306,142 @@ func (o *Orchestrator) RunWorkflow(ctx context.Context, wid workflow.ID) error {
...
+		if fetchedWF.DatasetID != wf.DatasetID {
+			return nil, fmt.Errorf("SaveWorkflow error: given workflow %q has a different DatasetID than the workflow on record", wf.ID)
+		}
+		if fetchedWF.OwnerID != wf.OwnerID {
+			return nil, fmt.Errorf("SaveWorkflow error: given workflow %q has a different OwnerID than the workflow on record", wf.ID)
+		}
+		if fetchedWF.Created != wf.Created {
+			return nil, fmt.Errorf("SaveWorkflow error: given workflow %q has a different Created time than the workflow on record", wf.ID)
+		}
+	}
```
ramfox, line 384: moved the responsibility of catching potential problems out of the `workflow.Store` and into `SaveWorkflow`

Combined `CreateWorkflow` and `UpdateWorkflow` into `SaveWorkflow`

lib/automation.go
```diff
@@ -126,3 +147,183 @@ func (automationImpl) Apply(scope scope, p *ApplyParams) (*ApplyResult, error) {
...
+	rollback = false
+}
+
+func (inst *Instance) run(ctx context.Context, streams ioes.IOStreams, w *workflow.Workflow, runID string) error {
+	ctxWithProfile := profile.AddToContext(ctx, w.OwnerID.String())
+	scope, err := newScope(ctxWithProfile, inst, "local")
+	if err != nil {
+		return err
+	}
+	ref, err := scope.ResolveReference(ctx, &dsref.Ref{InitID: w.DatasetID})
```
b5, line 301: Bug here: the `ref` variable is actually the "source" that the reference resolved from, which will always be `""` b/c the scope uses a local resolver.  Instead use the `Human()` method on the reference passed into `ResolveReference`:

```go
ref := &dsref.Ref{InitID: w.DatasetID}
	if _, err = scope.ResolveReference(ctx, ref); err != nil {
		return fmt.Errorf("resolving InitID %q: %w", w.DatasetID, err)
	}
	p := &SaveParams{
		Ref: ref.Human(),
		Dataset: &dataset.Dataset{
			ID: w.DatasetID,
			Commit: &dataset.Commit{
				RunID: runID,
			},
		},
```

be sure to rebase from master to make sure you're getting the right error.

lib/automation.go
```diff
@@ -126,3 +147,183 @@ func (automationImpl) Apply(scope scope, p *ApplyParams) (*ApplyResult, error) {
...
+		Dataset: p.Dataset,
+		Apply:   false,
+	}
+	// TODO(ramfox): bandaid! remove when save can handle having saving a dataset with no body/structure
+	if p.Dataset.ID == "" && p.Dataset.BodyFile() == nil {
+		saveParams.Apply = true
+	}
+	ds, err := datasetImpl{}.Save(scope, saveParams)
+	if err != nil && !errors.Is(err, dsfs.ErrNoChanges) {
+		log.Errorw("deploy save dataset", "error", err)
```
b5, line 217: use `Debug` level instead. Feels counterintuitive, but CLI users see ugly errors if we leave them at Error level

lib/automation.go
```diff
@@ -126,3 +147,183 @@ func (automationImpl) Apply(scope scope, p *ApplyParams) (*ApplyResult, error) {
...
+
+	go func(dp *event.DeployEvent) {
+		if err := scope.Bus().PublishID(scope.Context(), event.ETDeploySaveWorkflowStart, ref, deployPayload); err != nil {
+			log.Debug(err)
+		}
+	}(&deployPayload)
+
+	wf, err = scope.Automation().SaveWorkflow(wf)
+	if err != nil {
+		log.Errorw("deploy save workflow", "error", err)
```
b5, line 246: debug level

lib/automation.go
```diff
@@ -126,3 +147,183 @@ func (automationImpl) Apply(scope scope, p *ApplyParams) (*ApplyResult, error) {
...
+		deployPayload.RunID = runID
+		go func(dp *event.DeployEvent) {
+			if err := scope.Bus().PublishID(scope.Context(), event.ETDeployRun, ref, deployPayload); err != nil {
+				log.Debug(err)
+			}
+		}(&deployPayload)
+
+		err := scope.Automation().RunWorkflow(scope.Context(), wf.ID, runID)
+		if err != nil {
+			log.Errorw("deploy run workflow", "error", err)
```
b5, line 275: debug level

lib/automation.go
```diff
@@ -126,3 +147,183 @@ func (automationImpl) Apply(scope scope, p *ApplyParams) (*ApplyResult, error) {
...
+	}
+
+	deployPayload.WorkflowID = wf.ID.String()
+	go func(dp *event.DeployEvent) {
+		if err := scope.Bus().PublishID(scope.Context(), event.ETDeploySaveWorkflowEnd, ref, deployPayload); err != nil {
+			log.Debug(err)
+		}
+	}(&deployPayload)
+
+	if p.Run {
```
b5, line 263: to avoid double-running while we work on saving without dataset bodies, let's skip running if `saveParams.Apply` is true:

```go
if p.Run && !saveParams.Apply
```

lib/automation.go
```diff
@@ -126,3 +147,183 @@ func (automationImpl) Apply(scope scope, p *ApplyParams) (*ApplyResult, error) {
...
+	vi := dsref.ConvertDatasetToVersionInfo(p.Dataset)
+	ref := vi.SimpleRef().String()
+	deployPayload := event.DeployEvent{
+		Ref:        ref,
+		DatasetID:  p.Dataset.ID,
+		WorkflowID: p.Workflow.ID.String(),
+	}
+	log.Debugw("deploy started", "payload", deployPayload)
+	if scope.Bus() != nil {
+		scope.Bus().PublishID(scope.Context(), event.ETDeployStart, ref, deployPayload)
```
b5, line 180: when making an API call, this scope's context is matched to the length of the request, which means it's likely done by this point. You'll have to use `scope.AppContext()` instead, possibly wrapping `scope.AppContext()` with the current user profile first.

This applies to all uses of `scope.Context()` within this function. It'll be more maintainable to build a `ctx` variable early in the function & use that

lib/automation.go
```diff
@@ -126,3 +147,183 @@ func (automationImpl) Apply(scope scope, p *ApplyParams) (*ApplyResult, error) {
...
+	}
+	rollback := true
+	defer func(dp *event.DeployEvent) {
```
b5, line 183: This function doesn't use the `dp` argument. remove

lib/automation.go
```diff
@@ -126,3 +147,183 @@ func (automationImpl) Apply(scope scope, p *ApplyParams) (*ApplyResult, error) {
...
+			_, err := datasetImpl{}.Remove(scope, rp)
+			if err != nil {
+				log.Debugw("deploy rollback", "error", err)
+			}
+		}
+
+	}(&deployPayload)
+
+	go func(dp *event.DeployEvent) {
+		if err := scope.Bus().PublishID(scope.Context(), event.ETDeploySaveDatasetStart, ref, deployPayload); err != nil {
```
b5, line 201: any reason we're publishing pointers instead of `DeployEvent` literals? I see no situation where the event receiver should be able to modify the value. It'll have the added bonus of not needing to allocate a pointer to pass to each goroutine.

lib/automation.go
```diff
@@ -126,3 +147,183 @@ func (automationImpl) Apply(scope scope, p *ApplyParams) (*ApplyResult, error) {
...
+		if rollback {
+			rp := &RemoveParams{
+				Ref: ref,
+				Revision: &dsref.Rev{
+					Field: "ds",
+					Gen:   1,
+				},
+			}
+			_, err := datasetImpl{}.Remove(scope, rp)
+			if err != nil {
```
b5, line 193:
```suggestion
			if _, err := datasetImpl{}.Remove(scope, rp); err != nil {
```

event/workflow.go
```diff
@@ -1,9 +1,87 @@
...
 
+import (
+	"github.com/qri-io/qri/profile"
+)
+
 const (
 	// ETWorkflowTrigger signals that a workflow has been triggered
-	// Payload will be a string representation of the `workflow.ID`
-	// of the workflow that should be run
+	// Payload will be a WorkflowTriggerEvent
```
b5, line 9: All of these events should be namespaces under `automation` instead of `workflow`. Match constant strings to the existing event convention of `subsystem:UpperCamelCase`

event/workflow.go
```diff
@@ -1,9 +1,87 @@
...
 	// This event should not block
 	ETWorkflowTrigger = Type("workflow:trigger")
```
b5, line 11:
```go
ETAutomationWorkflowTrigger = Type("automation:WorkflowTrigger")
```

Payload names should stay as is. Generalize this to all event names in this file & rename file to `automation.go`

event/workflow.go
```diff
@@ -1,9 +1,87 @@
...
+	// ETWorkflowStarted signals that a workflow run has begun
+	// Payload will be a *WorkflowStartedEvent
+	// This event should not block
+	ETWorkflowStarted = Type("workflow:started")
```
b5, line 15:
```go
ETAutomationWorkflowStarted = Type("automation:WorkflowStarted")
```

event/workflow.go
```diff
@@ -1,9 +1,87 @@
...
+	ETWorkflowStopped = Type("workflow:stopped")
+	// ETDeployStart signals that a deploy has started
+	// Payload will be a DeployEvent
+	// This event should not block
+	ETDeployStart = Type("deploy:start")
+	// ETDeploySaveDatasetStart signals that we have started the save
+	// dataset portion of the deploy
+	// Payload will be a DeployEvent
+	// This event should not block
+	ETDeploySaveDatasetStart = Type("deploy:save dataset start")
```
b5, line 28:
```go
ETAutomationDeploySaveDatasetStart = Type("automation:DeploySaveDatasetStart")
```

verbose? yep. But at least it's clear

automation/orchestrator.go
```diff
@@ -191,41 +209,91 @@ func (o *Orchestrator) Shutdown() {
 	o.cancel()
 }
 
+// advanceTrigger may admit log errors
```
b5, line 212: admit -> emit

automation/orchestrator.go
```diff
@@ -191,41 +209,91 @@ func (o *Orchestrator) Shutdown() {
...
+	return wf
+}
+
 // handleTrigger calls `RunWorkflow` when an `event.ETWorkflowTrigger` event is fired
 // it expects the payload for the `event.ETWorkflowTrigger` to be a workflow.ID
 // represented as a string
 func (o *Orchestrator) handleTrigger(ctx context.Context, e event.Event) error {
 	if e.Type == event.ETWorkflowTrigger {
-		wid, ok := e.Payload.(string)
+		wtp, ok := e.Payload.(*event.WorkflowTriggerEvent)
```
b5, line 241: if we switch payloads from pointers to literals, be sure to update assertions like these. It's a good measure of the test suite: such a change should blow up tests

b5 (commented):
looking good!

lib/automation.go
ramfox: yea saw this and force pushed when I looked with fresh eyes

lib/automation.go
ramfox: that's v helpful, thank you!

lib/automation.go
ramfox: ah, no reason, this is just a weird hold over that I need to fix. When I was publishing events in the `defer` func, I needed to pass in a pointer to the defer func, and I just had it in my head that that was necessary each time.

automation/orchestrator_test.go
```diff
@@ -61,74 +48,179 @@ func TestIntegration(t *testing.T) {
...
+	expectedWorkflowStoppedEvent := event.WorkflowStoppedEvent{
+		DatasetID:  got.DatasetID,
+		OwnerID:    got.OwnerID,
+		WorkflowID: got.WorkflowID(),
+		RunID:      runID,
+		Status:     string(run.RSSucceeded),
+	}
+	var gotWorkflowStartedEvent event.WorkflowStartedEvent
+	var gotWorkflowStoppedEvent event.WorkflowStoppedEvent
+	workflowStoppedEventFired := make(chan string)
```
ramfox, line 159: realized we have 2 separate async events we need to make sure are happening in order to test properly
1) did the runFunc actually get triggered
2) did the workflow events fire properly. 

We need to wait for verification that the workflow events finished firing before we can test that they fired with all the correct information

b5: agreed. I think the better way to enforce this is to list out an expected order of events in a slice, build up a slice of all events received as you receive them, and compare the two

b5 (approved):
lovely. let's merge this & figure out what we missed 😄

automation/orchestrator_test.go
ramfox: ah, yes I'll definitely refactor to collect in a slice, but I phrased the above wrong. By "two separate events" I meant "two separate actions" 1) that the runFunc gets called and 2) that we have the correct events. But now looking at it, checking for the run func was only necessary before we put the events in place. Going to refactor this whole section.

PR #1828, opened 2021-07-08
Support docs examples from source comments, add json struct tags to lib input parameters

b5:
see docs by navigating to project root & running `make serve-api-docs`, then load `http://localhost:2502` in a browser.

after this PR, comments in the lib package have two new rules:
* comments can have at most one semicolon, which separates comments into `[description];[example]` sections
* a single example value is now supported when prefixed with 'e.g.', example values should match their syntax type (string examples should be "quoted", boolean examples should be true, numeric should be 2.145)

note that _only_ `e.g.` will work to create an example, not `eg.`, which I continually mistype

I also put these new changes to immediate use, annotating a slew of input structs with JSON tags to clean up their documentation & narrow field name case sensitivity

### DIFF
--- a/docs/api_docs.go
+++ b/docs/api_docs.go
@@ -11,6 +11,7 @@ import (
 	"io/ioutil"
 	"path/filepath"
 	"reflect"
+	"regexp"
 	"strings"
 	"text/template"
 
@@ -22,7 +23,14 @@ import (
 type docs struct {
 	QriVersion string
 	LibMethods []libMethod
+	MethodSets []methodSet
 	Types      []qriType
+}
+
+type methodSet struct {
+	Name        string
+	Doc         string
+	MethodCount int
 }
 
 type libMethod struct {
@@ -51,6 +59,7 @@ type field struct {
 	Doc          string
 	Tags         string
 	Comment      string
+	Example      string
 }
 
 type response struct {
@@ -65,15 +74,19 @@ func OpenAPIYAML() (*bytes.Buffer, error) {
 		return nil, err
 	}
 
-	var methods []libMethod
+	var (
+		methods    []libMethod
+		methodSets []methodSet
+	)
 
 	var nilInst *lib.Instance
-	for _, methodSet := range nilInst.AllMethods() {
-		msetType := reflect.TypeOf(methodSet)
-		methodAttributes := methodSet.Attributes()
+	for _, mSet := range nilInst.AllMethods() {
+		msetType := reflect.TypeOf(mSet)
+		methodAttributes := mSet.Attributes()
 
 		// Iterate methods on the implementation, register those that have the right signature
 		num := msetType.NumMethod()
+		validMethodCount := 0
 		for k := 0; k < num; k++ {
 			i := msetType.Method(k)
 			f := i.Type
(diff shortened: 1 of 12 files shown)


lib/datasets.go
```diff
@@ -224,9 +224,10 @@ func scriptFileSelection(ds *dataset.Dataset, selector string) (qfs.File, bool)
 // ActivityParams defines parameters for the Activity method
 type ActivityParams struct {
 	ListParams
-	// Reference to data to fetch history for
-	Ref  string
-	Pull bool
+	// Reference to data to fetch history for; eg b5/world_bank_population
```
dustmop, line 227: e.g. "b5/world_bank_population"

lib/datasets.go
```diff
@@ -365,11 +368,11 @@ func (m DatasetMethods) Pull(ctx context.Context, p *PullParams) (*dataset.Datas
 
 // PushParams encapsulates parmeters for dataset publication
 type PushParams struct {
-	Ref    string `schema:"ref" json:"ref"`
-	Remote string
+	Ref    string `json:"ref" schema:"ref" `
```
dustmop, line 371: Remove the space after `"ref"` and before the closing backtick.

docs/api_docs.go
```diff
@@ -451,11 +498,39 @@ func readASTFile(fset *token.FileSet, filepath string) (*ast.File, error) {
...
+	return f.Names[0].String()
+}
+
+func parseFieldComment(f *ast.Field) (description, example string) {
+	strs := strings.Split(f.Doc.Text(), ";")
+	switch len(strs) {
+	case 1:
+		return strs[0], ""
+	case 2:
+		example = strings.TrimSpace(strs[1])
```
dustmop, line 524: Should we check here if the `example` now starts with "e.g."? Maybe something like this:

```
if !strings.HasPrefix(example, "e.g.") {
  log.Errorf("example must start with 'e.g.', got %q", example)
  return strs[0], ""
}
```

lib/datasets.go
```diff
@@ -224,9 +224,10 @@ func scriptFileSelection(ds *dataset.Dataset, selector string) (qfs.File, bool)
 // ActivityParams defines parameters for the Activity method
 type ActivityParams struct {
 	ListParams
-	// Reference to data to fetch history for
-	Ref  string
-	Pull bool
+	// Reference to data to fetch history for; eg b5/world_bank_population
+	Ref string `json:"ref"`
+	// if true, pull any datasets that aren't stored locally; eg false
```
dustmop, line 229: e.g. false

lib/datasets.go
```diff
@@ -582,13 +585,13 @@ type RenderParams struct {
 	Ref string `json:"ref"`
 	// Optionally pass an entire dataset in for rendering, if providing a dataset,
 	// the Ref field must be empty
-	Dataset *dataset.Dataset
+	Dataset *dataset.Dataset `json:"dataset"`
 	// Optional template override
-	Template []byte
+	Template []byte `json:"template"`
 	// If true,
```
dustmop, line 591: Just noticed this incomplete comment. I think `UseFSI` was removed everywhere, though I'm not 100% sure if render needs it. Maybe just add a TODO to investigate.

dustmop (approved):
LGTM, with a few fix suggestions

PR #1830, opened 2021-07-09
fix(dataset): File components store content as text, instead of scriptBytes

### LINKED ISSUE #1826: readme component returned by `Get` should contain text of the file
One of the outstanding items on our fixit is concerning `qri get`: "readme isn’t populated". The root cause of the issue is that "file components", such as transform, viz, and readme, keep their files as "script" objects that only contain the path to the underlying file. Fixing this could be solved by just inlining the script contents for each of these components.

Is there any reason we don't want to always inline these 3 components in the `dataset.Get` method? It seems like all of these components are going to contain small files most of the time (we're probably a ways away from having large transform scripts showing up).

My proposal would be to modify `dataset/transform.go` and `dataset/readme.go` and `dataset/viz.go` so that each component has a `Text string` field that contains the plain text of the file.

Fixing this should also solve a number of related issues:
* https://github.com/qri-io/qri/issues/1578
* https://github.com/qri-io/qri/issues/1339

dustmop:
Depends on: https://github.com/qri-io/dataset/pull/262

Fixes #1826


go.mod
```diff
@@ -41,7 +41,7 @@ require (
 	github.com/olekukonko/tablewriter v0.0.4
 	github.com/pkg/errors v0.9.1
 	github.com/qri-io/dag v0.2.3-0.20210628012720-e8a2affbb114
-	github.com/qri-io/dataset v0.3.1-0.20210628012656-9f6991c2892e
+	github.com/qri-io/dataset v0.3.1-0.20210708233901-4c6ecaaf505d
```
b5, line 44: be sure to rebase this after merging the dataset PR to point to latest master commit

b5 (approved):
LGTM

PR #1831, opened 2021-07-09
feat(get): Limit size of body that `get` may return when using `all`

dustmop:


b5 (approved):
looks great! Just needs a rebase

PR #1834, opened 2021-07-12
feat(resolver): resolving a reference w an initID completes the ref

b5:
adding in human elements like username & name, along with HEAD path. We need this now because the automation subsystem is the first one to properly store _just_ initIDs


ramfox (approved):
LGTM!

PR #1835, opened 2021-07-14
fix(lib): adjust `Deploy` scope

ramfox:
We accounted for context cancellations when publishing the deploy
events, BUT we did not acccount for context cancellations when using the
scope itself as a parameter for the `Save` and `WorkflowSave` functions
that the `Deploy` function itself calls.
We need a scope that will outlast the deploy request. We construct a
scope using the instance, source, and the original scope's `AppContext`.


b5 (approved):
LGTM

PR #1837, opened 2021-07-15
refactor(starlark): Convert Dataset to a true starlark.Value

dustmop:


b5 (approved):
looks good to me. I'm assuming we'll use this approach to re-work the starlark API

PR #1838, opened 2021-07-18
feat(trigger): Add `CronListener` and `CronTriggers`

ramfox:
This PR also:
- starts the orchestrator listening for triggers when we run `qri connect`
- a new `MemTriggerStore` that we can use to track triggers for Listeners (gives a unified way to store ephemeral triggers, an done place to test/fix "Listen")
- adjusts `AssertListener` spec test - need to give space to `Advance` the trigger. Depending on how a `Listener` listens, we may need to "advance" the trigger so that the listener doesn't trigger off the same trigger multiple times. This is usually done by the orchestrator, so we need to mock that in the test
- adds the `CronListener` to our `DefaultOrchestratorOptions`


b5 (approved):
ended up formatting my review as a commit. feel free to have a look through! LGTM

PR #1843, opened 2021-07-27
use new collection subsystem locally

b5:
Apologies for the giant PR. 

This PR makes the collection subsystem the default and only source for listing operations when running qri locally. It migrates all info from the repo refstore subsystem, which can be deprecated after all users have migrated.

### Key changes:

* logbook `WriteDatasetInit` now accepts a username parameter
* local collection is passing nearly all tests that the refstore had to pass, this time using events
* many event payloads have been adjusted to remove duplicated fields and provide details the local collection relies on
* dataset "pushed & "pulled" events have been added with `VersionInfo` payloads. These are in addition to `ETDatasetPush/PullCompleted`, which are oriented toward progress monitoring

### Shortcomings:

* the collection set is multi-tenant, and relies on the event bus, which is not multi-tenant. There are numerous situations where we're adding or removing datasets for _all_ users when we should only be affecting a single user's collection. I'm confident we can get this sorted shortly.
* I've skipped a few tests 😢. Only in places where I know the subsystem being tested is going away soon
* restore cleanup: there are lots of place where we're hard-importing a refstore and updating it. We should remove those places
* the local collection does subscribe to automation events & update itself, but these subscriptions aren't tested.
* generally: event contracts aren't tested, and really need to be. It's also important that we make sure event publications that the local collection set relies on never fire in goroutines. local collection needs a chance to block the publishing function to update its state & write to disk

### Once this lands we can:
* make collection a `refResolver` and stop using logbook as a first line of defense for completing dataset references locally

### DIFF
--- a/base/dataset_test.go
+++ b/base/dataset_test.go
@@ -8,6 +8,9 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/qri-io/dataset/dstest"
 	"github.com/qri-io/qri/base/dsfs"
+	"github.com/qri-io/qri/collection"
+	"github.com/qri-io/qri/dsref"
+	"github.com/qri-io/qri/event"
 )
 
 func TestListDatasets(t *testing.T) {
@@ -88,9 +91,20 @@ func TestRawDatasetRefs(t *testing.T) {
 
 	ctx := context.Background()
 	r := newTestRepo(t)
-	addCitiesDataset(t, r)
+	s, err := collection.NewLocalSet(ctx, event.NilBus, "")
+	if err != nil {
+		t.Fatal(err)
+	}
 
-	actual, err := RawDatasetRefs(ctx, r)
+	ref := addCitiesDataset(t, r)
+	ws := s.(collection.WritableSet)
+	vi := dsref.NewVersionInfoFromRef(ref)
+	vi.InitID = "AnInitID"
+	if err := ws.Put(ctx, r.Profiles().Active(ctx).ID, vi); err != nil {
+		t.Fatal(err)
+	}
+
+	actual, err := RawDatasetRefs(ctx, r.Profiles().Active(ctx).ID, s)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/collection/migrate.go
+++ b/collection/migrate.go
@@ -2,38 +2,40 @@ package collection
 
 import (
 	"context"
-	"fmt"
+	"errors"
 
-	"github.com/qri-io/qri/event"
 	"github.com/qri-io/qri/repo"
 )
 
 // MigrateRepoStoreToLocalCollectionSet constructs a local collection.Set from
 // legacy repo data
-func MigrateRepoStoreToLocalCollectionSet(ctx context.Context, bus event.Bus, repoDir string, r repo.Repo) (Set, error) {
+func MigrateRepoStoreToLocalCollectionSet(ctx context.Context, s Set, r repo.Repo) error {
+	ws, ok := s.(WritableSet)
+	if !ok {
+		return errors.New("cannot migrate to CollectionSet. Provided CollectionSet is not writable")
+	}
+
 	datasets, err := repo.ListVersionInfoShim(r, 0, 1000000)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
 	book := r.Logbook()
 	for i, vi := range datasets {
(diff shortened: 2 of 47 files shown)


base/dataset_test.go
```diff
@@ -88,9 +91,20 @@ func TestRawDatasetRefs(t *testing.T) {
 
 	ctx := context.Background()
 	r := newTestRepo(t)
-	addCitiesDataset(t, r)
+	s, err := collection.NewLocalSet(ctx, event.NilBus, "")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ref := addCitiesDataset(t, r)
```
Arqu, line 99: This comes up oftern and feels a bit unergonomic, maybe we can push this whole chunk into the addCitiesDataset given we can push the collection into repo or get rid of repo.

event/dataset.go
```diff
@@ -1,24 +1,20 @@
 package event
 
-import (
-	"github.com/qri-io/qri/dsref"
-)
-
 const (
 	// ETDatasetNameInit is when a dataset is initialized
-	// payload is a DsChange
+	// payload is a dsref.VersionInfo
```
Arqu, line 5: 👍

remote/registry/regclient/client_test.go
```diff
@@ -48,7 +48,7 @@ func NewTestRunner(t *testing.T) (*TestRunner, func()) {
 		AcceptSizeMax: -1,
 		Enabled:       true,
 		AllowRemoves:  true,
-	}, r.Logbook())
+	}, r.Logbook(), node.Repo.Bus())
```
Arqu, line 51: qq: what's the difference between `registry/...` and `remote/registry/...` since they are near identical copies. I focused my efforts on the `registry/...` piece not sure if/why I would bring that over.

lib/collection.go
```diff
@@ -71,7 +72,35 @@ func (collectionImpl) List(scope scope, p *ListParams) ([]dsref.VersionInfo, err
 			Offset:  p.Offset,
 			Limit:   p.Limit,
 		}
-		return s.List(scope.ctx, scope.ActiveProfile().ID, lp)
+		infos, err := s.List(scope.ctx, scope.ActiveProfile().ID, lp)
+		if err != nil {
+			return nil, err
+		}
+
+		if p.EnsureFSIExists {
```
Arqu, line 80: I really dislike this param on the `ListParams`, feels very out of place and is tied to internal implementation details and should probably be something configuration specific not request param specific. I would at least rename it to `OnlyLocal` or something similar.

lib/collection.go
```diff
@@ -71,7 +72,35 @@ func (collectionImpl) List(scope scope, p *ListParams) ([]dsref.VersionInfo, err
...
+			// TODO(b5): adding a defensive check for the existence of the FSI
+			// subsystem in the hopes that cloud code won't follow this path. Haven't
+			// confirmed that's the case
+			if scope.FSISubsystem() != nil {
```
Arqu, line 84: AFAIK this has no bearing on cloud using FSI. FSI seems to be set and initialized on every instance creation including cloud. Cloud just simply didn't need/encounter it thus far. We should add some config check on `scope.FSISubsystem()` and return `nil` if configured so to make this viable.

collection/migrate.go
```diff
@@ -2,38 +2,59 @@ package collection
...
 
-	s, err := NewLocalSet(ctx, bus, repoDir)
-	if err != nil {
-		return nil, err
-	}
-
-	ws := s.(WritableSet)
-	if err = ws.Put(ctx, r.Profiles().Owner().ID, datasets...); err != nil {
-		return nil, err
+	// remove any datasets that couldn't be resolved
```
Arqu, line 52: Does this mean, old maybe broken datasets are now going to be completely lost from a user perspective as they can no longer list them. Maybe this is just a nice cleanup of what's broken anyways and we just bite the bullet?

Arqu (commented):
This is a lot! But super prep for the next batch of cleaning.
I'm just leaving this up for a discussion, but happy to move this forward.

lib/collection.go
b5: I fully agree and intend to remove this entirely in a follow up PR that removes FSI before the next release

collection/migrate.go
b5: yep, that was kinda my thinking. Why carry over broken stuff?

PR #1849, opened 2021-07-29
feat(lib): add lib `Run` and lib `Workflow` methods and endpoints

### LINKED ISSUE #1863: feat(orchestrator): `NewTestOrchestratorOptions`
```
orchestrator_test

// NewTestOrchestratorOptionsFromDir loads workflow test cases from the directory. If no `dirPath` is
// given, the workflow test cases will load from the default directory location
func NewTestOrchestratorOptionsFromDir(bus event.Bus, dirPath string) (OrchestratorOptions, []*wftest.TestCase, error)
```

```
package wftest

// TestCase is a workflow test case, usually built from a
// directory of files for use in tests
// Each workflow for each test case must have an associated
// dataset, whose `dataset.ID` matches the `workflow.DatasetID`
type TestCase struct {
    // Path to the directory on the local filesystem this test case is loaded from
    Path string
    // Name is the casename, should match the directory name
    Name string
    // Input is intended file for test input
    // loads from input.workflow.json
    Input *workflow.Workflow
    // Expect should match the test output
    // loads from expect.workflow.json
    Expect *workflow.Workflow
    Dataset *TestCase
}

var testCaseCache = map(map[string]*TestCase)

// LoadTestCases loads a directory of case directories
func LoadTestCases(dir string) (tcs []*TestCase, err error)

// NewTestCaseFromDir creates a test case from a directory of static test files
// dir should be the path to the directory to check
func NewTestCaseFromDir(dir string) (tc *TestCase, err error)

// ReadWorkflow grabs a workflow for a given dir for a given filename
func ReadWorkflow(dir, filename string) (*workflow.Workflow, error)

type TestRunner interface {
    Instance() *lib.Instance
    Owner() *profile.Profile
    Context() context.Context
}

// MustSaveDatasetToInstance adds the dataset in the `workflow.TestCase` to the
// instance,  updating the workflow to have the proper `workflow.DatasetID`
func MustSaveDatasetToInstance(t *testing.T, tr TestRunner, tc *TestCase) {
    var err error
    owner := tr.Owner()
    inst := tr.Instance()
    ctx := tr.Context()
    ds := tc.Dataset.Input
    wf := tc.Input
    ref := fmt.Sprintf("%s/%s", owner.Name, ds.Name)

    ds, err = inst.Dataset().Save(ctx, &lib.SaveParams{ Ref: ref, Dataset: ds })
    if err != nil {
        t.Fatalf("error saving dataset to instance: %s", err)
    }
    wf.DatasetID = ds.ID
    if _, err = inst.AutomationOrchestrator().SaveWorkflow(ctx, wf); err != nil {
        t.Fatalf("error saving workflow to instance: %s", err)
    }
}

```

Our major hurtle here is that in order to be able to call `lib.Run` successfully, we need the workflow to have an associated dataset that is saved properly to the instance. We also need that associated `dataset.ID` to be updated on the `Workflow`, so that we can properly fetch and run the `Workflow` via the `Workflow.DatasetID`. I'm proposing we use this `wftest.MustSaveDatasetToInstance` function that a test runner can use to make sure everything is aligned properly.

ramfox:
closes #1863 
closes #1849 

----

to reviewers: the last two commits could really have been in their own PR, but I didn't like the idea of adding these endpoints without ensuring they were tested in the API spec tests!

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -21,6 +21,7 @@ import (
 	ma "github.com/multiformats/go-multiaddr"
 	manet "github.com/multiformats/go-multiaddr/net"
 	apispec "github.com/qri-io/qri/api/spec"
+	"github.com/qri-io/qri/automation"
 	"github.com/qri-io/qri/base/dsfs"
 	"github.com/qri-io/qri/config"
 	testcfg "github.com/qri-io/qri/config/test"
@@ -135,9 +136,13 @@ func testConfigAndSetter() (cfg *config.Config, setCfg func(*config.Config) erro
 }
 
 func newTestInstanceWithProfileFromNode(ctx context.Context, node *p2p.QriNode) *lib.Instance {
+	return newTestInstanceWithProfileFromNodeAndOrchestratorOpts(ctx, node, nil)
+}
+
+func newTestInstanceWithProfileFromNodeAndOrchestratorOpts(ctx context.Context, node *p2p.QriNode, o *automation.OrchestratorOptions) *lib.Instance {
 	cfg := testcfg.DefaultConfigForTesting()
 	cfg.Profile, _ = node.Repo.Profiles().Owner().Encode()
-	return lib.NewInstanceFromConfigAndNode(ctx, cfg, node)
+	return lib.NewInstanceFromConfigAndNodeAndBusAndOrchestratorOpts(ctx, cfg, node, event.NilBus, o)
 }
 
 type handlerTestCase struct {
--- a/api/spec/testdata/automation.json
+++ b/api/spec/testdata/automation.json
@@ -17,5 +17,91 @@
         "Content-Type": "application/json"
       }
     }
+  },
+  {
+	  "endpoint": "/auto/deploy",
+	  "method": "POST",
+	  "headers": {
+		  "Content-Type": "application/json"
+	  },
+	  "body": {
+		  "ref": "peer/now",
+		  "dataset": {
+			  "name": "now",
+			  "peername": "peer",
+			  "transform": {
+				"steps": [
+					{
+					  "syntax":"starlark",
+					  "category":"setup",
+					  "name":"setup",
+					  "script":"# load starlark dependencies\nload(\"encoding/csv.star\", \"csv\")\nload(\"time.star\", \"time\")\n"
+					},
+					{
+					  "syntax":"starlark",
+					  "category":"transform",
+					  "name":"transform",
+					  "script":"# set the body\ndef transform(ds, ctx):\n  currentTime = time.now()\n  body = [\n    ['timestamp']\n  ]\n  body.append([str(currentTime)])\n  theCSV = csv.write_all(body)\n\n  ds.set_body(theCSV, parse_as='csv')\n"
+					}
+				]
+			}
+		},
+		"workflow": {
+		}
+	  },
+	  "expect": {
+		  "code": 200,
+		  "headers": {
+			  "Content-Type":"application/json"
+		  }
+	  }
+  },
+  {
+	  "endpoint": "/auto/run",
+	  "method": "POST",
+	  "headers": {
+		  "Content-Type": "application/json"
+	  },
+	  "body": {
+		  "workflowID": "1d79b0ff-a133-4731-9892-5ee01842ca81"
+	  },
+	  "expect": {
+		  "code": 200,
+		  "headers": {
+			  "Content-Type":"application/json"
+		  }
+	  }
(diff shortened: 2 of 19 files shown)


Arqu:
I've rebased to latest master as it makes my life branch surfing easier.

api/api_test.go
```diff
@@ -135,9 +136,13 @@ func testConfigAndSetter() (cfg *config.Config, setCfg func(*config.Config) erro
 }
 
 func newTestInstanceWithProfileFromNode(ctx context.Context, node *p2p.QriNode) *lib.Instance {
+	return newTestInstanceWithProfileFromNodeAndOrchestratorOpts(ctx, node, nil)
+}
+
+func newTestInstanceWithProfileFromNodeAndOrchestratorOpts(ctx context.Context, node *p2p.QriNode, o *automation.OrchestratorOptions) *lib.Instance {
```
ramfox, line 142: WE NEED TO KILL ALL PLACES THAT CALL THESE. We don't need it any more, now that we can add all these different parameters using `NewInstance` via options.

I already felt like I was adding too much to this pr with the `wftest` package, so I just followed the current patterns to get this to work with `APITestRunner`.

automation/workflow/wftest/wftest.go
```diff
@@ -0,0 +1,263 @@
...
+	// Expect should match the test output
+	// loads from expect.workflow.json
+	Expect  *workflow.Workflow
+	Dataset *dstest.TestCase
+}
+
+// LoadDefaultTestCases loads test cases from this package
+// the OwnerID is the first key in the `auth/key/test/keys.go`, which is the
+// primary key used in the default test configs
+func LoadDefaultTestCases() ([]*TestCase, error) {
```
ramfox, line 54: The default way to get workflow test cases!

The test cases currently in the `wftest/testdata` directory were primarily created for testing within the `wftest` pacakge and as a model for other packages that want to load from a directory.

lib/automation.go
```diff
@@ -97,6 +99,52 @@ func (m AutomationMethods) Deploy(ctx context.Context, p *DeployParams) error {
...
+
+// WorkflowParams are parameters for the Workflow command
+type WorkflowParams struct {
+	WorkflowID string `json:"workflowID"`
+	InitID     string `json:"initID"`
+}
+
+// Validate returns an error if WorkflowParams fields are in an invalid state
+func (p *WorkflowParams) Validate() error {
+	if p.WorkflowID == "" && p.InitID == "" {
```
dustmop: If both are provided, the implementation will use WorkflowID and ignore InitID. Should we make it a validation error to use both at once?

automation/orchestrator.go
```diff
@@ -165,6 +165,18 @@ func DefaultOrchestratorOptions(bus event.Bus, repoPath string) (OrchestratorOpt
 	}, nil
 }
 
+// DefaultMemOrchestratorOptions is primarily for use in tests
+// it returns options for an orchestrator whose Stores are in memory implimentations
```
dustmop, line 169: implimentations -> implementations

api/spec/testdata/automation.json
```diff
@@ -17,5 +17,91 @@
         "Content-Type": "application/json"
       }
     }
+  },
+  {
```
Arqu, line 21: Really happy to have this for my own testing

Arqu (approved):
Way to go on having s propper flow for workflow testing. 
🚀

automation/workflow/wftest/wftest.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+	}
+
+	preserve := TestCase{}
+	copier.Copy(&preserve, &tc)
+	testCaseCache = append(testCaseCache, tc)
+	return
+}
+
+// ReadWorkflow grabs a workflow for a given dir for a given filename
```
dustmop, line 182: I'd suggest changing the input parameter to be the full path to the file, and moving `filepath.Join` to the callers. It simplifies the behavior of this function in isolation, and greatly simplifies the comment describing this function.

api/api_test.go
dustmop: It's fine having these as temporary holdovers to make the tests work. Could you please add a TODO in the comment to remove them?

lib/automation.go
ramfox: Man I debate this all the time. I think the answer is to be as clear and specific as possible, so yes? It sometimes feels like overkill, but at least that way there is no confusion. 

ty for catching this

PR #1852, opened 2021-07-30
refactor(workflow): supply context to all workflow store ops

Arqu:
Adds a context to all store operations. This is required for the cloud side of implementation and probably any alternative store would also benefit from it. Useful also for later on if we want to branch this stuff out from a single node.


ramfox (approved):
LGTM! glad to have a fresh set of eyes to catch things we missed :)

PR #1853, opened 2021-07-30
fix(automation): update `Listeners` after the `Workflow` has been changed

ramfox:
Also, adjusts `CronTrigger.Advance` method to have the `nextRunStart` be one period after the current time.

### DIFF
--- a/automation/trigger/cron.go
+++ b/automation/trigger/cron.go
@@ -77,11 +77,8 @@ func (CronTrigger) Type() string { return CronType }
 // Advance sets the periodicity and nextRunStart to be ready for the next run
 func (ct *CronTrigger) Advance() error {
 	ct.periodicity = ct.periodicity.NextRep()
-	if ct.nextRunStart != nil {
-		*ct.nextRunStart = ct.periodicity.After(*ct.nextRunStart)
-		return nil
-	}
-	*ct.nextRunStart = ct.periodicity.After(NowFunc())
+	next := ct.periodicity.After(NowFunc())
+	ct.nextRunStart = &next
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


automation/trigger/cron.go
```diff
@@ -77,11 +77,8 @@ func (CronTrigger) Type() string { return CronType }
 // Advance sets the periodicity and nextRunStart to be ready for the next run
 func (ct *CronTrigger) Advance() error {
 	ct.periodicity = ct.periodicity.NextRep()
-	if ct.nextRunStart != nil {
-		*ct.nextRunStart = ct.periodicity.After(*ct.nextRunStart)
-		return nil
-	}
-	*ct.nextRunStart = ct.periodicity.After(NowFunc())
+	next := ct.periodicity.After(NowFunc())
+	ct.nextRunStart = &next
```
ramfox, line 81: So this is working under these assumptions:

1) the periodicity `Start`, stays the same, and does not change each time we
2) `nextRunStart` is the next run after the current time, it's the current time + the periodicity

We can change this to have the periodicity change it's start time each iteration, so that the periodicity is both the tracker for the period duration, and also keep the record of the next start time. 

I personally like that we separate the concerns for keeping track of when and if we have another iteration and the next time the trigger should fire, and would vote to keep it as is.

b5: Agree we should keep them separate. two reasons:

1. I like the fact that `NextRunStart` is a standard timestamp. It's much easier to code against "is the current time after this datestamp?"
2. it's not lossy. As you mentioned, the other version would have us overwriting the time the trigger was set to start.


That said, In a follow up PR we should _also_ be calling `NextIteration` on the interval so if a user types `R3/PT1D` it'll count down the three iterations accurately. But that'll require a test to make sure the that an interval that expires zeros itself out and doesn't fire again

b5 (approved):
LGTM

automation/trigger/cron.go
ramfox: It already does this in `NextRep`!

PR #1854, opened 2021-08-02
refactor(workflow): replace datasetID with initID

Arqu:
Depends on https://github.com/qri-io/qri/pull/1852


ramfox:
This looks good, in that I can't see any problems. But I don't understand the need for the change. Does `DatasetID` mean something different than `InitID`? The only place I see us talking about `InitID` is in `logbook`, in `dataset` we just call it `ID`, so it made sense to me to call it `DatasetID`. To me, having `InitID` on the `workflow` struct might seem like there may be a workflow `InitID`.

Arqu:
Refs contain InitIDs and they're ubiquitous by now. It doesn't do anything other than consistency.

ramfox (approved):
LGTM!

PR #1855, opened 2021-08-02
refactor(workflow): replace datasetID with initID

Arqu:


Arqu:
Merging straight as it's just https://github.com/qri-io/qri/pull/1854 which was a merge into a non master branch.

PR #1859, opened 2021-08-02
feat(scope): Scope has profileID. ReplaceParentContext method

dustmop:
Always add the profileID to the scope's context. This is used by automation and will soon be used by the event.Bus as well. Allow scope to detach their parent context, so that they can be used for long-running operations that should not be cancelled by their parents.

### DIFF
--- a/lib/automation.go
+++ b/lib/automation.go
@@ -182,8 +182,8 @@ func deploy(scope scope, p *DeployParams) {
 		WorkflowID: p.Workflow.ID.String(),
 	}
 
-	log.Debugw("deploy started", "payload", deployPayload)
-	sendEventAsync(scope, event.ETAutomationDeployStart, ref, deployPayload)
+	log.Debugw("deploy started", "ref", vi.SimpleRef().String(), "payload", deployPayload)
+	scope.sendEvent(event.ETAutomationDeployStart, ref, deployPayload)
 
 	rollback := true
 	defer func() {
@@ -201,7 +201,7 @@ func deploy(scope scope, p *DeployParams) {
 		}
 	}()
 
-	sendEventAsync(scope, event.ETAutomationDeploySaveDatasetStart, ref, deployPayload)
+	go scope.sendEvent(event.ETAutomationDeploySaveDatasetStart, ref, deployPayload)
 
 	saveParams := &SaveParams{
 		Ref:     vi.SimpleRef().String(),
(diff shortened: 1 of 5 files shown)


lib/scope.go
```diff
@@ -91,20 +116,38 @@ func (s *scope) CollectionSet() collection.Set {
 	return s.inst.collectionSet
 }
 
+// ReplaceParentContext returns a copy of the scope bound to a new parent
+// context
+func (s *scope) ReplaceParentContext(newParent context.Context) scope {
```
b5, line 121: I've changed this to not use `context.Background`, and instead accept a new parent param. We should almost always be re-rooting these onto the application context, so workflow execution has a chance to clean up on process shutdown

lib/scope.go
```diff
@@ -74,14 +97,16 @@ func (s *scope) Config() *config.Config {
 	return s.inst.cfg
 }
 
-// Context returns the context for this scope. Though this pattern is usually discouraged,
-// we're following http.Request's lead, as scope plays the same role. The lifetime of a
-// single scope matches the lifetime of the Context; this ownership is not long-lived
+// Context returns the context for this scope. Though this pattern is usually
+// discouraged, we're following http.Request's lead, as scope plays the same
+// role. The lifetime of a single scope matches the lifetime of the Context;
+// this ownership is not long-lived
```
b5, line 103: I've added some doc tweaks to this file. Code ruler should be set to 80 characters for comments

lib/automation.go
```diff
@@ -164,36 +163,28 @@ func (automationImpl) Deploy(scope scope, p *DeployParams) error {
...
 	}
-	log.Debugw("deploy started", "payload", deployPayload)
-	go func() {
-		if err := scope.Bus().PublishID(scope.Context(), event.ETAutomationDeployStart, ref, deployPayload); err != nil {
-			log.Debug(err)
-		}
-	}()
+
+	log.Debugw("deploy started", "ref", vi.SimpleRef().String(), "payload", deployPayload)
+	scope.sendEvent(event.ETAutomationDeployStart, ref, deployPayload)
```
b5, line 186: love the `sendEventAsync` function. I've moved it into a method of `scope` and dropped the required goroutine. IMHO go programs should use the `go` keyword to indicate async at the call site. I've also taken this opportunity to have some events _not_ fire async, which gives other subsystems a chance to fully handle the event before proceeding.

b5 (approved):
I've layered my feedback into a rebase, looks good!

lib/scope.go
dustmop: You are completely correct that this should be using `AppContext` instead of `context.Background`, and thanks for catching that bug in my original change. However, I don't think this should be `ReplaceParentContext(newParent)` and would rather it be `DetachParentContext`. Here's my reasoning:

For `DetachParentContext`, there's a very specific use-case it's covering: "this kicks off a job that runs in the parent, so my parent context's cancellation should not affect that job". For the new version `ReplaceParentContext`, I can't think of a case where some arbitrary context (aside from the AppContext) would be used to replace what is in the scope. It's too general of a tool, and so introduces a foot-gun. It makes it easy to reproduce my original error, by letting new code incorrectly call `ReplaceParentContext(context.Background())` when it should be using the `AppContext`. 

A general principle I like to follow, when there's code that looks like this:

```
obj.method(obj.thing(), other_params)
```

That is, a situation where `obj` has a `method` and some data that can be passed to that method. If it's the case that `obj.thing()` is the only valid thing to pass into `method`, and using some other value would be an error, it's preferable to just to eliminate that parameter, and have the implementation get the `thing()` in its body. In order words, do this:

```
obj.method(other_params)

...

func (o *object) method(params) {
  thing := o.thing()
  ...
}
```

This is exactly what you did with `sendEvent` in this same PR, by moving the `bus` into the implementation. There's no need to allow an arbitrary bus parameter when the one owned by the scope is the correct thing to use.

Now I could be mistaken, and we may want to have arbitrary contexts get passed to `ReplaceParentContext` at some point in the future. But (1) I can't think of that use-case off the top of my head, and (2) I'd rather avoid the potential errors until we observe evidence of that use-case.

PR #1860, opened 2021-08-03
refactor(wf_store): pass in ownerID to enable filtering

Arqu:
The current local store implementations don't really need to filter by ownerID, but other implementations might. This is just passing it in to enable the store to surface the option. Business logic whether that's used lives in `lib`.

### DIFF
--- a/automation/orchestrator.go
+++ b/automation/orchestrator.go
@@ -14,6 +14,7 @@ import (
 	"github.com/qri-io/qri/automation/run"
 	"github.com/qri-io/qri/automation/trigger"
 	"github.com/qri-io/qri/automation/workflow"
+	"github.com/qri-io/qri/base/params"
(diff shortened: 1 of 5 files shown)


automation/orchestrator.go
```diff
@@ -203,7 +203,7 @@ func (o *Orchestrator) handleContextClose(ctx context.Context) {
 // startListeners passes a list of deployed Workflows to configured trigger
 // Listeners
 func (o *Orchestrator) startListeners(ctx context.Context) error {
-	wfs, err := o.workflows.ListDeployed(ctx, -1, 0)
+	wfs, err := o.workflows.ListDeployed(ctx, "", -1, 0)
```
Arqu, line 206: Will need to work these out properly in a follow up PR.

b5 (approved):
👍

PR #1861, opened 2021-08-03
fix(jwt): Switch to golang-jwt as dependabot recommends

dustmop:


dustmop:
Fixes https://github.com/advisories/GHSA-w73w-5m7g-f7qc

Arqu (approved):
👍

PR #1862, opened 2021-08-03
fix(orchestrator): pass in OrchestratorOptions to create an Orchestrator

Arqu:


ramfox (approved):
LGTM

Future work, but just noting: depending on how we solve the `runFunc` and `applyFunc` situation on cloud, those may have to be `InstanceOptions` as well.

Arqu:
I guess this is going to move 1 PR at a time as we settle things down.

PR #1864, opened 2021-08-05
refactor(dsfs): replace qfs.AddingFS with qfs.MerkleDagStore, support save over HTTP-backed IPFS connection

b5:
reorganize the dsfs package while we're at it, moving much closer to save support without a body component. `MerkleDagStore` should be implementable over HTTP.

~~Test in cmd package still need a _lot_ of love. Once I've worked through those, I'd call this mergeable~~

Ok, everything is in a good state IMHO. I've smoke tested this locally & things are looking orderly. All the broken hashes in cmd come from fixing a bug in commit message construction: we used to erroneously detect body changes when none existed


base/dsfs/load.go
```diff
@@ -0,0 +1,302 @@
...
+	return DerefCommit(ctx, store, ds)
+}
+
+// LoadBody loads the data this dataset points to from the store
+func LoadBody(ctx context.Context, fs qfs.Filesystem, ds *dataset.Dataset) (qfs.File, error) {
+	return fs.Get(ctx, ds.BodyPath)
+}
+
+// DerefCommit derferences a dataset's Commit element if required
+// should be a no-op if ds.Structure is nil or isn't a reference
```
dustmop, line 93: ds.Structure -> ds.Commit

base/dsfs/load.go
```diff
@@ -0,0 +1,302 @@
...
+	data, err := fileBytes(fs.Get(ctx, path))
+	if err != nil {
+		log.Debug(err.Error())
+		return nil, fmt.Errorf("loading commit file: %s", err.Error())
+	}
+	return dataset.UnmarshalCommit(data)
+}
+
+// DerefMeta derferences a dataset's transform element if required
+// should be a no-op if ds.Structure is nil or isn't a reference
```
dustmop, line 118: ds.Structure -> ds.Meta

base/dsfs/load.go
```diff
@@ -0,0 +1,302 @@
...
+			log.Debug(err.Error())
+			return fmt.Errorf("loading dataset commit: %w", err)
+		}
+		cm.Path = ds.Commit.Path
+		ds.Commit = cm
+	}
+	return nil
+}
+
+// loadCommit assumes the provided path is valid
```
dustmop, line 107: Comment here should reflect what the function does from the perspective of a caller. How about "loadCommit loads and unmarshals the commit data at the given path"

Same for loadMeta, loadReadme, loadStats, loadTransform, loadStructure, loadViz

base/dsfs/load.go
```diff
@@ -0,0 +1,302 @@
...
+		log.Debug(err.Error())
+		return nil, fmt.Errorf("error loading transform raw data: %s", err.Error())
+	}
+
+	return dataset.UnmarshalTransform(data)
+}
+
+// LoadTransformScript loads transform script data from a dataset path if the given dataset has a transform script specified
+// the returned qfs.File will be the value of dataset.Transform.ScriptPath
+// TODO - this is broken, assumes file is JSON. fix & test or depricate
```
dustmop, line 239: Is this comment still relevant or is it stale? Looks like it was added by me 2 years ago when dsfs moved into base from dataset.

base/dsfs/commit.go
```diff
@@ -25,86 +24,64 @@ var Timestamp = func() time.Time {
...
 
-			if cff, ok := wfs.body.(*computeFieldsFile); ok {
-				updateScriptPaths(ds, added)
+		updateScriptPaths(dst, ds, added)
 
-				if err := confirmByteChangesExist(cff.ds, cff.prev, added, wfs.body.FullPath(), cff.sw.ForceIfNoChanges); err != nil {
-					return nil, fmt.Errorf("error saving: %w", err)
-				}
+		if err := confirmByteChangesExist(ds, prev, sw.ForceIfNoChanges, dst, added); err != nil {
+			return fmt.Errorf("saving: %w", err)
```
dustmop, line 61: If this error is seen by a user, it reads as ambiguous whether a save actually happened. `saving` is meant to be read like a category with "no changes" as an error, but it could be seen as a verb with "no changes" as purely informational. Consider changing to `saving failed: no changes`.

cmd/cmd_test.go
```diff
@@ -279,6 +279,9 @@ func TestSaveTwoComponents(t *testing.T) {
 
 	// Save a version, then same another with two components at once
 	run.MustExec(t, "qri save --file=testdata/movies/ds_ten.yaml me/test_ds")
+	got1 := run.MustLoadDataset(t, run.GetPathForDataset(t, 0))
+	fmt.Println(got1.Stats.Path)
```
dustmop: Is this print needed? Put another way, is it always going to be the same value, and can we turn it into an assertion instead?

cmd/testdata/expect/TestSaveThenOverrideMetaAndTransformAndViz.json
```diff
@@ -4,21 +4,21 @@
     "author": {
       "id": "QmeL2mdVka1eahKENjehK6tBxkkpk5dNQ1qMcgWi7Hrb4B"
     },
-    "message": "meta:\n\tupdated title\nviz added\ntransform added\nbody:\n\tadded row /",
```
dustmop, line 7: It is awesome that we are fixing this. I always wondered why this showed up in cases where it clearly wasn't true, thanks for the fix!

base/archive/zip_test.go
```diff
@@ -92,9 +92,13 @@ func TestWriteZipFullDataset(t *testing.T) {
...
-	tmppath := filepath.Join(os.TempDir(), "exported.zip")
-	// defer os.RemoveAll(tmppath)
-	t.Log(tmppath)
+	path, err := ioutil.TempDir("", "TestArchiveFullDataset")
+	if err != nil {
+		t.Fatal(err)
+	}
+	tmppath := filepath.Join(path, "exported.zip")
+	defer os.RemoveAll(tmppath)
+	// t.Log(tmppath)
```
Arqu, line 101: delete dead code

b5: yar, this comment is here as an indication on how to un-break the test when the golden file needs updating. I've added a clarifying comment

base/dsfs/commit.go
b5: agree. I just didn't like leading with "error", added in your suggestion

base/dsfs/load.go
b5: function is unused! I've removed it entirely

base/dsfs/load.go
b5: I've elected to remove the comment entirely

PR #1867, opened 2021-08-09
fix(api): make sure endpoints are of type APIEndpoint

Arqu:


Arqu:
Merging as is without review as it's a simple type omission.

PR #1869, opened 2021-08-09
chore(sql): remove sql support

b5:
sql is simply too slow to be useful in it's current state.

BREAKING CHANGE:
support for SQL directly within qri is removed


dustmop (approved):
LGTM

PR #1870, opened 2021-08-09
feat(collection): Delete only for the user who published delete event

dustmop:


event/event.go
```diff
@@ -157,10 +159,13 @@ func (b *bus) publish(ctx context.Context, typ Type, sessionID string, payload i
 		return ErrBusClosed
 	}
 
+	profileid := profile.IDFromCtx(ctx)
```
b5, line 162: crazy how big an effect this small change is!

b5 (approved):
LGTM!

PR #1871, opened 2021-08-09
fix: deploy panics, double version removal, collection clean up

ramfox:
1) update to latest qfs to fix nil ctx bug
2) ensure we have a default source for `deploy`, this extends to the deferred `Remove` that gets triggered if something goes wrong with the deploy
3) adjust the deploy path to only attempt to remove the just saved version if we have an error while deploying after the dataset has saved. If the error occurs during the save, the `Save` path has its own clean up. Also, ensure we are only removing the previous version if this current deploy has saved a previous version. If there were no changes to save, don't erroneously remove a version.
4) Collection needs to be updated when a dataset creation fails. The collection catches the dataset creation information very early in the save process, so we need to also be updated if the creation fails, so that we can remove that dataset from the collection. We are now doing this using the `ETDatasetCreateFail` event

### DIFF
--- a/lib/automation.go
+++ b/lib/automation.go
@@ -32,7 +32,7 @@ func (m AutomationMethods) Name() string {
 func (m AutomationMethods) Attributes() map[string]AttributeSet {
 	return map[string]AttributeSet{
 		"apply":    {Endpoint: qhttp.AEApply, HTTPVerb: "POST"},
-		"deploy":   {Endpoint: qhttp.AEDeploy, HTTPVerb: "POST"},
+		"deploy":   {Endpoint: qhttp.AEDeploy, HTTPVerb: "POST", DefaultSource: "local"},
 		"run":      {Endpoint: qhttp.AERun, HTTPVerb: "POST"},
 		"workflow": {Endpoint: qhttp.AEWorkflow, HTTPVerb: "POST"},
 		"remove":   {Endpoint: qhttp.AERemoveWorkflow, HTTPVerb: "POST"},
@@ -231,6 +231,7 @@ func (automationImpl) Deploy(scope scope, p *DeployParams) error {
 
 	// Because deploy runs as a background task, re-root execution context atop
 	// the application context
+	log.Debugw("app context", "ctx", scope.AppContext())
 	scope = scope.ReplaceParentContext(scope.AppContext())
 	// TODO(ramfox): if we decide that you can interact with the automation subsystem when
 	// qri connect is NOT running, we need a `wait` flag in DeployParams, that, when `true`,
@@ -251,9 +252,12 @@ func deploy(scope scope, p *DeployParams) {
 	log.Debugw("deploy started", "ref", vi.SimpleRef().String(), "payload", deployPayload)
 	scope.sendEvent(event.ETAutomationDeployStart, ref, deployPayload)
 
+	go scope.sendEvent(event.ETAutomationDeploySaveDatasetStart, ref, deployPayload)
+
+	changesSaved := false
 	rollback := true
 	defer func() {
-		if rollback {
+		if rollback && changesSaved {
 			rp := &RemoveParams{
 				Ref: ref,
 				Revision: &dsref.Rev{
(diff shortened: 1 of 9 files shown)


lib/datasets.go
```diff
@@ -975,6 +975,7 @@ func (datasetImpl) Save(scope scope, p *SaveParams) (*dataset.Dataset, error) {
 			if err := scope.Logbook().RemoveLog(ctx, ref); err != nil {
 				log.Errorf("couldn't cleanup unused reference: %q", err)
 			}
+			go scope.sendEvent(event.ETDatasetCreateFail, ref.InitID, ref.InitID)
```
dustmop, line 978: Is there a reason this needs to be a separate type of event? It appears that this does the same thing as `ETDatasetDeleteAll`, could we just reuse it here?

lib/automation.go
```diff
@@ -251,22 +252,6 @@ func deploy(scope scope, p *DeployParams) {
 	log.Debugw("deploy started", "ref", vi.SimpleRef().String(), "payload", deployPayload)
 	scope.sendEvent(event.ETAutomationDeployStart, ref, deployPayload)
 
-	rollback := true
-	defer func() {
```
dustmop, line 255: Moving this defer function later means it won't run if an error & return happens before it gets reached. That doesn't currently matter because if `Save` fails there's nothing to `Remove`, but it makes it easier to introduce bugs later on if this code path changes. Generally speaking, defer'd rollbacks should be setup early, such that they handle all failure cases that might occur.

Rather, to achieve the effect you're going for, I think it would be better to add `changesSaved := false` on the line above (which reads as "no changes have been saved yet"), keep your `if rollback && changesSaved` condition, and then after `Save` happens, make that assignment into: `changesSaved = !errors.Is(err, dsfs.ErrNoChanges)`

lib/datasets.go
ramfox: Yea.... I'm of two minds on this. This pr shows which side I ended up falling on. On one hand, adding an additional event for a course of action that is so similar (and currently has identical consequences) to an action we already have defined seems like it has the potential for bugs. On the other hand, the events talk about two different circumstances, in one case we are alerting "hey the user said remove this entire dataset", in the other "hey we failed creating a dataset", and that might actually have different consequences in the future (although, tbh I can't think of any rn).

I've been thinking of events as a way of the code path saying "hey this is an important moment, thought you should know it happened", rather than "I know I need you to do this thing and this is the event that tells you to do it", so I leaned toward making a new event that describes the circumstance. That could also be the totally wrong lens!

lib/automation.go
ramfox: good point! thought I was being *clever* lol

lib/datasets.go
b5: > although, tbh I can't think of any rn

There's your answer. If you don't know of a way things _will_ be separate in the very near future, avoid introducing new code. We should be policing the list of event types _very_ hard, and have a small number of well-defined event types that re-use wherever appropriate. This is a good example.

In a situation like this the important thing to do is to read the comment describing the event you're about to overlay onto, and check if your use case fits within the description. If not, modify the description and/or the event name itself until it accurately describes _both_ your use case & the existing uses.

dustmop:
An interesting discussion has happened around how to name, think of, and reuse events. To that end I tweaked some of the comments describing our dataset change events, but also wanted to lay out some general principles around event patterns.

The entire goal of using events is that they allow for looser coupling of subsystems. The publisher and listener shouldn't know too much about each other, all they know are event types and the associated payloads. Contrast this to a function call, wherein the caller knows explicitly what functionality it is invoking at a callsite. Events avoid this, at the loss of some clarity, but to the benefit of having less dependency on exactly how other code works and the manner in which it is implemented.

Event objects should only be as specific as they must be, from the perspective of the subsystem that is listening for them. Specifically, the listener doesn't need to all the details that caused an event to fire; as long as it can do the work needed to update its own state, it shouldn't need to concern itself with unnecessary details. This leads to a system with greater flexibility, better isolation of components, and better reuse of its parts.

PR #1873, opened 2021-08-11
fix(lib): fix deploy if a ds has no changes

ramfox:
1) fix time comparison, should be using `time.Equal` rather than `==`
2) When a dataset has no changes, the `lib.Save` command returns a `nil` dataset. However, we are still relying on the dataset to provide us with an init id. On top of that, the dataset sent over to us by the user does not need to have an init id in order for the deploy request to go through, so we are not guaranteed that the submitted dataset will have what we need. In order to ensure we have an init id, we must resolve the reference.


lib/automation_test.go
```diff
@@ -187,6 +179,21 @@ def transform(ds,ctx):
...
+	p.Workflow = gotWF
+	if err := tr.Instance.WithSource("local").Automation().Deploy(ctxCancelable, p); err != nil {
+		t.Fatalf("deploy unexpected error: %s", err)
+	}
+	cancel()
+
+	// Wait to make sure the workflow runs without error
+	errMsg = <-done
+	if errMsg != "" {
+		t.Errorf(errMsg)
```
b5, line 194: use `t.Fatal` to stop test execution here on error

lib/automation.go
```diff
@@ -289,13 +289,33 @@ func deploy(scope scope, p *DeployParams) {
...
 	}
-	changesSaved = !errors.Is(err, dsfs.ErrNoChanges)
+	if errors.Is(err, dsfs.ErrNoChanges) {
+		ds = p.Dataset
+		changesSaved = false
+		if ds.ID == "" {
+			r := &dsref.Ref{
+				Username: ds.Peername,
+				Name:     ds.Name,
+			}
```
b5, line 299: I know it says it's deprecated, but use `dsref.ConvertDatasetToVersionInfo(ds).SimpleRef()` instead. We should un-deprecate that function b/c it's totally fine to use, and even necessary in cases like this.

lib/automation.go
```diff
@@ -289,13 +289,33 @@ func deploy(scope scope, p *DeployParams) {
...
+			_, err := scope.ResolveReference(scope.Context(), r)
+			if err != nil {
```
b5, line 301: merge into one line:

```go
if _, err := scope.ResolveReference(scope.Context(), r); err != nil {
```

b5 (approved):
a few tiny nits

lib/automation.go
ramfox: fab

PR #1875, opened 2021-08-11
fix(lib): `lib.Run` should mirror `lib.Deploy` return ASAP behavior

ramfox:
So the frontend can correctly track runs, and so the frontend's http
requests do not time out if a run lasts very long, return the runID, and
only error if there is something wrong with the request params

### DIFF
--- a/lib/automation.go
+++ b/lib/automation.go
@@ -105,7 +105,6 @@ type RunParams struct {
 	Ref        string `json:"ref"`
 	InitID     string `json:"initID"`
 	WorkflowID string `json:"workflowID"`
-	RunID      string `json:"runID"`
 }
 
 // Validate returns an error if RunParams fields are in an invalid state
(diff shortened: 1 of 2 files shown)


lib/automation.go
```diff
@@ -365,7 +365,11 @@ func (automationImpl) Run(scope scope, p *RunParams) (string, error) {
 		}
 		p.WorkflowID = wf.WorkflowID()
 	}
-	return scope.AutomationOrchestrator().RunWorkflow(scope.AppContext(), workflow.ID(p.WorkflowID), p.RunID)
+	if p.RunID == "" {
```
b5, line 368: I feel like I'm blanking on a conversation we've already head. Shouldn't we _always_ be creating a new RunID here? What if I supply my own (bad) run ID?

b5: ok so either drop this on this PR, or add a TODO

PR #1876, opened 2021-08-11
refactor(logbook): support multi tenancy

b5:
making moves in the right direction for multi tenancy.


b5:
I'm going to merge this so we can keep moving. It's too big a PR to properly review

PR #1877, opened 2021-08-11
refactor(collection): rely on `ETAutomationWorkflowCreated` events

ramfox:
We were previously using the `ETAutomationDeploy` events, but unfortunately they don't actually have enough specificity for us to know exactly what is going on when we catch the events in the collection.

We need to follow the same patterns we are using to add to the collection when a dataset is added or removed: rely on events emitted at those exact times.

We need an `ETAutomationWorkflowCreated` and an `ETAutomationWorkflowRemoved` event to let us know that a workflow has been added or removed.

### DIFF
--- a/collection/collection.go
+++ b/collection/collection.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 
 	logger "github.com/ipfs/go-log"
+	"github.com/qri-io/qri/automation/workflow"
 	"github.com/qri-io/qri/base/params"
 	"github.com/qri-io/qri/dsref"
 	"github.com/qri-io/qri/event"
@@ -314,9 +315,10 @@ func (s *localSet) subscribe(bus event.Bus) {
 		event.ETRegistryProfileCreated,
 
 		// automation events
-		event.ETAutomationDeployStart,
 		event.ETAutomationWorkflowStarted,
 		event.ETAutomationWorkflowStopped,
+		event.ETAutomationWorkflowCreated,
+		event.ETAutomationWorkflowRemoved,
 
 		// fsi
 		event.ETFSICreateLink,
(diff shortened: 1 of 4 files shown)


collection/collection.go
```diff
@@ -384,49 +385,45 @@ func (s *localSet) handleEvent(ctx context.Context, e event.Event) error {
...
+	case event.ETAutomationWorkflowStopped:
+		if evt, ok := e.Payload.(event.WorkflowStoppedEvent); ok {
 			err := s.updateOneAcrossAllCollections(evt.InitID, func(vi *dsref.VersionInfo) {
-				vi.WorkflowID = evt.WorkflowID
-				if evt.Error != "" {
-					vi.WorkflowID = ""
-				}
+				vi.RunStatus = evt.Status
 			})
 			if err != nil {
```
ramfox, line 403: Okay so this is a bit odd.

1) what "kind" of run need to be represented in the collection? If a user does a save with the apply flag, should that run id and run status be captured? If so, relying on any kind of workflow or deploy event will not capture this. We should instead rely on the actually transform events. However, as of right now, we can't differentiate between a dry run or a saved run at the transform event level, which means any dry run information would be captured here as well.

2) Even if we are okay leaving out "save w/ the apply flag" type runs, we are not properly capturing any "run status" information in the collection during a deploy, because we currently disable any use of the "RunWorkflow" section, since that will cause "double runs" for dataset with transforms, since any dataset with a transform is being "applied" until we are able to save without a structure or body. Once we are able to do that, we will capture any run information in the deploy path correctly, because we will actually be using `RunWorkflow` to apply the transform.

So right now, any run information is going to be unreliable. Not sure where to settle this or what compromises we are okay with for now.

b5: Great articulation of the ambiguities of what "running" is. I vote you copy-paste this comment into a new issue & add a blurb about how we need more distinction for "dry runs" (apply command on the backend).

That said, this code moves us closer to something workable in the short term, so let's merge it & work out a proper fix in the new issue discussion

PR #1882, opened 2021-08-12
fix(ctx): add ctx to keystore

Arqu:


profile/store.go
```diff
@@ -30,37 +30,37 @@ var (
 // that must have an associated private key
 type Store interface {
 	// Owner is a single profile that represents the current user
-	Owner() *Profile
+	Owner(ctx context.Context) *Profile
```
b5, line 33: @Arqu & I chatted about this, and we both agree the `Owner` getter method should _not_ accept a context, but the setter should.

We're keeping it here for now because we don't have a strong enough guarantee that all mutations to the owner profile will make it to the profile store, so we're going to keep the context until we can get that assertion into place

b5 (approved):
LGTM

PR #1883, opened 2021-08-13
fix(collection): maintain `WorkflowID` and run information in collection

ramfox:
fixes bug that meant any time we added a new version of the dataset, we would also override the workflow id, run status, and previous run id.


dustmop (approved):
Would definitely love a test for this, but that can be in a separate PR. LGTM!

PR #1885, opened 2021-08-13
logbook write fix

Arqu:
As said before the concern with this line is whether it invokes double writes on the core side. For the cloud side this properly invokes the opstore write.


Arqu:
closed wit the new PutLog interface.

PR #1886, opened 2021-08-13
feat(logbook): add `logPutter` interface for transactional saves

b5:
this is an incomplete interface, but gets us a short-term solution for opstores that need an interface for transactional logbook updates. To make use of this an posture implementation should add a new method, where the implementation should be consistent with `MergeLog` semantics. The following should work:

```
func (s *Opstore) PutLog(ctx context.Context, l *oplog.Log) error {
  return  s.MergeLog(l)
}
```

This interface is **not** implemented for all write operations yet, but provides a mechanism to do so in follow up changes

### DIFF
--- a/logbook/logbook.go
+++ b/logbook/logbook.go
@@ -193,7 +193,7 @@ func (book *Book) Owner() *profile.Profile {
 func (book *Book) initialize(ctx context.Context) error {
 	log.Debug("intializing book", "owner", book.owner.ID.Encode())
 	// initialize owner's log of user actions
-	userActions := oplog.InitLog(oplog.Op{
+	ownerOplog := oplog.InitLog(oplog.Op{
 		Type:      oplog.OpTypeInit,
 		Model:     UserModel,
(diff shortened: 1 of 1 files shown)


logbook/logbook.go
```diff
@@ -215,11 +215,26 @@ func (book *Book) ReplaceAll(ctx context.Context, lg *oplog.Log) error {
...
+	PutLog(ctx context.Context, l *oplog.Log) error
+}
+
+// save writes the book to book.fsLocation, if a non-nil authorLog is provided
+// save tries to write a transactional update
+func (book *Book) save(ctx context.Context, authorLog *UserLog) (err error) {
+	if authorLog != nil {
+		if lp, ok := book.store.(logPutter); ok {
+			if err := lp.PutLog(ctx, authorLog.l); err != nil {
+				return err
```
Arqu, line 233: 👍

Arqu (approved):
I'm happy with this!

PR #1888, opened 2021-08-13
refactor(transform): Single dataset, move conflict check up to lib.Save

dustmop:
Instead of a pair of read/write datasets on the starlark ds object, have only a single dataset. For checking conflicts, have Apply track which components are modified by a transform script. Compare that result against the components provided to the save path, but do so in a much more obvious way, up in lib/

### DIFF
--- a/transform/apply.go
+++ b/transform/apply.go
@@ -38,9 +38,10 @@ const (
 
 // Transformer holds dependencies needed for applying a transform
 type Transformer struct {
-	appCtx context.Context
-	loader dsref.Loader
-	pub    event.Publisher
+	appCtx  context.Context
+	loader  dsref.Loader
+	pub     event.Publisher
+	changes map[string]bool
 }
 
 // NewTransformer returns a new transformer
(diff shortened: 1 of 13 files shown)


transform/apply.go
```diff
@@ -53,7 +54,7 @@ func NewTransformer(appCtx context.Context, loader dsref.Loader, pub event.Publi
 }
 
 // Apply applies the transform script to a target dataset
-func (t Transformer) Apply(
+func (t* Transformer) Apply(
```
b5, line 57: gofmt plz. Type should own the star

transform/startf/exec_step.go
```diff
@@ -24,20 +24,20 @@ var log = golog.Logger("startf")
...
-	eventsCh  chan event.Event
-	thread    *starlark.Thread
+	starCtx    *starctx.Context
+	dsLoader   dsref.Loader
+	target     *dataset.Dataset
+	globals    starlark.StringDict
+	bodyFile   qfs.File
+	eventsCh   chan event.Event
+	thread     *starlark.Thread
+	changeList map[string]bool
```
b5, line 34: again, use a `map[string]struct{}`

b5 (approved):
negative diffstat! Some small nits & then looks good to me

dustmop:
Ok, now switched to using PathMap to get the components that exist in the incoming dataset. Switched the `changeSet` (previously changeList) to use full component names now, which removes the need for the `shortNameToComponent` function.

PR #1889, opened 2021-08-16
test(collection): tests for workflow & new version creation

ramfox:
add test to prove that if one user `pulls` a dataset, it doesn't effect another user's collection

### DIFF
--- a/collection/spec/collection.go
+++ b/collection/spec/collection.go
@@ -336,54 +336,34 @@ func AssertCollectionEventListenerSpec(t *testing.T, constructor Constructor) {
 		t.Skip("TODO (b5): add a third dataset")
 	})
 
-	t.Run("user_2_automated_datasets", func(t *testing.T) {
+	t.Run("user_2_automation_datasets", func(t *testing.T) {
 		// miss piggy's collection should be empty. Kermit's collection is non-empty,
 		// proving basic multi-tenancy
 		if _, err := c.List(ctx, missPiggy.ID, params.ListAll); err == nil {
 			t.Fatalf("listing without providing any keyIDs should error: %q", err)
 		}
 
-		muppetTweetsInitID := "muppetTweetsInitID"
-		muppetTweetsName := "muppet_tweets"
+		// automated dataset tests
+		missPiggyDatasetInitID := "missPiggyDatasetInitID"
+		missPiggyDatasetName := "miss_piggy_dataset"
 
-		t.Skip("TODO (b5): need user-scoped events to make this work")
-
-		// // simulate name initialization, normally emitted by logbook
-		// mustPublish(ctx, t, bus, event.ETDatasetNameInit, event.DsChange{
-		// 	InitID:     muppetTweetsInitID,
-		// 	Username:   kermit.Peername,
-		// 	ProfileID:  kermit.ID.Encode(),
-		// 	PrettyName: muppetTweetsName,
-		// 	Info: &dsref.VersionInfo{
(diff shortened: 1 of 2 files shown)


collection/spec/collection.go
```diff
@@ -344,9 +344,9 @@ func AssertCollectionEventListenerSpec(t *testing.T, constructor Constructor) {
 		}
 
 		muppetTweetsInitID := "muppetTweetsInitID"
-		muppetTweetsName := "muppet_tweets"
+		// muppetTweetsName := "muppet_tweets"
 
-		t.Skip("TODO (b5): need user-scoped events to make this work")
+		// t.Skip("TODO (b5): need user-scoped events to make this work")
```
ramfox, line 349: not sure which events are intended to be user scoped to work, and not sure what the next lines (351-361) are supposed to be checking.

Ultimately, scoped the context to `missPiggy` for use in all subsequent events, but again, not fully sure what the intended usage should be. Should all events be ensuring correct scope in the collection, or just certain ones?

dustmop: I am also not sure what exactly the test is doing and what behavior it expects. I missed looking to this specific test when doing my recent scoped events work, but can at least provide what might be helpful info: only `ETDatasetDeleteAll` and `ETDatasetPulled` are currently using info from the context, which they obtain from the Event object creation in `publish` in event/event.go.

Perhaps others should also be doing the same thing, but I'm not totally clear on the design of collection and which events are actually intended to be global (things like "rename" probably should be?). I kept the scope of my recent work small and only modified handler that explicitly had TODOs to use scoped user info from context. Except `pushed` because that is (1) actually on the receiving end and therefore is not even using user-specific code paths and (2) not currently in use anywhere so not testable.

b5: great question. The thing we're trying to test: when miss piggy pulls a dataset, It shouldn't end up in Kermit's list of datasets.

collection/spec/collection.go
```diff
@@ -343,66 +343,75 @@ func AssertCollectionEventListenerSpec(t *testing.T, constructor Constructor) {
 			t.Fatalf("listing without providing any keyIDs should error: %q", err)
 		}
 
+		scopedCtx := profile.AddIDToContext(ctx, missPiggy.ID.Encode())
```
dustmop, line 346: A recommendation for doing this particular piece differently: if you look at the test user_3_pull_dataset, `scopedCtx` gets constructed inside a plain `{ ... }` block (no explicit control flow) and then immediately is used, rather than being kept as a local var that gets reused in the rest of the test. This accomplishes two subtle things: (1) better simulates the way this object is used in real code and (2) really makes references to `scopedCtx` stand out. There's no need to carefully read other parts of the code to see whether or not `ctx` or `scopedCtx` is being used for each call, because the plain block guarantees that the scope is not visible outside of it.

dustmop (commented):
There is an existing test is this file that covers the case you're describing: user_3_pull_dataset has a user pull a dataset and then validates that the dataset is in their collect. However, it doesn't do the converse: check that the dataset is *not* is an uninvolved user's collection. Furthermore, this test that you're modifying does more, specifically it is creating workflows and changing changing head commits, those facts should be part of the description of the test.

ramfox:
Okay, actually moved around the tests so they better matched the given descriptions! `automation_datasets` now only really deals with workflow datasets & the `pull_datasets` test makes sure that a 2nd user's collection is not affected when the first user pulls a dataset.

b5 (approved):
LGTM!

PR #1890, opened 2021-08-17
feat(collection): expand `WritableSet` interface, add option to pass in custom `WritableSet`

ramfox:
Because we need to repeat many of these handler/event patterns in `cloud`, we are dividing up the `Collection` to make it more reusable. 

The `Collection` is now a struct that contains a `WritableSet`. The `Collection` is responsible for listening to events and calling various `Put`-like functions on the `WriteableSet` when certain events are handled. 

To keep things moving, rather than refactor a ton about the `collection` package, first I'm keeping these `Put`-like functions as close as possible to the methods that already existed on `localSet`. These can most likely be narrowed in future prs.

The `spec` package as been remove, and the tests moved into the `collection` packages itself. Although, `AssertWritableCollectionSpec` with some tweaking and additional tests should probably be moved back into a `spec` package, especially once the interface settles.

### DIFF
--- a/automation/orchestrator.go
+++ b/automation/orchestrator.go
@@ -68,8 +68,7 @@ type Orchestrator struct {
 	running      bool
 }
 
-// NewOrchestrator constructs an orchestrator, whose only responsibility, right
-// now, is to create a workflow store, run store, & listen for trigger events
+// NewOrchestrator constructs an orchestrator
 func NewOrchestrator(ctx context.Context, bus event.Bus, runFactory RunFactory, applyFactory ApplyFactory, opts OrchestratorOptions) (*Orchestrator, error) {
 	log.Debugw("NewOrchestrator", "opts", opts)
 
--- a/collection/collection.go
+++ b/collection/collection.go
@@ -28,50 +28,36 @@ var (
 	log         = logger.Logger("collection")
 )
 
-// Collection is a list of datasets, with each list scoped to a user profile
-// A user's collection may consist of datasets they have created and datasets
-// added from other users. Collections are the canonical source of truth for
-// listing a user's datasets in a qri instance. While a collection owns the list
-// the fields in a collection item are cached values gathered from other
-// subsystems, and must be kept in sync as subsystems mutate their state.
-type Collection struct {
-	set WritableSet
+// SetMaintainer maintains a set of collections, each scoped to a user profile
+// It keeps the collection set in sync as subsystems mutate their state.
+type SetMaintainer struct {
+	set Set
 }
 
-// NewCollection constructs a collection
-func NewCollection(ctx context.Context, bus event.Bus, set WritableSet) (*Collection, error) {
-	log.Debugw("NewCollection")
+// NewSetMaintainer constructs a SetMaintainer
+func NewSetMaintainer(ctx context.Context, bus event.Bus, set Set) (*SetMaintainer, error) {
+	log.Debugw("NewSetMaintainer")
 
 	if bus == nil {
 		return nil, fmt.Errorf("bus of type event.Bus required")
 	}
 	if set == nil {
-		return nil, fmt.Errorf("set of type collection.WritableSet required")
+		return nil, fmt.Errorf("set of type collection.Set required")
 	}
-	c := &Collection{
+	c := &SetMaintainer{
 		set: set,
 	}
 	c.subscribe(bus)
 	return c, nil
 }
 
-// List lists a collection as a slice of `dsref.VersionInfo`s, scoped to a user
-func (c *Collection) List(ctx context.Context, pid profile.ID, lp params.List) ([]dsref.VersionInfo, error) {
-	return c.set.List(ctx, pid, lp)
+// Set returns the SetMaintainer's underlying `Set`
+func (sm *SetMaintainer) Set() Set {
+	return sm.set
 }
 
-// Get fetches dataset information as a `dsref.VersionInfo`, scoped by user
-func (c *Collection) Get(ctx context.Context, pid profile.ID, initID string) (dsref.VersionInfo, error) {
-	return c.set.Get(ctx, pid, initID)
-}
-
-// Put adds a list of `dsref.VersionInfo`s to a user's collection
-func (c *Collection) Put(ctx context.Context, pid profile.ID, item ...dsref.VersionInfo) error {
-	return c.set.Put(ctx, pid, item...)
-}
-
-func (c *Collection) subscribe(bus event.Bus) {
-	bus.SubscribeTypes(c.handleEvent,
+func (sm *SetMaintainer) subscribe(bus event.Bus) {
+	bus.SubscribeTypes(sm.handleEvent,
 		// save events
 		event.ETDatasetNameInit,
 		event.ETDatasetCommitChange,
@@ -95,7 +81,7 @@ func (c *Collection) subscribe(bus event.Bus) {
 	)
 }
 
-func (c *Collection) handleEvent(ctx context.Context, e event.Event) error {
+func (sm *SetMaintainer) handleEvent(ctx context.Context, e event.Event) error {
 	switch e.Type {
 	case event.ETDatasetNameInit:
 		if vi, ok := e.Payload.(dsref.VersionInfo); ok {
@@ -104,7 +90,7 @@ func (c *Collection) handleEvent(ctx context.Context, e event.Event) error {
 				log.Debugw("parsing profile ID in name init", "err", err)
 				return err
 			}
-			if err := c.set.Put(ctx, pid, vi); err != nil {
+			if err := sm.set.Add(ctx, pid, vi); err != nil {
 				log.Debugw("putting one:", "err", err)
 				return err
 			}
@@ -113,7 +99,7 @@ func (c *Collection) handleEvent(ctx context.Context, e event.Event) error {
 	case event.ETDatasetCommitChange:
 		// keep in mind commit changes can mean added OR removed versions
 		if vi, ok := e.Payload.(dsref.VersionInfo); ok {
-			c.set.UpdateOneAcrossAllCollections(ctx, vi.InitID, func(m *dsref.VersionInfo) {
+			sm.set.UpdateEverywhere(ctx, vi.InitID, func(m *dsref.VersionInfo) {
 				// preserve fsi path
 				vi.FSIPath = m.FSIPath
 				// preserve workflow id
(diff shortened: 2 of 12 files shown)


automation/orchestrator.go
```diff
@@ -68,8 +68,7 @@ type Orchestrator struct {
 	running      bool
 }
 
-// NewOrchestrator constructs an orchestrator, whose only responsibility, right
-// now, is to create a workflow store, run store, & listen for trigger events
+// NewOrchestrator constructs an orchestrator
```
ramfox, line 71: this is weird commentary from when i was iteratively adding bits to the orchestrator

automation/orchestrator.go
```diff
@@ -68,8 +68,7 @@ type Orchestrator struct {
 	running      bool
 }
 
-// NewOrchestrator constructs an orchestrator, whose only responsibility, right
-// now, is to create a workflow store, run store, & listen for trigger events
```
ramfox, line 72: this is weird commentary from when i was iteratively adding bits to the orchestrator

collection/collection.go
```diff
@@ -28,22 +28,217 @@ var (
...
 	Delete(ctx context.Context, profileID profile.ID, initIDs ...string) error
+	// UpdateUsernameAcrossAllCollections updates all instances of a username
+	// in all collections
+	UpdateUsernameAcrossAllCollections(ctx context.Context, changedUsernameProfileIDString, newUsername string) error
+	// AddOneAcrossAllCollections adds a new dataset's information to all
+	// collections
+	AddOneAcrossAllCollections(ctx context.Context, add dsref.VersionInfo) error
+	// UpdateOneAcrossAllCollections updates a dataset's information in all
+	// collections
+	UpdateOneAcrossAllCollections(ctx context.Context, initID string, mutate func(vi *dsref.VersionInfo)) error
```
ramfox, line 237: These `AcrossAllCollections` methods are part of what makes this interface expansive. Pros to keeping these in the interface itself: no need to lock & unlock for each iteration. The con is new implementations need to implement a greater number methods.

collection/collection.go
```diff
@@ -28,22 +28,217 @@ var (
...
+			if err := c.set.DeleteFromCollection(ctx, profileID, initID); err != nil {
+				log.Debugw("removing dataset from collection", "profileID", profileID, "initID", initID, "err", err)
+			}
+		}
+	case event.ETRegistryProfileCreated:
+		if p, ok := e.Payload.(event.RegistryProfileCreated); ok {
+			c.set.UpdateUsernameAcrossAllCollections(ctx, p.ProfileID, p.Username)
+		}
+	case event.ETDatasetPushed:
+		// TODO(b5): need user-scoped events for this so we can know *who* pulled
```
dustmop, line 145: I realize this is not newly written code, just moved from elsewhere, but this TODO has been on my mind. Note that this event is published on the receiver's end of a push. Since we know only local nodes use events and cloud doesn't, and currently non-cloud nodes can't push to each other's collections, I think it would be better to just log an error here and not even call `AddOneAcrossAllCollections`. Once that situation changes, it'll be easier to figure out what the handler should be doing here if it showing an error that is loud and explicit, rather than having to figure out what this incorrect code is doing.

collection/collection.go
```diff
@@ -28,22 +28,217 @@ var (
...
+
+// Set maintains lists of datasets, with each list scoped to a user profile.
 type Set interface {
 	List(ctx context.Context, pid profile.ID, lp params.List) ([]dsref.VersionInfo, error)
+	Get(ctx context.Context, pid profile.ID, initID string) (dsref.VersionInfo, error)
 }
 
 // WritableSet is an extension interface on Set that adds methods for
 // adding and removing items
 type WritableSet interface {
```
dustmop, line 223: As mentioned when chatting, I'm not sure if we gain anything by having WritableSet be a different interface, rather than just having Set contain methods for adding. If a Set is meant to be read-only, then those methods could just return errors. This suggestion would of course lose compile-time guarantees, causes error handling to be needed at runtime, but that's also true if lots of places in code need to type convert Sets into WritableSets.

lib/lib.go
```diff
@@ -88,7 +88,7 @@ type InstanceOptions struct {
 	keyStore                key.Store
 	profiles                profile.Store
 	bus                     event.Bus
-	collectionSet           collection.Set
+	collectionSet           collection.WritableSet
```
dustmop, line 91: Lines like this make me especially think that `collection.Set` should just have methods for adding rather than having a separate `collection.WritableSet` interface.

collection/collection.go
```diff
@@ -75,7 +270,7 @@ var (
 // persist collections, serializing to a directory of "profileID.json" files,
 // with one for each profileID in the set of collections. providing an empty
 // repoDir value will create an in-memory collection
-func NewLocalSet(ctx context.Context, bus event.Bus, repoDir string, options ...LocalSetOptionFunc) (Set, error) {
+func NewLocalSet(ctx context.Context, repoDir string, options ...LocalSetOptionFunc) (WritableSet, error) {
```
dustmop, line 273: I am a fan of removing the Bus, since this implementation doesn't need it, and returning a WritableSet to avoid the need for a type cast later on.

lib/lib.go
```diff
@@ -624,12 +623,20 @@ func NewInstance(ctx context.Context, repoPath string, opts ...Option) (qri *Ins
...
 			o.MigrateRepo = inst.repo
 		})
 		if err != nil {
 			return nil, err
 		}
+		o.collectionSet = set
+	}
+
+	if o.collectionSet != nil {
+		inst.collection, err = collection.NewCollection(ctx, inst.bus, o.collectionSet)
```
dustmop, line 636: Ok, I was confused before but I think I can see what's going on now. I'll leave a comment at top-level concerning the biggest thing I think we need to sort out.

lib/scope.go
```diff
@@ -119,8 +119,8 @@ func (s *scope) AppContext() context.Context {
 }
 
 // Collection returns the Instance collection subsystem
-func (s *scope) CollectionSet() collection.Set {
-	return s.inst.collectionSet
+func (s *scope) CollectionSet() *collection.Collection {
```
dustmop, line 122: Method is named `Collection.Set` but returns a `collection.Collection` rather than a `collection.Set`.

dustmop (commented):
Ok, I was somewhat confused while reading this, and I think its due to my tendency to focus on names. But I think it's important to do so, because it's the biggest part of collective understanding and communication around how our codebase works. Let me summarize:

1) A `collection` is a list of dsrefs that a user has or cares about, that they build using `save` and `pull`, and can see by calling `list`.
2) A `set` is a group of collections that exist on a single qri node. Locally these are stored in `~/.qri/collections`. A single collection from a set can be obtained using `set.List` and when a set updates some data it can do so across all collections
3) A `collection.Collection` wraps a `set` and also (locally) handles events that it routes to the set to update data within.

We gotta rename either (1) or (3), otherwise it'll cause endless confusion. For example, maybe (3) becomes a `collection.Coordinator` or `collection.Maintainer` or something like that. Or perhaps (1) becomes a `reflist` to describe what the return value is. I'm not a huge fan of that because I think we want to be able to talk about "what datasets are in a user's collection" and that is invoking the meaning behind (1) rather than (2).

I'm not super familiar with how this is going to be used in cloud, and obviously the design of this is going to influence that. I'm going to think on it a bit more and then post some thoughts of how to better arrange these names and concepts.

dustmop:
Alright, I have some suggestions on how this could work. Feel free to update / tweak / modify / ignore as needed.

1) The newly introduced wrapper called `collection.Collection` should be renamed to `SetMaintainer` (I am not at all tied to this name, there is probably something better) as what it is doing is hooking up the event bus to a Set. Move it out of the `collection` package and into some other directory (maybe `lib/collection_set.go`? I am not sure). This is done because `SetMaintainer` is one specific implementation rather than generic collection functionality. After that, the `collection` package does not know about `event` at all, which is nice for cloud.

2) lib/lib.go constructs this `SetMaintainer` and holds onto it, but lib/scope.go should just have a method called `CollectionSet` that returns a `collection.Set`. The users of scope don't know that their set is subscribed to events.

3) Combine the `Set` and `WritableSet` interfaces to get this:

```
// Set is a grouping of all users' collections that this node knows of
type Set interface {
	// List the collection of a single user
	List(ctx context.Context, pid profile.ID, lp params.List) ([]dsref.VersionInfo, error)
	// Get info about a single dataset in a single user's collection
	Get(ctx context.Context, pid profile.ID, initID string) (dsref.VersionInfo, error)
	// Put adds a list of `dsref.VersionInfo`s to a user's collection
	PutList(ctx context.Context, pid profile.ID, item []dsref.VersionInfo) error
	// Add adds a single dataset to a user's collection
	Add(ctx context.Context, pid profile.ID, add dsref.VersionInfo) error
	// RenameUsername changes a user's name
	RenameUsername(ctx context.Context, pid profile.ID, newUsername string) error
	// UpdateEverywhere updates a dataset in all collections that contain it
	UpdateEverywhere(ctx context.Context, initID string, mutate func(vi *dsref.VersionInfo)) error
	// Delete removes a single dataset from a single user's collection
	Delete(ctx context.Context, pid profile.ID, removeID string) error
}
```

changes of note:
* `Put` changed to `PutList` to make it explicit it is bulk adding refs. We already have `Add` (renamed from `AddOneToCollection`) for a single ref, might as well make the difference explicit
* `AddOneAcrossAllCollections` is removed, since that is only be used in response to the `ETDatasetPushed` event, which I noted earlier is buggy behavior that isn't needed
* `Delete` used to take a list of initIDs, but was not being called by anything, so that behavior isn't needed. Replace it with `DeleteFromCollection` which only deletes one.
* Shortened the other names while still trying to accurately describe their behavior.
* Method parameters should prefer `profile.ID` rather than strings, but the implementations should still validate these by calling `pid.Validate()` to avoid storing corrupted data.

4) Note that this interface matches pretty closely to how SQL operates. Hopefully this means it ends up being a good match for how cloud wants to operate.

One final mention, operations like `RenameUsername`, which used to be `UpdateUsernameAcrossAllCollections` makes me think we should be denormalizing data. That is, each reference in the collection set shouldn't contain usernames themselves. Rather, they should be using profileIDs that are then mapped using another table or storage mechanism to the usernames. Then a username rename means modifying only a single piece of data rather than iterating over every element to modify just one field. This is why I avoid the `Everywhere` suffix in the method name. I don't think this is how things work right now, but I do think it's something we should aspire to.

Arqu:
I'm down with all of this, will leave final call to Dustin as he reviewed initially.
I love the cleanup, and particularly the transition to profile.ID. Clouds arch nemesis is working off of usernames and refs.
As far as I can see and compare, this should be a pretty smooth transition on the cloud side.

collection/collection.go
```diff
@@ -28,22 +28,208 @@ var (
...
+// be kept in sync as subsystems mutate their state
+type Set interface {
+	// List the collection of a single user
+	List(ctx context.Context, pid profile.ID, lp params.List) ([]dsref.VersionInfo, error)
+	// Get info about a single dataset in a single user's collection
+	Get(ctx context.Context, pid profile.ID, initID string) (dsref.VersionInfo, error)
+	// PutList adds a list of `dsref.VersionInfo`s to a user's collection
+	PutList(ctx context.Context, pid profile.ID, items []dsref.VersionInfo) error
+	// Add adds a single dataset to a user's collection
+	Add(ctx context.Context, pid profile.ID, add dsref.VersionInfo) error
```
b5, line 226: Not a fan of this change. Drop `PutList` and flip `Add` to accept a variadic arg:

```go
// Add adds one or more versionInfos to a user's collection
Add(ctx context.Context, pid profile.ID, add ...dsref.VersionInfo) error`
```

Add should batched into a single transaction under the hood, having more than one way to add a dataset means more than one place where bugs can arise, and adds one more requirement to implementers

collection/collection.go
```diff
@@ -28,22 +28,208 @@ var (
...
+	// RenameUsername changes a user's name
+	RenameUsername(ctx context.Context, pid profile.ID, newUsername string) error
```
b5, line 228: `RenameUser`

collection/migrate.go
```diff
@@ -12,11 +11,6 @@ import (
 // MigrateRepoStoreToLocalCollectionSet constructs a local collection.Set from
 // legacy repo data
 func MigrateRepoStoreToLocalCollectionSet(ctx context.Context, s Set, r repo.Repo) error {
-	ws, ok := s.(WritableSet)
-	if !ok {
-		return errors.New("cannot migrate to CollectionSet. Provided CollectionSet is not writable")
-	}
```
b5, line 18: gotta love the little wins

b5 (commented):
fantastic set of changes. I've requested some edits, mostly superficial we-should-do-this-while-we're-here type stuff

b5:
@dustmop I've flipped the assignee to you, lmk if you want to work through deets!

PR #1891, opened 2021-08-17
fix(logbook): pass branch writes through logPutter

Arqu:


b5 (approved):
okok rubber stamp! merge!

PR #1893, opened 2021-08-19
fix(run): add ctx to run.Store

Arqu:
Also added an `owner profile.ID` to `ListByStatus` as cloud will have to scope it down and from a multi tenancy standpoint makes sense to control for it. And I want to avoid hacks we used to do where we magically pull out the relevant user from context.

### DIFF
--- a/automation/run/store.go
+++ b/automation/run/store.go
@@ -1,6 +1,7 @@
 package run
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -10,6 +11,7 @@ import (
 	"github.com/qri-io/qri/automation/workflow"
 	"github.com/qri-io/qri/base/params"
 	"github.com/qri-io/qri/event"
+	"github.com/qri-io/qri/profile"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


automation/run/store.go
```diff
@@ -25,26 +27,26 @@ var (
...
 	// workflow.ID
-	GetStatus(wid workflow.ID) (Status, error)
+	GetStatus(ctx context.Context, wid workflow.ID) (Status, error)
 	// ListByStatus returns a list of run.State entries with a given status
 	// looking only at the most recent run of each Workflow
-	ListByStatus(s Status, lp params.List) ([]*State, error)
+	ListByStatus(ctx context.Context, owner profile.ID, s Status, lp params.List) ([]*State, error)
 	// Shutdown closes the store
-	Shutdown() error
+	Shutdown(ctx context.Context) error
```
b5, line 49: wait, do we _really_ need a context on shutdown? Strikes me as a bad idea

b5 (commented):
I'm not _so_ into adding contexts everywhere, but see how this is an important stop-gap in the short term. If we can nix it from the shutdown method, looks good

automation/run/store.go
Arqu: Probably not, cloud should have a separate shutdown mechanic anyways (and does).

Arqu:
Updated the PR with a bit more stuff, the `step.Duration` can exceed a 32bit int and was giving me some trouble, so aligned that and also removed `ctx` from the `AddEvent` handler as that needs to run on a separate context from the event caller.

b5 (approved):

Issue #1823, opened 2021-06-30
Need stats in /get


chriswhong:
The frontend needs stats in `/get` to populate the Dataset Headers

ramfox:
We will be adding a `/collection/get` endpoint, that takes a `Ref` or `InitID`, and returns a `VersionInfo` with the additional fields needed to populate the `DatasetHeader`

Check out this comment for more details: https://github.com/qri-io/qri/issues/1880#issuecomment-910422967

PR #1900, opened 2021-09-02
refactor(*): add "aggregate" fields to `VersionInfo` & listen for cloud event in `collection`

### LINKED ISSUE #1880: discussion: dataset history and cloud metadata & its storage
```go
type Uberdata struct {
  InitID string
  WorkflowID string
  Title string // title at HEAD
  RunCount int // number of runs
  HeadSize int64
  CommitCount int
  DownloadCount int
  FollowerCount int
}
```

Other field candidates (based on what we expect on the dataset page)
- `Automated` boolean (can be infered if there is a workflowID)
- `Private` boolean (privacy always starts a sprawling discussion since we have many definitions of what "private" will mean at some point in time)

1) Does this belong in the `collection`
Our easiest bet would be to stick this struct in the collection. My main concern is that for the most part the `collection` is a cache for information we have available stored in other places. If the `collection` is only intended to be a cache, then we need to make sure all of these other pieces of information are safely stored elsewhere. This isn't a problem for fields that exist in other stores, but potentially an issue for the cloud related fields.

2) Which leads me to a Cloud related fields question, what do these mean in a local context? Are they zero? `FollowerCount` and `DownloadCount` may not make sense locally, but for a remote (non-cloud remote), these might be things we can keep track of. If so, we need to have systems that track and store these as well. Is that just the `Uberdata` store, or should we be expecting a flurry of `remote` systems that note when someone follows or downloads a dataset.

4) What are we naming this thing? Candidates so far: `AggregateData`, `Uberdata`, `DatasetHistoryData`, `DatasetAggData`

5) Another cloud related feature, but does something like `Issue Count` also belong here, when issues are implemented?

ramfox:
closes #1880 #1823

### DIFF
--- a/cmd/stringers.go
+++ b/cmd/stringers.go
@@ -138,12 +138,12 @@ func (vis versionInfoStringer) String() string {
 	} else {
 		fmt.Fprintf(w, ", %d errors", vis.NumErrors)
 	}
-	if vis.NumVersions == 0 {
+	if vis.CommitCount == 0 {
 		// nothing
-	} else if vis.NumVersions == 1 {
-		fmt.Fprintf(w, ", %d version", vis.NumVersions)
+	} else if vis.CommitCount == 1 {
+		fmt.Fprintf(w, ", %d version", vis.CommitCount)
 	} else {
-		fmt.Fprintf(w, ", %d versions", vis.NumVersions)
+		fmt.Fprintf(w, ", %d versions", vis.CommitCount)
 	}
 
 	fmt.Fprintf(w, "\n\n")
--- a/event/remote.go
+++ b/event/remote.go
@@ -46,6 +46,18 @@ const (
 	// (logbook + versions) remove completed
 	// payload will be a RemoteEvent
 	ETRemoteClientRemoveDatasetCompleted = Type("remoteClient:RemoveDatasetCompleted")
+	// ETRemoteDatasetFollowed indicates that the dataset has been followed by a user
+	// payload is an `InitID` string
+	ETRemoteDatasetFollowed = Type("remote:DatasetFollowed")
+	// ETRemoteDatasetUnfollowed indicates that the dataset has been unfollowed by a user
+	// payload is an `InitID` string
+	ETRemoteDatasetUnfollowed = Type("remote:DatasetUnfollowed")
+	// ETRemoteDatasetIssueOpened indicates that an issue has been opened for this dataset
+	// payload is an `initID` string
+	ETRemoteDatasetIssueOpened = Type("remote:DatasetIssueOpened")
+	// ETRemoteDatasetIssueClosed indicates that an issue has been closed for this dataset
+	// payload is an `initID` string
+	ETRemoteDatasetIssueClosed = Type("remote:DatasetIssueClosed")
 )
 
 // RemoteEvent encapsulates the push / pull progress of a dataset version
--- a/lib/collection.go
+++ b/lib/collection.go
@@ -35,6 +35,7 @@ func (m CollectionMethods) Attributes() map[string]AttributeSet {
 	return map[string]AttributeSet{
 		"list":        {Endpoint: qhttp.AEList, HTTPVerb: "POST"},
 		"listrawrefs": {Endpoint: qhttp.DenyHTTP},
+		"get":         {Endpoint: qhttp.AECollectionGet, HTTPVerb: "POST"},
 	}
 }
 
@@ -57,6 +58,29 @@ func (m CollectionMethods) ListRawRefs(ctx context.Context, p *EmptyParams) (str
 		return res, err
 	}
 	return "", dispatchReturnError(got, err)
+}
+
+// CollectionGetParams defines parameters for looking up the head of a dataset from the collection
+type CollectionGetParams struct {
+	Ref    string `json:"ref"`
+	InitID string `json:"initID"`
+}
+
+// Validate returns an error if CollectionGetParams fields are in an invalid state
+func (p *CollectionGetParams) Validate() error {
+	if p.Ref == "" && p.InitID == "" {
+		return fmt.Errorf("either ref or initID are required")
+	}
+	return nil
+}
+
+// Get gets the head of a dataset as a VersionInfo from the collection
+func (m CollectionMethods) Get(ctx context.Context, p *CollectionGetParams) (*dsref.VersionInfo, error) {
+	got, _, err := m.d.Dispatch(ctx, dispatchMethodName(m, "get"), p)
+	if res, ok := got.(*dsref.VersionInfo); ok {
+		return res, err
+	}
+	return nil, dispatchReturnError(got, err)
 }
 
 // collectionImpl holds the method implementations for CollectionMethods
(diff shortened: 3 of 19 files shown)


repo/ref/convert.go
```diff
@@ -39,7 +39,7 @@ func ConvertToVersionInfo(r *DatasetRef) dsref.VersionInfo {
 		build.CommitMessage = ds.Commit.Message
 	}
 	if ds != nil {
-		build.NumVersions = ds.NumVersions
+		build.CommitCount = ds.NumVersions
```
ramfox, line 42: Okay so we've changed NumVersions on verison info to CommitCount, but it's still NumVersions on the dataset.Dataset struct itself.

Now that collection is doing what it is supposed to be doing, do we need NumVersions on the dataset anymore? I assumed that was a bandaid until we got something like collection up and running.

lib/collection.go
```diff
@@ -142,7 +166,7 @@ func (collectionImpl) List(scope scope, p *ListParams) ([]dsref.VersionInfo, err
 			infos[i] = reporef.ConvertToVersionInfo(&r)
 		}
 	} else if listProfile.Peername == "" || reqProfile.Peername == listProfile.Peername {
-		infos, err = base.ListDatasets(scope.Context(), scope.Repo(), p.Term, restrictPid, p.Offset, p.Limit, p.Public, p.ShowNumVersions)
+		infos, err = base.ListDatasets(scope.Context(), scope.Repo(), p.Term, restrictPid, p.Offset, p.Limit, p.Public, true)
```
ramfox, line 169: lines 123-191 here are fallback in case we don't have a collection.

Do we still need this? Or is the idea that we should really just be relying on the collection /list?

lib/collection.go
```diff
@@ -198,3 +222,19 @@ func (collectionImpl) ListRawRefs(scope scope, p *EmptyParams) (string, error) {
 	}
 	return base.RawDatasetRefs(scope.Context(), scope.ActiveProfile().ID, scope.CollectionSet())
 }
+
+// Get gets the head of a dataset as a VersionInfo from the collection
+func (collectionImpl) Get(scope scope, p *CollectionGetParams) (*dsref.VersionInfo, error) {
```
ramfox, line 227: Name? 

`GetFromCollection`?
`Header`?
`DatasetDetails`?
`GetDatasetInfo`?

lib/http/api.go
```diff
@@ -33,6 +33,8 @@ const (
 
 	// AEList lists all datasets in your collection
 	AEList APIEndpoint = "/list"
+	// AECollectionGet returns info on a head dataset in your collection
+	AECollectionGet APIEndpoint = "/collection/get"
```
ramfox, line 37: Okay, so, wanted to get some feedback on this endpoint name.

Right now, it would be the only endpoint prefixed with `/collection`, since the main collection endpoint is just `/list`

Some other ideas:
`/header`
`/details`
`/getfromcollection`

event/remote.go
```diff
@@ -46,6 +46,18 @@ const (
...
+	ETRemoteDatasetFollowed = Type("remote:DatasetFollowed")
+	// ETRemoteDatasetUnfollowed indicates that the dataset has been unfollowed by a user
+	// payload is an `InitID` string
+	ETRemoteDatasetUnfollowed = Type("remote:DatasetUnfollowed")
+	// ETRemoteDatasetIssueOpened indicates that an issue has been opened for this dataset
+	// payload is an `initID` string
+	ETRemoteDatasetIssueOpened = Type("remote:DatasetIssueOpened")
+	// ETRemoteDatasetIssueClosed indicates that an issue has been closed for this dataset
+	// payload is an `initID` string
+	ETRemoteDatasetIssueClosed = Type("remote:DatasetIssueClosed")
```
ramfox, line 60: @Arqu took these from issue #1880, but wanted to get a double check on if these work for cloud, if you think we should rename, etc!

repo/ref/convert.go
dustmop: Would love to get rid of `dataset.NumVersions`. It was added way back when (Feb 2019) here: https://github.com/qri-io/dataset/pull/177, in order to support `qri list --num-versions`, but these newer pieces of tech like collections and VersionInfo are a better solution.

lib/collection.go
dustmop: It's inefficient for `base.ListDatasets` to get the number of versions, it has to traverse the `PreviousPath` field for each version. That's why it was made opt-in when calling `qri list --num-versions`. Could even see defaulting this value to `false` instead, as long as no api consumer depends on this behavior. It's fine either way, since this is hopefully going away soon as we move entirely over to collections.

cmd/stringers.go
```diff
@@ -133,17 +133,17 @@ func (vis versionInfoStringer) String() string {
 	} else {
 		fmt.Fprintf(w, ", %d entries", vis.BodyRows)
 	}
-	if vis.NumErrors == 1 {
+	if vis.CommitCount == 1 {
```
dustmop, line 136: Changing `NumErrors` to `CommitCount` seems like a mistake

lib/collection.go
dustmop: I'm good with `Get`, as it is doing an analogous action to `datasets.Get`. The only other option I would go for might be `Head`, as long as the collection is only ever tracking head versions, and not used to provided historical information as well.

If hypothetically we landed on a struct name like `AggInfo` for "aggregated information" I could see changing this to a method name to match, but that's a discussion for later. In general, if a method name sounds like it is referring to a specific type of structure, it should return that same kind of structure, so that it reads like you're getting "the collection aggInfo for this InitID". Similar reasoning for why using `Head` might make sense.

lib/http/api.go
dustmop: See my earlier comment about the method name for the same answer here (spoiler: `/collection/get` is good)

event/remote.go
Arqu: These look great, cloud is a bit out of date on the formatting/naming so have some clean up to do. Made an issue to keep it on my plate.

repo/ref/convert.go
ramfox: filed issue https://github.com/qri-io/dataset/issues/267

lib/collection.go
ramfox: What's the deciding line that we have officially moved everything over to collections?

lib/collection.go
ramfox: okay in keeping with speed/moving forward, going to keep this as `Get` and we can rename in a later pr when the feature is more settled.

lib/http/api.go
ramfox: 👍 going to keep for now, we can change in a future pr if we rename to something we find more sensible

event/remote.go
ramfox: got the verbal okay from @Arqu that these looked good

lib/collection.go
dustmop: Definitely not using dscache at all, and I think repo will go away once we're confident that collection is working as intended. To simplify my previous comment, I think this parameter should be `false` instead of `true`.

dustmop (approved):
LGTM

PR #1901, opened 2021-09-02
refactor(startf): use stepfile to convert flat files to steps

b5:
requires https://github.com/qri-io/dataset/pull/266 first, but this moves us over to a single codepath for all starlark transform execution.


dustmop:
Even though it's still draft, looks good at the moment, and am very glad to see these two code paths get combined.

b5:
superseded by #1905

PR #1902, opened 2021-09-03
refactor(transform): add `transform.Mode` enum to track "apply" modes

ramfox:
We can use the `Mode` field in the `event.TransformLifecycle`, `event.TransformStepLifecyle`, and `event.TransformMessage` to track whether or not a transform is an "apply" run, or a run whose output is not saved, or a "commit" run, or a run whose output is saved as a new dataset version.

We can use these events to track "RunCount" on a dataset in the collection.

----

Added `Mode` to `TransformLifecycle`, `TransformStepLifecycle`, and `TransformMessage`

Event Types that have a `Mode` field in the payload:
	`ETTransformStart = Type("tf:Start")`
	`ETTransformStop = Type("tf:Stop")`
	`ETTransformStepStart = Type("tf:StepStart")`
	`ETTransformStepStop = Type("tf:StepStop")`
	`ETTransformStepSkip = Type("tf:StepSkip")`
	`ETTransformError = Type("tf:Error")`

Event Types that do NOT have a `Mode` field in the payload:
	
`ETTransformPrint = Type("tf:Print")`
This event is only called via the `print` starlark command, not sure we need to be adding a "mode" to this, as I believe it's handled differently than other events in the frontend.

`ETTransformDatasetPreview = Type("tf:DatasetPreview")`
This event has an event of type `dataset.Dataset`, and so we would need to add a new transform event to incorporate a `Mode` field into this event payload. However, I believe that it, too, is treated differently than other events in the frontend, and so we may not need to add a "mode" to this as well.


base/dataset.go
```diff
@@ -228,9 +227,15 @@ func ListDatasets(ctx context.Context, r repo.Repo, term, profileID string, offs
...
+type Lister interface {
+	// List the dsref.VersionInfos of a single user
+	List(ctx context.Context, pid profile.ID, lp params.List) ([]dsref.VersionInfo, error)
+}
+
 // RawDatasetRefs converts the dataset refs to a string
-func RawDatasetRefs(ctx context.Context, pid profile.ID, s collection.Set) (string, error) {
-	res, err := s.List(ctx, pid, params.ListAll)
+func RawDatasetRefs(ctx context.Context, pid profile.ID, l Lister) (string, error) {
+	res, err := l.List(ctx, pid, params.ListAll)
```
ramfox, line 238: there was a import cycle when importing `transform` into `collection`. 

Narrowing down the interface and adding it to `base` seemed the best way to go.

transform/apply.go
```diff
@@ -36,6 +36,25 @@ const (
...
+	// RMCommit indicates the transform was executed as a "commit"
+	// meaning, the transform was run with the intension to save the
+	// output dataset
+	RMCommit = Mode("commit")
+)
+
+func (m Mode) String() string {
+	return string(m)
+}
+
```
ramfox, line 57: Since we cannot import `transform` into the `event` package, added this helper func to convert `Modes` into `strings`.

However, having `Mode` be its own type may be overkill, based on how we are using it.

May be better to just have some const strings `ApplyMode = "apply"` & `CommitMode = "commit"`.

b5: I think your post-PR intuition is correct: we don't really need an extra type for the `RunMode` enumeration. Looking through this PR a few things jump out as a consequence of the extra type in the `transform` package:

* collection now needs to import transform
* seeing lots of `.String()` to-and-froming
* forces the `base` package to grow. We want to avoid having base be a catch-all dumping ground if we can

So, I think this comment is spot-on. Let's have `RMApply` and `RMCommit` be string constants `apply` and `commit` respectively

transform/apply.go
```diff
@@ -61,6 +80,7 @@ func (t *Transformer) Apply(
 	wait bool,
 	scriptOut io.Writer,
 	secrets map[string]string,
+	mode Mode,
```
b5, line 83: Given that we're switching away from an "enumeration" (we don't  have true enums in go), I'd like a stronger mechanism for checking that we're using this correctly. Given that we're only talking about two possible modes, the right way to approach this is with a set of two exported methods that both call the same unexpected method with different `RunMode` strings:

```go

// Apply applies the transform script to a target dataset
func (t *Transformer) Apply(
	ctx context.Context,
	target *dataset.Dataset,
	runID string,
	wait bool,
	scriptOut io.Writer,
	secrets map[string]string,
) error {
	return t.apply(ctx, target, runID, wait, scriptOut, secrets, RMApply)
}

// Commit applies the transform script to a target dataset, associating all 
// events with the "commit" RunMode
func (t *Transformer) Commit(
	ctx context.Context,
	target *dataset.Dataset,
	runID string,
	wait bool,
	scriptOut io.Writer,
	secrets map[string]string,
) error {
	return t.apply(ctx, target, runID, wait, scriptOut, secrets, RMCommit)
}

func (t *Transformer) apply(
	ctx context.Context,
	target *dataset.Dataset,
	runID string,
	wait bool,
	scriptOut io.Writer,
	secrets map[string]string,
	runMode string,
) error {
	log.Debugw("applying transform", "runID", runID, "wait", wait)

	if target.Transform == nil {
		return errors.New("apply requires a transform component")
	}
      // ...
```

here the old `Apply` now becomes an unexported function that both reference, and within `lib/datasets.go/Save` we'd call `transformer.Commit` instead of apply. In the future we can invert this & have the transformer call save by adding to the `Commit` method.

Adding this second method to `transformer` means we should also rename this file from `apply.go` to `transform.go`

ramfox: perf, love it

event/transform.go
```diff
@@ -65,6 +68,7 @@ const (
 
 // TransformMessage is the payload for print and error events
 type TransformMessage struct {
-	Lvl TransformMsgLvl `json:"lvl"`
-	Msg string          `json:"msg"`
+	Lvl  TransformMsgLvl `json:"lvl"`
+	Msg  string          `json:"msg"`
+	Mode string          `json:"mode,omitempty"`
```
b5, line 73: ah ok, so we're able to do modes on individual messages as well in the end. Not totally sure this'll be necessary, but might be useful for an "ignore all events with mode === `apply`" on the frontend

b5 (approved):
_delightful_

event/transform.go
ramfox: mentioned above, but this is currently only used for `Error` event types, which use the message payload. For the other two message types right now, `Print` & `DatasetPreview`, we don't have a way of plumbing down the mode without adding complexity, so I didn't add them for now.

PR #1905, opened 2021-09-09
Overhaul Transform API

b5:
After numerous design discussions, we've arrived at a new API for running transforms. This commit makes a number of breaking changes at once:

* `download(ctx)` and `transform(ds,ctx)` special functions are **removed**
* a new `dataset` module is loaded by default into all transform scripts, the module is scoped to the  dataset the transform script is bound to
* a new module function `dataset.latest()` takes no arguments and returns the HEAD dataset version, returns an empty dataset if no history exists
* a new module function `dataset.commit(ds)` requires a dataset argument and **must be called by the user exactly once** in a transform script to update dataset. The dataset passed to commit is passed directly to save, without merging into the prior version
* calling `dataset.commit` halts the program
* calling `dataset.commit` _zero_ times produces a warning, but is not an error (will error with no changes in practice)
* starlark `context` module is removed
* `config` is an immutable dict-like type replaces `ctx.get_config(key)` with `config.get(key[, default])`. `config` is loaded into the transform script by default. config may inherit more dictionary-type properties in the future.
* `secrets` is an immutable dict-like type, replaces `ctx.get_sercret(key)` with `secrets.get(key[, default])`. secrets is loaded into the transform script by default. secrets may inherit more dictionary-type properties in the future.
* `dataset.TransformStep.Category` is no longer used
* starlark steps are no longer frozen after execution, allowing symbol manipulation across steps

An example conversion from old to new:

OLD:
```py
load("http.star", "http")
---
def download(ctx):
  csvDownloadUrl = ctx.get_config("url")
  return http.get(csvDownloadUrl).body()
---
def transform(ds, ctx):
  # ctx.download is whatever download() returned
  csv = ctx.download
  # set the dataset body
  ds.set_body(csv, parse_as='csv')
```

NEW:
```py
load("http.star", "http")
ds = dataset.latest()
---
csvDownloadUrl = config.get("url")
res = http.get(csvDownloadUrl).body()
---
ds.set_body(res, parse_as='csv')
dataset.commit(ds)
```

BREAKING CHANGES:
We've overhauled the starlark API to use a cell-based approach. All transform scripts will require updating


b5:
@dustmop, after a multi-hour rebase nightmare, I've got most of the changes we discussed re-worked atop the new dataframe api. tests in the `cmd` package aren't passing, but while I get those fixed up I'd love a review.

transform/startf/testdata/load_ds.star
```diff
@@ -1,6 +1,6 @@
 load('assert.star', 'assert')
 
-movies = load_dataset("peer/movies")
+movies = load_version("peer/movies")
```
dustmop, line 3: Is this function staying as `load_version`. If so, it reads weird to pass it the ref to a dataset, without any version information. Would rather keep it as `load_dataset`.

transform/startf/transform.go
```diff
@@ -168,229 +144,194 @@ func ExecScript(ctx context.Context, target *dataset.Dataset, opts ...func(o *Ex
 	resolve.AllowSet = o.AllowSet
 	resolve.AllowLambda = o.AllowLambda
 	resolve.AllowNestedDef = o.AllowNestedDef
+	resolve.LoadBindsGlobally = true
+	resolve.AllowRecursion = true
```
dustmop, line 148: Do we need AllowRecursion? Enabling that is equivalent to enabling unbounded loops (while loops) which makes analysis much more complicated in the general case. Would love if we could keep this disabled.

transform/startf/transform.go
```diff
@@ -168,229 +144,194 @@ func ExecScript(ctx context.Context, target *dataset.Dataset, opts ...func(o *Ex
...
-	t.globals, err = starlark.ExecFile(thread, pipeScript.FileName(), pipeScript, t.locals())
-	if err != nil {
-		if evalErr, ok := err.(*starlark.EvalError); ok {
-			return fmt.Errorf(evalErr.Backtrace())
-		}
-		return err
+// RunStep runs the single transform step using the dataset
+func (r *StepRunner) RunStep(ctx context.Context, ds *dataset.Dataset, st *dataset.TransformStep) error {
+	r.globals["print"] = starlark.NewBuiltin("print", r.print)
+	r.globals["load_version"] = starlark.NewBuiltin("load_version", r.LoadVersionFunc(ctx, ds))
```
dustmop, line 177: So `load_version` is staying despite us keeping the name `Dataset` instead of `Version`?

transform/startf/transform.go
```diff
@@ -168,229 +144,194 @@ func ExecScript(ctx context.Context, target *dataset.Dataset, opts ...func(o *Ex
...
 	}
 }
 
-// ModuleLoader sums all loading assets to resolve a module name during transform execution
-func (t *transform) ModuleLoader(thread *starlark.Thread, module string) (dict starlark.StringDict, err error) {
-	if module == skyqri.ModuleName && t.skyqri != nil {
-		return t.skyqri.Namespace(), nil
+func (r *StepRunner) print(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var message starlark.String
+	if err := starlark.UnpackArgs("print", args, kwargs, "message", &message); err != nil {
```
dustmop, line 283: Something that the built-in starlark `print` function does is support for auto stringification. That means as long as a type implements `starlark.Stringer` it can be passed as the first argument to `print` and will be converted to a string automatically. As written, this function instead requires this:

```
df = dataframe.DataFrame()
print('%s' % df)
```

rather than just:

```
df = dataframe.DataFrame()
print(df)
```

Would be nice if we could add this functionality as well. Doesn't have to happen in this refactoring PR, but a TODO would be nice.

dustmop (commented):
First pass looks pretty great! Just want to confirm the intention of moving `load_dataset` to `load_version`?

transform/startf/testdata/load_ds.star
b5: ooops, should totally be `load_dataset`, will change

transform/startf/transform.go
b5: woah I completely forgot how hard this makes things on the static analysis side. Yes will definitely disable

transform/startf/transform.go
b5: ugh good point. our print is terrible. fixed.

PR #1909, opened 2021-09-13
fix(collection): use `logbook` to fill in missing run and commit count data during `collection` migration

### LINKED ISSUE #1904: bug(collection): migration does not fill in all necessary fields
Currently the migration of a collection only checks the repo for information.

To get all the necessary data here is where the migration would need to check:

WorkflowID: `workflow.Store.Get` using the InitID
RunCount: `logbook.BranchLog` filtering for `RunModel`, adding each `RunModel`
RunStatus: `logbook.BranchLog` get latest `RunModel` op. Status found in the `op.Note` field
RunDuration: same  op as above , `op.Size` field
RunID: same as above, `op.Ref` field

ramfox:
closes #1904

### DIFF
--- a/collection/migrate.go
+++ b/collection/migrate.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/qri-io/qri/base/dsfs"
 	"github.com/qri-io/qri/dsref"
+	"github.com/qri-io/qri/logbook"
 	"github.com/qri-io/qri/repo"
 )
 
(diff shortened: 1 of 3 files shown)


collection/migrate.go
```diff
@@ -42,6 +43,42 @@ func MigrateRepoStoreToLocalCollectionSet(ctx context.Context, s Set, r repo.Rep
...
+		dlog, err := ulog.Log(datasets[i].InitID)
+		if err != nil {
+			log.Errorf("can't get dataset log for dataset %s, %s", datasets[i].InitID, err)
+			continue
+		}
+		if len(dlog.Logs) < 0 {
+			log.Errorf("no branch logs for dataset log %s, %s", datasets[i].InitID, err)
+			continue
+		}
+		blog := dlog.Logs[0]
```
ramfox, line 61: This would be much "cleaner" if logbook easily returned the "main" branch log based on an InitID, and then that `BranchLog` had a method `CommitOps` and `RunOps`, that returned a list of ops of just the `CommitModel` and `RunModel` respectively.

However, this is the first time we've needed something like that, and the place we need it is in migration code, whose whole purpose is to pull the information in a way that is not typically expected.

I'm trying to spell out that I don't mind that there isn't a convenient way to get a specific Op from logbook, since the whole point of `collection` is to retain the important information about the dataset, rendering the need to pull it from logbook mostly mute (except in this specific case).

collection/migrate.go
```diff
@@ -42,6 +43,42 @@ func MigrateRepoStoreToLocalCollectionSet(ctx context.Context, s Set, r repo.Rep
...
+		commitCount := 0
+		runCount := 0
+		mostRecentRunRecorded := false
+		for _, op := range blog.Ops {
+			if op.Model == logbook.CommitModel {
+				commitCount++
+				continue
+			}
```
b5, line 69: this is close, but you need to account for removed and amended versions as well. Here's an example, but instead of constructing new VersionInfos, you'd be adjusting commitCount to decrement on remove type actions: 

https://github.com/qri-io/qri/blob/4e63b2f8ecafab531370a8ea45232e64df45002c/logbook/logbook.go#L1234-L1257

collection/migrate.go
b5: agreed. In situations like this, create an unexported function in this file and call it here, making all of this a subroutine that we can consider moving into logbook if it's needed elsewhere

b5 (commented):
just need to handle deletes, then we're good

collection/migrate.go
ramfox: oh nice thank you! I completely forgot that remove ops are also model type `CommitModel`.

b5 (approved):
LGTM!

PR #1910, opened 2021-09-15
refactor(startf): transforms should return a dataset with the `Structure.Entries`

### LINKED ISSUE #1907: bug(apply): need to calculate `ds.Structure.Entries` for the applied dataset
Fixes qri-io/frontend#254

First, this will add some time to the amount of time it takes for a dryrun to occur, since we would need to iterate over the entire dataset to learn the proper row count. However, my personal opinion is this is one of the main things a user would be looking to when doing a dry run as a sanity check.

To fix this, we need a very paired down equivalent of the `dsfs.calculateComputedFields` method, `calculateEntries` that uses a `dsio.EntryReader` to iterate over the rows of the dataset & track the row count.

We can call this in the `transform.Apply` function, after we have done the actual transform apply because we split `Apply` into `Apply` and `Commit`.

`Commit` is used exclusively in the "Save" path (which does it's own computed fields calculation), so keeping this work inside `Apply` means we will never "double calculate" the number of entries.

ramfox:
closes #1907 

also closes qri-io/frontend#254

### DIFF
--- a/transform/startf/ds/dataset.go
+++ b/transform/startf/ds/dataset.go
@@ -387,5 +387,37 @@ func (d *Dataset) AssignBodyFromDataframe() error {
 	if err != nil {
 		return err
 	}
+	// adding `Entries` here allows us to know the entry count for
+	// transforms that are "applied" but not "commited"
+	// "commited" dataset versions get `Entries` and other stats
+	// computed at the time the version is saved
+	d.ds.Structure.Entries = df.NumRows()
+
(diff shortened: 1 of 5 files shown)


ramfox:
@dustmop lmk if the comments add the correct context & if I should add/adjust any of it.

dustmop (approved):
LGTM, great set of comments to explain the behavior being added!

PR #1914, opened 2021-09-16
feat(key): add key-to-base64 encoder and decoder functions

b5:
our keys-to-string serialization is a mess. this helps standardize the use of `base64.StdEncoding` across the codebase for key-to-string serialization.

@Arqu I've written this specifically to avoid adding a DecodeKey func to the cloud utils package


auth/key/key.go
```diff
@@ -43,3 +44,53 @@ func IDFromPubKey(pubKey crypto.PubKey) (string, error) {
...
+		return nil, fmt.Errorf("decoding base64-encoded private key: %w", err)
+	}
+
+	pk, err = crypto.UnmarshalPrivateKey(data)
+	if err != nil {
+		return nil, fmt.Errorf("invalid private key: %w", err)
+	}
+
+	return pk, nil
+}
```
b5, line 96: Most of this PR is clean up. This is the func I'd like to see us making use of on cloud

Arqu (approved):
Ah, very nice. Also this is just one more example of our issues with serializations we have. 
Dustin recently worked on making `profile.ID` conversions explicit with `.Encode()`. `key.ID` suffers from the same with `.Pretty()` and pub/priv keys here are probably the better of the current cases we have.

I got a mental note to do a higher order upgrade of our key management once we get to key types. Should probably work on a spec for this.

PR #1916, opened 2021-09-16
fix(transform): transforms default to CSV format

### LINKED ISSUE #1915: transform scripts should default to CSV format
with our latest switch to using `ds.body` assignment, we've lost the `parse_as` argument that we formerly relied on to set data format. As a short-term fix we'd like to _default_ to CSV data format.

In a follow-up we'll detect the number of dimensions in the body and error if it's higher than 2.

b5:
with our latest switch to using `ds.body` assignment, we've lost the `parse_as` argument that we formerly relied on to set data format. As a short-term fix we'd like to _default_ to CSV data format.

In a follow-up we'll detect the number of dimensions in the body and error if it's higher than 2.

closes #1915


dustmop (approved):
LGTM

PR #1917, opened 2021-09-17
overhaul crypto generator API, support Ed25519 in JWT token construct/verify

b5:
Still lots more to do to make sure we _actually_ support ED keys, but this is a good start. To that end, this PR adds Ed25519 test keys. Any `testkeys.GetKeyData(11)` or higher is an ED key. Next step is to use them all over in tests and see if anything explodes.

I also need to do a little research into public key in-lining, and weather we want this or not.

This PR breaks the `key.CryptoGenerator` interface by removing two methods. I'm not marking this as a breaking change because it's an internal API.


auth/auth_test.go
```diff
@@ -17,6 +18,7 @@ import (
 )
 
 func TestAuthAndGetProfile(t *testing.T) {
+	golog.SetLogLevel("token", "debug")
```
Arqu, line 21: Do we need to keep this line?

Arqu (approved):
Neat

auth/auth_test.go
b5: ah good catch. thank you!

PR #1918, opened 2021-09-17
fix(get): extend pagination on /get

Arqu:
Fix for https://github.com/qri-io/frontend/issues/299
Currently only supports `limit` & `offset` params, but user facing endpoints should take in `page` & `pageSize` too.


ramfox (approved):
LGTM!

PR #1921, opened 2021-09-23
feat(websocket): add process to authenticate websocket connections

### LINKED ISSUE #1912: feat(lib): scope websocket connections
Not every websocket connection should receive all events.

Now that events can be scoped, we should only send certain events on certain websocket connections.

To do that, we must associate each connection with a scope (profile id) and only send events with that profile id over the connection.

---
in `lib/websocket.go`
The `wsHandler.conns` should be `map[Profile.ID]*websocket.Conn`

`wsHandler.WSConnectionHandler` should parse the token in the request. If it's valid, it should continue with accepting a websocket connection. The connection should be saved to `wsHandler.conns` indexed by the profile id.

`wsHandler.wsMessageHandler` - if the event does not have a scope, it should be sent over every connection (???). If the event does have a scope, it should be sent over the associated connection.

---

The frontend would need to delay it's websocket connection "upgrade" until the user has logged in. Non-logged in users should not be able to receive any messages over the websocket. (this follows nicely with out current set up, we require a user to create an account or sign in before running anything).

potential issues to research:
1) anything coming through dispatch is properly scoped. We also made a point to scope automated runs. Are there potentially other parts of the codebase that aren't scoped?
2) Is there are, do they need to be scoped? If something is a system wide message, for example, should we also be passing events that have no scope over all connections?
3) should the TTL of the token be the TTL of the connection? In which case, we need to keep track of TTLs and close any outstanding connections. How would a token-refresh play into any of this. Is there a way to notify the websocket handler of any refreshes, so we can extend the TTL without dropping connections? If there is a dropped connection, it should be the responsibility of the frontend to request a new connection.

ramfox:
closes #1912 

also moves `websocket` work into its own package


lib/websocket.go
```diff
@@ -20,7 +43,16 @@ type WebsocketHandler interface {
...
 	// Collect all websocket connections
-	conns []*websocket.Conn
+	conns         map[string]*WSConn
+	connsLock     sync.Mutex
+	tokenSource   key.Store
+	subscriptions map[string]string
+	subLock       sync.Mutex
+}
+
+type WSConn struct {
```
ramfox, line 53: terrible name

Not sure what to call this. It's here to help us associate profileIDs and with connIDs in the "other" direction. maybe just calling it a `Conn` is okay? or would that be too confusing with the other kinds of conns?

lib/websocket.go
```diff
@@ -58,14 +103,140 @@ func (h *wsHandler) wsMessageHandler(_ context.Context, e event.Event) error {
...
+
+	claims, ok := tok.Claims.(*token.Claims)
+	if !ok || claims.Subject == "" {
+		return fmt.Errorf("cannot get profile.ID from token")
+	}
+	// TODO(b5): at this point we have a valid signature of a profileID string
+	// but no proof that this profile is owned by the key that signed the
+	// token. We either need ProfileID == KeyID, or we need a UCAN. we need to
+	// check for those, ideally in a method within the profile package that
+	// abstracts over profile & key agreement
```
ramfox, line 144: this TODO is pulled from `inst.activeProfile`, where we do the same mining for the profileID using the token. Should probably be pulled into its own method in... token? it would return a (string, error), erroring if the token is invalid or if the profileID does not exist on the token

lib/websocket.go
```diff
@@ -20,7 +44,16 @@ type WebsocketHandler interface {
...
+	conns         map[string]*wsConn
+	connsLock     sync.Mutex
+	keystore      key.Store
+	subscriptions map[string]string
+	subLock       sync.Mutex
+}
+
+type wsConn struct {
+	profileID string
+	conn      *websocket.Conn
```
ramfox, line 56: terrible name

Not sure what to call this. It's here to help us associate profileIDs and with connIDs in the "other" direction. maybe just calling it a Conn is okay? or would that be too confusing with the other kinds of conns?

lib/websocket.go
```diff
@@ -58,14 +102,154 @@ func (h *wsHandler) wsMessageHandler(_ context.Context, e event.Event) error {
...
+
+	claims, ok := tok.Claims.(*token.Claims)
+	if !ok || claims.Subject == "" {
+		return fmt.Errorf("cannot get profile.ID from token")
+	}
+	// TODO(b5): at this point we have a valid signature of a profileID string
+	// but no proof that this profile is owned by the key that signed the
+	// token. We either need ProfileID == KeyID, or we need a UCAN. we need to
+	// check for those, ideally in a method within the profile package that
+	// abstracts over profile & key agreement
```
ramfox, line 144: this TODO is pulled from inst.activeProfile, where we do the same mining for the profileID using the token. Should probably be pulled into its own method in... token? it would return a (string, error), erroring if the token is invalid or if the profileID does not exist on the token

lib/websocket.go
```diff
@@ -58,14 +102,154 @@ func (h *wsHandler) wsMessageHandler(_ context.Context, e event.Event) error {
...
+type message struct {
+	Type    msgType         `json:"type"`
+	Payload json.RawMessage `json:"payload"`
+}
+
+// subscribeMessage is the expected structure of an incoming "subscribe"
+// message
+type subscribeMessage struct {
+	Token string `json:"token"`
+}
```
ramfox, line 255: slightly modeled after how we do `events`, but with the understanding that we are communicating using json

lib/websocket_test.go
```diff
@@ -26,23 +35,106 @@ func TestWebsocket(t *testing.T) {
...
 	}
 
-	wsCancel()
+	// add connection
+	randIDStr := "test_connection_id_str"
+	SetIDRand(strings.NewReader(randIDStr))
+	connID := newID()
+	SetIDRand(strings.NewReader(randIDStr))
+
+	wsh.WSConnectionHandler(mockWebsocketWriterAndRequest())
```
ramfox, line 77: this tests that the `wsHandler` is setting up a connection & placing it in the `conns` map.

However, it doesn't set us up to be able to test the `wsHandler.read` or `wsHandler.handleMessage` methods. May need to expand on the `mockHijacker` struct below  & create methods that allow us to write on the connection.

lib/websocket.go
b5: all of these `ws` prefixes are a clue this should be moved into a sub package `github.com/qri-io/qri/lib/websocket`. Rename `WebsocketHandler` to `websocket.Handler`, and keep only the interface and constructor exported.

Also, this struct will need an `ID` field in the end so we can transition back to "anonomyous" connections

lib/websocket.go
```diff
@@ -29,7 +62,11 @@ var _ WebsocketHandler = (*wsHandler)(nil)
 // can connect to in order to get realtime events
 func NewWebsocketHandler(ctx context.Context, inst *Instance) (WebsocketHandler, error) {
```
b5, line 63: making this a subpackage means we can't accept an `*Instance` argument, and that's a good thing. Accept a keystore and bus instead.

lib/websocket.go
```diff
@@ -2,15 +2,39 @@ package lib
...
+
+// SetIDRand sets the random reader that NewID uses as a source of random bytes
+// passing in nil will default to crypto.Rand. This can be used to make ID
+// generation deterministic for tests. eg:
+//    myString := "SomeRandomStringThatIsLong-SoYouCanCallItAsMuchAsNeeded..."
+//    workflow.SetIDRand(strings.NewReader(myString))
+//    a := NewID()
+//    workflow.SetIDRand(strings.NewReader(myString))
+//    b := NewID()
+func SetIDRand(r io.Reader) {
```
b5, line 34: no need to export this

lib/websocket.go
```diff
@@ -20,7 +44,16 @@ type WebsocketHandler interface {
 // and serves to maintain the list of connections
 type wsHandler struct {
 	// Collect all websocket connections
```
b5, line 46: drop this outdated comment

lib/websocket.go
```diff
@@ -20,7 +44,16 @@ type WebsocketHandler interface {
 // and serves to maintain the list of connections
 type wsHandler struct {
```
b5, line 45: I think this can be renamed to `connections`. let's change the comment to `connections maintains the set of active websocket connections & associated connection metadata`

lib/websocket.go
```diff
@@ -58,14 +102,154 @@ func (h *wsHandler) wsMessageHandler(_ context.Context, e event.Event) error {
...
-			}
-		}(k, c)
+	profileIDString := e.ProfileID
+	if profileIDString == "" {
+		log.Debugf("Event with SessionID %q has no scope. Not sending event over websocket.", e.SessionID)
+		return nil
+	}
+	connID, ok := h.subscriptions[profileIDString]
+	if !ok {
+		return fmt.Errorf("no websocket connection ID found for profile %q", profileIDString)
```
b5, line 112: no need to leak that it's an ID that's not found:

```suggestion
		return fmt.Errorf("no connection found for profile %q", profileIDString)
```

lib/websocket.go
```diff
@@ -58,14 +102,154 @@ func (h *wsHandler) wsMessageHandler(_ context.Context, e event.Event) error {
...
-	for k, c := range h.conns {
-		go func(k int, c *websocket.Conn) {
-			err := wsjson.Write(ctx, c, evt)
-			if err != nil {
-				log.Errorf("connection %d: wsjson write error: %s", k, err)
-			}
-		}(k, c)
+	profileIDString := e.ProfileID
+	if profileIDString == "" {
+		log.Debugf("Event with SessionID %q has no scope. Not sending event over websocket.", e.SessionID)
```
b5, line 107: This'll cause logging pollution. Unstead of debug logging on _every_ message that lacks a profile ID, debug log when we're sending to a known connection, which you're doing below.

lib/websocket.go
```diff
@@ -20,7 +44,16 @@ type WebsocketHandler interface {
 // and serves to maintain the list of connections
 type wsHandler struct {
 	// Collect all websocket connections
-	conns []*websocket.Conn
+	conns         map[string]*wsConn
+	connsLock     sync.Mutex
+	keystore      key.Store
+	subscriptions map[string]string
```
b5, line 50: an interesting side effect of this design: each profile can have at most one websocket connection, which will cause two browser tabs from the same profile to fail. Let's add a TODO to support multiple connections for the same profile, which will require an array of connection ID strings as the value instead of a single string.

We need it someday, but not to get an MVP out the door

lib/websocket.go
```diff
@@ -38,15 +75,22 @@ func NewWebsocketHandler(ctx context.Context, inst *Instance) (WebsocketHandler,
...
 	}
-	h.conns = append(h.conns, c)
+	connID := newID()
+	wsc := &wsConn{
+		conn: conn,
+	}
+	h.connsLock.Lock()
+	defer h.connsLock.Unlock()
+	h.conns[connID] = wsc
+	go h.read(connID)
```
b5, line 93: pass `r.Context()` to the read method here

lib/websocket.go
```diff
@@ -58,14 +102,154 @@ func (h *wsHandler) wsMessageHandler(_ context.Context, e event.Event) error {
...
+	}
+	h.connsLock.Lock()
+	defer h.connsLock.Unlock()
+	delete(h.conns, connID)
+}
+
+// read listens to the given connection, handling any messages that come through
+// stops listening if it encounters any error
+func (h *wsHandler) read(id string) error {
+	ctx := context.Background()
```
b5, line 189: accept a context argument instead of using `context.Background()`

lib/websocket.go
```diff
@@ -58,14 +102,154 @@ func (h *wsHandler) wsMessageHandler(_ context.Context, e event.Event) error {
...
+	msg := &message{}
+	var err error
```
b5, line 191: I'm a fan of using named return values to avoid needing lines like this, so long as you always explicitly declare returns (eg: `return err`, never `return`)

lib/websocket.go
```diff
@@ -58,14 +102,154 @@ func (h *wsHandler) wsMessageHandler(_ context.Context, e event.Event) error {
...
+	wsc, ok := h.conns[id]
+	if !ok {
+		return fmt.Errorf("connection for connection ID %q not found", id)
+	}
+
+	for {
+		err = wsjson.Read(ctx, wsc.conn, msg)
```
b5, line 198: plumbing the request context here _should_ mean the context gets closed when the websocket connection is closed. I haven't looked at the source, but hopefully `wsjson` picks up on closed contexts and returns an error. That'll mean connections are pruned here instead of leaking & running forever

lib/websocket.go
```diff
@@ -58,14 +102,154 @@ func (h *wsHandler) wsMessageHandler(_ context.Context, e event.Event) error {
...
+		if err != nil {
+			// all websocket methods that return w/ failure
+			// close the connection
+			h.removeConn(id)
+			return err
+		}
+		go h.handleMessage(id, msg)
```
b5, line 205: no need for a goroutine here. we want this connection to get a response before doing anything else

lib/websocket/websocket.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+	id := newID()
+	c := &conn{
+		id:   id,
+		conn: wsc,
+	}
+	h.connsLock.Lock()
+	defer h.connsLock.Unlock()
+	h.conns[id] = c
+	go h.read(r.Context(), id)
```
ramfox, line 107: hmmm, adding the request context here is causing the connection to be canceled very quickly. 

The examples in the `nhooyr/websocket` repo use the request context in this way, so I'm not sure what is different about our request context that is making this context cancel prematurely.

b5: hrm, ok. Let's file a follow-on issue after this is merged to look into it. Shouldn't be a show-stopping thing

b5 (approved):
LGTM

lib/websocket/websocket.go
ramfox: filed issue referencing this convo https://github.com/qri-io/qri/issues/1925

PR #1922, opened 2021-09-23
fix(run): run store deserialization updated

Arqu:
@ramfox This made a lot of issues in parsing the runs from the run store and introduces some follow up behavior issues. 
The only gotcha is that we have to make sure all the payloads do actually conform to the `event.TransformMessage` payload type which I'm still not a 100% on as it's a bit hard to track down.

### DIFF
--- a/automation/run/run_test.go
+++ b/automation/run/run_test.go
@@ -1,6 +1,7 @@
 package run
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 	"testing"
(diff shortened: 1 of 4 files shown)


automation/run/run_test.go
```diff
@@ -123,6 +150,16 @@ func TestStateAddTransformEvent(t *testing.T) {
...
 
+			serializedRun, err := json.Marshal(got)
+			if err != nil {
+				t.Fatal(err)
+			}
+			deserializedRun := &State{}
+			err = json.Unmarshal(serializedRun, deserializedRun)
+			if err != nil {
+				t.Fatal(err)
+			}
```
b5, line 161: this test should check that state can round-trip to JSON and be equal to input:

```go
if diff := cmp.Diff(got, deserializedRun); diff != "" {
  t.Errorf("serialization mismatch (-want +got):\n %s")
}
```

b5 (commented):
Want to confirm what you mean by this:

> The only gotcha is that we have to make sure all the payloads do actually conform to the event.TransformMessage payload type which I'm still not a 100% on as it's a bit hard to track down.

Do you mean all payloads of event type `event.ETTransformPrint`? If so I'm ok with requiring the payload of that event always have the same type, but want to confirm other events submitted to the run store (like ETTransformStart) aren't included in this requirement

Arqu:
@b5 sorry, bad wording on my end. What I mean is that the handler for the specific payload used to accept a plain `string`, and now I've forced it to be an `event.TransformMessage`. It's hard to find all the places we send those to make sure we don't send any more strings. But yes, all of this is contained strictly to the `event.ETTransformPrint, event.ETTransformError` events so damage is scoped. I can make it more resilient by trying to deserialize either tho, but would rather enforce the correct default behavior.

b5 (approved):
LGTM

PR #1926, opened 2021-09-24
fix(websocket): request context causes early cancellation

ramfox:
Using the http request context as the context for `read` causes the connection to close early. The examples used in the `nhooyr/websocket` repo do not cause this issue, so not totally sure what the error is here. However, in order to have the context survive, we need to use the background context. Thankfully, if the connection closes we are able to catch it in the `read` function. It will error and we can prune the connection. This will have to do until we can devote more time to understand what is going wrong.


b5 (approved):
Lgtm

PR #1929, opened 2021-09-27
fix(lib): hack - deploy path needs to create a `runID`

### LINKED ISSUE #1928: bug(lib): deploy need to create its own `runID` early so frontend can coordinate
layed out this in comment https://github.com/qri-io/frontend/issues/258#issuecomment-928196331

But basically, until we can save without needing a body/structure, during a "Deploy" we have to "run" transforms as part of the "save" path, rather than separately as part of an optional "run".

However, frontend needs to have the `RunID` ASAP in order to coordinate events.

For now, as soon as we know that a dataset is going to have a transform as part of the `Deploy`, we need to generate a `runID` and pass it to the frontend through the `ETAutomationDeployStart` event.

ramfox:
Until we can save a dataset without a body or structure, the `Deploy`
path is going to handle the "running" of any transforms. In order for
the frontend to coordinate events, it needs to know this `runID`. If we
detect that we will be "apply"ing a transform during a "save" in a
"deploy", we must create a `runID` early & pass it to the frontend
through the deploy events ASAP. When we can save a dataset without a
structure or a body, then these pieces can be removed.

closes #1928 and unblocks qri-io/frontend#258

### DIFF
--- a/lib/automation.go
+++ b/lib/automation.go
@@ -244,10 +244,23 @@ func (automationImpl) Deploy(scope scope, p *DeployParams) error {
 func deploy(scope scope, p *DeployParams) {
 	vi := dsref.ConvertDatasetToVersionInfo(p.Dataset)
 	ref := vi.SimpleRef().String()
+
+	// TODO(ramfox): bandaid! remove when save can handle having saving a dataset with no body/structure
+	runID := ""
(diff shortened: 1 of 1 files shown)


b5 (approved):
Works for me

PR #1930, opened 2021-09-28
fix(lib): scope the `lib.apply` context

ramfox:
Missed this on my first pass because I figured this scoping was taken care of using the `newScopeFromWorkflow` function. However, this only scopes the `scope.Context` not the `scope.AppContext`, which is what we need in `lib.apply`.

Since the `lib.run` and `lib.apply` use the `scope` for the `newScopeFromWorkflow` differently, I thought it best to adjust the scoping issue in `lib.apply` only, rather than `newScopeFromWorkflow`. If there comes a time when `scope.Context()` for both workflow runs and workflow applies are both expected to last the context of the app itself, we may want to make this adjustment in `newScopeFromWorkflow` instead.

### DIFF
--- a/lib/automation.go
+++ b/lib/automation.go
@@ -15,6 +15,7 @@ import (
 	"github.com/qri-io/qri/dsref"
 	"github.com/qri-io/qri/event"
 	qhttp "github.com/qri-io/qri/lib/http"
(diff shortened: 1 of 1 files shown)


dustmop (approved):
LGTM, this seems correct to me!

PR #1932, opened 2021-09-30
feat(edkeys): added ed key support

Arqu:


auth/key/gen.go
```diff
@@ -18,29 +17,44 @@ type CryptoGenerator interface {
 	GeneratePrivateKeyAndPeerID() (string, string)
 }
 
-// cryptoGenerator is a source of cryptographic info
-type cryptoGenerator struct{}
+// cryptoGenerator is a source of cryptographic info for RSA keys
+type cryptoGenerator struct {
+	algo int
```
b5, line 22: this should be of type `crypto_pb.KeyType`, or at least an `int32`. Irritating that they don't alias the Key Type in `crypto`, you could alias it in this package as well

b5 (approved):

PR #1935, opened 2021-10-08
refactor(ws): add support for multiple connections

Arqu:
Enables users to have multiple ws connections open. This was a nuisance to users and also killed e2e testing.


Arqu:
@ramfox The only think I'm a bit unsure here is if I'm handling unsubscribe correctly. Not sure where from it might be called externally and what the end goal in that case might be. Does the request mean terminate ALL connections for the profile or just that specific one?

PR #1949, opened 2021-10-27
feat(automation): `CancelRun` will cancel run associated with the `runID`, if that is the run that is running

### LINKED ISSUE #1946: feat(automation): cancel runs early
We want a user to be able to cancel runs early, both "apply" and "commit" runs. We also want to be able to cancel `Deploys` & `Workflows`.

However, can we cancel saves, mid save? Let's look at the code path.

Because the concerns happen in different places. If we are trying to cancel a "deploy", it may be that we are canceling during a "save", "workflow update", or "run".

Just Runs

- Add `cancelRunCh`, channel living on `Orchestrator`
- In `runWorkflow` and `ApplyWorkflow`, wrap the context in a context enclosure. In a go routine, select:  for messages coming off the `cancelRunCh`, if it matches the runID, call `cancel`, and for the ctx.Done, close the routine after either happens.
- Add `orchestrator.CancelRun` that takes a runID string & sends it over the `cancleRunCh` channel.
- add `CancelRun` method in `lib.Automation` that takes a runID string
- add method endpoint `AECancelRun`
- Testing this is going to be annoying. New test? `ApplyWorkflow` and `RunWorkflow` where the `apply` & `run` functions delay for 4 seconds. Test that after we call `CancelRun`, we timeout within less time.

ramfox:
closes #1946 

Also fixes some `orchestrator.runLock` errors.

### DIFF
--- a/automation/orchestrator.go
+++ b/automation/orchestrator.go
@@ -56,7 +56,11 @@ type Orchestrator struct {
 	// the workflows in the expected order, running only as many at once as configured, and
 	// allows communication back to the user about where they are in the run queue, allows for
 	// cancelling runs that haven't happened yet
-	runLock      sync.Mutex
+	runLock sync.Mutex
+	// TODO(ramfox): `cancelRunCh` is the current shim to ensure we can
+	// cancel the currently running run. Like the above `runLock`, this should be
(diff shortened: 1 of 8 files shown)


dustmop (approved):
LGTM!

PR #1951, opened 2021-10-28
feat(transform): Control flow graphs built by starlark analyzer

dustmop:


b5 (approved):
great progress! In review I've mainly just read & tried to catch up to your approach. I'm able to follow the way you've laid out the program.

Biggest next step IMHO is laying out the public API contract before diving into handling all the conditional edges

dustmop:
> Biggest next step IMHO is laying out the public API contract before diving into handling all the conditional edges

I think the next thing I want to add is the actual dataflow analysis, at least some basic version of it that works in an interesting case, since I have a version of that sitting in a branch. After that, completely agree it's time to focus on the public facing API, and after that handling all the edge cases and other TODOs.

b5:
ah lovely! given that you have a branch yes let's flip the order & move to that next. Public API can wait

PR #1952, opened 2021-10-28
fix(automation): validate write access

Arqu:


lib/automation.go
```diff
@@ -210,6 +210,9 @@ func (automationImpl) Apply(scope scope, p *ApplyParams) (*ApplyResult, error) {
 			return nil, err
 		}
 	}
+	if err := scope.Logbook().ProfileCanWrite(scope.Context(), ref.InitID, scope.ActiveProfile()); err != nil {
```
b5, line 213: can we get at least one test that proves this denies access locally?

Arqu: oh, right... mb rushing stuff lately.

b5 (approved):
LGTM

PR #1957, opened 2021-11-02
chore(automation): add logging for run lock and unlock

ramfox:


automation/orchestrator.go
```diff
@@ -322,14 +322,24 @@ func (o *Orchestrator) RunWorkflow(ctx context.Context, wid workflow.ID, runID s
 	return runID, o.runWorkflow(ctx, wf, runID)
 }
 
+func (o *Orchestrator) lock() {
+	o.runLock.Lock()
```
dustmop, line 326: Could we get two log lines here, one before the call to `o.runLock.Lock()`, and one after? When the lock is already in use, that call will block, meaning it could be useful to see the time between them.

dustmop (approved):
LGTM

PR #1960, opened 2021-11-03
feat(transform): Recover from transform panics, return as error

dustmop:


Arqu (approved):
LGTM. I'm going to go ahead and merge this.

PR #1961, opened 2021-11-03
fix(events): fix event synchronization and add logging

Arqu:


Arqu:
Battle tested yesterday, going to move forward and merge.

PR #1962, opened 2021-11-03
feat(automation): orchestrator queue events

Arqu:


automation/orchestrator.go
```diff
@@ -349,6 +349,18 @@ func (o *Orchestrator) listenForCancelationAndUnlock(ctx context.Context, cancel
 }
 
 func (o *Orchestrator) runWorkflow(ctx context.Context, wf *workflow.Workflow, runID string) error {
+	go func() {
+		if err := o.bus.Publish(ctx, event.ETAutomationRunQueuePush, &runID); err != nil {
+			log.Debug(err)
+		}
+	}()
+	defer func() {
```
Arqu, line 357: Unsure about how this flies, especially with ctx cancelations. Also real ugly, but expect you to pull this into the queue interface soon.

event/event.go
```diff
@@ -152,8 +152,8 @@ func (b *bus) PublishID(ctx context.Context, typ Type, sessionID string, payload
 
 func (b *bus) publish(ctx context.Context, typ Type, sessionID string, payload interface{}) error {
 	log.Debugw("publish", "type", typ, "payload", payload)
-	b.lk.Lock()
-	defer b.lk.Unlock()
+	b.lk.RLock()
```
Arqu, line 155: Relaxed these as they weren't a real issue and allow us to actually queue multiple jobs, otherwise we will still hang on every handler run for it to complete for the short term.

automation/orchestrator.go
ramfox: So "Push" just means it goes on the queue, and "Pop" just means we've pulled it off the queue (not that the entire run as finished). With that in mind, "Push" is in the right place, but "Pop" should go right after the "lock" (and not in a defer), since that is what is simulating "coming off the queue" at the moment.

automation/orchestrator.go
```diff
@@ -422,6 +434,18 @@ func (o *Orchestrator) ApplyWorkflow(ctx context.Context, wait bool, scriptOutpu
...
+		go func() {
+			if err := o.bus.Publish(ctx, event.ETAutomationApplyQueuePop, &runID); err != nil {
+				log.Debug(err)
+			}
+		}()
+	}()
```
ramfox, line 448: same comment as above

automation/orchestrator.go
Arqu: ahhh, right right! ty

ramfox (approved):
LGTM

PR #1965, opened 2021-11-08
fix(lib): go routine to assure instance context cancellation now closes properly

ramfox:
whelp, intro-ed a bug and now fixing it. Go routine to log that the instance context has canceled never returned properly.


Arqu (approved):
👍 nice catch

PR #1967, opened 2021-11-08
refactor(api): clean up and move util funcs

Arqu:


lib/http/api.go
```diff
@@ -152,14 +153,18 @@ const (
...
 
 // DsRefFromPath parses a path and returns a dsref.Ref
-func DsRefFromPath(path string) (dsref.Ref, error) {
-	refstr := PathToQriPath(path)
-	return dsref.ParsePeerRef(refstr)
+func DsRefFromPath(prefix, path string) (dsref.Ref, error) {
+	refstr := PathToQriPath(prefix, path)
+	if !strings.Contains(refstr, "/") {
+		return dsref.Ref{Username: refstr}, nil
+	}
```
b5, line 160: I'm highly suspicious of this line. This PR is more about hygiene tho, so I'm happy to let it go under the presumption that we'll add more testing to the api package in general in the future

b5 (approved):
LGTM

lib/http/api.go
Arqu: ehhh, I screwed this up with copy pasta should just be `dsref.ParsePeerRef(refstr)` as it was only updated for the added prefix handling.

PR #1969, opened 2021-11-08
feat(dispatch): Cursor for collection.List pagination

dustmop:
A working Cursor implementation that can work generically for many methods. For now, use it just for collection.List. This includes an http response with the json data needed to fetch the next page.

### DIFF
--- a/api/util/responses.go
+++ b/api/util/responses.go
@@ -7,10 +7,9 @@ import (
 
 // Response is the JSON API response object wrapper
 type Response struct {
-	Data       interface{}  `json:"data,omitempty"`
-	Meta       *Meta        `json:"meta,omitempty"`
-	NextPage   *NextPageReq `json:"nextpage,omitempty"`
-	Pagination *Page        `json:"pagination,omitempty"`
+	Data     interface{}  `json:"data,omitempty"`
+	Meta     *Meta        `json:"meta,omitempty"`
+	NextPage *NextPageReq `json:"nextPage,omitempty"`
 }
 
 // Meta is the JSON API response meta object wrapper
@@ -24,8 +23,8 @@ type Meta struct {
 type NextPageReq struct {
 	Method      string `json:"method"`
 	URL         string `json:"url"`
-	ContentType string `json:"contenttype"`
-	JSONBody    string `json:"jsonbody"`
+	ContentType string `json:"contentType"`
+	JSONBody    string `json:"jsonBody"`
 }
 
 // WriteResponse wraps response data in an envelope & writes it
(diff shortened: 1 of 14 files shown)


api/util/responses.go
```diff
@@ -7,9 +7,10 @@ import (
 
 // Response is the JSON API response object wrapper
 type Response struct {
-	Data       interface{} `json:"data,omitempty"`
-	Meta       *Meta       `json:"meta,omitempty"`
-	Pagination *Page       `json:"pagination,omitempty"`
+	Data       interface{}  `json:"data,omitempty"`
+	Meta       *Meta        `json:"meta,omitempty"`
+	NextPage   *NextPageReq `json:"nextpage,omitempty"`
+	Pagination *Page        `json:"pagination,omitempty"`
```
b5, line 13: hmm... we either need to remove `Pagination` or move `NextPage` into the `Page` struct

api/util/responses.go
```diff
@@ -7,9 +7,10 @@ import (
 
 // Response is the JSON API response object wrapper
 type Response struct {
-	Data       interface{} `json:"data,omitempty"`
-	Meta       *Meta       `json:"meta,omitempty"`
-	Pagination *Page       `json:"pagination,omitempty"`
+	Data       interface{}  `json:"data,omitempty"`
+	Meta       *Meta        `json:"meta,omitempty"`
+	NextPage   *NextPageReq `json:"nextpage,omitempty"`
```
b5, line 12: we use `lowerCamelCase` for field names. should be `nextPage`

api/util/responses.go
```diff
@@ -19,6 +20,14 @@ type Meta struct {
 	Error   string `json:"error,omitempty"`
 }
 
+// NextPageReq is the request to get the next page of results
+type NextPageReq struct {
+	Method      string `json:"method"`
+	URL         string `json:"url"`
+	ContentType string `json:"contenttype"`
+	JSONBody    string `json:"jsonbody"`
```
b5, line 28: json representation should be `contentType`, `jsonBody`. Do we _really_ need the content type field, or can we rely on the convention of "supply content type if you want the non-default"?

api/util/responses.go
dustmop: I think removing Pagination is the better choice. Nothing is relying on that aspect of the http api?

api/util/responses.go
dustmop: I think including the ContentType in the response is useful because it makes it incredibly apparent how to call the api. It's not strictly necessary, but if a consumer were to look at the json response and see this:

```
"nextPage": {
  "method": "POST",
  "url": "/list",
  "contentType": "application/json",
  "jsonBody": "{\"offset\":50,\"limit\":25}"
}
```

then that's basically telling the exact `curl` command to use:

```
curl -X POST -H "Content-Type: application/json" -d '{"offset":50,"limit":25}' http://localhost:2503/list
```

which is nice for quality of life.

lib/cursor.go
```diff
@@ -0,0 +1,79 @@
...
+			if pos != -1 {
+				jsonName = jsonName[:pos]
+			}
+			lowerName = strings.ToLower(jsonName)
+		}
+		v := target.Field(i)
+		if v.IsZero() {
+			continue
+		}
+		result = append(result, fmt.Sprintf("%s=%v", lowerName, v))
```
b5, line 76: did you consider using [`url.Values`](https://pkg.go.dev/net/url#Values) to build the response? it'll have the advantage of sorting by key on call to `values.Encode()`

api/util/responses.go
b5: I'm realizing now that I've double-raised this issue in review. Answering here for completeness

This convenience is coming at the cost of wire bloat. This change is de-facto sending a constant `"method":"POST","contentType":"application/json",` per request. As UTF8 that's at least 49 bytes of _required_ overhead that we could just cover with documentation. Coupled with the fact that 99.99% of responses will not be read by humans, I think we need to err on the side of performance.

lib/cursor.go
dustmop: Unfortunately, url.Values is not very convenient, as it uses `map[string][]string`, when we really want `map[string]string` instead (non-list values).

dustmop:
Pushed a commit handling these suggestions.

b5 (approved):
LGTM

PR #1972, opened 2021-11-12
feat(transform): Detect terminal size, pass it to transform to affect DataFrame rendering

dustmop:
Depends on https://github.com/qri-io/starlib/pull/146

This PR adds terminal size detection to `qri apply`. The width and height of the terminal is passed down to `lib/` and eventually to `transform/` and `starlib/dataframe/`, where it affects how DataFrames get stringified. Users of the http api can also send this same info via normal arguments, in order to match the correct size of whatever html element they are going to display upon.

The first commit refactors the automation orchestrator a bit, to make passing this info a bit easier. Some new nomenclature is introduced here: rather than `Run` for runs that commit and `Apply` for runs that do not commit, this change uses `RunAndCommit` and `RunEphemeral`. The idea is that `apply` as a piece of terminology was intended as a high level command to contrast with `qri save`. When deeper in the code base, with a different context, it isn't quite as obvious which of these is and isn't making a new commit, especially since those parts of the code are referring to `Workflows` which are being `run` in either case.

Some parts of this are missing functionality. In particular, `qri save --apply` is not handling terminal size information. That can be done in a follow-up PR, the groundwork laid here will make it much easier to implement.

### DIFF
--- a/automation/orchestrator.go
+++ b/automation/orchestrator.go
@@ -40,8 +40,8 @@ type OrchestratorOptions struct {
 // only non-test implementation is lib.Instance, but this interface is used
 // to avoid a direct dependency
 type WorkflowRunner interface {
-	RunEphemeral(context.Context, string, *workflow.Workflow, *dataset.Dataset, bool, WorkflowRunParams) error
-	RunAndCommit(context.Context, string, *workflow.Workflow, ioes.IOStreams, WorkflowRunParams) error
+	RunEphemeral(ctx context.Context, runID string, wf *workflow.Workflow, ds *dataset.Dataset, wait bool, params WorkflowRunParams) error
+	RunAndCommit(ctx context.Context, runID string, wf *workflow.Workflow, streams ioes.IOStreams, params WorkflowRunParams) error
 }
 
 // WorkflowRunParams are additional parameters for a workflow run
--- a/automation/orchestrator.go
+++ b/automation/orchestrator.go
@@ -40,8 +40,15 @@ type OrchestratorOptions struct {
 // only non-test implementation is lib.Instance, but this interface is used
 // to avoid a direct dependency
 type WorkflowRunner interface {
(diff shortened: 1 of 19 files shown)


automation/orchestrator.go
```diff
@@ -36,44 +36,43 @@ type OrchestratorOptions struct {
...
-
-// Apply executes an ephemeral workflow transform
-type Apply func(ctx context.Context, wait bool, runID string, w *workflow.Workflow, ds *dataset.Dataset, secrets map[string]string) error
+// WorkflowRunner is for running workflows using some execution engine
+// only non-test implementation is lib.Instance, but this interface is used
+// to avoid a direct dependency
+type WorkflowRunner interface {
+	RunEphemeral(context.Context, string, *workflow.Workflow, *dataset.Dataset, bool, WorkflowRunParams) error
+	RunAndCommit(context.Context, string, *workflow.Workflow, ioes.IOStreams, WorkflowRunParams) error
+}
```
b5, line 45: As a rule all interface method declarations should include variable names to help the reader

I'm into this change, but we should cc @ramfox

automation/orchestrator.go
```diff
@@ -36,44 +36,43 @@ type OrchestratorOptions struct {
...
 
-// ApplyFactory is function that produces an Apply function
-type ApplyFactory func(ctx context.Context) Apply
+// WorkflowRunParams are additional parameters for a workflow run
+type WorkflowRunParams struct {
+	Secrets      map[string]string
+	OutputWidth  int
+	OutputHeight int
```
b5, line 51: I'm into the declaration of the `WorkflowRunParams` struct. Another possible place to put these fields is on `ioes.IOStreams`

automation/orchestrator.go
ramfox: v into this

ramfox (approved):
Love these orchestrator changes! 
LGTM

PR #1975, opened 2021-11-22
fix(transform): Log errors recovered from transforms

dustmop:


Arqu (approved):
LGTM, but since it's a new logger, can you add this one as an option to the log-all mechanics on instance creation, think we'll be adding more of these with time.

dustmop:
Yes, good catch! Will add that.

PR #1977, opened 2021-11-23
feat(staticlark): Dataflow analysis for sensitive data

dustmop:
An early implementation of dataflow analysis for starlark. Some caveats:
* control flow not being used, only linear flow is analyzed
* no lexical scoping nor inner functions
* axioms need to be populated

### DIFF
--- a/transform/staticlark/analyze.go
+++ b/transform/staticlark/analyze.go
@@ -24,10 +24,18 @@ func AnalyzeFile(filename string) ([]Diagnostic, error) {
 	globals := newSymtable(starlark.Universe)
 	// Build a graph of all calls, using top level calls and pre-defined globals
 	callGraph := buildCallGraph(funcs, topLevel, globals)
+
+	// Trace sensitive data using dataflow analysis
+	dataflowDiags, err := analyzeSensitiveDataflow(callGraph, nil)
+	if err != nil {
+		return nil, err
+	}
+
 	// Return any unused functions
 	// TODO(dustmop): As more analysis steps are introduced, refactor this
 	// into a generic interface that creates Diagnostics
-	return callGraph.findUnusedFuncs(), nil
+	unusedDiags := callGraph.findUnusedFuncs()
+	return append(dataflowDiags, unusedDiags...), nil
 }
 
 // Diagnostic represents a diagnostic message describing an issue with the code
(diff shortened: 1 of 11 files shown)


transform/staticlark/analyze.go
```diff
@@ -24,10 +24,18 @@ func AnalyzeFile(filename string) ([]Diagnostic, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
 	// Return any unused functions
 	// TODO(dustmop): As more analysis steps are introduced, refactor this
 	// into a generic interface that creates Diagnostics
-	return callGraph.findUnusedFuncs(), nil
+	unusedDiags := callGraph.findUnusedFuncs()
+	return append(dataflowDiags, unusedDiags...), nil
```
b5, line 38: delighted to see the `Diagnostic` struct already paying dividends here

transform/staticlark/environment.go
```diff
@@ -0,0 +1,109 @@
...
+	"strings"
+
+	"go.starlark.net/syntax"
+)
+
+const sensitiveVarName = "$"
+
+// environment is a collection of variable bindings, and the
+// list of other bindings that it derives value from
+// TODO(dustmop): Handle lexical scoping
```
b5, line 15: what's the plan for scoping? make the environment struct recursive, or a flat struct that applies scopes to string keys?

transform/staticlark/dataflow_test.go
```diff
@@ -0,0 +1,117 @@
...
+			sensitiveReturn: true,
+		},
+		"dangerous": &funcNode{
+			dangerousParams: []bool{true, false},
+			reasonParams: []reason{
+				reason{lines: []string{"assume it is dangerous"}},
+				reason{},
+			},
+		},
+	}
```
b5, line 30: This test was super helpful for getting my head around axioms

b5 (approved):
This is great. I don't want to interrupt with design critique b/c this feels like a midpoint more than a refactoring moment. I find your selected terminology great, really helps lever into the complexity of the topic

transform/staticlark/environment.go
dustmop: Yes, add a stack to the environment, pushing and pulling as scopes are entered and left. Each binding needs a unique slot, and the name corresponds to that slot based upon which scope it is looked up in. Same type of structure used in a compiler or interpreter.

PR #1978, opened 2021-11-24
feat(api): POST a body file to the new `/ds/save/form` endpoint to create a dataset

ramfox:
Added `/ds/save/upload` endpoint as a "sugar" endpoint to our API. The `SaveByUploadHandler` assumes that all parts of the `lib.SaveParams` will come from the multipart form.

To test via curl:

```
curl -X POST -F ref=peername/dataset_name -F bodyFile=@my_new_body_file.csv localhost:2503/ds/save/upload
```


api/endpoints.go
```diff
@@ -24,4 +24,6 @@ const (
 	AEGetCSVShortRef qhttp.APIEndpoint = "/ds/get/{username}/{name}/body.csv"
 	// AEUnpack unpacks a zip file and sends it back
 	AEUnpack qhttp.APIEndpoint = "/ds/unpack"
+	// AESaveByForm is the route used to save a dataset using a multipart form file in the request
+	AESaveByForm qhttp.APIEndpoint = "/ds/save/form"
```
ramfox, line 28: alternative name or endpoint suggestions?

api/handlers.go
```diff
@@ -161,6 +172,169 @@ func UnpackHandler(routePrefix string) http.HandlerFunc {
...
+	bodyfile, bodyHeader, err := r.FormFile("body")
+	if err == http.ErrMissingFile {
+		err = nil
+	} else if err != nil {
+		err = fmt.Errorf("error opening body file: %w", err)
+		return
+	}
+	if bodyfile != nil {
+		ds.BodyPath = bodyHeader.Filename
+		ds.SetBodyFile(qfs.NewMemfileReader(bodyHeader.Filename, bodyfile))
```
ramfox, line 332: This is the biggest departure from previous iterations:

Rather than opening the file & consuming the reader to inline the bytes, we instead just attach the reader as a script file to the body (or viz, or transform, or readme)

api/handlers.go
```diff
@@ -161,6 +172,172 @@ func UnpackHandler(routePrefix string) http.HandlerFunc {
...
+			p.Dataset = &dataset.Dataset{}
+		}
+		if err := formFileDataset(r, p.Dataset); err != nil {
+			util.WriteErrResponse(w, http.StatusBadRequest, err)
+			return
+		}
+
+		ds, err := inst.Dataset().Save(r.Context(), p)
+		if err != nil {
+			// what should the error response be
```
dustmop, line 204: I would suggest using `RespondWithError` from `api/util/errors.go`. It does some work to inspect the error and respond with an appropriate http status code. As long as `Dataset.Save` is returning known errors, the response code will be correct, and if it isn't, at least fixing that can be centralized to the `ResponseWithError` method.

api/endpoints.go
ramfox: I like this rename!

api/handlers.go
```diff
@@ -161,6 +172,172 @@ func UnpackHandler(routePrefix string) http.HandlerFunc {
...
+			}
+		case ".json":
+			if err = json.NewDecoder(datafile).Decode(ds); err != nil {
+				err = fmt.Errorf("error decoding json file: %w", err)
+				return
+			}
+		}
+	}
+
+	if peername := r.FormValue("peername"); peername != "" {
```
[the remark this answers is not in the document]
ramfox: thank you for catching this

api/handlers.go
```diff
@@ -161,6 +172,172 @@ func UnpackHandler(routePrefix string) http.HandlerFunc {
...
+		ds.Peername = peername
+	}
+	if name := r.FormValue("name"); name != "" {
+		ds.Name = name
+	}
+	if bp := r.FormValue("body_path"); bp != "" {
+		ds.BodyPath = bp
```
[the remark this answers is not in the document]
ramfox: Yea, so `Save` is expecting a `ds.BodyPath` & it uses that path as a way to generate a dataset name if no name is given. But this is probably not the place to do it. Going to remove the assignment here, move it to assigning to the `SaveParams`, and add a comment.

api/handlers.go
ramfox: thank you I forgot this existed

dustmop (approved):
LGTM!

PR #1980, opened 2021-11-26
fix(lib): `/get` endpoint should return preview if no `selector` given

### LINKED ISSUE #1973: bug(lib): `Get` should return a dataset preview if no selectors are specified
This was supposed to be the behavior of the `/get` endpoint that never actually made it into lib.

This change will allow the frontend to only make one request to fill the Dataset Preview page

ramfox:
closes #1973

### DIFF
--- a/api/testdata/expect/TestDatasetGet.test_ds.json
+++ b/api/testdata/expect/TestDatasetGet.test_ds.json
@@ -1,4 +1,11 @@
 {
+  "body": [
+    ["toronto", 40000000, 55.5, false], 
+    ["new york", 8500000, 44.4, true],
+    ["chicago", 300000, 44.4, true],
+    ["chatham", 35000, 65.25, true],
+    ["raleigh", 250000, 50.65, true]
+  ],
   "bodyPath": "/mem/QmVYgdpvgnq3FABZFVWUgxr7UCwNSRJz97vBU9YX5g5pQ4",
   "id": "6yn4jmjjsndpwf4qllvjcunrzni3o3vcejlkvdsvhayfcvrgrpsq",
   "commit": {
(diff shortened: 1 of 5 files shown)


Arqu (commented):
Code looks fine to me. I'm only worried about this changing the default behavior. Not sure of where it's all used so far and how it might impact existing users/clients.

For example:
- direct downloads - should be fine as I think most/call calls have a selector (but not a 100% due to the sugar routes)
- API usage - others have used the get endpoint to integrate a dataset into their blog/app/something
- Workflows/transforms loading a dataset might now get trimmed datasets
- Frontend expecting to load full body but getting a preview?

Would be good to do a decent pass to limit exposure here before merging this.

ramfox:
Thank you for bringing this up & making this list, it was definitely necessary to think through

> - Workflows/transforms loading a dataset might now get trimmed datasets
> - Frontend expecting to load full body but getting a preview?
In frontend, for workflows and in the history, we do a separate call to fetch the body, so frontend will not be negatively affected by this change
In backend, we only call `lib.Get` (really `instance.Datasets().Get()`) in `api`, `cmd`, and in `lib` for testing, so there is no case where some internal call will cause workflow or transform to return unexpected values.

> - direct downloads - should be fine as I think most/call calls have a selector (but not a 100% due to the sugar routes)
double checked, you are right, all download sugar routes have a selector. The only sugar route that does not it is the `Get` endpoint + the ref in the URL sugar route, and in that case the expected behavior is to return a preview.

> - API usage - others have used the get endpoint to integrate a dataset into their blog/app/something
This is the only one that I can't account for. Other folks who use qri to integrate a dataset in something will definitely be affected by this!

Should we do some notifying? a post on discord? It should definitely go in any release notes when we do a release.

Arqu (approved):
Great stuff!

Don't think the API users should be really affected was mostly just trying to enumerate some cases. We don't really have a ton of integrations with qri datasets right now and those that do as far as I know either use the `/body` selector or the `?format=zip` modifier, either of which should be a pretty safe route.

PR #1981, opened 2021-11-27
feat(List): order by "name", ascending, and "updated", descending

### LINKED ISSUE #1974: feat(collection): `List` should allow for ordering by "name", "updated" & "created"
We want the `/list` endpoint to replace the `dataset_summary/{username}` endpoint on cloud that returns a list of that user's datasets. In order to match the behavior we need the list to be able to be sorted.

The expected sorting behavior on frontend is `updated` (which is the default) & `name`. The cloud endpoint `dataset_summary/{username}` allows sorting by `created` (default) & `name`. In the `collection.localSet` (our core implementation of `collection.Set`), we need to examine the list params & order the list response before slicing for limit & offset. Let's allow for the union of options: `update`, `created`, & `name`.

ramfox:
depends on #1980 

closes #1974 

NOTES:

Unable to get a reliable way to sort by "created" in the core implementation without adding access to `logbook` for the core `Collection`. We would also need to add a `Created` field to `VersionInfo`. If/when we create a `UberData` struct, this field should be moved there. 

Initially, I thought we could rely on the natural ordering of the collection, and assume that each dataset is added to the collection in the order it was created. However, this does not hold for datasets that you pull from a remote, as these can be created at any time & pulled at any later time. However, once we pull a dataset, we have access to it's history & can determine the time of creation via the `logbook`, hence the need to for the core implementation to access `logbook` to determine the "created" time.

Further:
We need to enforce a more detailed & standard way to sort lists in qri core & to merge some of our different "list params". However, I think we should wait for the `cursor` work to land before implementing changes.

### DIFF
--- a/collection/collection.go
+++ b/collection/collection.go
@@ -8,8 +8,10 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"sync"
+	"time"
 
 	logger "github.com/ipfs/go-log"
 	"github.com/qri-io/qri/automation/workflow"
@@ -350,6 +352,8 @@ func NewLocalSet(ctx context.Context, repoDir string, options ...LocalSetOptionF
 	return s, err
 }
 
+// List currently supports ordering by name (ascending) and last update (descending)
+// default is name
 func (s *localSet) List(ctx context.Context, pid profile.ID, lp params.List) ([]dsref.VersionInfo, error) {
 	s.Lock()
 	defer s.Unlock()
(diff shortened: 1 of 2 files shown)


collection/collection.go
```diff
@@ -369,6 +373,18 @@ func (s *localSet) List(ctx context.Context, pid profile.ID, lp params.List) ([]
...
+		switch lp.OrderBy[0] {
+		case "updated":
+			sortedCol := make([]dsref.VersionInfo, len(col))
+			copy(sortedCol, col)
+			sort.Sort(byUpdated(sortedCol))
+			col = sortedCol
+		case "name":
+		default:
+		}
+	}
```
Arqu, line 386: We could have a map of order keys/sort func's to make this slimmer but not worth it for just a handful of keys.

Arqu (approved):
LGTM but need to change the PR base to the "preview" PR or wait for it to merge and rebase.

PR #1982, opened 2021-11-29
feat(lib): run info endpoint

Arqu:
Simply exposes a way to get stuff from the run store, which enables cloud to also serve its own.
Part of https://github.com/qri-io/frontend/issues/265


ramfox (approved):
LGTM

PR #1983, opened 2021-11-30
fix(api): Restore the WritePageResponse method

dustmop:
Partially revert https://github.com/qri-io/qri/pull/1969. Cloud needs the WritePageResponse. Put it back, with a TODO to remove it once more methods switch over to using cursors for pagination.


Arqu (approved):
👍 thx

PR #1987, opened 2021-12-02
fix(lib): use `username` parameter in `lib.List`

### LINKED ISSUE #1985: bug(collection): list always forces the current user
If you go to https://qri.cloud/chriswhong/ and either view the dataset list logged in as a non `chriswhong` user or logged out, you will get your own list of datasets or an empty list respectively. 

This is due to us simply passing the `scope.ActiveProfile().ID` on the call which should be fine for the CLI version, but doesn't account for multiple profiles or the case for cloud. This was not apparent as we used the old previews endpoints prior to this on cloud.

ramfox:
Also, allows fetching a dataset from the collection that is not in your own collection

closes #1985


Arqu (approved):
LGTM

Issue #1989, opened 2021-12-06
calls to /auto/runinfo are returning Timestamp and Type keys


chriswhong:
Run -> RunStep.output should be `EventLogLine`
```
export interface EventLogLine {
  type: EventLogLineType
  ts: number
  sessionID: string
  data: Record<string, any>
}
```

Calls to /auto/runinfo are returning steps where output has keys `Timestamp` and `Type` with capitalized letters.  Need to make it match the expected interface.

Related to https://github.com/qri-io/frontend/pull/579

PR #1990, opened 2021-12-07
refactor(event): align serialization properties with the rest

Arqu:
Event's are breaking out of our default pattern for naming fields when serialized. However due to a lot of these events being user and/or stored elsewhere we need to be able to deserialize either for backwards compatibility.

Fixes https://github.com/qri-io/qri/issues/1989

### DIFF
--- a/automation/run/run.go
+++ b/automation/run/run.go
@@ -243,6 +243,7 @@ type _event struct {
 	Type      event.Type      `json:"type"`
 	Timestamp int64           `json:"timestamp"`
 	SessionID string          `json:"sessionID"`
+	ProfileID string          `json:"profileID"`
 	Payload   json.RawMessage `json:"payload"`
 }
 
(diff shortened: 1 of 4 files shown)


ramfox:
There is one other place we should consider changing:

https://github.com/qri-io/qri/blob/e499860785d74ecd30f5b8d675096b6f134fb1a3/lib/websocket/websocket.go#L106-L114

This is how we serialize events over the websockets. I'm proposing we change these to match the way we serialize our other events (namely, change "ts" => "timestamp"), that way all our event serialization uses the same fields.

Arqu:
Agree with Kasey here, we should continue the same naming conventions to frontend as are on the back end.
I'll ping/file and issue on frontend to align this and make sure it lands at the same time.

Arqu:
I went ahead and did it solo, but unsure if I caught it all or if this is good at all. 
https://github.com/qri-io/frontend/pull/645

PR #1993, opened 2021-12-09
fix(collection): make event handler use independent ctx

Arqu:
This makes the collection event handler independent of the normal request execution which sometimes caused canceled context due to the event being sent from a go routine attached to a http request.


ramfox (approved):
Looks good to me!

PR #1994, opened 2021-12-14
refactor(lib, params): allow using url queries to add listing parameters

ramfox:
All endpoints that use listing can now give pagination parameters in the url:

try it using our json post endpoints:

`curl -X 'POST' localhost:2503/list?Limit=5`

sugar endpoints:
`curl 'localhost:2503/ds/get/username/name?Limit=5&Offset=5`

----

note to self: 
Before merging, squash all commits and use "breaking changes" commit notation


api/parse_request.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/qri-io/qfs"
+	"github.com/qri-io/qri/api/util"
+	"github.com/qri-io/qri/base/fill"
+	"github.com/qri-io/qri/dsref"
+	"github.com/qri-io/qri/lib"
+	"gopkg.in/yaml.v2"
+)
+
+// parseGetParamsFromRequest parse the form from the request to create `lib.GetParams`
+func parseGetParamsFromRequest(r *http.Request, p *lib.GetParams) error {
```
ramfox, line 21: co-locates all methods that pull params from request, except for the method defined on `params.List`

Definite argument can be made for getting rid of the `params.ListParams` interface, and moving the `ListParamsFromRequest` method here.

lib/datasets.go
```diff
@@ -72,32 +71,29 @@ func (m DatasetMethods) Attributes() map[string]AttributeSet {
...
 	Ref string `json:"ref"`
 	// a component or nested field names to extract from the dataset; e.g. "body"
 	Selector string `json:"selector"`
-	// number of results to limit to. only applies when selector is "body"
-	Limit int `json:"limit"`
-	// number of results to skip. only applies when selector is "body"
-	Offset int `json:"offset"`
 	// TODO(dustmop): Remove `All` once `Cursor` is in use. Instead, callers should
 	// loop over their `Cursor` in order to get all rows.
+	// TODO(ramfox): are we in a place to remove All?
```
ramfox, line 81: Have we done enough with `Cursor` to remove the `All` field?

lib/datasets.go
```diff
@@ -248,13 +201,25 @@ func scriptFileSelection(ds *dataset.Dataset, selector string) (qfs.File, bool)
 
 // ActivityParams defines parameters for the Activity method
 type ActivityParams struct {
-	ListParams
+	params.List `json:"list,omitempty"`
+	// Term filters for "history" or "run" activity type
+	Term string `json:"term,omitempty"`
```
ramfox, line 206: Filtering is pretty up in the air right now. It seems to only really be used in two cases:
1) in search
2) in Activity (between "history" & "run")

Side stepping this problem for now, by just not using `Filter` and instead staying with using `Q` field in `SearchParams` and `Term` field in `ActivityParams`.

However, one potential way to use the filter field in `SearchParams`:
`/search?filter=q:my_search_term`

`ActivityParams`:
`/log?filter=type:history`
`/log?filter=type:run`

lib/http.go
```diff
@@ -80,10 +82,41 @@ func DecodeParams(r *http.Request, p interface{}) error {
...
 
+func listParamsFromRequest(r *http.Request, p interface{}) error {
+	pa, ok := p.(params.ListParams)
+	if !ok {
+		return nil
+	}
+	lp, err := pa.ListParamsFromRequest(r)
+	if err != nil {
+		return err
+	}
```
ramfox, line 102: I could make a strong argument against having an interface `params.ListParams` &  just use reflection to determine if a struct embeds the `params.List` struct. However, reflection is pretty costly, so I figured it might be better to use a less costly route (type assertion) to determine whether we even want to move on to using reflection in the first place.

But this is also a bit odd, so looking for feedback here.

lib/peers.go
```diff
@@ -88,8 +98,17 @@ func (m PeerMethods) Disconnect(ctx context.Context, p *ConnectParamsPod) error
...
+
+// SetNonZeroDefaults sets a default limit and offset
+func (p *ConnectionsParams) SetNonZeroDefaults() {
+	if p.Offset < 0 {
+		p.Offset = 0
+	}
+	if p.Limit <= 0 {
+		p.Limit = params.DefaultListLimit
+	}
 }
```
ramfox, line 112: embedding `params.List` has the advantage of us being able to write unique `SetNonZeroDefaults`, in case we want different limits for different endpoints.

lib/search.go
```diff
@@ -28,9 +29,18 @@ func (m SearchMethods) Attributes() map[string]AttributeSet {
 
 // SearchParams defines paremeters for the search Method
 type SearchParams struct {
-	Query  string `json:"q"`
-	Limit  int    `json:"limit,omitempty"`
-	Offset int    `json:"offset,omitempty"`
+	params.List `json:"list,omitempty"`
+	Query       string `json:"q"`
```
ramfox, line 33: could replace `Query` with `params.List.Filter`. Filter hasn't been fully thought through yet, though, so was holding off.

lib/cursor_test.go
```diff
@@ -5,11 +5,12 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/qri-io/qri/base/params"
 )
 
 func TestCursorToParams(t *testing.T) {
-	lp := ListParams{
-		OrderBy: "created",
+	lp := params.List{
```
ramfox, line 12: Question about `Cursor`: should the `ToParams` method add the `params.List` information into the `nextPage.url` string?

base/params/list.go
```diff
@@ -10,11 +28,48 @@ var ListAll = List{Limit: -1}
 // List defines input parameters for listing operations
 type List struct {
 	Filter  []string
-	OrderBy []string
+	OrderBy OrderBy
```
dustmop, line 31: Can we add json struct tags to serialize all of the fields in this struct so that they must be lower-case? Makes it more consistent with how everything else works.

lib/http.go
dustmop: Yeah, I don't think the `ListParams` interface is really pulling its weight. For one, the method `ListParamsFromRequest` doesn't really use its owner at all, `p` in this case. Rather, it just wants to know if `p` has an embedded `params.List`, and if so, the relevant fields from the request should be assigned to it. I'm pretty sure you could do a type check on p instead of needing reflection. Something like:

```
  lp, ok := p.(params.List)
  if !ok {
    return
  }
  lp.limit = ...
```

Furthermore, reflection is only "slow" when compared to accessing fields directly. It's going to perform similarly to type assertions. Plus, if there's only a small number of reflection calls per http request, for example only like 4 or 5, the http overhead is going to dominate the runtime.

lib/collection.go
```diff
@@ -42,8 +42,29 @@ func (m CollectionMethods) Attributes() map[string]AttributeSet {
 // ErrListWarning is a warning that can occur while listing
 var ErrListWarning = base.ErrUnlistableReferences
 
+// CollectionListParams defines parameters for listing a user's collection
+type CollectionListParams struct {
+	params.List `json:"list,omitempty"`
```
dustmop, line 47: I could be wrong here, I forget how it works exactly, but will adding this json tag add another level to the serialized json? Would be cool if we could avoid that, if possible. We would like to get:

```
{
  "limit": ...
  "offset": ...
  "username": ...
  ...
}
```

rather than

```
{
  "list": {
    "limit": ...
    "offset": ...
  },
  "username": ...
  ...
}
```

Pretty sure that leaving off this json tag will get the former instead of the latter.

base/params/list.go
```diff
@@ -31,6 +86,13 @@ func (lp List) Validate() error {
 	return nil
 }
 
+// IsEmpty determines if the List struct is empty
+func (lp List) IsEmpty() bool {
+	return lp.Limit == 0 && lp.Offset == 0 &&
+		(lp.Filter == nil || len(lp.Filter) == 0) &&
```
dustmop, line 92: No need to check for nil. `len` will do that, and return 0 for a nil slice.

lib/datasets.go
dustmop: It is okay to remove for `collections.List`, but not `Get`. The `dataset.Get` method doesn't return a cursor yet, and `cmd` isn't trying to consume a cursor.

lib/datasets.go
dustmop: I am fine with putting something temporarily in place, as I agree it isn't well defined. `Term` looks good for now!

lib/collection.go
ramfox: you are 100% correct, and thank you for catching this!!!!!

base/params/list.go
ramfox: thank you, I was having nil pointer dereference trauma

base/params/list.go
ramfox: yessss thank youuuuu

lib/http.go
ramfox: Thank you this is so beyond valuable.

I set up a little go playground script to see how using type assertion to get at embedded fields would work, and it looks like it doesn't let us cast structs as the embedded field types: https://go.dev/play/p/_g50UfRvEMt

Your advice about the speed of reflection is very helpful. It makes me more confident that the right call is to nix the interface, and just use reflection to judge if something has list params.

Thank youuuuuuu

ramfox:
Looks like the only breaking changes are on CLI, not API, so we are good to merge without disrupting frontend work!

base/params/list.go
```diff
@@ -1,7 +1,25 @@
...
+	// ErrListParamsNotEmpty indicates that there are list values in the params
+	ErrListParamsNotEmpty = fmt.Errorf("list params not empty")
+	// DefaultListLimit is the default value for a list
+	DefaultListLimit = 25
+)
+
+// ListParams can pull list param information from an http.Request
+type ListParams interface {
+	// ListParamsFromRequest pulls list params from the URL
+	ListParamsFromRequest(r *http.Request) error
```
dustmop, line 21: The only issue I have is with this interface. I don't think it's strictly necessary, and may just be more confusing than useful. But, I don't have any strong suggestions currently on how to replace it, so it's not a big deal.

dustmop (approved):

dustmop:
LGTM, approved!

PR #1995, opened 2021-12-14
feat(transform): Analyze `if` statements by unioning each branch

dustmop:
In dataflow analysis, each branch of an `if` statement is analyzed separately. Then, the environments from those branches are merged together, since assignments from either one can influence bindings. Furthermore, variables in the conditional itself influences the values of assignments within those branches.

First commit adds a `join` field to `if` statements, which indicates where the two branches join up. This make it much easier to recursively analyze each branch of the control structure.

### DIFF
--- a/transform/staticlark/call_graph_test.go
+++ b/transform/staticlark/call_graph_test.go
@@ -55,6 +55,8 @@ top_level_func
    print
 branch_elses
  print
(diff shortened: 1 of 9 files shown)


b5 (approved):
Looking great!

PR #1996, opened 2021-12-14
fix(transform): Bump starlib, fix how OutputConfig is used

dustmop:
There was a mistake in the spelling of OutputConfig, but this bug was covered up by other bugs in starlib/dataframe. Bump our dependency to fix these bugs, and use the function SetOutputSize instead. Also fix the stack trace display.


b5 (approved):
LGTM

PR #2002, opened 2021-12-23
feat(whatchanged): Restore the `whatchanged` command

dustmop:
The command `whatchanged` returns what components changed at a specific version of in the history of the dataset.

This command was originally part of FSI, which was removed by commit eaeba31aeb6242bd4aacf79f7c20f9579585d549. However, this command is kind of its own thing, and does not have a hard dependency on FSI functionality. This change adds the command only, without the similar `status` command, which *did* depend on FSI.

### DIFF
--- a/base/component_status.go
+++ b/base/component_status.go
@@ -1,0 +1,205 @@
+package base
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/qri-io/dataset"
+	"github.com/qri-io/qfs/muxfs"
+	"github.com/qri-io/qri/base/component"
+	"github.com/qri-io/qri/base/dsfs"
+	"github.com/qri-io/qri/dsref"
+)
+
+var (
+	// STUnmodified is "no status"
+	STUnmodified = "unmodified"
+	// STAdd is an added component
+	STAdd = "add"
+	// STChange is a modified component
+	STChange = "modified"
+	// STRemoved is a removed component
+	STRemoved = "removed"
+	// STParseError is a component that didn't parse
+	STParseError = "parse error"
+	// STMissing is a component that is missing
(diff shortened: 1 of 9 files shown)


dustmop:
Example usage on command-line:

```
qri whatchanged me/some_ds@/ipfs/QmNvbzZXKSpobfmWC13FBYVSx4LouHdMu1cQKwBV5ANHKk
```

HTTP API:
```
curl -X POST localhost:2503/ds/whatchanged -H 'Content-Type: application/json' -d '{"ref": "dustmop/some_ds@/ipfs/QmNvbzZXKSpobfmWC13FBYVSx4LouHdMu1cQKwBV5ANHKk"}'
```

base/component_status.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+// StatusItem is the status of a component of a dataset, and whether it
+// was changed at a specific version in history
+type StatusItem struct {
+	SourceFile string    `json:"sourceFile"`
+	Component  string    `json:"component"`
+	Type       string    `json:"type"`
+	Message    string    `json:"message"`
+	Mtime      time.Time `json:"mtime"`
```
ramfox, line 43: mtime is coming up empty for all fetches

Not sure if it's a bug, or just not needed any longer because we no longer rely on fsi?

ramfox (approved):
Just one question about `mtime` from seeing all zero time values in each request!

Everything else looks good to me!

base/component_status.go
dustmop: That is a good point. I believe `whatchanged` never returned mtime, because it would always match the commit's timestamp. It makes sense then to remove this field, I am also going to add a unit test.

dustmop: Also removed `SourceFile` and `Message`, as those are no longer needed without file system integration.

